PR #307, opened 2015-01-05
Aggregate updates in the feed panel #294

### LINKED ISSUE #294: Aggregate updates in the feed panel

xeluna:
Fixes #294


damithc:
Kudos for starting to use pull requests @xeluna 
Try to resist commenting out code. Be brave and delete. We can always recover from past versions.
@dariusf to review and merge.

dariusf:
Alright. Should we still include this in 0.9.0?

damithc:
yes, i'd like to play around with it if possible.

src/model/TurboFeed.java
```diff
@@ -23,123 +18,149 @@
 public class TurboFeed implements Listable {
```
dariusf: `TurboFeed` is a bit of an ambiguous name. Maybe `TurboFeedItem`, or just `FeedItem`? We might need to evaluate which classes get the `Turbo*` prefix in future.

I'm not sure about implementing `Listable`. `getListName` only allows for a string representation, which is probably inadequate for what we need (arbitrary layouts). Also it's been around for a while and we might not need it any more.

src/model/TurboFeed.java
```diff
@@ -23,123 +18,149 @@
...
 	
 	public static final FilterExpression EMPTY = Qualifier.EMPTY;
```
dariusf: This isn't used.

src/model/TurboFeed.java
```diff
@@ -23,123 +18,149 @@
...
-        	return "";
-        } else {
-    		return pt.format(this.issueEvent.getCreatedAt()) + "\n" + message;
-//    				+ "\n" + this.issueEvent.getEvent();
-        }
+    	return message;
     }
-
-	@Override
-	public void copyValues(Object other) {
```
dariusf: Having to implement this method is one of reasons `Listable` is problematic.

src/service/ServiceManager.java
```diff
@@ -312,76 +332,99 @@ public int getRequestLimit(){
...
 		String user = getRepoOwner();
 		String repo = getRepoName();
+
+		// Pull Requests
+		int reqCount = 0;
+		List<PullRequest> requests = pullReqService.getPullRequests(
+				new SearchRepository(user, repo), IssueService.STATE_OPEN);
+		if (requests != null && !requests.isEmpty()) {
+			for (PullRequest request : requests) {
+				if (request != null) {
```
dariusf: This part is quite deeply nested. You could ensure that the list is non-null beforehand (it's okay if it's empty), then filter it for null elements before going into the `for` loop logic. That would eliminate 2 levels of nesting.

src/ui/feedmanagement/FeedManagementComponent.java
```diff
@@ -32,39 +32,43 @@ public VBox initialise() {
...
 		return layout;
 	}
-	
+
+	private void showIssue() {
+		TurboFeed selectedIssue = listView.getSelectionModel()
+				.selectedItemProperty().get();
+		if (selectedIssue != null) {
+			ui.triggerEvent(new IssueSelectedEvent(selectedIssue.getIssueNum(),
+					-1));
```
dariusf: `-1` shouldn't be used for the column index here. It *will* break a number of things. Since we're moving the feed into a column it shouldn't be that big a problem, but it should be done as soon as we can.

src/model/TurboFeed.java
```diff
@@ -23,123 +18,149 @@
...
+		return currIssue.getLabels();
+	}
+
 	@Override
 	public String getListName() {
+    	return getFeed();
+    }
+
+	@Override
+	public void copyValues(Object other) {
```
dariusf: Having to implement this method is one of reasons Listable is problematic.

dariusf:
I'm not sure we should merge this yet. I've made a build with these changes included and put it on the previous milestone temporarily ([here](https://github.com/HubTurbo/HubTurbo/releases/download/V0.8.1/resource-v0.9.0.jar)). Will release, tag, enable auto-update, etc. formally tomorrow.

damithc:
release v0.9 formally, and move onto v1.0. Reschedule any 0.9 stuff that
are not done yet.

Issue #299, opened 2015-01-03
Add logging to HT


damithc:
This can help in debugging memory leaks etc.

damithc:
I think we need a "rolling" log file that keeps last x Mb or n entries (so that the log file does not keep growing infinitely). Not sure if built-in logging of Java can do that. if not, we might need something like log4j

damithc:
@dariusf you may have to set up the logging infrastructure for others to start using it.

dariusf:
We are actually using log4j already, but logging is only done in a few places. It's also probably not configured optimally yet.

damithc:
DEBUG level log messages should be verbose enough to tell us the sequence
of actions that happened.

dariusf:
Logging has been added for almost all of the UI elements and operations (may have missed some). It's still needed for local file updates and syncing.

PR #311, opened 2015-01-06
Add logging to HT #299

dariusf:
#299


damithc:
No changed needed to log4j config file?
Try to limit each branch to only one issue. In case we want to undo a fix later, we should be able to do that by reversing the corresponding merge commit.

Issue #314, opened 2015-01-06
Sync-related logging


dariusf:
Log:

- changes to local files/cache
- changes to issues (if possible)
- overall changes after sync (x issues updated/added)

PR #317, opened 2015-01-06
Sync-related logging #314

taychunmei:
#314

### DIFF
--- a/src/model/Model.java
+++ b/src/model/Model.java
@@ -282,7 +282,7 @@ public class Model {
 		Platform.runLater(new Runnable() {
 			@Override
 			public void run() {	
-				logger.info(issueList.size() + " issues changed/added since last sync");
+				logger.debug(issueList.size() + " issues changed/added since last sync");
 				for (int i = issueList.size() - 1; i >= 0; i--) {
 					Issue issue = issueList.get(i);
 					TurboIssue newCached = new TurboIssue(issue, selfRef.get());
@@ -297,7 +297,7 @@ public class Model {
 		TurboIssue tIssue = getIssueWithId(issue.getId());
 		if(tIssue != null){
 			tIssue.copyValues(issue);
-			logger.info("Updated issue: " + issue.getId());
+			logger.debug("Updated issue: " + issue.getId());
 		}else{		
 			issues.add(0, issue);
 			logger.info("Added issue: " + issue.getId());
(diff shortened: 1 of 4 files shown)


src/model/Model.java
```diff
@@ -296,8 +297,10 @@ public void updateCachedIssue(TurboIssue issue){
 		TurboIssue tIssue = getIssueWithId(issue.getId());
 		if(tIssue != null){
 			tIssue.copyValues(issue);
-		}else{
+			logger.info("Updated issue: " + issue.getId());
```
damithc: This kind of detailed logs can be done at `debug` level. Furthermore, try to record what exactly was changed (i.e. the new values). Can be helpful in debugging cases of 'missing labels' etc.

src/storage/DataCacheFileHandler.java
```diff
@@ -91,7 +91,7 @@ public void setIssues(List<TurboIssue> issues) {
 	}
 	
 	public void writeToFile(String repoIdString, String issuesETag, String collabsETag, String labelsETag, String milestonesETag, String issueCheckTime, ObservableList<TurboUser> collaborators, ObservableList<TurboLabel> labels, ObservableList<TurboMilestone> milestones, ObservableList<TurboIssue> issues) {
-		logger.info("About to write to file...");
+		logger.info("About to write to file for repo: " + repoIdString);
```
damithc: Are you logging the 'last checked time' property somewhere?

damithc:
Added a couple of comments after a quick look. @dariusf to review in detail.

src/model/Model.java
```diff
@@ -384,6 +387,15 @@ private void updateCachedList(List list, List newList, String repoId){
...
 	        	list.removeAll(removed);
+  	
+	        	Listable listItem = (Listable)newList.get(0);
+	        	if (listItem.getClass() == TurboMilestone.class) {
+	        		logNumOfUpdates(newList, "milestone(s)");
+	        	} else if (listItem.getClass() == TurboLabel.class) {
+	        		logNumOfUpdates(newList, "label(s)");
+	        	} else if (listItem.getClass() == TurboUser.class) {
+	        		logNumOfUpdates(newList, "collaborator(s)");
+	        	}
```
dariusf: Using `instanceof` is probably better, as it's less brittle (it'll still work if we subclass any of the `Turbo*` classes).

Issue #323, opened 2015-01-07
Issues are wrongly indicated as pull requests


dariusf:

taychunmei:
![image](https://cloud.githubusercontent.com/assets/2648455/5641865/34206c92-967b-11e4-8e95-724738e5dacc.png)
Just wondering, are you guys also getting this weird symbol instead of the pull request img?

damithc:
I get the correct symbol (for V1.0)

PR #324, opened 2015-01-07
Issues are wrongly indicated as pull requests #323

xeluna:
#323


src/model/TurboIssue.java
```diff
@@ -82,6 +82,10 @@ public void setPullRequest(PullRequest pr){
 		this.pullRequest = pr;
 	}
 	
+	public boolean hasPullRequest() {
```
damithc: `isPullRequest`? not sure which one suits better.

dariusf:
We just discussed it and decided on `isPullRequest`.

Issue #322, opened 2015-01-07
`updated` filter qualifier


dariusf:
It filters by number of hours for now.

damithc:
Doesn't work for me. In fact if I use `updated:<24` in a filter, the whole filter doesn't work. I'm using V1.1.0

dariusf:
You may have to clear the cache for it to work (force refresh, failing that
delete .hubturbocache). Caching doesn't work very well with changes to the
data model currently.

PR #326, opened 2015-01-07
`updated` filter qualifier #322

dariusf:
#322


src/filter/Parser.java
```diff
@@ -289,17 +296,34 @@ private FilterExpression parseRangeOperator(String name, Token token) {
 				assert false : "Should not happen";
 				return null;
 			} else {
-				// Number
+				// May a number or something else
```
damithc: typo?

PR #370, opened 2015-01-12
A flood of refresh triggered after computer is awaken from sleep/hibernation #369

### LINKED ISSUE #369: A flood of refresh triggered after computer is awaken from sleep/hibernation
It also causes spike in memory usage. #282

taychunmei:
Fixes #369

### DIFF
--- a/src/service/ServiceManager.java
+++ b/src/service/ServiceManager.java
@@ -157,7 +157,7 @@ public class ServiceManager {
 				UIReference.getInstance().getUI().triggerEvent(new RefreshDoneEvent());
 			}
 		};
-		pollHandler = pollExecutor.scheduleAtFixedRate(pollTask, 0, SECS_BETWEEN_POLLS, TimeUnit.SECONDS);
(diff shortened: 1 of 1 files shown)


damithc:
Does it solve the problem?

taychunmei:
I tested with my computer. It seemed to work

damithc:
@dariusf for review

dariusf:
fixedRate will run tasks at a constant rate, no matter how long the tasks take to finish.

fixedDelay will wait till a task finishes, then schedule the next one from that point.

I'm not sure that this addresses this exact issue (wasn't it fixed with the change from timer to thread pool?), but fixedDelay is the one we want.

damithc:
How big is the thread pool?
Also consider the behavior when the connection to the server is not
available. The network connection may not be available for a short while
soon after the computer wakes up, which might cause this kind of problems.

dariusf:
It's 1 large. It's just meant to run the periodic update over and over. The reason for changing it was that the executor classes are more robust than Timer.

This shouldn't be affected by network issues. If the network is available it just doesn't update on that tick now.

PR #381, opened 2015-01-16
Refactoring in preparation for future fixes #380

### LINKED ISSUE #380: Refactoring in preparation for future fixes
Groundwork for a number of concurrency- and memory-related issues. Won't completely fix but slightly improves the situation for #276, #379, #282.

- Reduced maximum memory usage and startup time.
- Code is simpler and less brittle. Fewer complications from data binding.
- #379 happens much less frequently.

dariusf:
Fixes #380


dariusf:
On second thought I don't think this should be merged yet. Noticed some regressions today. We should probably have more tests before a big change like this.

damithc:
yup, best to add more tests before refactoring.

dariusf:
Superseded by #401

PR #385, opened 2015-01-19
Some updates missing from cards in the feed #365

### LINKED ISSUE #365: Some updates missing from cards in the feed
image

xeluna:
Fixes #365


src/service/ServiceManager.java
```diff
@@ -433,35 +432,9 @@ public Milestone editMilestone(Milestone milestone) throws IOException{
...
-							}
-						}
-					}
-				}
-				if (!toContinue) {
-					break;
-				}
-			}
-		}
-		return eventList;
```
damithc: Glad to see this method deleted. A very bad case of 'arrowhead code' :-p

PR #386, opened 2015-01-20
IssuePanel will show Comment Count

rd1992:
Only shows comment count if comments > 0


damithc:
I prefer if the comment count is shown in a non-obtrusive way, without taking much space. For example, similar to how GitHub does it.
image

dariusf:
We have the Octicon font already, which has a [comment](https://octicons.github.com/icon/comment/) icon. We could just right-align the text element and use the unicode character for the comment icon to have what GitHub shows.

rd1992:
ohh misunderstood the requirement. I thought this was what was being referred to. But i will fix it.
hubturbo1

PR #408, opened 2015-02-03
Give a keyboard shortcut to jump from the filter box to the first issue in the panel #315

### LINKED ISSUE #315: Give a keyboard shortcut to jump from the filter box to the first issue in the panel
Most of the time we are on the keyboard when typing a filter. It's nice if we can remain in the keyboard when moving to the issue list from there.

rd1992:
Hardcoded to CTRL + R in IssuePanel.java
Fixes #315

### DIFF
--- a/src/ui/components/NavigableListView.java
+++ b/src/ui/components/NavigableListView.java
@@ -157,4 +157,12 @@ public class NavigableListView<T> extends ScrollableListView<T> {
 	public void setOnItemSelected(IntConsumer callback) {
 		onItemSelected = callback;
 	}
+	
+	public void selectFirstItem(){
+		requestFocus();
+		if (getItems().size() == 0) return;
+		selectedIndex = Optional.of(0);
+		onItemSelected.accept(selectedIndex.get());
+		
+	}
 }
--- a/src/ui/issuecolumn/IssueColumn.java
+++ b/src/ui/issuecolumn/IssueColumn.java
@@ -59,7 +59,7 @@ public abstract class IssueColumn extends Column {
 	public static final FilterExpression EMPTY = filter.expression.Qualifier.EMPTY;
 	private Predicate<TurboIssue> predicate = p -> true;
 	private FilterExpression currentFilterExpression = EMPTY;
-	private FilterTextField filterTextField;
+	protected FilterTextField filterTextField;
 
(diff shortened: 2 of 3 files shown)


damithc:
Why R?
perhaps Ctrl+ down arrow is more intuitive?

damithc:
Actually, Ctrl+down arrow makes sense only if the focus is on the filter box. If we want a general shortcut to go to the first issue, we might have to choose another shortcut. Ctrl+up to go to first and Ctrl+down to go to last?

damithc:
@dariusf why this is not reviewed yet?

rd1992:
@damithc i am making changes on it. need to do select last issue. only first issue selection is done.

dariusf:
I thought it was still ongoing. It's still Ctrl+R now. Are we changing this?

Also, do we need a hotkey for going to the last issue? Home and End should work after we jump to the issue list from the filter box.

rd1992:
@dariusf yeah am changing it. Keeping it to Shift + down arrow from filter box.

damithc:
In that case, no need for a hot key to jump to the last one.
I noticed that home and end keys make the selection jump to top and bottom, but the issues don't load in bView.

damithc:
Shift is used to navigate without loading issues. So I'm not sure if Shift
is suitable for jumping to first issue if the jump results in loading the
issue.

damithc:
Perhaps Ctrl+down to jump and load, Shift+down just to jump without loading?

rd1992:
Then shift + down seems redundant. If the user wants to view the first issue, he will have to do either control + down (which is straightforward) or shift + down followed by down arrow and then up arrow. Until and unless we keep ENTER as the way of loading the issue in bView? then even after a user presses shift + down or home or end key followed by ENTER the issue loads in bView.

damithc:
Shift+down is unlikely to be used a lot, except when the user wants to load
an issue other than the first one. e.g. type the filter, press enter,
notices the third issue as the one he wants to view, press shift+down three
times to get there, and press enter to load.

rd1992:
Well i have set it to control + down for now. Shift + Down functionality is not clear to me so havent implemented it.

rd1992:
@dariusf can you merge this?

Issue #302, opened 2015-01-04
BrowserView gets disconnected randomly


damithc:
Clicking an issue should load the issue in the BView, but this stopped working at some point in the middle of an HT session. I had to restart HT to get it working again. First time I thought it was a fluke, but happened again today in a different computer.

damithc:
Here's the tail end of the log file. Hard to see which part is relevant because there is no timing information in the log file.

v0.8.0.jar:?]
	at ui.labelmanagement.LabelManagementComponent.populateTree(LabelManagementComponent.java:140) [resource-v0.8.0.jar:?]
	at ui.labelmanagement.LabelManagementComponent.createTreeView(LabelManagementComponent.java:81) [resource-v0.8.0.jar:?]
	at ui.labelmanagement.LabelManagementComponent.initialise(LabelManagementComponent.java:49) [resource-v0.8.0.jar:?]
	at ui.sidepanel.SidePanel.createLabelsTab(SidePanel.java:353) [resource-v0.8.0.jar:?]
	at ui.sidepanel.SidePanel.tabLayout(SidePanel.java:226) [resource-v0.8.0.jar:?]
	at ui.sidepanel.SidePanel.changeLayout(SidePanel.java:191) [resource-v0.8.0.jar:?]
	at ui.sidepanel.SidePanel.refreshSidebarLabels(SidePanel.java:143) [resource-v0.8.0.jar:?]
	at ui.labelmanagement.LabelManagementComponent$2.lambda$0(LabelManagementComponent.java:105) [resource-v0.8.0.jar:?]
	at ui.labelmanagement.LabelManagementComponent$2$$Lambda$435/839842443.run(Unknown Source) [resource-v0.8.0.jar:?]
	at com.sun.javafx.application.PlatformImpl.lambda$null$164(Unknown Source) [jfxrt.jar:?]
	at com.sun.javafx.application.PlatformImpl$$Lambda$47/394370609.run(Unknown Source) [jfxrt.jar:?]
	at java.security.AccessController.doPrivileged(Native Method) [?:1.8.0_20]
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$165(Unknown Source) [jfxrt.jar:?]
	at com.sun.javafx.application.PlatformImpl$$Lambda$46/1555093762.run(Unknown Source) [jfxrt.jar:?]
	at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(Unknown Source) [jfxrt.jar:?]
[...]

PR #415, opened 2015-02-04
BrowserView gets disconnected randomly #302

rd1992:


damithc:
Use the following format for pull request titles:
Issue title #issue_number

src/browserview/BrowserComponent.java
```diff
@@ -187,11 +199,17 @@ public void newMilestone() {
...
-				driver.get(GitHubURL.getPathForNewIssue());
-			}
-		});
+		if (isBrowserActive()) {
+			runBrowserOperation(() -> {
+				if (!driver.getCurrentUrl().equals(GitHubURL.getPathForNewIssue())) {
+					driver.get(GitHubURL.getPathForNewIssue());
+				}
+			});
+		}
```
damithc: Use this format `} else {`

chromedriver-log.log
```diff
@@ -0,0 +1,197 @@
+[1.052][INFO]: COMMAND InitSession {
```
damithc: Is this file necessary?

damithc:
Commented on some style issues. Didn't review the code.

src/browserview/BrowserComponent.java
```diff
@@ -161,11 +161,16 @@ private void hidePageElements() {
 	 */
 	public void newLabel() {
 		logger.info("Navigating to New Label page");
-		runBrowserOperation(() -> {
-			if (!driver.getCurrentUrl().equals(GitHubURL.getPathForNewLabel())) {
-				driver.get(GitHubURL.getPathForNewLabel());
-			}
-		});
+		if (isBrowserActive()) {
```
dariusf: We could move this check into `runBrowserOperation`. That way we'd only have to update it in one place every time. Plus we're guaranteed that uses of `driver.getCurrentUrl()` won't be null/empty in the body of tasks, etc.

src/browserview/BrowserComponent.java
```diff
@@ -235,6 +265,34 @@ public void showIssue(int id) {
...
+				default:
+					break;
+				}
+				return false;
+			}
+	}
+	
+	private void resetBrowser(){
+		logger.info("Chrome not responding.");
+		quit();
```
dariusf: `quit` was meant to be called only once, on application exit. Since `driver.quit()` kills the process, I guess it's no longer needed.

Issue #272, opened 2014-12-23
F5 doesn't refresh if nothing is selected


damithc:
e.g. immediately after a project switch (when nothing on the UI is selected), F5 does not seem to refresh (at least the count down does not reset).

PR #423, opened 2015-02-08
F5 doesn't refresh if nothing is selected #272

rd1992:


dariusf:
This will trigger a refresh on every keystroke. Probably should check if the key pressed is F5.

Also, `columns.refresh()` is not really needed any more after the recent refactoring; it's enough to do [this](https://github.com/HubTurbo/HubTurbo/blob/master/src/ui/UI.java#L163-L164). I forgot to change [MenuControl](https://github.com/HubTurbo/HubTurbo/blob/master/src/ui/MenuControl.java#L245-L246).

rd1992:
Ohh shit took out the if condition. my bad.

rd1992:
Fixed.

PR #453, opened 2015-02-16
User can jump to comment box using a hotkey #449

### LINKED ISSUE #449: User can jump to comment box using a hotkey
When an issue is selected in HT (and the issue is loaded in bView), we can use the 'C' key to jump to the comment box of the issue.

rd1992:
Fixes #449


damithc:
Note that the jump should not happen for a 'C' typed while the focus is in a filter box.

rd1992:
yeah it doesnt. i tested that out.

PR #462, opened 2015-02-17
Wrong Chrome window brought to front #450

### LINKED ISSUE #450: Wrong Chrome window brought to front
In my Win8 computer, I had another Chrome window. HT brings that window to the front instead of the one spawned by HT.

rd1992:
Fixes #450

### DIFF
--- a/src/browserview/BrowserComponent.java
+++ b/src/browserview/BrowserComponent.java
@@ -134,8 +134,13 @@ public class BrowserComponent {
 		driver.manage().window().setSize(new Dimension(
 				(int) availableDimensions.getWidth(),
 				(int) availableDimensions.getHeight()));
-		browserWindowHandle = user32.GetForegroundWindow();
+		browserWindowHandle = getHandle();
 		return driver;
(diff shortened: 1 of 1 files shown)


damithc:
@dariusf release a new version with this if this fix is ok.

dariusf:
It looks okay to me. Should we include all the currently-open PRs in the new version as well? They're all more or less standalone.

damithc:
No. For this release, merge this one only.
In future, this is what we should do between a dev release and the public
release of the same: we keep adding new features to the master. Any fixes
to the dev release should be done in a separate branch (e.g. V1.9-hotfixes)
branching off from the dev release commit. That way, we fix the dev release
without bringing in new features to the next public release. hot fix branch
should merge to master frequently so that the current master contains both
fixes to dev release and new features for future releases. If not clear, we
can discuss next time we meet.

dariusf:
Okay, I'll merge only this one. Not exactly sure of the branching workflow but I'll clarify next week.

PR #463, opened 2015-02-18
Wrong Chrome window brought to front #450

rd1992:


rd1992:
[https://github.com/HubTurbo/HubTurbo/pull/451]
this one should also be merged. otherwise the chromewindow wont come to front when minimized.

PR #473, opened 2015-02-22
progress bar goes into a loop during sync #444

### LINKED ISSUE #430: Shows issues from wrong project
This happened when I loaded powerpointlabs repo for the first time. A hard refresh fixed the problem.
image

### LINKED ISSUE #444: progress bar goes into a loop during sync
The progress bar goes in a loop during the 2nd phase of the sync (when the views are being refreshed to to match the loaded data). Is this intended? I didn't notice it in 1.8

dariusf:
Fixes #444, fixes #430.

Another attempt to fix concurrency issues. I've rewritten the sync-related parts such that thread-safety is delegated to a separate timer class: if the timer is thread-safe, our operations will (mostly) be too.

Could you all help me to eyeball that class [here](https://github.com/HubTurbo/HubTurbo/blob/Issue444/src/service/updateservice/TickingTimer.java) and see if there are any thread-safety problems? The usage of the timer is limited to [here](https://github.com/HubTurbo/HubTurbo/blob/Issue444/src/service/ServiceManager.java#L408-L512), which might be useful if you want to see it in context.

### DIFF
--- a/src/model/Model.java
+++ b/src/model/Model.java
@@ -451,7 +451,6 @@ public class Model {
 				TurboIssue newCached = new TurboIssue(issue, Model.this);
 				updateCachedIssue(newCached);
 			}
-			triggerModelChangeEvent();
 			latch.countDown();
 		});
 	}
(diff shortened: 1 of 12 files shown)


damithc:
Seems to be fixed in V2.0

PR #483, opened 2015-02-25
Upon regaining focus, sync only if the bView has changed #457

### LINKED ISSUE #457: Upon regaining focus, sync only if the bView has changed
This should reduce the annoyance of syncing everytime HT gains focus.
One simple minded way to o detect if the bView has changed: HT can keep a copy of the HTML loaded to bView and compare it with the current HTML.

rd1992:
Will only sync when bview url changes
Fixes #457

### DIFF
--- a/src/ui/UI.java
+++ b/src/ui/UI.java
@@ -69,9 +69,6 @@ public class UI extends Application implements EventDispatcher {
 	private MenuControl menuBar;
 	private BrowserComponent browserComponent;
 	private RepositorySelector repoSelector;
-	
-	// Bview changes
-	private String pageContentOnLoad = "";
 
 	// Events
 
@@ -174,7 +171,7 @@ public class UI extends Application implements EventDispatcher {
 					// A refresh is triggered if:
 					// 1. Repo-switching is not disabled (meaning an update is not in progress)
 					// 2. The repo-switching box is not in focus (clicks on it won't trigger this)
-					boolean shouldRefresh = isRepoSwitchingAllowed() && !repoSelector.isInFocus() && hasBviewChanged();
+					boolean shouldRefresh = isRepoSwitchingAllowed() && !repoSelector.isInFocus() && browserComponent.hasBviewChanged();
 					if (shouldRefresh) {
 						logger.info("Gained focus; refreshing");
 						ServiceManager.getInstance().updateModelNow();
@@ -182,14 +179,6 @@ public class UI extends Application implements EventDispatcher {
 				});
 			}
 		});
-	}
-	
-	private boolean hasBviewChanged() {
-		if (browserComponent.getCurrentPageSource().equals(pageContentOnLoad)){
-			return false;
-		}
-		pageContentOnLoad = browserComponent.getCurrentPageSource();
-		return true;
 	}
 
 	private static void initialiseJNA() {
--- a/src/ui/UI.java
+++ b/src/ui/UI.java
@@ -71,7 +71,7 @@ public class UI extends Application implements EventDispatcher {
 	private RepositorySelector repoSelector;
 	
 	// Bview changes
-	private String currentUrl = "";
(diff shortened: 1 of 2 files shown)


damithc:
Checking the URL may not be enough. For example, adding a label doesn't change the URL. Compare the full text instead?

rd1992:
switching to comparing html source of the page. will leave some test comments

rd1992:
test comment

rd1992:
test comment

rd1992:
test comment

rd1992:
another test comment

rd1992:
Has been fixed

src/ui/UI.java
```diff
@@ -69,6 +69,9 @@
 	private MenuControl menuBar;
 	private BrowserComponent browserComponent;
 	private RepositorySelector repoSelector;
+	
+	// Bview changes
+	private String currentPageSource = "";
```
damithc: better name: `pageContentAtIssueLoading` ?  Otherwise line 188 looks odd.

rd1992: alright. changing to a better name.

rd1992:
test comment

src/ui/UI.java
```diff
@@ -69,6 +69,9 @@
 	private MenuControl menuBar;
 	private BrowserComponent browserComponent;
 	private RepositorySelector repoSelector;
+	
+	// Bview changes
+	private String pageContentOnLoad = "";
```
dariusf: Could this be kept track of in the browser component instead? That way it's concerned with its own state. The UI would call `browserComponent.hasPageChanged()` or something.

rd1992: good idea. i have updated it. you can merge it.

damithc:
Seems to work only if the focus switching is done via alt+tab, but not via clicking.
e.g.
1.Select issue in HT
2. click on the page loaded in bView
3. Click on HT
At this point there should not be a sync, but it does trigger a sync.

damithc:
Note that bView updated automatically after a while (e.g. timing info such as '5 minutes ago'). This means when HT gains focus after a long absence of focus, it immediately goes into a sync because the current bView doesn't match the previous snapshot. To counter that, we might have to refine the 'sync trigger check'. I'm not sure how though. On the other hand, this may not be a big problem for users who don't switch projects too often.

rd1992:
@damithc: lets say you loaded the issue the first time and changed focus to bView. The check is only done when HT main window comes back to focus. So HT doesn't know yet that the url has changed in the bview. However any subsequent focus changes (for the same url) wont trigger a sync.

rd1992:
as per your second comment. we sync every minute. Also if HT comes to focus after a long time i feel it should trigger a sync. (use case: Lets say i add an issue while your HT window was not in focus. So to see the issue a sync is required).

rd1992:
the first issue you raised can be solved pretty easily. i probably misunderstood the requirement and implemented it this way.

damithc:
Yes it is good to sync at that point. The problem is that project switching is disabled during sync. Many a time I click on HT on the task bar to bring it to front with the intention of switching projects, but I have to wait till the sync finishes. We disabled project switching during syncs as a workaround to a data corruption issue. It was a dirty fix which is causing grief in other ways. At some point we need to go back to the root cause (data corruption if project changes during a sync) and fix it.

damithc:
The fact of the matter is syncs are good and we should do them as frequently as possible. It's just that we should do it without hampering other things.

rd1992:
alright i have pushed the latest code. it does fix the issue you described

dariusf:
The thing about the current design is that it assumes everything happens synchronously -- it wasn't a problem before we had caching as there was no state to be corrupted. We'll need to rewrite those parts (and design them better, probably with atomic updates, work queues for downloading, writing to file, etc.) to solve the problem at its root.

PR #485, opened 2015-02-25
bView disconnects randomly #471

### LINKED ISSUE #471: bView disconnects randomly
This happened again (V1.9.4). The workaround doesn't seem to work. It works if I close the bView, but doesn't work for the 'disconnect' scenario

rd1992:
Going for a reset if the browser stops responding for whatever reason.
Fixes #471

### DIFF
--- a/src/browserview/BrowserComponent.java
+++ b/src/browserview/BrowserComponent.java
@@ -251,12 +251,17 @@ public class BrowserComponent {
 	}
 	
 	private boolean isBrowserActive(){
-		if (driver.getCurrentUrl().isEmpty() && driver != null){
-			logger.warn("Unable to read url from bview. Resetting.");
-			return false;
-		}
-		else if (driver == null){
+		if (driver == null){
 			logger.warn("chromedriver process was killed !");
+			return false;
+ 		}
+		try {
+			String url = driver.getCurrentUrl();
+			if(url.isEmpty() || url == null){
+				return false;
+			}
+		} catch (WebDriverException e){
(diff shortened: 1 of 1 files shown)


src/browserview/BrowserComponent.java
```diff
@@ -252,9 +252,11 @@ public void jumpToComment(){
 	
 	private boolean isBrowserActive(){
 		if (driver.getCurrentUrl().isEmpty() && driver != null){
+			logger.warn("Unable to read url from bview. Resetting.");
 			return false;
 		}
 		else if (driver == null){
+			logger.warn("chromedriver process was killed !");
 			return false;
 		}
```
dariusf: The logic here could be structured like this:

```java
if (driver == null) {
    // chromedriver process killed
}
// driver != null from here on
else if (driver.getCurrentUrl() == null /* extra null check */ || driver.getCurrentUrl().isEmpty()) {
    // unable to read url from browser
}
```

Also, is it possible that `getCurrentUrl()` may throw `WebDriverException`? I looked at the docs and there was no mention of it, but then `WebDriverException` is unchecked, so...

damithc:
Can we wrap this up before today's dev release?

src/browserview/BrowserComponent.java
rd1992: hmm it is possible that it throws this exception. shall impement a try catch block here

rd1992:
I have taken @dariusf 's suggestions and pushed the changes.

PR #487, opened 2015-02-26
bView stops responding after alerts in window #484

### LINKED ISSUE #484: bView stops responding after alerts in window
to reproduce: try deleting a comment on an issue. Alert pops up for confirmation. HT throws an exception. is also tied to https://github.com/HubTurbo/HubTurbo/issues/471

rd1992:
fixed by catching the exception and updating chromedriver
Fixes #484

### DIFF
--- a/src/browserview/BrowserComponent.java
+++ b/src/browserview/BrowserComponent.java
@@ -288,6 +288,8 @@ public class BrowserComponent {
 					case NoSuchWindow:
 						resetBrowser();
 						runBrowserOperation(operation); // Recurse and repeat
+					case UnexpectedAlert:
+						break;
 					case NoSuchElement:
 						logger.info("Warning: no such element! " + e.getMessage());
 						break;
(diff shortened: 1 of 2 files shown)


damithc:
@dariusf am I supposed to get this new chromedriver when I run HT using the resources.jar?

dariusf:
Yup, it's packaged in the jar. You might have to delete the existing `chromedriver.exe`, so the new one gets extracted.

damithc:
How is it going to work for regular users? We don't have a way to ask them
to delete the exe file.

dariusf:
A temporary measure could be to have HT remove it on startup. That's what we wanted to do for the cache files, in case the data model changed.

The long-term solution could be to improve the updater to the point where the chromedriver can been distributed separately (so only the 12mb~ jar file and right chromedriver per platform, not all 3, are downloaded). We could then update it separately as well.

PR #500, opened 2015-03-07
Implement keyboard Shortcuts (Navigational Shortcuts) #488

### LINKED ISSUE #488: Implement keyboard Shortcuts (Navigational Shortcuts)
Implement the coming soon keyboard shortcuts in the Navigation section documented [here](https://github.com/HubTurbo/HubTurbo/wiki/Keyboard-Shortcuts)

rd1992:
Fixes #488

### DIFF
--- a/src/ui/components/FilterTextField.java
+++ b/src/ui/components/FilterTextField.java
@@ -104,6 +104,7 @@ public class FilterTextField extends TextField {
 				confirmEdit();
 			} else if (e.getCode() == KeyCode.ESCAPE) {
 				revertEdit();
+				getParent().requestFocus();
 			}
 		});
 	}
--- a/src/ui/components/FilterTextField.java
+++ b/src/ui/components/FilterTextField.java
@@ -88,7 +88,7 @@ public class FilterTextField extends TextField {
 				// Can't find out the characters deleted...
 			} else if (Character.isAlphabetic(typed)) {
 				performCompletion(e);
-			} else if (typed == ' '){
+			} else if (typed == ' ' && getText().isEmpty()){
 				e.consume();
 			}
 		});
--- a/src/ui/components/FilterTextField.java
+++ b/src/ui/components/FilterTextField.java
@@ -88,7 +88,7 @@ public class FilterTextField extends TextField {
 				// Can't find out the characters deleted...
 			} else if (Character.isAlphabetic(typed)) {
 				performCompletion(e);
-			} else if (typed == ' ' && getText().isEmpty()){
+			} else if (typed == ' ' && (getText().isEmpty() || getText().endsWith(" "))){
 				e.consume();
 			}
 		});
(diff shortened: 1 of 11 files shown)


damithc:
which shortcuts are implemented?

rd1992:
- double space to jump between issue list and filterbox
- f/b to move forward/backward between panels
- ESC to cancel edits and shift focus away from filtertext
- j/k to move between issues and load them in bview

src/ui/components/FilterTextField.java
```diff
@@ -102,6 +104,7 @@ private void setup() {
 				confirmEdit();
 			} else if (e.getCode() == KeyCode.ESCAPE) {
 				revertEdit();
+				getParent().requestFocus();
```
dariusf: `setOnCancel` could be used here. Each panel would hook into the escape/cancel event and focus itself or something when it triggers.

src/ui/issuecolumn/ColumnControl.java
```diff
@@ -223,4 +235,35 @@ public double getColumnWidth() {
...
+					handleKeys(e.getCode() == KeyCode.F);
+					assert currentlySelectedColumn.isPresent() : "handleKeys doesn't set selectedIndex!";
+					break;
+				default:
+					break;
+				}
+		});
+	}
+
+	private void handleKeys(boolean isDownKey) {
```
dariusf: Old variable name :p

dariusf:
Tried it out and it works well, just a few small issues:

- f/b don't scroll the selected panel into view. It may need something similar to ScrollableListView, likely simpler since we don't need to scroll over time.
- f/b cause the next filter box to be focused. Maybe it would be better if the panel got focus instead, so we could press ffff to go through a bunch of panels?


- The double spaces in the filter box get consumed when it's empty, but not otherwise.
- Pressing space 3 times quickly will move focus back to the filter box. The double-space detection could maintain some state to tell when it's been triggered, in which case it could reset the timer or something, so it only triggers again on the 4th press.

On hindsight it may be easier to subclass FilterTextField and implement the navigation-specific hotkeys there. Depends on what the order of events will be if we use `setOnKeyTyped` in both a class and a subclass, and if we can consume it from the subclass implementation?

damithc:
Yes, ffff... should cycle through panels in the forward direction.

rd1992:
@dariusf thanks for the insight : especially the case with three spaces.
To clarify about the filterbox being in focus when a user presses f or b. What should be in focus then? 
And the double space not being consumed: you mean the user can only type one space between filters? For eg. is:open priority:high vs is:open         priority:high?

I'll fix the other two. As for subclassing I want to do it too but would rather do it once we have a good amount of keyboard shortcuts implemented.

dariusf:
The parent of the filter box I guess, same as when pressing Esc.

More than one space is fine, but since double space is a navigation shortcut, it just seems weird that it would change the state of the filter box.

rd1992:
thanks for clarifying. So we should think of a shortcut to make the user bring focus from parent to either the filter box or the issue list. I'll comment on the internal doc we have going through

rd1992:
alright finally fixed scrolling !
1. double space now does three things : jump from filterbox to issue list. jumps from issuelist to filterbox. takes focus to filterbox of selected column
2. user cannot keep typing consecutive spaces in the filterbox. only one space is allowed.
@dariusf give the latest commit a try before merging?

PR #514, opened 2015-03-14
Non-consecutive spaces trigger double space hotkey

### LINKED ISSUE #513: Non-consecutive spaces trigger double space hotkey
For example, typing " && " (space, `&&` operator, space) quickly will shift focus to the issue list.

rd1992:
Fixes #513

### DIFF
--- a/src/ui/issuepanel/IssuePanel.java
+++ b/src/ui/issuepanel/IssuePanel.java
@@ -134,12 +134,13 @@ public class IssuePanel extends IssueColumn {
 					listView.selectFirstItem();
 				}
 				if(event.getCode() == KeyCode.SPACE){
-					if (KeyPress.isDoublePress(event, System.currentTimeMillis())) {
+					if (KeyPress.isDoublePress(event, System.currentTimeMillis()) && KeyPress.getLastKeyPressedCode() == KeyCode.SPACE) {
 						KeyPress.setTimeKeyDown(0);
 						event.consume();
 						listView.selectFirstItem();
 					}
 				}
+				KeyPress.setLastKeyPressedCode(event.getCode());
 			}
 		});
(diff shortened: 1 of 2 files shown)


dariusf:
Hmm, could the public interface for KeyPress be simplified? isDoublePress could take a key code (instead of an event, which it doesn't do anything with now) and return true or false to indicate if that event should be ignored.

KeyPress would keep track of the time of the last press, as well as the last key code. There wouldn't be any way for it to get into an inconsistent state if we forgot to call one of the setters after making a change.

rd1992:
So the only change you are suggesting is to send the keycode instead of the whole event to the keyPress class? because what you described later the class already does that.

dariusf:
Not really, I'm thinking we can move all this into KeyPress instead of having it in IssuePanel:

```java
if (KeyPress.isDoublePress(...) && KeyPress.getLastKey() == ...) {
    KeyPress.setTimeKeyDown(...);
}
KeyPress.setLast(...);
```

That way only the application logic is left in IssuePanel, and KeyPress hides all the details of checking for a double key press.

```java
if (KeyPress.isDoublePress(...) {
    // here we're assured that the double press is legitimate:
    // not a triple press, consecutive keys, etc.
}
```

rd1992:
ohh. yeah but that would be very specific to double press. if we want to extend the class to include shortcuts such as gl, gi, gp then need to keep track of the first keypress g followed by p.

dariusf:
Yes, we would add another method for that. The idea is that the class would contain all logic related to detecting key presses.

A benefit of this is that if we wanted to extend the implementation (for example, switching from a lastKeyPressed field to a buffer, to support gi, gl, etc.) the application logic wouldn't change at all.

Also locality of related code helps future maintainers, i.e. they can understand all of it by just looking at one area.

rd1992:
@dariusf i have simplified the interface. Could you take a look and merge this? I will extend the same class for the consecutive keypresses

PR #515, opened 2015-03-15
Clicking of some areas of the panel does not gain focus to the panel

### LINKED ISSUE #512: Clicking of some areas of the panel does not gain focus to the panel
For example, clicking on the area indicated below does not make the panel gain focus
image

rd1992:
Fixes #512
Also added css effect to highlight selected column

### DIFF
--- a/src/ui/issuecolumn/IssueColumn.java
+++ b/src/ui/issuecolumn/IssueColumn.java
@@ -77,12 +77,11 @@ public abstract class IssueColumn extends Column {
 		focusedProperty().addListener(new ChangeListener<Boolean>() {
 			@Override
 			public void changed(ObservableValue<? extends Boolean> unused, Boolean wasFocused, Boolean isFocused) {
-				if (!isFocused) {
-					if (getStyleClass().contains("issue-panel-focused"))
-						getStyleClass().remove("issue-panel-focused");
-					return;
+				if (isFocused) {
+				    getStyleClass().add("issue-panel-focused");
+				} else {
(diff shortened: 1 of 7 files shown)


src/ui/issuecolumn/IssueColumn.java
```diff
@@ -68,6 +70,21 @@ public IssueColumn(UI ui, Stage mainStage, Model model, ColumnControl parentColu
...
+			public void changed(ObservableValue<? extends Boolean> unused, Boolean wasFocused, Boolean isFocused) {
+				if (!isFocused) {
+					if (getStyleClass().contains("issue-panel-focused"))
+						getStyleClass().remove("issue-panel-focused");
+					return;
+				}
+				getStyleClass().add("issue-panel-focused");
+			}
+		});
 	}
```
dariusf: Is there a need to check for the presence of the style class, i.e. could the following work?

```java
if (isFocused) {
    getStyleClass().add("issue-panel-focused");
} else {
    getStyleClass().remove("issue-panel-focused");
}
```

rd1992: not really. just being defensive. if by some odd reason that css rule is set somewhere else.

dariusf: Maybe use an assertion? So if the style class gets on there for some reason we would be alerted.

rd1992: actually i realized no need for an assertion also. the remove() function does not throw exceptions if the element is not present. so no need for the if check

PR #524, opened 2015-03-17
f/g keys don't work when the focus is in the issue list 

### LINKED ISSUE #522: f/g keys don't work when the focus is in the issue list
They should still work right?

rd1992:
Fixes #522


dariusf:
Just curious, why change the switch statements to ifs? I found the earlier version neater, personally.

rd1992:
keeping style consistent. since i have combination keys (ctrl+ x) and single key presses (c, l, m).

PR #534, opened 2015-03-22
NPE when closing panel

### LINKED ISSUE #533: NPE when closing panel
Happens every time a panel is closed by clicking the 'x'. Probably due to some interaction with panel selection
```
4147508 22 Mar 2015 18:28:44 INFO UI [JavaFX Application Thread] - About to trigger event ColumnClickedEvent
4147508 22 Mar 2015 18:28:44 INFO UI [JavaFX Application Thread] - Triggered event ColumnClickedEvent
4147509 22 Mar 2015 18:28:44 ERROR UI [JavaFX Application Thread] - null
java.lang.NullPointerException
	at ui.issuecolumn.IssueColumn.lambda$1(IssueColumn.java:75) ~[bin/:?]
```

rd1992:
Fixes #533

### DIFF
--- a/src/ui/issuecolumn/IssueColumn.java
+++ b/src/ui/issuecolumn/IssueColumn.java
@@ -73,7 +73,7 @@ public abstract class IssueColumn extends Column {
 		this.setOnMouseClicked(e-> {
 			ui.triggerEvent(new ColumnClickedEvent(columnIndex));
 			try {
-				getParent().requestFocus();
+				requestFocus();
 			} catch (Exception e1) {
 				logger.warn("Nothing to focus on as panel was closed");
 			}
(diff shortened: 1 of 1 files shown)


dariusf:
The issue is that both click handlers (['x' label](https://github.com/HubTurbo/HubTurbo/blob/Issue533/src/ui/issuecolumn/IssueColumn.java#L157) and panel) are triggered when the 'x' is clicked. The 'x' click causes the panel to be removed from its parent container. The event then bubbles up to the panel, which is why at that point getParent() is null.

You could make it so the event is consumed when the panel is closed. That would solve the underlying problem.

`try {} catch (Exception e) {}` is a bit of a band-aid. It will mask any exception that comes up through that code path, so it's not the best way to fix this.

PR #536, opened 2015-03-24
A panel should gain focus after startup/project switching

### LINKED ISSUE #507: A panel should gain focus after startup/project switching
Currently, nothing has focus so I can't using the keyboard shortcuts unless I click on something first

rd1992:
Fixes #507

### DIFF
--- a/src/ui/UI.java
+++ b/src/ui/UI.java
@@ -135,7 +135,9 @@ public class UI extends Application implements EventDispatcher {
 				repoSelector.setDisable(false);
 				repoSelector.refreshComboBoxContents(ServiceManager.getInstance().getRepoId().generateId());
 				triggerEvent(new BoardSavedEvent());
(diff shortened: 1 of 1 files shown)


src/ui/UI.java
```diff
@@ -135,6 +135,7 @@ private void getUserCredentials() {
 				repoSelector.setDisable(false);
 				repoSelector.refreshComboBoxContents(ServiceManager.getInstance().getRepoId().generateId());
 				triggerEvent(new BoardSavedEvent());
+				getMenuControl().scrollTo(columns.getCurrentlySelectedColumn().get(), columns.getChildren().size());
```
dariusf: There may not be a selected column at this point. You should check `columns.getCurrentlySelectedColumn().isPresent()`, or assert it.

rd1992: alright.

PR #537, opened 2015-03-24
Save login credentials to avoid typing password everytime

### LINKED ISSUE #491: Save login credentials to avoid typing password everytime
The web ui doesn't require users to login every time. May be we can do something similar in HT?

rd1992:
Fixes #491 

There is no easy way of doing this. Have to instantiate browser before our mainstage.

### DIFF
--- a/src/storage/SessionConfiguration.java
+++ b/src/storage/SessionConfiguration.java
@@ -1,16 +1,25 @@
 package storage;
 
+import java.security.InvalidKeyException;
 import java.security.Key;
+import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import javax.crypto.BadPaddingException;
 import javax.crypto.Cipher;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.NoSuchPaddingException;
 import javax.crypto.spec.SecretKeySpec;
 
+import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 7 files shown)


src/ui/UI.java
```diff
@@ -198,7 +198,7 @@ private static void initialiseJNA(String windowTitle) {
 		return commandLineArgs;
 	}
 
-	private void quit() {
+	void quit() {
```
dariusf: quit can be made public, to be more explicit.

src/storage/SessionConfiguration.java
```diff
@@ -73,4 +79,42 @@ public String getLastLoginUsername() {
...
+		String result = "";
+		try {
+			String key = "HubTurboHubTurbo";
+		    Key aesKey = new SecretKeySpec(key.getBytes(), "AES");
+	        Cipher cipher = Cipher.getInstance("AES");
+	        cipher.init(Cipher.DECRYPT_MODE, aesKey);
+	        result = new String(cipher.doFinal(lastLoginEncrypted));
+	      }catch(Exception e) {
+	         e.printStackTrace();
+	      }
```
dariusf: Catching `Exception` is bad because it lets any other exception type slip through. It may be a bit verbose in this case but I feel being more explicit is better.

Also, we should log the error instead of only printing the stack trace.

damithc: New versions of Java allows having multiple exceptions in a single catch
clause: catch (Exception1 | Exception2){ ...}

dariusf:
Turned out to be a problem with the way we configured Gson. I've fixed that (static fields should never be serialised and now won't be) and made the logger static.

PR #549, opened 2015-03-25
making sure shortcuts work in filterbox

### LINKED ISSUE #548: Ctrl+X and Ctrl+D don't work
Supposed to maximize and resize, but don't seem to work. Ctrl+N works though. I'm thinking if we should have just one shortcut for this, to cycle between the three sizes. Ctrl+R for resize? May be we can reassign Ctrl+N as an alternative for 'new panel'?

rd1992:
Fixes #548

### DIFF
--- a/src/ui/issuepanel/IssuePanel.java
+++ b/src/ui/issuepanel/IssuePanel.java
@@ -33,7 +33,10 @@ public class IssuePanel extends IssueColumn {
 	private NavigableListView<TurboIssue> listView;
 	private final KeyCombination keyCombBoxToList = new KeyCodeCombination(KeyCode.DOWN, KeyCombination.CONTROL_DOWN);
 	private final KeyCombination keyCombListToBox = new KeyCodeCombination(KeyCode.UP, KeyCombination.CONTROL_DOWN);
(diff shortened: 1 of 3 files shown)


src/ui/issuepanel/IssuePanel.java
```diff
@@ -218,6 +230,15 @@ public void handle(KeyEvent event) {
...
 				}
+				if (maximizeWindow.match(event)) {
+					ui.maximizeWindow();
+				}
+				if (minimizeWindow.match(event)) {
+					ui.minimizeWindow();
+				}
+				if (defaultSizeWindow.match(event)) {
+					ui.setDefaultWidth();
+				}
```
dariusf: Could we factor these (and the hotkeys above) out into another function? That way the hotkeys for the panel and the filter box would always stay the same.

Also, using `if (...) {...} else if (...) {...}` seems better here, as it makes it clear that only one of these conditions can occur at a time.

dariusf:
I'm getting merge conflicts on this now. Guess the files have changed substantially in the meantime.

rd1992:
Yeah probably. I'll fix them soon

Issue #545, opened 2015-03-25
cannot navigate using f/d after closing a panel


damithc:
after closing the current panel using Ctrl+w, I expect to be able to navigate panels using f/d, but there is no easy way to go to the panel navigation mode.

rd1992:
this ties in with #489

PR #551, opened 2015-03-25
Ctrl + W updates index of selected column

### LINKED ISSUE #489: Ctrl+W sometimes doesn't work
Supposed to close the current panel. Sometimes it works, sometimes it doesn't.

rd1992:
Fixes #489 and #545

### DIFF
--- a/src/ui/issuecolumn/ColumnControl.java
+++ b/src/ui/issuecolumn/ColumnControl.java
@@ -229,8 +229,19 @@ public class ColumnControl extends HBox {
 	
 	public void closeCurrentColumn() {
 		if (currentlySelectedColumn.isPresent()) {
-			closeColumn(currentlySelectedColumn.get());
-			currentlySelectedColumn = (getChildren() == null || getChildren().size() == 0) ? Optional.empty(): Optional.of(0);
+			int columnIndex = currentlySelectedColumn.get();
+			System.out.println("Going to close column " + columnIndex);
+			closeColumn(columnIndex);
+			if (getChildren() == null || getChildren().size() == 0) {
+				currentlySelectedColumn = Optional.empty();
+			} else {
+				int newColumnIndex = (columnIndex > getChildren().size() - 1)
+									 ? columnIndex - 1
+									 : columnIndex;
+				System.out.println("jumping to column " + newColumnIndex);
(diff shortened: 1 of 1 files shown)


dariusf:
Hmm, should the selection jump to the closest panel instead? For example, when you close tabs in Chrome, selection jumps to the next tab, unless it's the last one in which case it goes to the previous.

rd1992:
@dariusf check it out now. See if this seems more to your liking

src/ui/issuecolumn/ColumnControl.java
```diff
@@ -229,8 +229,19 @@ public void setCurrentlyDraggedColumnIndex(int i) {
 	
 	public void closeCurrentColumn() {
 		if (currentlySelectedColumn.isPresent()) {
-			closeColumn(currentlySelectedColumn.get());
-			currentlySelectedColumn = Optional.empty();
+			int columnIndex = currentlySelectedColumn.get();
+			System.out.println("Going to close column " + columnIndex);
+			closeColumn(columnIndex);
+			if (getChildren() == null || getChildren().size() == 0) {
```
dariusf: I don't think `getChildren()` will ever be null. This can be safely omitted, or made into an assertion.

src/ui/issuecolumn/ColumnControl.java
```diff
@@ -229,8 +229,19 @@ public void setCurrentlyDraggedColumnIndex(int i) {
 	
 	public void closeCurrentColumn() {
 		if (currentlySelectedColumn.isPresent()) {
-			closeColumn(currentlySelectedColumn.get());
-			currentlySelectedColumn = Optional.empty();
+			int columnIndex = currentlySelectedColumn.get();
+			System.out.println("Going to close column " + columnIndex);
```
dariusf: We should use a logger instead. `System.out.println` has no guarantees of thread safety (and so messages may appear out of order in concurrent scenarios), and won't write to log files.

rd1992: ohh wait. i dint intend to put this in !! sorry

rd1992:
ok i checked in some debugging print statements. Removed the redundant code

dariusf:
No worries!

PR #568, opened 2015-03-29
Log files count does not stop at 5

### LINKED ISSUE #544: Log files count does not stop at 5
The number of log files keeps growing beyond 5

rd1992:
Fixes #544


rd1992:
Earlier the log limit was applied daily. So maximum number of logs created was 5 __daily__

Issue #574, opened 2015-03-30
'space accepts current suggestion' can result in unexpected results.


damithc:
e.g I was trying to type `set up files` but ended up with `set updated files` :-p This can become quite annoying.

rd1992:
This can go either way. To accept auto complete then the user will have to press another key to accept the suggestion before pressing SPACE or ENTER. Or the user presses Backspace to reject the filter and keep typing normally. Both ways involve the user pressing two keys.

damithc:
True. User has to type an extra key either to accept the suggestion or reject the suggestion. At the moment I prefer to require the extra key for accepting the suggestion. Reaon: When the user wants to type `up` and expand it to `updated`, the user is consciously using a short form and aware of the longer form. He can add the extra key without thinking much. Whereas typing `up` and having to type an extra key to avoid `updated` can be seen as an unexpected annoyance. We can't expect users to lookout for unexpected suggestions every time they type a word in the filter.

damithc:
No longer relevant as space does not accept the current suggestion any more.

PR #575, opened 2015-03-30
Double-space does not transfer focus from filter box to issue list

rd1992:
removed all the code handling space key from filterTextField. This is because it will be handled in #574


dariusf:
`resetLastKeyPressedCodeAndTime` is another stateful function which a maintainer has to know *exactly* when to call, so it isn't so good for code locality. Could I suggest moving the logic into `isDoublePress` as well?

```java
if(event.getCode() == KeyCode.SPACE){
    event.consume();
    if (KeyPress.isDoublePress(event.getCode())) {
        listView.selectFirstItem();
    }
} else {
    KeyPress.resetLastKeyPressedCodeAndTime();
}
```
becomes
```java
if (KeyPress.isDoublePress(KeyCode.SPACE, event.getCode())) {
    event.consume();
    listView.selectFirstItem();
}
```

rd1992:
Hmm. i want to consume the single space actually. This is because double space occurs at _three_ places -> column, filterbox and listview. Not consuming the space in the filterbox is what was causing the issue in the first place. If the space in the box is not consumed, the space press will pass to the parent (issuecolumn) and it will keep triggering the double space event there. Then it only works from issuecolumn to filterbox and not the other way around.
I do agree with the stateful function though. changing the key press code to move the logic there :)

dariusf:
Right, good point. Just tried it out and it works well.

PR #580, opened 2015-04-06
Make keyboard shortcuts work for the whole column

### LINKED ISSUE #539: Pressing L while bView doesn't have an issue loaded transfers focus to bView
e.g.
press gp (bView loads the pull requests page)
press L by mistake (nothing should happen, but it seems focus is transferred to bView)

### LINKED ISSUE #542: Make keyboard shortcuts work for the whole column
Currently only works when an issue is selected

rd1992:
Fixes #542, fixes #539

### DIFF
--- a/src/browserview/BrowserComponent.java
+++ b/src/browserview/BrowserComponent.java
@@ -528,10 +528,7 @@ public class BrowserComponent {
 	}
 
 	public boolean isCurrentUrlIssue() {
-		if (driver.getCurrentUrl().matches("https://github.com/(.+)/(.+)/(issues|pull)/([0-9]+)")) {
-			return true;
-		}
-		return false;
+		return GitHubURL.isUrlIssue(driver.getCurrentUrl());
 	}
 
 }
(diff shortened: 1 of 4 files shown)


src/browserview/BrowserComponent.java
```diff
@@ -526,4 +526,12 @@ public void showContributors() {
...
 		});
 	}
+
+	public boolean isCurrentUrlIssue() {
+		if (driver.getCurrentUrl().matches("https://github.com/(.+)/(.+)/(issues|pull)/([0-9]+)")) {
+			return true;
+		}
+		return false;
+	}
+
```
dariusf: This method should go in `GitHubURL`.

Also, minor problem, but we should use `([^/]+)` instead of `(.+)`. That would reject urls like `https://github.com/////issues/1`.

rd1992: Hmm. good catch on the url. However i still need to check for url via browsercomponent as i need to know what the url is. Check out the latest commit :)

dariusf:
Yes, that's fine.

PR #589, opened 2015-04-09
The panels that is selected at startup is not highlighted

### LINKED ISSUE #521: The panels that is selected at startup is not highlighted
The new highlighting feature does not seem to apply for the panel that is selected at startup or after a project switch

### LINKED ISSUE #558: Mid view auto scroll does not work after project switch

### LINKED ISSUE #563: panel navigation keys should work right after start up
It is typical that the user wants to navigate panels right after startup.

### LINKED ISSUE #572: Parent panel does not get highlighted when using ESC to move focus from filter to parent panel
The focus moves to the parent panel correctly, but the UI does not highlight the panel.

rd1992:
Fixes #521, fixes #558, fixes #563, fixes #572

### DIFF
--- a/src/ui/UI.java
+++ b/src/ui/UI.java
@@ -137,6 +137,7 @@ public class UI extends Application implements EventDispatcher {
 				triggerEvent(new BoardSavedEvent());
 				if(columns.getCurrentlySelectedColumn().isPresent()) {
 					getMenuControl().scrollTo(columns.getCurrentlySelectedColumn().get(), columns.getChildren().size());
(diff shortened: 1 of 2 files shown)


dariusf:
Looks good. Since both parts are the same, could we factor them out into a method? They won't get out of sync that way.

rd1992:
yeah i was gonna leave a comment. I wanted to fix #572 also. Will refactor too :)

rd1992:
@dariusf done.

PR #595, opened 2015-05-13
Architectural changes

dariusf:
**Goals**

1. Multiple projects loaded and updating concurrently
2. Multiple projects per panel
3. Testability
4. Swappable back end: easy to stub or change entirely

**So far:**

- **Structural changes**: clean separation of components. Caching is invisible to application logic. Both storage and repository back ends are accessed through a unified interface, and can be swapped out for stubs/alternative back ends with only local changes.
- **IO handled via a work queue**: user actions put tasks on the queue, and tasks notify the UI when they are done. This is to allow multiple repositories to download/update in parallel without blocking the UI.
- **Immutable data**: things are immutable as much as possible. Resources are value types, so we don't have to maintain pointers to other objects.

Currently issues can be downloaded and updated concurrently. Nothing is rendered yet. I'll be moving more existing functionality over to the new framework.


dariusf:
The initial transition to the new architecture is almost done. It's not working yet, but the pieces are all in place and just need to be wired together.

I came up against a few issues that we should probably work out:

- How to we treat multiple repositories? What we discussed was the notion of a **primary repo**, which is the one you input on login. The rest are opened as needed when a filter is qualified with a repo name. Is there still a need for the repo-switching combobox at the top left, or some other UI to switch primary repo/load others without using a filter?
- Filtering multiple repos. I'm thinking of syntax like `repo:owner/name` for specifying which repo's issues to show. If the owner is left out (`repo:name`) your login name will be used for it. If the entire thing is left out the primary repo will be used.
    + If multiple repos are open, what is displayed in the issue list? Everything from all?
- How do we save preferences? In the past each repo had its own set of preferences which would apply when it was opened.
    + We could save them based on the primary repo, but doing that makes operations involving multiple repos asymmetric: opening repo A, then B would have different preferences compared to B, then A.
    + Maybe we shouldn't have a primary repo, and just one view (with many boards). In this case what would an unqualified filter show?
- What should the progress UI for loading and updating many repos at once look like? One progress bar is no longer sufficient to convey everything that is happening.

damithc:
May be **default repo** is a better name than primary repo?
Yes, that syntax is fine. Note that in some cases it is `org/repo`, note `owner/repo`.
A panel will show issues that fit the filter. If no project is specified in the filter, issues from the default project are shown. 
I think we should still save preferences for each repo. When we deal with a specific issue, we use the preferences for the repo that owns the issue. 
One progress bar for each repo?

dariusf:
Here are the preferences we save right now:

- Username and password
- Previously-opened repos
- Boards (used to be per-repo, should we share these across all repos now? Considering that asymmetric behaviour is a bit unintuitive)
- Previously-open panels (previously per-repo, is there now just one set of panels?)

Do all progress bars go in the status bar? What I'm thinking of is a bar for each one that looks like [this](http://stackoverflow.com/a/14061347).

dariusf:
Another question: when a `repo` qualifier is omitted from the filter, should we filter against the default repo, or all repos? In the former case, we wouldn't have a way to filter across all repos. Do we need extra syntax for that maybe? (`repo:all` is ambiguous though)

dariusf:
This is finally ready. There are still a few nonessential features missing (among others):

- Feed isn't working yet
- No progress UI
- No more drag-and-drop

but it's almost back to the old level of usability. The missing features can be re-added gradually. I'm merging to master and making a release so we can try it.

PR #600, opened 2015-05-15
Bring browser window to the front (mac osx)

### LINKED ISSUE #442: Bring browser window to the front and keep it in focus (mac osx)
See issue [https://github.com/HubTurbo/HubTurbo/issues/240]

lejolly:
Partially fixes #442 

I'm only able to bring HT's chrome window to the front by using AppleScript to check and bring up the chrome window that matches the page title
Assumes that HT's chrome driver only has one tab and window, and is in the same space
Works with multiple instances of chrome (even if they are in the same space)

Original script is below:
```applescript
tell application "System Events" to repeat with p in (every application process whose name contains "Chrome" and name does not contain "Helper")
    if (title of window of p as string) contains "{insert page name here}" then tell process p to perform action "AXRaise" of window 1 of p
end repeat
```

### DIFF
--- a/src/browserview/BrowserComponent.java
+++ b/src/browserview/BrowserComponent.java
@@ -12,6 +12,10 @@ import java.util.concurrent.Executors;
 
 import javafx.concurrent.Task;
 
+import javax.script.ScriptEngine;
+import javax.script.ScriptEngineManager;
+import javax.script.ScriptException;
+
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringEscapeUtils;
 import org.apache.logging.log4j.LogManager;
@@ -62,6 +66,7 @@ public class BrowserComponent {
 	private static final int SWP_NOMOVE = 0x0002;
 	private static HWND browserWindowHandle;
 	private static User32 user32;
+	private static ScriptEngine scriptEngine;
 	
 	static {
 		setupJNA();
@@ -97,7 +102,7 @@ public class BrowserComponent {
 		driver = createChromeDriver();
 		logger.info("Successfully initialised browser component and ChromeDriver");
 	}
-	
+
 	/**
 	 * Called when application quits. Guaranteed to only happen once.
 	 */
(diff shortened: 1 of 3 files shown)


src/browserview/BrowserComponent.java
```diff
@@ -418,6 +427,18 @@ private void bringToTop(){
...
+
+	public void bringToFront() {
+		if (PlatformSpecific.isOnMac()) {
+			try {
+				scriptEngine.eval("tell application \"System Events\" to repeat with p in (every application process whose name contains \"Chrome\" and name does not contain \"Helper\")\nif (title of window of p as string) contains \"" + driver.getTitle() + "\" then tell process p to perform action \"AXRaise\" of window 1 of p\nend repeat");
+			} catch (ScriptException e) {
+				logger.info("Bring browser window to front script exception! " + e.getLocalizedMessage());
+			} catch (Exception e) {
+				logger.info("Bring browser window to front exception! " + e.getLocalizedMessage());
+			}
```
dariusf: I'm not sure catching `Exception` is needed here. The only other exception ScriptEngine is documented as throwing is NullPointerException, which we generally don't want to catch as it's indicative of programmer error.

dariusf:
I get an NPE when running this (on Yosemite): `scriptEngine` is null in `bringToFront`. Maybe the interface to AppleScript has changed?

lejolly:
ah yes, i've implemented the workaround found here: https://netbeans.org/bugzilla/show_bug.cgi?id=242171

src/browserview/BrowserComponent.java
lejolly: it's more for debugging purposes (because i think applescript is notoriously flaky), should i remove them then?

dariusf:
I get this now:

```
205795 18 May 2015 14:21:59 INFO BrowserComponent [JavaFX Application Thread] - Bring browser window to front script exception! Can’t make «class titl» of window of item 2 of every «class pcap» of application "System Events" whose name contains "Chrome" and not 64 into type string.
```
I tried it 5 times or so, and it worked once, but then it brought up my other Chrome window.

src/browserview/BrowserComponent.java
dariusf: Yes, I think it's safe to remove. We're dealing with AppleScript through the interface provided by ScriptEngine, and we're already handling ScriptException. Other errors coming up through that code path will likely be our own (or a JDK bug).

lejolly:
moved changes over to the main repo, so this pull request is now invalid

PR #603, opened 2015-05-20
UI Optimisations

lejolly:
Trying to make the UI faster/smoother. 

First fix seems to help with the scrolling performance issue by having the user avatars load in the background. I assume before the fix, they were being loaded on the UI thread and thus affected the scrolling. 

More fixes to come if I can find/solve them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ chromedriver_linux
 chromedriver.exe
 .hubturbocache/
 
+.idea
+HubTurbo.iml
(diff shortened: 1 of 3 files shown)


lejolly:
Next possible area of optimisation: `IssuePanelCell` which is a custom `ListCell` implementation

Reason: `updateItem` for each cell that is displayed is called basically every time the list changes in any way, e.g. scrolling, even if the data within each cell stays the same. This affects performance because `updateItem` contains the expensive `setGraphic` operation.

PR #668, opened 2015-06-04
Checkstyle integration

dariusf:
Merged #667, but it seems that warnings also cause checkstyle to fail, stopping the whole CI build. Reverted the merge.

We have a few options (not mutually exclusive):

- `./gradlew check --continue`, so the build doesn't just halt when checkstyle fails
- Fix more violations
- Relax checkstyle rules for things which can't be fixed easily (javadocs on every method)


dariusf:
Changes to Google style guide:

 - Removed requirements for Google copyright notice, named TODOs
 - Allowed such `__METHOD_NAMES__`
 - Changed import order requirements to match Eclipse's default
 - Removed requirement for `if` and `else` blocks to have braces. This is useful for guard clauses and automatically-generated equals methods.
 - Increased max line length to 120.
 - Removed javadoc checks. There doesn't seem to be a way to have optional javadocs, yet have those which are present be checked.

PR #676, opened 2015-06-08
90% Coverage for tests

lejolly:


damithc:
Try to minimize commented out code.

lejolly:
@damithc so do we just remove the code?

damithc:
Yes. If you think the code may be need later, you can do it as a separate commit and write a good commit message so that we can easily locate it later.

lejolly:
ok, got it.

PR #687, opened 2015-06-10
Selection should not change when switching back from bView to pView #684

### LINKED ISSUE #684: Selection should not change when switching back from bView to pView
Currently, the selection seems to be based on index rather than a unique identifier. For example, if the 3rd issue was selected originally, the 3rd issue will be selected after switching to pView even if the originally selected issue is no longer the 3rd issue.

lejolly:
Fixes #684


lejolly:
@damithc could you check the branch out and see if it fixes the problem

damithc:
Is there anyway to write a test for this?

lejolly:
We could try to expose the currently selected index, change the last updated time of an item, force the ui to refresh, and see if the index changes (it should assuming the current filter allows for it). 






On Wed, Jun 10, 2015 at 3:29 AM -0700, "Damith C. Rajapakse" <notifications@github.com> wrote:












Is there anyway to write a test for this?

damithc:
Yes, let's try that. Preferably, write the test in a separate branch, make sure it fails, and then merge it to this branch and see if it passes.

lejolly:
hmm, having some trouble trying to reproduce the bug in test mode

damithc:
Can repro manually?

damithc:
I used the filter `updated:5 sort:updated` to repro manually: select the top issue (there should be more than 1 issue in the panel), change to bView, do some change to the selected issue (e.g add comment/label), switch back to pView.

lejolly:
yes i can reproduce it when using it with a real repo

lejolly:
test now passes with the fix (and fails without the fix), works on travis too.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -113,6 +114,15 @@ public void restoreSelection() {
         }
     }
 
+    // sample generic findIndex, most of the time using the equals method
+    // will not give the desired behaviour because an item's properties might
+    // change but the item is still considered to be the same
+//    boolean areItemsEqual(T item1, T item2) {
+//        return item1.equals(item2);
+//    }
```
damithc: Clean up needed here.

Issue #683, opened 2015-06-10
Issues should be sorted by latest updated time first in updated view


acruis:
Currently, if no sorting order is specified, issues are sorted by descending ID (i.e. latest creation time first).

This is true even when the `updated:` filter is used. Intuitively, when the `updated:` filter is used, issues that are updated the latest should be shown on top (unless another sorting order is explicitly specified through the `sort:` filter).

damithc:
Agreed.

PR #690, opened 2015-06-12
Ignore user's own updates when using the updated filter #679

### LINKED ISSUE #679: Ignore user's own updates when using the updated filter
May need to be broken down into sub-issues. 

Scenario:
1. user types in `updated:XX sort:date` in filter box
2. user clicks on first issue in panel
3. user comments on said issue

4a. when user focuses back onto HT, currently it detects a change in the page and refreshes the repo/list, and since it was updated, the issue jumps to the bottom, issue selection in the list also changes
4b. what should happen is that when going back to HT, it should still detect a change in the page and refresh, but should not change the order of the list by ignoring the latest update done by the user

acruis:
Fixes #679

User's own updates (i.e. events) will not be considered when sorting issues by date in conjunction with the `updated:` filter.

More code to be added to ensure that the comment icon does not light up for own comments.

### DIFF
--- a/src/main/java/filter/expression/Qualifier.java
+++ b/src/main/java/filter/expression/Qualifier.java
@@ -366,13 +366,13 @@ public class Qualifier implements FilterExpression {
         }
     }
 
-    public Comparator<TurboIssue> getCompoundSortComparator(IModel model) {
+    public Comparator<TurboIssue> getCompoundSortComparator(IModel model, boolean metadataRefresh) {
         if (sortKeys.isEmpty()) {
             return (a, b) -> 0;
         }
         return (a, b) -> {
             for (SortKey key : sortKeys) {
-                Comparator<TurboIssue> comparator = getSortComparator(model, key.key, key.inverted);
+                Comparator<TurboIssue> comparator = getSortComparator(model, key.key, key.inverted, metadataRefresh);
                 int result = comparator.compare(a, b);
                 if (result != 0) {
                     return result;
@@ -382,7 +382,8 @@ public class Qualifier implements FilterExpression {
         };
     }
 
-    public static Comparator<TurboIssue> getSortComparator(IModel model, String key, boolean inverted) {
+    public static Comparator<TurboIssue> getSortComparator(IModel model,
+                                                           String key, boolean inverted, boolean metadataRefresh) {
         Comparator<TurboIssue> comparator = (a, b) -> 0;
 
         boolean isLabelGroup = false;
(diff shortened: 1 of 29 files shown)


src/main/java/filter/expression/Qualifier.java
```diff
@@ -398,6 +399,15 @@ private static boolean isMetaQualifier(Qualifier q) {
             case "date":
                 comparator = (a, b) -> a.getUpdatedAt().compareTo(b.getUpdatedAt());
                 break;
+            case "nonselfupdate":
```
damithc: Use camel case? Easier to read that way.

src/main/java/ui/issuepanel/IssuePanel.java
```diff
@@ -36,7 +37,11 @@
         new KeyCodeCombination(KeyCode.N, KeyCombination.CONTROL_DOWN);
     private final KeyCombination defaultSizeWindow =
         new KeyCodeCombination(KeyCode.D, KeyCombination.CONTROL_DOWN);
+
+    // At the moment, apparently, whenever the screen refreshes, all previously lit up comments will
+    // lose the highlight.
```
damithc: Is this PR supposed to fix that problem or is this just an observation? In other words, what's the purpose of this comment?

src/main/java/filter/expression/Qualifier.java
```diff
@@ -398,6 +399,15 @@ private static boolean isMetaQualifier(Qualifier q) {
...
+                // Purposefully inverted a and b to make latest issues show up first.
+                if (metadataRefresh) {
+                    comparator = (a, b) ->
+                        b.getMetadata().getNonSelfUpdatedAt().compareTo(a.getMetadata().getNonSelfUpdatedAt());
+                } else {
+                    comparator = (a, b) -> b.getUpdatedAt().compareTo(a.getUpdatedAt());
+                }
+                break;
```
dariusf: Adding a case here has the side effect of making `sort:nonselfupdate` valid syntax, which is probably not what we want.

Since there is only one instance of "nonselfupdate" as a key and `metadataRefresh` is constant there, maybe you could create another method to handle it? `getUpdateComparator`?

Also, since the comparator brought on by `updated` can never compose with another `sort` comparator, there's no need to modify `getCompoundSortComparator`.

src/main/java/ui/issuepanel/IssuePanel.java
acruis: It is an observation, or rather what I believe is the current behaviour of the program. I was actually going to confirm whether this was actually the current behaviour. If yes, then we can make an issue to fix it.

src/main/java/filter/expression/Qualifier.java
acruis: Okay.

src/main/java/filter/expression/Qualifier.java
acruis: We can compose sort comparators currently, right? The current case is that if there is an explicitly specified sort comparator, it will override `nonSelfUpdate`, which might not be what we want, in which case the user should be given the control to maintain the `nonSelfUpdate` order alongside other sort comparators, too. For example, sorting by repo, and then by `nonSelfUpdate`.

But if we allow the `nonSelfUpdate` order to be lost when another sort order is explicitly specified then extracting it to another method without touching `getCompoundSortComparator` does make sense.

dariusf: Oh, so it was intentional to add it as syntax? My bad then.

Do add tests for it. :p

src/main/java/filter/expression/Qualifier.java
```diff
@@ -480,17 +491,50 @@ private boolean idSatisfies(TurboIssue issue) {
...
+
+        if (!issue.getCreator().equalsIgnoreCase(currentUser)) {
+            // But on the second time being filtered, we consider only if creator is different
+            if (updatedRange.encloses(hoursSinceCreation)) return true;
+        }
+
+        // Creation time is also outside of range, we do not need to show this issue at all.
+        return false;
+
+//        if (numberRange.isPresent()) {
```
lejolly: remove?

acruis: Why?

lejolly:
ongoing until status bar metadata info is fixed.

lejolly:
@acruis when i go from updated:X to updated:(a number less than X), the filter seems to stop working

acruis:
Fixes #683 too.

lejolly:
seems good to go.

PR #694, opened 2015-06-15
Clicking on the selected issue should not reload the bView if the issue is already loaded #686

### LINKED ISSUE #686: Clicking on the selected issue should not reload the bView if the issue is already loaded
A user may click on the selected issue as a means of switching the focus back to the pView. This should not reload the issue in the bView if the bView is already showing that issue.

lejolly:
Fixes #686


acruis:
Looks good.

PR #697, opened 2015-06-15
Logout doesn't work #624

### LINKED ISSUE #624: Logout doesn't work
Logout quits the app, after reopening, it still automatically logs me in

HyungJon:
Fixes #624 

Fixed by invoking setLastLoginCredentials method in Preferences class to set saved credentials to empty strings. Re-running the jar after logging out will no longer cause automatic login.

I attempted to add another code to automatically bring back the login screen after logging out, but I think it can't be done without changing the accessibility of one or more of the methods in UI class, which I was wary of doing.

### DIFF
--- a/src/main/java/ui/MenuControl.java
+++ b/src/main/java/ui/MenuControl.java
@@ -68,8 +68,8 @@ public class MenuControl extends MenuBar {
     }
 
     private Menu createPreferencesMenu() {
-    	Menu preferences = new Menu("Preferences");
-    	
+        Menu preferences = new Menu("Preferences");
+        
         MenuItem logout = new MenuItem("Logout");
         logout.setOnAction(e -> {
             logger.info("Logging out of HT");
@@ -79,8 +79,8 @@ public class MenuControl extends MenuBar {
         
         MenuItem quit = new MenuItem("Quit");
         quit.setOnAction(e -> {
-        	logger.info("Quitting HT");
-        	ui.quit();
+            logger.info("Quitting HT");
+            ui.quit();
         });
         
         preferences.getItems().addAll(logout, quit);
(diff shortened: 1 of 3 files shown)


HyungJon:
Update: one possible way to make this work seems to be calling the start method in UI class, which will invoke the getUserCredentials method and bring up the login screen. However I don't think MenuControl will have access to the same Stage parameter required.

damithc:
@lejolly or @acruis can one you assign yourself to this PR and mentor Jon through this issue?

lejolly:
@HyungJon for now we'll just re-enable the old logout functionality, but you are more than welcome to open a new issue to try to show the login screen after the logout (probably won't be that easy to implement though).

lejolly:
@HyungJon the travis build is failing, due to `guitests.UseGlobalConfigsTest` failing. the test relies on the logout function not wiping the credentials which after your fix it now does. what i recommend you do is to create a new `quit` menu button that only calls `UI.quit()` and use that instead for the test.

HyungJon:
Made those edits and resolved an issue I had. It seems the addition of an untested part has caused a test coverage problem though. I will try to write a test for logout too.

lejolly:
@HyungJon you can only launch the app once per test class

HyungJon:
I have been trying to write a test for logout, but I'm having difficulty because both quit and logout terminate the application and I didn't manage to find a way to launch the app again somehow.

lejolly:
@HyungJon see previous comment, so you can always create a new test class to just test for logout. also, please make sure the test passes locally before pushing it (to the ci) because the ci adds it's own set of problems (e.g. random x11 errors, etc).

HyungJon:
I guess that's the only way. I was reluctant to do it because it makes the scope of the test classes a bit messy, but no helping it.

Tests passed locally, but I'm still getting a X11 error. Looking to fix it.

acruis:
Just rerun the travis build if you get an X11 error =)

acruis:
You can change the indentation settings of Eclipse/IntelliJ to 4 spaces instead of tab so that whenever you press <kbd>TAB</kbd> the IDE will insert 4 spaces instead of the tab character.

PR #701, opened 2015-06-16
Specify keyboard shortcuts in a different file #598

### LINKED ISSUE #598: Specify keyboard shortcuts in a different file
Currently keyboard shortcuts are hardcoded in various files.

lejolly:
Fixes #598 by creating a central class `KeyboardShortcuts` to specify all keyboard shortcuts throughout HT. Also implements saving and loading of certain keyboard shortcuts from `GlobalConfig`.


lejolly:
@damithc currently only the keyboard shortcuts for mark as read/unread are customisable, which keyboard shortcuts should we allow users to specify?

src/main/java/ui/components/KeyboardShortcuts.java
```diff
@@ -0,0 +1,116 @@
...
+ * Classes that currently have keyboard shortcut code:
+ * ui.components.NavigableListView
+ * ui.issuecolumn.ColumnControl
+ * ui.issuepanel.IssuePanel
+ * ui.MenuControl
+ *
+ * Utility Class:
+ * util.KeyPress
+ */
+public class KeyboardShortcuts {
```
damithc: Can this handle cases where more than one key does the same thing? i.e. there are multiple shortcuts to achieve the same task

lejolly: currently no, what should happen if a user decides to use the same key for more than one action? should we reset all the shortcuts to the default, just ignore the actions with the same key or reset those particular actions to the default keys?

damithc: I meant cases like both E and R can be used to mark an issue as read. I guess we should not allow users to use the same shortcut for multiple things. In such cases, we go by the order in which they are specified and take the one that comes later in the order. i.e. every time a shortcut is specified, we override the one specified previously.

lejolly: For example, if i map the key <kbd>L</kbd> to both navigate to the left panel and open a new issue, and the new issue mapping occurs later in the json, so it will take on the user defined shortcut key. my question is, what happens to the key mapping of navigating to the left panel? do we use the default because this will have its own problems too, like if the default key is mapped to some other action.

lejolly:
default keyboard shortcut for scroll to top has been changed to <kbd>i</kbd> to avoid overlapping with mark as unread.

lejolly:
conflict resolution: based on the order of `getKeyboardShortcutsFromHashMap()` in `KeyboardShortcuts`, the first shortcut will get it's shortcut set first, if it's missing or invalid then the default will be used. if the key has already been assigned to something else then it will use the default too.

damithc:
I prefer to keep U as 'scroll to top' and use something else for 'Unread'. That's because the former is based on key location while the latter is rarely used and not based on key location. We can even use `e` as a toggle for both read and unread.

damithc:
How about we just don't allow conflicts at all? If there is a conflict, exit after an explanatory error message.

lejolly:
my rationale was <kbd>u</kbd> for unread makes the most sense (without having a toggle key), while <kbd>i</kbd> is more comfortable to reach with the right hand especially when switching back and forth between <kbd>i</kbd> and <kbd>n</kbd> for scroll to top/bottom. exiting upon detecting a conflict also seems a bit harsh/jarring, maybe a dialog box popping up when a conflict is detected might be a better way to inform the user, while letting the user continue on using the default keys (for the conflicting shortcuts)?

damithc:
Ok, can keep i for scroll to top.
I suggested exiting because the current way of falling back on defaults
seem complicated as each fallback can result in more fallbacks. Yes, it is
better to allow user to continue if can be done without too much
complications.

lejolly:
when an invalid number of shortcut keys are specified (either more of less, when it is zero, the defaults are used)
warning

when an invalid key is specified or that key has already been used by some other shortcut
warning

when a user defined shortcut could not be found
warning

all this takes place at the start of the launch, even before the login dialog

acruis:
Looks good to go!

lejolly:
let's go!

acruis:
Okay, can merge.

PR #702, opened 2015-06-16
 Re-enable or remove Force Refresh #670 

### LINKED ISSUE #670: Re-enable or remove Force Refresh
It was added as a stop-gap measure in the past for users to recover from cache corruption. Ideally we shouldn't need it any more.

HyungJon:
Fixes #670 
Removed the menu item and trigger for force refresh, as well as a reference to said menu item on the menu. I believe these are all the codes related to force refresh.


lejolly:
@HyungJon we're trying to limit to one `difficulty.beginner` issue per person just so that everyone can get a feel of the workflow to submit a pull request, etc. so i'll close this for now, but you're welcome to finish up your other pull requests and after that to work on harder issues.

PR #704, opened 2015-06-17
When using the updated filter, every IssueSelectedEvent re-downloads the metadata #703

### LINKED ISSUE #703: When using the updated filter, every IssueSelectedEvent re-downloads the metadata
Leading to high api usage.

acruis:
Fixes #703

### DIFF
--- a/src/main/java/filter/expression/Qualifier.java
+++ b/src/main/java/filter/expression/Qualifier.java
@@ -366,13 +366,13 @@ public class Qualifier implements FilterExpression {
         }
     }
 
-    public Comparator<TurboIssue> getCompoundSortComparator(IModel model, boolean metadataRefresh) {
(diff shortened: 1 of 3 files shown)


src/main/java/filter/expression/Qualifier.java
```diff
@@ -366,13 +366,13 @@ private static boolean isMetaQualifier(Qualifier q) {
         }
     }
 
-    public Comparator<TurboIssue> getCompoundSortComparator(IModel model, boolean metadataRefresh) {
+    public Comparator<TurboIssue> getCompoundSortComparator(IModel model, boolean nonSelfSortable) {
```
damithc: booleans should be named as such e.g. `isNotSelfSortable`

lejolly: ok i'll create an issue to fix it

PR #710, opened 2015-06-18
Marking issues as read and unread re-downloads metadata #709

### LINKED ISSUE #709: Marking issues as read and unread re-downloads metadata
Related to #703

acruis:
Fixes #709


damithc:
If we have good test coverage, a change to functional code should have a
corresponding change in test code :-)

-- 
Sent from a mobile device.

PR #716, opened 2015-06-19
updated guitests to use push #715

### LINKED ISSUE #715: update guitests to use push
and to take into account new mark as read keyboard shortcut

lejolly:
Fixes #715


acruis:
Looks good to me.

PR #719, opened 2015-06-19
Refactoring of the GUI logic #685

### LINKED ISSUE #685: Refactoring of the GUI logic
Currently there is a lot of logic in GUI, especially when the issue panels are repopulated as part of an update. The flow of execution can be unclear at times due to code in multiple classes in GUI being carried out as a result of repopulation of data, especially when events are triggered to carry data between the Logic component to said GUI classes.

This logic can be abstracted out into a "GUI Controller", so to say, which manages the flow of execution and serves relevant data to each GUI component.

acruis:
Fixes #685

This PR keeps track of the refactoring done so far on the GUI logic.


src/main/java/ui/GUIController.java
```diff
@@ -0,0 +1,297 @@
...
+    /**
+     * The handler method for a ModelUpdatedEvent. It extracts the issues of the multimodel carried by the event,
+     * and then filters and sorts the issues to place each panel in ColumnControl based on these issues.
+     *
+     * The filtering process also produces a list of issues to request metadata for, if e.hasMetadata is false
+     * and their respective panels specify the display of metadata (through the UPDATED filter).
+     *
+     * Based on the hasMetadata state of the model, as specified by the event, the issues will then be displayed
+     * on their respective panels, or held back until their metadata requests have been fired and the downloaded
+     * metadata come back as a subsequent ModelUpdatedEvent (with e.hasMetadata being true).
```
lejolly: Do you want to immediately show the matching issues and then in the event timeline just put a placeholder message `downloading metadata` or something and then refresh again when you get the metadata? In this way it's obvious to the user that they should wait for the metadata to come in.

acruis: That works too, but then issues will jump around as the sorting order can be different before and after receiving metadata.

lejolly: I think that's fine, at least the user isn't held up. When the metadata comes in and the list is refreshed with the events, the user might not even realise the change in order =P

acruis: I don't think the user is held up. If you mean that the GUI might be frozen, the metadata retrieval is carried out in a different thread so that's OK.

Prior to getting metadata, we're not fully filtering the issues by non-self updates, which means some issues might show up even if they are not of relevance to the user (when the only updates in the specified time window are own updates). Plus, issues can move around quite frequently, as each metadata retrieval is triggered whenever Logic.refresh is called.

I like the `downloading metadata` idea, though! It's always good to give more visual feedback to the user.

lejolly: Good point, my focus is to present something usable to the user asap? if issues disappear/get rearranged i guess that's fine too? Maybe we can expand the placeholder message to `downloading metadata, list subject to change`?

acruis: I agree that we need to present something usable to the user, though if the user specified the `updated` filter, my guess is that they will want all the relevant data before deciding to work on it, which means the user might just wait until the `list subject to change` is gone.

On the other hand, the panel is only blank when we are starting the program and downloading metadata before showing the user the issues. Subsequent refreshes won't make the issues already in the panel to disappear until metadata has arrived, so the user does have something to look at.

lejolly: Say you open a new panel and it lists all the issues there, after entering the 'updated:24' filter, it would be nice to see a shorter list first instead of having the scroll through X number of issues?

acruis: I think the problem here is that `updated` can be interpreted either as a filter (which should happen quickly) or a functional keyword (which triggers a metadata update, for which it is acceptable to tell the user to wait).

If `updated` is interpreted as a filter, then it makes more sense to give the user immediate visual feedback, but if it's interpreted as a command ("download metadata for me") then as a user, I think it's OK to wait for a while.

Though, since it's a filter box, it's better to interpret it primarily as a filter keyword. I'll make a separate issue for this.

lejolly:
@dariusf any issues with this?

lejolly:
@acruis comment when you're done with removing whatever you wanted to remove

dariusf:
- Naming: `Column` is a bit of a relic from when we didn't know what to call them. Could we standardise `Panel`? Then:

Column -> AbstractPanel
IssueColumn -> FilterPanel (provides filtering of issues)
IssuePanel -> ListPanel (displays issues as a list)

Names are off the top of my head, could be better. Point is we should take this chance to improve names.

- Concurrency: `modelUpdated` should be wrapped in `Platform.runLater`, since it modifies shared state from an arbitrary thread. I didn't check every single method, but since this is a *GUI* controller, all of it can be confined to the UI thread. Things would be a lot simpler if we made sure that was the case.

Nice work overall, this does make things simpler.

acruis:
Thanks! I was actually thinking of subclassing IssueColumn to put the filter storage functionality in there, but I realized there was no more use to IssueColumn itself in that case. Simply renaming it to FilterPanel does make a lot more sense.

About Platform.runLater: yup, I'll do that.

acruis:
Wrapping `modelUpdated` in `Platform.runLater` causes an NPE upon startup of tests (but only upon startup of tests and not when running the actual UI). Might be because we're bypassing the login dialog during most tests. I'll wait until #738 is done and then work with `Platform.runLater` later.

It's good that we've tried though, there might be a hidden race condition here, too.

lejolly:
ok looks good.

PR #726, opened 2015-06-23
Create assertions in KeyboardShortcutsTest and MenuControlTest #723

### LINKED ISSUE #723: Create assertions in KeyboardShortcutsTest and MenuControlTest

lejolly:
Fixes #723

### DIFF
--- a/src/main/java/ui/UI.java
+++ b/src/main/java/ui/UI.java
@@ -33,6 +33,7 @@ import util.Utility;
 import util.events.*;
 import util.events.Event;
 import util.events.testevents.UILogicRefreshEventHandler;
+import util.events.testevents.WindowResizeEvent;
 
 import java.awt.*;
 import java.util.HashMap;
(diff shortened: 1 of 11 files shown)


acruis:
OK, looks good to go.

src/main/java/ui/UI.java
```diff
@@ -431,6 +432,9 @@ public MenuControl getMenuControl() {
     }
 
     public void setDefaultWidth() {
+        if (isTestMode()) {
+            triggerEvent(new WindowResizeEvent(WindowResizeEvent.EventType.DEFAULT_SIZE_WINDOW));
+        }
```
damithc: let's try to minimize this kind of code. It reduces the credibility of tests when the code behaves differently in test mode.

damithc:
Let's call these 'tests' not 'assertions', to avoid confusing with java `assert`s

PR #727, opened 2015-06-23
auto-complete is getting in the way of editing filters #651

### LINKED ISSUE #651: auto-complete is getting in the way of editing filters
When I'm editing an existing filter (e.g fix a typo), the auto-complete jumps in and inserts stuff that I don't need. Perhaps hold back auto-complete if there is a letter on the right?
I also noticed the cursor jumping back to the beginning of the filter at unexpected moments.

lejolly:
Fixes #651

Original PR by @dariusf


acruis:
OK, looks good to me.

PR #730, opened 2015-06-24
Hide login dialog using auto-login #560

### LINKED ISSUE #560: Hide login dialog using auto-login
We don't need to show that dialog during auto-login right?

lejolly:
Fixes #560


lejolly:
YOU SHALL ~~NOT~~ PASS

src/main/java/ui/LoginDialog.java
```diff
@@ -219,6 +225,9 @@ private void populateSavedFields() {
             passwordField.setText(lastLoginPassword);
         }
 
+        loadCredentialsFromFile();
+        resolveCredentials();
```
acruis: Since `loadCredentialsFromFile` already calls `resolveCredentials`, is there a need to call `resolveCredentials` here?

lejolly: good catch, will fix it.

acruis:
OK, good to go.

PR #731, opened 2015-06-24
 unexpected default project at start up #665 

### LINKED ISSUE #665: unexpected default project at start up
At times, the default project at start up doesn't seem to be the one I had selected when I shut down HT the previous time.

HyungJon:
Fixes #665 

Changes:
1. Repository is properly added to list when selected from drop-down list now
2. Opening repository from combo box and column filter now invoke separate methods to avoid conflicting conditions.


HyungJon:
Newly emerged issue: opening a repository through the filter results in HubTurbo repeatedly opening the repo (essentially refreshing the column) in very short intervals. Please do not merge until this is resolved.

HyungJon:
Bugs fixed.

src/main/java/backend/Logic.java
```diff
@@ -117,6 +140,26 @@ public void refresh() {
             }
         });
     }
+    
+    // Dummy method for refactoring above two methods. Currently not used yet.
```
damithc: Get rid of dead code. You can recover it later from Git later.

damithc:
Normally, a change to functional code should come with changes to test code. If a change to functional code does not break an existing test, that means the test coverage is weak for that functionality. Go ahead and add a test in that case.

HyungJon:
Understood. Adding test cases for default repository settings.

HyungJon:
Since UseGlobalConfigsTest already contains checks for last viewed repos I added a code in it for testing repo selection from combo box. It seemed impossible to make the cursor automatically select a repo so I just implemented it by making the cursor move and click in fixed ways, but since the repository selector has fixed dimensions I don't think it would be a problem even in other machines with different screen resolutions. It results in coverage test to still fail though.

lejolly:
Try not to use hardcoded movements of the cursor, instead why not use `comboBox.show()` instead?

HyungJon:
I might be missing something, but invoking repositorySelector.show() is failing to display the list. It does try to, based on the arrow button changing its shade, but the list is not appearing. Is there something more to be done?

lejolly:
`RepositorySelector` isn't the `comboBox`?

HyungJon:
Is this approach 

ComboBox<String> repositorySelector = find("#repositorySelector");

wrong in this case?

lejolly:
That's correct. Have you tried running it on the FX thread?

HyungJon:
I'm not familiar with it yet. This is under ```UseGlobalConfigsTest```, what needs to be added (imported, coded, etc) to make it ```show()``` properly?

lejolly:
Usually it just involves just wrapping it with a `Platform.runLater()` when trying to manipulate the UI, the corresponding error that is indicative of this is `Exception in thread "main" java.lang.IllegalStateException: Not on FX application thread;`

HyungJon:
Done. I think it needs some fix though. How do I ensure that the code after ```Platform.runLater()``` will wait until the runnable is done running instead of running concurrently? I put a sleep(200) as a rudimentary measure, but it's not quite smooth enough.

lejolly:
increase the delay, yeah there are `sleep(X)` littered all around the tests to prevent race conditions

HyungJon:
Okay. I found 200 to be a good enough waiting time.

As for refactoring, for the sake of having proper names this is what I had in mind. This will require some name changes in several places though, so I'm a bit hesitant to make the change myself.
cap 2015-06-25 00-23-31-210

HyungJon:
Bumping for more testing, criticism and opinions on the refactoring.

acruis:
Do take a look at the changes in #749 (outlined by #732) to see how your PR fits in with the new behaviour of the repo selector.

HyungJon:
Understood. Concept-wise there should be no conflict (the only difference made here is that ```prefs.addToLastViewedRepositories``` now has a broader condition), but I'll look into the code to make sure there are no unforeseen clashes.

HyungJon:
I believe moving the condition check out of the refactored ```openRepository``` back to ```openPrimaryRepository``` and ```openRepositoryFromFilter``` will reconcile both versions. Like this.

cap 2015-07-02 17-02-24-630

I think the best approach will be to close this pull request and repeat these changes on a branch that has #749 merged.

lejolly:
conceptually, wouldn't the current default repo already be open, if so what is the difference between the two methods? also, you can now merge the latest master into your branch since #749 has already been merged and then see how your changes can fit into it. 

p.s. try not to post screenshots of code, rather just copy it here like this:
```java
System.out.println("Hello world");
```

HyungJon:
Good point. I think ```isAlreadyOpen(repoId)``` in ```openPrimaryRepository``` would be a redundant check and only default repo check is sufficient. As for why the default check is added, that's actually the cause of #665 . Selecting a repo from the combo box was not adding the repo to the viewed repos list because the items found in the combo box are those that have already been opened. 

The reason for this separation is that if the same conditions are used, then opening a repo from the filter will result in endless loop of refreshing. So it appeared that opening from the filter should have the original condition.

HyungJon:
Actually if I recall correctly, I think the ```isAlreadyOpen(repoId)``` was left there because when logging in (i.e. when no repos are open yet) checking only the default repo may cause problems. I'll check again.

acruis:
After you're done, do check out and try out the version on the current `master` and see how you plan to deal with #665.

HyungJon:
I've already merged the current master and revised the code. Resolving conflicts and modifying tests now.

HyungJon:
I'm having more trouble with the merge than I expected. I will close this pull request, delete the branch and restart from the new master. Since the new tests have already been written in #749 it seems I only have to rewrite the code for opening the repositories.

PR #746, opened 2015-06-27
No error notification or recovery on corrupted repo JSON #700

### LINKED ISSUE #700: No error notification or recovery on corrupted repo JSON
A JSON storage file of mine was somehow corrupted. When I loaded HT, it detected that the JSON file exists for the repo, and tries to load it but fails silently, resulting in only a thrown exception and a blank panel where the repo's issues are supposed to be.

Ideally, the user should be notified of the problem through the status bar, and HT should automatically attempt to re-download the repo from GitHub.

screen shot 2015-06-16 at 11 42 06 am

ndt93:
Fixes #700 

This only resolves the case in which the JSON data is arbitrarily corrupted. If the file is changed but still syntactically correct and fit into a Model object (likely from deliberate manipulation), the program won't be able to detect the mismatch.

Lol, the coveralls thing is so demanding :P

### DIFF
--- a/src/main/java/backend/RepoIO.java
+++ b/src/main/java/backend/RepoIO.java
@@ -52,32 +52,36 @@ public class RepoIO {
 
     public CompletableFuture<Model> openRepository(String repoId) {
         if (repoStore.isRepoStored(repoId)) {
-            return loadRepositoryFromStore(repoId);
+            return loadRepoFromStoreAsync(repoId)
+                    .exceptionally(e -> {
+                        return downloadRepoFromSourceBlocking(repoId);
+                    });
         } else {
-            return downloadRepositoryFromSource(repoId);
+            return downloadRepoFromSourceAsync(repoId);
         }
     }
 
-    private CompletableFuture<Model> loadRepositoryFromStore(String repoId) {
+    private CompletableFuture<Model> loadRepoFromStoreAsync(String repoId) {
         return repoStore.loadRepository(repoId)
-                .thenCompose(this::updateModel)
-                .exceptionally(ex -> {
-                    try {
-                        return downloadRepositoryFromSource(repoId).get();
-                    } catch (ExecutionException | InterruptedException e) {
-                        logger.info("Error while downloading " + repoId);
-                        logger.error(e.getLocalizedMessage());
-
-                        return new Model(repoId);
-                    }
-                });
+                .thenCompose(this::updateModel);
     }
 
-    private CompletableFuture<Model> downloadRepositoryFromSource(String repoId) {
+    private CompletableFuture<Model> downloadRepoFromSourceAsync(String repoId) {
         UI.status.displayMessage("Downloading " + repoId);
         return repoSource.downloadRepository(repoId)
                 .thenCompose(this::updateModel)
                 .exceptionally(withResult(new Model(repoId)));
+    }
+
+    private Model downloadRepoFromSourceBlocking(String repoId) {
+        try {
+            return downloadRepoFromSourceAsync(repoId).get();
+        } catch (ExecutionException | InterruptedException e) {
+            logger.info("Error while downloading " + repoId);
+            logger.error(e.getLocalizedMessage());
+
+            return new Model(repoId);
+        }
     }
 
     public CompletableFuture<Model> updateModel(Model model) {
--- a/src/main/java/backend/RepoIO.java
+++ b/src/main/java/backend/RepoIO.java
@@ -64,7 +64,7 @@ public class RepoIO {
                 .exceptionally(ex -> {
                     try {
                         return downloadRepositoryFromSource(repoId).get();
-                    } catch (ExecutionException|InterruptedException e) {
+                    } catch (ExecutionException | InterruptedException e) {
                         logger.info("Error while downloading " + repoId);
                         logger.error(e.getLocalizedMessage());
 
--- a/src/main/java/backend/RepoIO.java
+++ b/src/main/java/backend/RepoIO.java
@@ -5,9 +5,12 @@ import static util.Futures.withResult;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 
 import org.apache.logging.log4j.Logger;
 
+import ui.UI;
+import util.HTLog;
 import backend.github.GitHubSource;
 import backend.interfaces.RepoSource;
 import backend.interfaces.RepoStore;
@@ -16,8 +19,6 @@ import backend.json.JSONStoreStub;
 import backend.resource.Model;
 import backend.resource.serialization.SerializableModel;
 import backend.stub.DummySource;
-import ui.UI;
-import util.HTLog;
 
 public class RepoIO {
 
@@ -51,14 +52,32 @@ public class RepoIO {
 
     public CompletableFuture<Model> openRepository(String repoId) {
         if (repoStore.isRepoStored(repoId)) {
-            return repoStore.loadRepository(repoId)
-                .thenCompose(this::updateModel)
-                .exceptionally(withResult(new Model(repoId)));
+            return loadRepositoryFromStore(repoId);
         } else {
-            return repoSource.downloadRepository(repoId)
+            return downloadRepositoryFromSource(repoId);
+        }
+    }
+
+    private CompletableFuture<Model> loadRepositoryFromStore(String repoId) {
+        return repoStore.loadRepository(repoId)
+                .thenCompose(this::updateModel)
+                .exceptionally(ex -> {
+                    try {
+                        return downloadRepositoryFromSource(repoId).get();
+                    } catch (ExecutionException|InterruptedException e) {
+                        logger.info("Error while downloading " + repoId);
+                        logger.error(e.getLocalizedMessage());
+
+                        return new Model(repoId);
+                    }
+                });
+    }
+
+    private CompletableFuture<Model> downloadRepositoryFromSource(String repoId) {
+        UI.status.displayMessage("Downloading " + repoId);
+        return repoSource.downloadRepository(repoId)
                 .thenCompose(this::updateModel)
                 .exceptionally(withResult(new Model(repoId)));
-        }
     }
 
     public CompletableFuture<Model> updateModel(Model model) {
--- a/src/main/java/backend/json/ReadTask.java
+++ b/src/main/java/backend/json/ReadTask.java
@@ -5,14 +5,17 @@ import java.util.concurrent.CompletableFuture;
 
 import org.apache.logging.log4j.Logger;
 
-import com.google.gson.Gson;
-import com.google.gson.reflect.TypeToken;
-
+import util.HTLog;
+import util.exceptions.JSONLoadException;
+import util.exceptions.RepoStoreException;
 import backend.interfaces.RepoStore;
 import backend.interfaces.StoreTask;
 import backend.resource.Model;
 import backend.resource.serialization.SerializableModel;
-import util.HTLog;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonParseException;
+import com.google.gson.reflect.TypeToken;
 
 class ReadTask extends StoreTask {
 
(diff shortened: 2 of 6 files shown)


src/main/java/backend/json/ReadTask.java
```diff
@@ -27,21 +30,39 @@ public ReadTask(String repoId, CompletableFuture<Model> response) {
...
+            response.complete(model);
+        } catch (RepoStoreException e) {
+            logger.error(HTLog.format(repoId, "Failed to load from store"));
+            response.completeExceptionally(e);
+        }
     }
 
-    private Model load(String repoId) {
+    /**
+     * Load repository data from RepoStore into a new Model.
```
damithc: `Load` --> `Loads`

lejolly:
please take note of the proper naming for pull requests.

ndt93:
Ok no problem.

src/test/java/tests/StoreTests.java
```diff
@@ -90,6 +93,44 @@ public void testStore() throws ExecutionException, InterruptedException {
...
+
+        JSONStore jsonStore = new JSONStore();
+        jsonStore.loadRepository("testrepo/testrepo").get();
+
+        File f = new File("store/test/testrepo/testrepo");
+        f.delete();
+    }
+
+    @Test(expected = ExecutionException.class)
+    public void testNonExistedJSON() throws InterruptedException, ExecutionException {
```
acruis: `NonExisted` here, `NonExist` below. Do keep your method naming consistent.

src/main/java/backend/RepoIO.java
```diff
@@ -51,14 +52,32 @@ public RepoIO(boolean isTestMode, boolean enableTestJSON) {
...
-            return repoSource.downloadRepository(repoId)
+            return downloadRepositoryFromSource(repoId);
+        }
+    }
+
+    private CompletableFuture<Model> loadRepositoryFromStore(String repoId) {
+        return repoStore.loadRepository(repoId)
+                .thenCompose(this::updateModel)
+                .exceptionally(ex -> {
+                    try {
```
acruis: Why is this try-catch block here? Exceptions from `downloadRepositoryFromSource` should have been caught by the `.exceptionally` call downloadRepositoryFromSource.

ndt93: `exceptionally`'s mechanism is different from a normal `try/catch`. It's a proper part of the stream's monadic chain and only deal with exceptions which has been wrapped using `CompletableFuture.completeExceptionally` by 1 of the `CompletableFuture` earlier in the chain.

As such, `exceptionally` handles a special control flow path in which it requires a unwrapped return value for which I use `CompletableFuture.get` to obtain, in order to preserve the same stream programming style. From the *[javadocs](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#get--)*, `CompletableFuture.get` will throw various exceptions (in the traditional way) in various scenarios.

To summarize, the code won't compile if the `try/catch` is removed.

src/test/java/tests/StoreTests.java
ndt93: Ok sure, I will change all to `NonExisted`

lejolly: I think the name should be `NonExistent`?

ndt93: Lol pardon my English, I think being consistent should be good enough for these functions for now? I will probably check the words' definitions probably next time :p

src/main/java/backend/RepoIO.java
dariusf: An alternative way to see why it won't compile is that `Runnable::run` is not declared as throwing any exceptions, and `ExecutionException` and `InterruptedException` are both checked.

The try/catch is unfortunately necessary, but it'd be nice to factor this out, possibly into something more generic. Maybe a new function, `withFuture`, so we can do something like this?

```java
return repoStore.loadRepository(repoId)
    .thenCompose(this::updateModel)
    .exceptionally(withFuture(downloadRepositoryFromSource(repoId)));
```
This way we can reuse existing error-handling from `downloadRepositoryFromSource`.

The implementation should be fairly similar to [withResult](https://github.com/HubTurbo/HubTurbo/blob/master/src/main/java/util/Futures.java#L60).

acruis: Ah, I see. So the try-catch block is there to handle `.get()`'s exceptions.

Then, the next question that arises is, why is `loadRepositoryFromStore` handling the task of downloading data from source? The name `loadRepositoryFromStore` is unclear as it doesn't imply that `loadRepositoryFromStore` will trigger the download.

So, you can either change the name to `loadRepositoryFromStore` to something clearer and longer (which is probably not desirable), or maybe you can do error handling at a higher level in the call chain, like in `openRepository`?

acruis: Just noticed @dariusf's comment. A `withFuture` does seem to make more sense than what I suggested.

src/main/java/backend/json/ReadTask.java
```diff
@@ -27,21 +30,40 @@ public ReadTask(String repoId, CompletableFuture<Model> response) {
 
     @Override
     public void run() {
-        Model model = load(repoId);
-        response.complete(model);
+        try {
+            Model model = load(repoId);
+            response.complete(model);
+        } catch (RepoStoreException e) {
+            logger.error(HTLog.format(repoId, "Failed to load from store"));
```
lejolly: Try not to use `failed` in error messages as it makes failed tests harder to find in the ci build logs.

src/main/java/backend/RepoIO.java
dariusf: The naming is a good point too.

On second thought, `withFuture` may need a second parameter for the value to default to (in this case, the empty model). This is less DRY, but I'm not sure the API is expressive enough to seamlessly add `downloadRepositoryFromSource` to the chain and have its result be used. What we need is `thenCompose` for handling errors. Maybe it can be implemented externally, like `withResult`?

ndt93: I actually struggle with quite a number of dilemmas for this fix since every solution seems to be either ugly in their own way or add more complications than I feel they deserve it. It's not really a problem with the chaining thing, but it's the lack of expressiveness in the current java's functional interface.

I wanted to do the handling in `openRepository` at first since that makes the most naturally sense. However, we have to implement part of the *Futures* chain in `loadRepo...` and the other part in `openRepository` since the chain returns immediately and may crash at arbitrary time, so a surrounding `try/catch` is technically useless. And it seems unsymmetrical for me b/w the `downloadRepo...` and `loadRepo...` (aka there is a mismatch in level of abstraction). When a developer uses these 2 functions, `downloadRepo...` handles everything nicely in a package while `loadRepo...` requires its user to handle exceptions separately. What makes it worse is that I cannot make `loadRepo...` *throws exceptions* loudly because the exceptions are produced within the monad chain's mechanism. So I justify that `loadRepo...` is handling something weird (that `RepoStore.isRepoStored` returns `true` but the repo is technically not stored) and should, as part of its behavior, handle this case behide the scene, and to the viewpoint of a user of this method, it's behaving as expected (well, except for being slower). This behavior could be made explicit by a docs for this method. We can also make things symmetrical by taking `exceptionally` out of both `downloadRepo...` and `loadRepo...` but then there is, uuh, DRY and everything :P. Anw, actually I don't have a very strong opinion on this one since all of the solutions that I think could work are ugly in some ways, so a vote or any althernative suggestion?

Now, the `try/catch` looks weird in the flow of the chain and we could try to generalize this like `withResult`. But then it seems that in this case, the exception is modifying the control flow in a quite dangerous way with the chain from `downloadRepo...` having deep side-effects (`updateModel`) and blocking. A generic exception handler would not be aware of the uncertain nature of its possibly multiple argument CompletableFutures and could be more restricting than making things more flexible (I guess that's why java doesn't allow thenm in `exceptionally` in the first place). So I prefer not to handle these *exceptions* in a generic way but feel more comfortable having all the *try/catch/throw/exception* smacked onto my face loudly where they belong :P. Having such a method would be a good structural convenience though, if we have more snippets that handle exceptions by having 1 wrapped value, 1 unwrapped value and a default error message with trivial execution footprints then it would be a good time to refactor.

Then a nicely packaged way is to extend java's functional CompletableFutures to be more expressive, but then I feel it does not yet deserve the additional complications given our purpose.

src/main/java/backend/json/ReadTask.java
ndt93: Ok sure.

src/main/java/backend/RepoIO.java
dariusf: There *are* lots of different factors at play here. We should pick something that achieves the right tradeoffs.

Personally I'm for either `withFuture` (where error-handling is left implicit) or a method like `downloadRepoNow` (which factors out the blocking `get`, keeping error-handling explicit). Either one keeps things at the same level of abstraction *within the method*.

While there may be some asymmetry in how `load` and `download` work *as methods*, this is already the highest level of abstraction in which they exist (`RepoIO` is supposed to abstract away differences between GitHub and cache), so I'm not sure we can do anything about it. Anything higher up is meant to be insulated from caching details.

Blocking and updating the model in `exceptionally` should be fine. Anything downstream in the chain of operations will receive the same data, just a bit later. In that sense there are no side effects.

Handling exceptions in a generic manner has tradeoffs, as you mentioned. There are more flexible alternatives to `exceptionally` which give you the Throwable to work with, though I'm not sure that much explicitness is necessary. Maybe we could move the `exceptionally` clause higher, to make it explicit that we're not handling the failure of `updateModel`?

```java
return repoStore.loadRepository(repoId)
    .exceptionally(/* download repo */);
    .thenCompose(this::updateModel)
    .exceptionally(withResult(new Model(repoId)));
```

Speaking of which, would updating the empty model handle the download, i.e. would this work?

```java
return repoStore.loadRepository(repoId)
    .exceptionally(withResult(new Model(repoId)));
    .thenCompose(this::updateModel)
    .exceptionally(withResult(new Model(repoId)));
```

ndt93: The first thing I tried is updating an empty Model since it require the least amount of code and almost everything can be kept as they are. But then the default constructor give the empty Model the timestamp as current time and nothing is updated. I considered explicitly assign the timestamp to some time in the past but it feels too much like a hack. Then I also want to resuse all the downloading new repo goodies like the progress bar without changing so much code so user knows why it takes so long in this case.

Having multiple exceptionally then a `thenCompose` doesn't help (like in the 1st code snippet, I tried a similar thing also) since according to the docs, if any CompletionStage got exception then every later stage also completes exceptionally (like in a unix pipe). I was trying to use `isCompleteExceptionally()` but then I didn't get anything out of it either.

I have also tried all other variants of exception handling in `CompletableFuture`. They are essentially buffed up `exceptionally` with additional info passed in but exactly the same restricting expressiveness so they don't help either.

The side-effects I talked about won't happen since we have this *pendingQueue* thing. But suppose we didn't have that one and tried to download same repos concurrently then the side-effects will become obvious.

So something that looks so simple turns out to be so annoying :P

ndt93: Hmm I think your `downloadRepoNow` is a great idea

dariusf: Sure, that's more explicit.

Just to confirm how multiple `exceptionally` clauses work: once an `exceptionally` triggers and yields some result, my understanding is that downstream actions get the result as if all went well upstream and continue, only jumping to the next `exceptionally` if there is *another* exception. That's what we're relying on for finer-grained error-handling, by having `exceptionally` at different levels of abstraction.

ndt93: Yup, I think that' the case. You can confirm it *[here](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html)*, it's the 4th paragraph of the intro description) but I may misinterpret something. What I was trying actually has a slightly difference purpose though. I wanted to use `thenCompose(downloadRepo...) with `exceptionally` and some additional stages before and after to handle the downloading so we can  eliminate the `try/catch` and blocking `get`` all together, but then that `thenCompose` complete exceptionally if 1 of the stage before complete exceptionally so it ends up in the last `exceptionally` and return empty model anyway

dariusf: Oh, right. Yes, this is the expressiveness problem from before: that `thenCompose` can't be used to wrap or replace `exceptionally`.

ndt93: Ok well then I guess I would refactor the thing to `downloadRepoBlocking` later if no one else has other opinions on this.

ndt93: I was bumping into something really interesting while refactoring that highlights the side-effects of things which I think would be beneficial to everyone to know.

So I'm trying to reuse `withResult` with the new `downloadRepoFromSourceBlocking(...)` and did something like this:

```java
...
return loadRepoFromStoreAsync(repoId)
        .exceptionally(withResult(downloadRepoFromSourceBlocking(repoId)));
...
```

The program ended up always downloading the repo no matter what. The reason is that java is not lazily evaluated and `downloadRepoFromSourceBlocking(repoId)` gets evaluated immediately as an argument to `withResult` and hence downloads the repo straightaway whenever `openRepository` is called.

acruis: Yep, I think it's better to put that in a new issue, too.

Good insight on lazy evaluation though! (or the lack of)

lejolly:
@ndt93 is this ready to be reviewed?

ndt93:
Yup, I have no more things to add for this at the moment, unless you guys find some other things to change

acruis:
OK, reviewed and ready to merge.

PR #755, opened 2015-07-02
Clicking on pView does not bring the bView to front #747

### LINKED ISSUE #747: Clicking on pView does not bring the bView to front
This bug did not crop up during the summer, until now. It used to be frequent during semester period. The problem did not go away even after a restart. 
Win 8.

lejolly:
Fixes #747


lejolly:
@damithc please check if this branch solves the issue

damithc:
Yes, seems it's working again. I guess it's difficult to add a test for this?

lejolly:
Yup because the ci runs on Linux.

lejolly:
Decided not to implement relaunching of chrome if the bview can't be reached on focus back to the pview. Chrome will only be relaunched when the user clicks on an issue/navigates to another url. @acruis thoughts?

acruis:
Sounds fine to me.

PR #749, opened 2015-06-29
Proposed behaviour for RepositorySelector #732

lejolly:
Fixes #732, #619 and #665.

PR #756, opened 2015-07-02
unexpected default project at start up #665

### LINKED ISSUE #665: unexpected default project at start up

HyungJon:
Fixes #665 
Re-written version of #731 so that it is compatible with #749 

Additional changes made from #731 and #749 
1. The boolean parameter ```isPrimaryRepository``` for method ```openRepository```has been removed, as condition checks were moved to ```openPrimaryRepository``` and ```openRepositoryFromFilter```, making the parameter redundant.
2. Checking ```isAlreadyOpen``` turned out to be not redundant.
3. Corrected the condition check for ```updateUI```. Previously, it would update only if the repository to be opened is already open or already pending.


src/main/java/backend/Logic.java
```diff
@@ -85,22 +85,26 @@ public void refresh() {
     }
 
     public CompletableFuture<Boolean> openPrimaryRepository(String repoId) {
-        return openRepository(repoId, true);
+        assert Utility.isWellFormedRepoId(repoId);
+        prefs.setLastViewedRepository(repoId);
+        if ((isAlreadyOpen(repoId) && repoId.equals(getDefaultRepo())) || models.isRepositoryPending(repoId)) {
```
lejolly: if i understand this correctly, the extra check here is the main change in this pr. so that we don't do anything if the primary repo remains the same right? my question is will there ever be a case where this method is called but the primary repo isn't changed?

lejolly: to add on to the above, could you provide a list of steps that can reproduce the issue that the extra check is meant to solve? (list of steps applied to the current master should demonstrate the issue, same list of steps applied to your branch should not demonstrate the issue)

HyungJon:
It appears #749 has fully fixed the issue I attempted with #731 through a different approach, hence this PR is mostly obsolete. I will close this PR, unless there are any other changes I made to the refactoring and the condition check for ```updateUI``` that appears should be merged.

acruis:
Go ahead and close this PR :)

PR #758, opened 2015-07-03
 'mark as read' does not remove issue from 'is:unread' filter #754 

### LINKED ISSUE #754: 'mark as read' does not remove issue from 'is:unread' filter
The issue eventually disappears from the panel but it takes 20 seconds (sometimes more) for that to happen. Previously, pressing `E` made the issue disappear immediately. Is the delay intentional? I don't mind such a delay if it was intentional and deterministic. In fact I prefer to have such a delay.
image

HyungJon:
Fixes #754 

Marking an issue as read automatically changes the focus to the issue below the read issue, unless the read issue is at the bottom.
No changes made to the delay in the read issue disappearing, as this is dependent on the deterministic factor of column refresh intervals.


acruis:
Looks good.

@damithc The read issues are cleared upon a refresh. Is this behaviour deterministic enough?

damithc:
Yes, that is fine. I'll open a separate issue if I don't like it after dog fooding for a while. Meanwhile, can merge this one.

HyungJon:
Automatic refresh upon marking an issue as read or after a certain interval to give time for rectifying mark as read by mistake can be implemented, but my opinion is that dismissing all read issues with a single scheduled refresh would be more suitable and result in less load on the system.

acruis:
The refresh to clear the read issues doesn't engage the repo I/O component of the program. It simply re-renders the `ListPanelCards` and thus shouldn't cause too much load on the system.

I think the current version is acceptable, though. It's more convenient to dogfood on `master`, anyway. Time to merge!

src/test/java/guitests/KeyboardShortcutsTest.java
```diff
@@ -98,10 +98,22 @@ public void keyboardShortcutsTest() {
...
         push(getKeyCode("MARK_AS_READ"));
+        assertEquals(issueIdExpected, selectedIssueId);
+        push(getKeyCode("UP_ISSUE")); //required since focus has changed to next issue
+        assertEquals(true, issuePanel.getSelectedIssue().isCurrentlyRead());
+        
+        push(KeyCode.END);
+        push(getKeyCode("MARK_AS_READ"));
+        assertEquals(1, selectedIssueId);
         assertEquals(true, issuePanel.getSelectedIssue().isCurrentlyRead());
+
```
damithc: 1. I wouldn't mind a bit more comments in tests, to explain the intention of each test case, what is expected etc. 
2. Tests can be more extensive. e.g. Corner cases such as there is only one issue in the list, no issues in the list, etc.

acruis: Noted.

Issue #740, opened 2015-06-26
Label Picker


lejolly:
Reduce interaction with browser
Autocomplete for shortest match
Will need to be broken up into smaller steps
Pop-up should be populated with existing labels

Later
Label groups: e.g. priority, if it has medium, by typing high, it should remove medium and put high

acruis:
Label groups can be either exclusive (`p.High`) or nonexclusive (`feature-filters`). The classification is determined by whether the separator in the label name is a `.` or a `-`.

damithc:
Shouldn't this match `enhancement` instead of `duplicate`?
![image](https://cloud.githubusercontent.com/assets/1673303/8612487/e18468e8-2703-11e5-8c92-0ca156428489.png)

damithc:
Pressing space without any other letters applies the first label in the list. May be we should not do that? i.e. not start selecting labels until user types a non-space letter.

lejolly:
ok i can disable selection when the textfield is empty, as for the previous one, currently selected labels are always placed in front for easy removal. you can use the up/down arrow keys to change the selection too.

lejolly:
Closed with #759

PR #759, opened 2015-07-03
Label Picker #740

lejolly:
Implements #740

### DIFF
--- a/src/main/java/backend/stub/DummyRepoState.java
+++ b/src/main/java/backend/stub/DummyRepoState.java
@@ -90,7 +90,7 @@ public class DummyRepoState {
         // Issue 6 is closed
         issues.get(6).setOpen(false);
 
-        // add more labels into repo
+        // Label groups for testing label picker exclusivity
         labels.put("p.low", new TurboLabel(dummyRepoId, "p.low"));
         labels.put("p.medium", new TurboLabel(dummyRepoId, "p.medium"));
         labels.put("p.high", new TurboLabel(dummyRepoId, "p.high"));
--- a/src/main/java/ui/components/pickers/LabelPickerDialog.java
+++ b/src/main/java/ui/components/pickers/LabelPickerDialog.java
@@ -29,7 +29,9 @@ public class LabelPickerDialog extends Dialog<List<String>> {
     private Map<String, Boolean> resultList = new HashMap<>();
     private FlowPane topPane;
     private FlowPane bottomPane;
-    private Optional<String> possibleAddition = Optional.empty();
+    private Optional<String> targetLabel = Optional.empty();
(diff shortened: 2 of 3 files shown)


lejolly:
some progress

edit_labels_for_issue__744_in_hubturbo_hubturbo

acruis:
Added backend support for replacing issue labels. When the GUI is ready, switch the current `replaceIssueLabels` method in RepoIO with the currently-commented out one below it.

lejolly:
final UI for label picker

edit_labels_for_issue__765_in_hubturbo_hubturbo

damithc:
Don't spend a lot of time polishing/testing this UI because it is not what
I want. But ok as a first version. In the UI I have in mind, label
selection will be done by typing part of the label rather than
navigate/toggle approach.

lejolly:
like this?

hubturbo

damithc:
Label dialog goes out of screen for some issue.
image

lejolly:
would you prefer it to appear in the middle of the screen?

damithc:
Enter key should apply the last label in-progress. e.g. `bu`<kbd>enter</kbd> should be as same as `bu`<kbd>space</kbd><kbd>enter</kbd>

lejolly:
ok

damithc:
I think the position depends on where the issue card is. Ideally, the dialog should appear as close to the issue card as possible, should not cover the issue card, and should not go out of the screen.

lejolly:
ok, i might tackle that after changing the `TextField` behaviour. But i can't reproduce the going out of the screen issue on OS X.

damithc:
I'm ok with the current UI. Add tests, update user docs, get it reviewed and merge :-)

lejolly:
ok will make some minor changes and then work on getting it ready to merge

lejolly:
@damithc 2 changes
1. labels in the picker will now show the full label names including the groups to reduce confusion for situations such as: `p.low` and `d.low`
2. searching by groups is now supported using `p.` should display `priority.low`, `priority.medium` and `priority.high`, thus typing `p.l` should select `p.low` or `priority.low`

damithc:
Nice work. May be sort by group? i.e.
aspect labels
[blank line] 
priority labels
[blank line]
...
[blank line]
regrouped labels

It not trivial to do, can do later as a separate issue.

lejolly:
@damithc could you see if the latest commit fixes #789?

damithc:
Yes, it is better now. Adding the tick now moved the next label a few pixels, not as much as before. The other time the whole label list shifts is when the list of labels (above the text box) spills over to the next line.

acruis:
Good to merge!

Issue #762, opened 2015-07-05
Update .classpath to be compatible with new libraries


HyungJon:
Some changes to code may also be desirable as the new setting results in several more warnings.

lejolly:
What warnings?

HyungJon:
Most are "FutureTask is a raw type. References to generic type FutureTask<T> should be parametrized" warning. That and some "The value is not used" warnings.

ndt93:
You can use the command `gradle eclipse` to generate eclipse config files from the updated *build.gradle*. There is also an official [eclipse extension](https://marketplace.eclipse.org/content/buildship-gradle-integration) for gradle or this more popular [eclipse tooling](https://marketplace.eclipse.org/content/gradle-integration-eclipse-0) for gradle.

Should we remove the eclipse files from the repo all together, and/or mention the above steps in the developer guide? Since it could be confusing for an eclipse user who hope to contribute to this repo and happen to think that the eclipse project files are supposed to be updating.

Eclipse users unite! :P

damithc:
Eclipse files should be updated. Eclipse should be the official IDE of this
project. Normally I insist on all devs using the official IDE, but in this
case I allow using IDEA because it is good for the dev team to have some
hands-on experience with other IDEs.

PR #763, opened 2015-07-05
 Update .classpath to be compatible with new libraries #762 

HyungJon:
Updated .classpath file.


HyungJon:
The build is successful on my machine but fails here. It appears just updating the .classpath file is not enough.

HyungJon:
I attempted to build test again, and it skipped all JUnit tests and returned build successful immediately. I don't think I can resolve this issue. Closing PR for now.

PR #766, opened 2015-07-07
Pressing F5 triggers two calls of Logic.refresh #742

### LINKED ISSUE #742: Pressing F5 triggers two calls of Logic.refresh
screen shot 2015-06-26 at 12 02 16 pm

There's no difference in the data presented to the user, but wasteful in terms of API calls.

Clicking "Refresh" from the top menu bar, changing the bView, and periodic refreshes don't cause this problem.

I'm not sure if this is a Mac problem. Do help me test on other machines and see if you can reproduce console messages like above :)

acruis:
Fixes #742


lejolly:
Looks good to go.

Issue #290, opened 2014-12-31
Down arrow goes to the wrong item if the item list changed immediately before


damithc:
For example, say 3rd issue is selected and a refresh adds another issue above the selected issue. Pressing the down arrow will jump to the originally in 5th position instead of the one at 4th position.

dariusf:
I've fixed the problem with an issue itself disappearing. The other problems are harder, due to the lack of control we have over the JavaFX list.

To summarise the issues:

- The model of selection we keep remains correct. Selection moving to wrong items is a visual glitch.
- It's hard to distinguish cases where selection jumps to the top and cases where it moves due to things being added.
- We know when the refresh is completed, but we don't know when the list actually finishes updating. Selecting an item on refresh doesn't work because the list may not actually have reflected the changes.

The alternatives mentioned just now are:

- Intercepting default arrow key behaviour and handling the events ourselves. Quick and easy implementation and almost works, but a major problem with this is that the list doesn't scroll naturally. There doesn't seem to be an easy way to override scrolling behaviour (big class hierarchy, complicated interdependencies, mostly private methods).
- Implementing selection on top of [this](https://github.com/TomasMikula/Flowless) low-level layout widget. More work, but the upsides of this are a greater degree of control, selection and jumping issues going away, ability to lazily scroll through large lists.

dariusf:
Fixed in b985fe9244e97a3eb5ca57f339396b18f07fd513 and d0b6501803b2ade7de5aadf5264fef0f13060bb2.

I found a way to scroll a listview programmatically, which allowed me to implement the first fix. Flowless may be a viable direction for the future but has its share of quirks too.

damithc:
Nice. I'm always a bit wary of third-party libraries, especially immature
ones.

damithc:
The issue is not completely solved. I observed in the following situation:
nth issue is selected. 
1.move to bView and modify that issue.
2. move to HT and press down arrow. selection goes to the next issue (correct).
nth issue disappears due to the changes done a moment ago. This is delayed because refreshing 3. takes  few seconds.
4. wrong issue is selected (selection changes to the wrong issue).

dariusf:
Okay, I'll fix this for the release later.

damithc:
Another situation:
The selected issue disappears from the panel. The next issue is selected automatically. However, that issue is not loaded in bView (i guess that's fine).
I click on the selected issue. It does not load in the bView. I press enter. Still it doesn't load in bView.

[...]

PR #776, opened 2015-07-09
pView does not auto-scroll to the selected card when the selection goes out of the visible area #774

### LINKED ISSUE #774: pView does not auto-scroll to the selected card when the selection goes out of the visible area
as per title. Let me know if not clear.

lejolly:
Fixes #774


lejolly:
it now calls scroll to item when the selection has been restored, the problem is that the scrolling only partially works.

lejolly:
scrolling issues can be solved by eliminating scrolling animations (which is a custom implementation by HT)

current implementation
flow1

native implementation of scrolling without any animations
flow2

@damithc what do you think?

lejolly:
the previous commit changes the scrolling to the native implementation, @damithc could you try it out and see if it works for you?

lejolly:
can be dogfood-ed to see if the new behaviour is acceptable.

damithc:
The current version 'scrolls selection to top' (the selected issue goes to
the top of the panel) which is not suitable. When the selection is visible,
no scrolling should happen. If the selection goes out of the visible area,
the list should auto-scroll just enough to bring the selected issue into
view. That is the 'normal' auto-scroll behaviour expected of a list panel.

damithc:
I was watching the GUI tests and noted that auto-scrolling seems to work fine in those tests where there is a long list of issues.

lejolly:
but in actual use i think the rate of success is a lot lower.

damithc:
It may be because I had a second monitor on the left (it went out of the
first monitor but still showed up in the 2nd monitor)

lejolly:
@damithc does the latest version still scroll if the selection is visible?

damithc:
Which branch should I try?

lejolly:
The Issue774 branch

damithc:
Better, but not there yet. Current situation: Let's say 10 cards are
visible. If I navigate from 10th to 11th, the 11th scrolls to the top and
panel shows cards 11-20. Desired: scroll by one card so that cards 2-11 are
visible.

damithc:
Actually, scrolling of the label picker branch is better. the only problem
is the success rate is not 100%. Sometimes scrolling is as expected,
sometimes not enough (the selected card does not become visible completely).

lejolly:
Yes the original version of scrolling works exactly the way you want it, the only problem is that it isn't entirely reliable, so it's really up to you which version you prefer. I haven't found a foolproof fix for the current version.

damithc:
Is it possible to tweak the original version to auto-scroll to the card
that is just below the selected card, instead of trying to scroll to the
selected card?

lejolly:
That's possible but I don't think it'll solve the problem, will try it out tomorrow.

damithc:
The current one (i'm on label picker branch) seems to work most of the time. We can also use Ctrl+down to scroll at times it doesn't work. while it is not ideal, it is not totally unusable either. See if you can improve the success rate a bit. If cannot, we can KIV this as it is not that important.

dariusf:
Some context on this problem in #290. `ScrollableListView` was actually supposed to address the shortcomings of the native `scrollTo`, and this issue of only being able to show a card by moving it to the top. It does need more polish though.

Would it be possible to compute Y_OFFSET dynamically (based on card height), instead of using a constant? That would solve the problem of cards not becoming completely visible.

damithc:
What happened to this PR?

lejolly:
um, it hasn't been fixed yet, but i'm closing this pr for now because all the stuff i did in the past is now invalid on the latest master. 

also, since this doesn't happen all the time and repeatedly pressing the same key eventually gives the desired result (for scroll to top/bottom), i guess this isn't such a high priority now.

PR #796, opened 2015-07-20
Change Eclipse development environment to use Gradle #792

### LINKED ISSUE #792: Change Eclipse development environment to use Gradle
[Gradle has joined the Eclipse Foundation and is working to make it a first-class citizen in Eclipse. ](https://gradle.org/press-release/eclipse-gradle/)

For now, official support is still optional and must be installed using this [plugin](http://marketplace.eclipse.org/content/buildship-gradle-integration). After installation, one can just import the repo as a Gradle project directly into Eclipse. It will then download all necessary dependencies, etc. 

It seems to work (although not all the gradle tasks are visible), but I've tested assembling, testing and even checks (which include checkstyle). For a fresh installation of Eclipse, this is an easier option because the source, test, and resources folders do not need to be defined by the user. 

So shall we remove all Eclipse related files as well as the jars? I can then update the developer documentation.

lejolly:
Fixes #792


lejolly:
@acruis when reviewing this could you test out if the instructions [here](https://github.com/HubTurbo/HubTurbo/wiki/Development-Environment) work for eclipse?

damithc:
Does it have to be Eclipse Mars? Luna should work too right? Potential developers are likely to have a slightly older version already installed. So we should not insist on the latest version if the older one is good enough.

lejolly:
Yup it should work, will update the docs.

lejolly:
@acruis and @damithc could you dogfood this to ensure that updating the EGit version doesn't break any existing functionality.

lejolly:
easiest way to test this is to just use the gradle wrapper and run `clean shadowJar` to just build the jar skipping all the tests.

acruis:
Looks good to me.

damithc:
Tried with the fresh clone. Everything worked as advertised.

damithc:
BTW, importing to eclipse creates a file `.project`. Add that file to `.ignore`?

lejolly:
ok, got it.

acruis:
OK, to merge.

PR #803, opened 2015-07-22
ChromeDriverTest failure #802

### LINKED ISSUE #802: ChromeDriverTest failure
Lines 68, 70 and 156 in ```ChromeDriverTest``` have ```assertEquals(GitHubURL.DOCS_PAGE, url);``` instead of ```GitHubURL.getPathForDocsPage```, causing the test to fail because it retrieves ```.../blob/%s/docs/...``` instead of ```.../blob/V2.9.0/docs/...```

Is this intended, or should this be changed?

lejolly:
Fixes #802

### DIFF
--- a/docs/Creating-a-Release.md
+++ b/docs/Creating-a-Release.md
@@ -24,22 +24,15 @@
 
 ## Create and upload JAR
 
-**Eclipse**
-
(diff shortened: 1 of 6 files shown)


acruis:
Good to go.

PR #807, opened 2015-07-22
User can name a panel #782

### LINKED ISSUE #782: User can name a panel
so that user can give a descriptive name to a panel e.g. `recent high priority updates` `PRs I'm reviewing` etc.

HyungJon:
Fixes #782

Note: running HubTurbo with this PR merged will delete all saved filters because the save function works differently now.

There are currently clashes with the way boards are saved, resulting in panel names being reset when opening a new board. I will fix it in a fresh issue.


HyungJon:
@lejolly Can you check if the test still fails locally for you?

lejolly:
@HyungJon yup it passes for me now.

HyungJon:
Waiting after creating panel did not fix ```NoNodesFoundException```.

HyungJon:
I strongly suspect there is something fundamental I did wrong, given how other tests that do similar things pass with no problem, but I can't figure out what that is.

HyungJon:
Increasing the sleep time like ```UseGlobalConfigsTest``` did not resolve the issue.

lejolly:
yeah it's not a problem with the delays, but rather you should look at `LabelPickerTests` to see how tests with dialogs are done. namely, try not to show more than one window at a time.

HyungJon:
I am not sure whether creating an Event for displaying rename dialog will fix it, but I'll try it.

HyungJon:
Now just trying to fix a bug that causes rename dialog of every panel to appear when event handling is attempted.

HyungJon:
All right, I prevented ```NoNodesException``` now. (In commit 10399e3. 93805c2 still causes it somehow.) It also does save the panels to json, although for some reason panel names are saved as default instead of new names. I'll continue working on it.

HyungJon:
Confirmed that filters are saved correctly, which means the rename function is not working on Travis build. Again, it does pass on my local machine, and I can't figure out which part of the function (dialog, event handler, or renaming code" is not working correctly under Travis. I apologize for the trouble again.

lejolly:
Entering text into a pop up dialog on Travis is extremely tricky. I think some other test just went to find the textfield in the dialog and manually set the text instead of typing it out.

HyungJon:
All right, I fixed all problems with my codes. However ```UseGlobalConfigsTest``` which is unrelated to my PR keeps failing randomly. Should I cater to it within this PR as well?

lejolly:
nope. could you include a screenshot of the changes your pr implements.

HyungJon:
It adds a bar on top of each panel that displays panel name and a rename button.
cap 2015-07-29 22-43-27-157

And upon exiting, it saves the panels' names and filters to json so that the names and filters can be restored for next session.
cap 2015-07-29 22-46-39-170

lejolly:
ok, could you also add one more screenshot of how the whole HT interface looks like now. @damithc what do you think?

HyungJon:
cap 2015-07-29 23-07-07-754

damithc:
hmm... good first effort but not up to the level of our current UI.
Suggestions:
Instead of the 'Rename' button, show it as a clickable edit icon (the usual pencil icon) at the end of the current name.
Can the editing be done in-place instead of as a separate dialog? If cannot, polish up the dialog. The current one doesn't look very neat.
Is there a limit to the name? If not, how does it look if the name is very long? 
What's the default name for a panel?

HyungJon:
Default is "panel". I set maximum name length to 48, which seemed to be the longest a name can be without pushing the rename button outside.
I was going to swap the close and rename buttons, but if using the pencil icon I guess they should be placed side-by-side?

I will look into how I can rename panels without using dialogs, since dialogs cause trouble for Travis anyway.

damithc:
Ok, can put the edit icon above the close icon.

HyungJon:
I shifted the buttons around. I feel the rename and quit buttons could stand out a bit more though.
cap 2015-07-31 14-00-59-464
And I couldn't find a way to make the name editable without using a dialog within my skills. I cleaned up the dialog screen. Is this better?
cap 2015-07-31 14-01-18-352

damithc:
Good progress. 
Some suggestions.
The pencil icon does not match the x icon. Find a better matching icon pair?
Remove the question mark from the dialog box. Can remove the horizontal line?

HyungJon:
I changed the dialog now. 
cap 2015-07-31 19-57-04-540

I couldn't find a better unicode for the rename button, so I tried changing the X instead. If it doesn't fit, I'll use an image for the rename button.
cap 2015-07-31 19-56-47-617

damithc:
Sorry, just realized that you are following the look and feel of the 'save
board' dialog. Go ahead and use the same. It's better to be consistent.
I'm ok with the proposed two icons.
Another thing, when user saves a board, now we need to save the panel names
as well.

HyungJon:
I was initially going to work on the compatibility with boards function in a separate PR because I felt it might be too big a change for a single PR, but if it's okay I'll work on it within this one then.

damithc:
Well the two are interdependent (one can't exist without the other), so
have to go in the same PR.

HyungJon:
This is going to change the format of json save files. Is it okay if the old ```global.json``` becomes unusable, and has to be deleted (or manually modified to the new format) in order for the new version to work? I think I could enable auto-upgrade, but I'm not sure whether I should write code that will be very messy and will become dead after the upgrade happens.

damithc:
I'm afraid we have no choice but to write code that can work even with the
old json format (i.e. auto-upgrade). We don't know when this feature will
be released and who will be using HT by that time.

HyungJon:
I fixed the error that was preventing startup due to save file incompatibility, although it still can't read the contents whose save formats became different (i.e. saved panels and boards). Is this sufficient, or would you like an updater that will actually update the previous json file to the new format and retain the information in it? I think the updater can also be expanded to support future modifications that affect json save format.

HyungJon:
Implemented updater. I'll push it if it's appropriate.

HyungJon:
Requesting this PR to be reviewed.

lejolly:
some things to take note of:

1. pressing `Cancel` or <kbd>Esc</kbd> results in a `NoSuchElementException`, since the result is stored in an `Optional` please do the proper checks first. 
2. i think that name variables after the javafx component is less confusing: e.g. `nameText` rather than `nameBox` for a `Text` object. 
3. lastly, i did a quick prototype to allow doubleclick-ing the `nameArea` to allow me to edit the name of the panel. i think this is less jarring than a dialog box?

panel_rename

HyungJon:
Will take care of 1 and 2. I'll also look into cleaning up some JavaFX codes I wrote, for alignment in particular.
I agree 3 is better.

HyungJon:
Failures in unrelated test.

src/main/java/prefs/GlobalConfig.java
```diff
@@ -22,11 +23,11 @@
 
     private static final Logger logger = LogManager.getLogger(GlobalConfig.class.getName());
 
-    private List<String> lastOpenFilters = new ArrayList<>();
+    private List<PanelInfo> openPanels = new ArrayList<>();
```
dariusf: This can be named better. It's the list of panels on quitting, right?

src/main/java/prefs/GlobalConfig.java
```diff
@@ -65,28 +66,44 @@ public void clearMarkedReadAt(String repoId, int issue) {
...
+        return lastOpenFilters;
+    }
+    
+    public List<String> getPanelNames() {
+        List<String> openPanelNames = new ArrayList<>();
+        for (int i = 0; i < openPanels.size(); i++) {
+            openPanelNames.add(openPanels.get(i).getPanelName());
+        }
+        return openPanelNames;
+    }
```
dariusf: These two methods seem unnecessary. We already have a way to get the `PanelInfo`. Since they're only used in tests, map over that in the tests instead.

src/main/java/prefs/GlobalConfig.java
```diff
@@ -65,28 +66,44 @@ public void clearMarkedReadAt(String repoId, int issue) {
...
+    
+    public void setPanelInfo(List<PanelInfo> panelInfo) {
+        this.openPanels = new ArrayList<PanelInfo>(panelInfo);
+    }
```
dariusf: No need to parameterise the ArrayList.

src/main/java/prefs/GlobalConfig.java
```diff
@@ -65,28 +66,44 @@ public void clearMarkedReadAt(String repoId, int issue) {
...
+    
+    public List<PanelInfo> getPanelInfo() {
+        return new ArrayList<PanelInfo>(openPanels);
```
dariusf: Same here.

src/main/java/prefs/PanelInfo.java
```diff
@@ -0,0 +1,34 @@
+package prefs;
+
+public class PanelInfo {
+
+    private String name;
+    private String filter;
```
dariusf: This class can be immutable (final fields, no getter/setter boilerplate). It won't be changed unless you create a new panel.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -67,6 +80,50 @@ public FilterPanel(UI ui, IModel model, PanelControl parentPanelControl, int pan
...
         filterTextField.setKeywords(all);
     };
+    
+    private Node createNameBar() {
+        nameText = new Text(panelName);
+    
+        renameButton = new Label(RENAME_PANEL);
+        renameButton.getStyleClass().add("label-button");
+        renameButton.setId(model.getDefaultRepo() + "_col" + panelIndex + "_renameButton");
+        renameButton.setOnMouseClicked(e -> {
```
dariusf: Try to separate the event handlers from the code which creates the elements. `createNameBar` should call another method which creates all the elements, and all it should do after that is bind the event handler. This will facilitate refactoring the views in future.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -67,6 +80,50 @@ public FilterPanel(UI ui, IModel model, PanelControl parentPanelControl, int pan
...
+            TextInputDialog renameDialog = new TextInputDialog(panelName);
+            renameDialog.getEditor().setId("panelrenameinput");
+            renameDialog.setTitle("Rename " + panelName);
+            renameDialog.setHeaderText("Enter a new name for this panel.");
+            Optional<String> result = renameDialog.showAndWait();
+            mainStage.show();
```
dariusf: What does this do? I don't think it's a good idea for panels to be coupled to the stage.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -67,6 +80,50 @@ public FilterPanel(UI ui, IModel model, PanelControl parentPanelControl, int pan
...
+            if (result.isPresent()) {
+                newName = result.get();
+            }
+            if (!newName.equals("")) {
+                panelName = newName;
+            }
+            if (panelName.length() > MAX_NAME_LENGTH) {
+                panelName = panelName.substring(0, MAX_NAME_LENGTH);
+            }
+
```
dariusf: This is unnecessarily complicated. Can probably do something along the lines of `result.orElse("").substring(0, MAX_NAME_LENGTH)`.

The substring thing can also be factored out into a utility function (`truncate`?). It could add ellipses if the input is too long.

src/main/java/ui/issuepanel/PanelControl.java
```diff
@@ -56,28 +60,28 @@ public void updateModel(IModel newModel) {
...
 
-        if (filters.isEmpty()) {
+        if (panelInfo.isEmpty()) {
             addPanel();
             return;
         }
 
-        for (String filter : filters) {
-            addPanel().filterByString(filter);
+        for (int i = 0; i < panelInfo.size(); i++) {
```
dariusf: Can replace with `forEach`.

dariusf:
Do squash noisy commits when you are done. Try to keep the abstraction level of commit messages high by describing features rather than individual changes.

src/main/java/ui/issuepanel/FilterPanel.java
HyungJon: The ```mainStage.show()```? That is to ensure the focus goes back to the main pView, so that Travis can work properly.

dariusf: Use an event then. Less coupling that way.

Does this happen for other instances of `TextInputDialog::showAndWait`? If it's absolutely necessary, you could add a comment there explaining why it's needed.

It might also be a good idea to factor the mouse click callback out into a private method (SLAP).

src/main/java/prefs/GlobalConfig.java
dariusf: On second thought, these being here makes the tests DRYer. Never mind then. Just replace the for loops with `map`s.

src/main/java/ui/issuepanel/FilterPanel.java
HyungJon: It is used with functions that bring out new dialogs, like ```LabelPicker``` because Travis fails to go back to the main stage after it's done with the dialog. I'll add comments to indicate this.

If events are used, I can move the event handler and ```mainStage.show()``` to ```PanelControl```. Is this okay in terms of coupling?

dariusf: I would put it in `UI`. The main use of the stage reference outside `UI` is to show or focus it; such uses should be replaced with the new event. Ideally we should abstract it away from child components.

For complicated cases like `Dialog` (which was written before we started using the new dialogs), it's okay to leave it there. It can be changed someday but it's not a pressing need.

HyungJon: I would actually like to move it to ```PanelControl``` rather than leaving it in ```FilterPanel``` because I did try using event handler here before, and found a dialog will pop for each panel instead of just the one I want to rename. I could fix it only by moving it to ```PanelControl```.

src/main/java/prefs/GlobalConfig.java
HyungJon: Will revert. Could I know why ```map``` is recommended here rather than ```foreach```? I'm not sure how to replace it with map.

HyungJon:
Oh... I'm having problems with squashing again. I found some noisy commits to squash and did a rebase, but I suspect it somewhat messed up the repository. My local branch is said to be 111 commits ahead of my remote, and I think my attempts to fix it caused more trouble.

I'm not sure if this can be fixed. Also, can I know how to squash the commits properly?

damithc:
Squashing commits is a general thing, not specific to HubTurbo. See if there are web resources explaining how to do it. If your local repo got messed up and can't figure out how to undo it, just clone from your fork into a new place?

src/main/java/prefs/GlobalConfig.java
lejolly: it allows you to do this:

```java
public List<String> getPanelNames() {
    return openPanels.stream()
        .map(PanelInfo::getPanelName)
}
```

HyungJon:
It would seem I didn't properly understand the instructions I found online.

I will redo the required improvements from another cloned repo then. Alternatively, can this PR be merged if there are no big changes required, and I make the required improvements in another issue?

damithc:
We have to do the code improvements in this PR itself. But you can omit the squashing this time if you are not comfortable with it yet.

ndt93:
Squashing commits after it has been pushed to remote branches could be messy and dangerous (luckily for yourself only). We will need to use this: `git push origin +branch_name` or this `git push --force origin branch_name` if the squashed commits have been pushed to remote before.

This screwed me up plenty of times in the past. Can be fixed without cloning but I forget the steps to do it already. May be research a bit more?

HyungJon:
The biggest difficulty I had both times I messed up my repo was the rebasing adding a large number of commits, by others included, instead of just squashing the ones I picked. I'll look into squashing commits on open projects later. I'll continue with the improvements now.

damithc:
Seems Source tree can do it on the GUI. Never tried it before though https://answers.atlassian.com/questions/213902/in-sourcetree-how-do-i-squash-commits

dariusf:
Don't worry, I checked out your branch to review it, and here's the SHA: 9da0469. If you're still having trouble, `git reset --hard 9da0469` will bring everything back to how it was right before I commented.

src/main/java/ui/issuepanel/FilterPanel.java
dariusf: Yes, don't leave it in `FilterPanel`, but put it in `UI` instead of `PanelControl`. There's also no need for the latter to know about the stage: it's only concerned with dispatching focus events for something to pick up on.

src/main/java/prefs/GlobalConfig.java
dariusf: The idea is that `map` communicates more precisely what the function is doing: transforming a collection into a new collection of the same size. A `for` loop is more general and has more potential for mistakes, say an off-by-one error due to a typo that goes unnoticed.

lejolly:
@HyungJon that mockup was done with just adding and removing children from `nameArea`

HyungJon:
I implemented rename function that uses in-place ```TextField```. Clicking on panel name brings out the rename textfield. Pushing Esc cancels the rename function.

I also changed the way lengthy names are dealt with: instead of cutting off excess characters, the textfield will now refuse any more characters if the length limit has been reached. Since no dialog is used ```mainStage.show()``` won't be needed, so I moved event handler back to ```FilterPanel```. I'll add more tests once the functionality is approved.

lejolly:
1. i think the edit button should be kept there
2. the activation area should be wider rather than just on the panel name itself
3. the caret after activation should be at the end of the name or the whole name should be selected
4. are you stripping out extra spaces in the name (either at the start/end)
5. text field should be the same width as the filter box?
6. y position of the ui elements below the text field shouldn't move after activation
7. not sure about this one, but maybe doubleclick is better for preventing unwanted activations?

also try to continue including screenshots either by updating the top post or including it in future comments.

HyungJon:
I don't experience elements moving up/down when text field comes up. I'll see if I can prevent it.

lejolly:
ah could be an OS X issue then. just increase the padding?

HyungJon:
Does Travis not use ```doubleClick``` properly? It's failing to rename the first panel.

HyungJon:
Okay, Travis build failure resolved by just calling the event. I will now add a few more tests including some corner cases.

lejolly:
1. i think the edit button is a bit too close to the close button now. 
2. the edit button should not open the rename dialog but rather behave as if i doubleclicked on the panel name. 
3. there's still a slight shift of the ui elements on OS X when activating the panel rename
4. maybe remove the edit button when in panel renaming mode unless there's a good reason to keep it there, e.g. clicking on it accepts the panel rename

HyungJon:
Is it okay if I either make the ```FilterTextField``` longer to accommodate the entire panel width since it no longer has the close button to its right? And make ```PanelRenameTextField``` a bit shorter than that.

lejolly:
sure, but do post a screenshot of it so that others can comment too.

HyungJon:
This is how the panel would look normally.
cap 2015-08-11 19-55-16-751

This is how it looks when renaming.
cap 2015-08-11 19-55-37-812

lejolly:
can we align the panel name with the icons? maybe increase the font size while we're at it.

HyungJon:
cap 2015-08-11 21-29-16-891
cap 2015-08-11 21-29-20-078
Is this okay now? I felt I should keep the textfield font size as it is, because otherwise the panel name bar will have to be rather big to prevent Y-axis movements of elements. Problem is that the maximum name length is like half the length of the textfield though.

lejolly:
@damithc how many characters should we cater for the panel name?

@HyungJon how many characters can we fit for the smaller font size and this larger one?

damithc:
As many as we can fit in the panel :-)

HyungJon:
Smaller font accommodates about 36 characters, larger font 24. If I add a function to set maximum based on character width rather than count, probably up to about 1.5 times more for each.

damithc:
For now, let's go with 36. We can optimize later.

HyungJon:
All right. Reverting to smaller font.

HyungJon:
I'll look into merging ```PanelInfoBasicTest``` with ```UseGlobalConfigsTest``` since the scope of the test largely overlaps with it. The new one I just pushed, ```PanelRenameTest```, is for corner cases specific to renaming.

HyungJon:
Random NoNodesVisibleException again. I seem to be the only one getting this, wouldn't it cause problems for others if this PR is merged as it is?

lejolly:
if the exception is not related to the pr then you don't have to fix it, if however you want to merge your tests with `UseGlobalConfigsTest` then the resulting test needs to pass at least most of the time.

HyungJon:
Will work on fixing both.

lejolly:
please do, since `guitests.UseGlobalConfigsTest > globalConfigTest` has failed 3 times in a row on travis.

HyungJon:
Why would it be failing on this PR only? Also because it might point to a solution.

lejolly:
probably because you merged `PanelInfoBasicTest` into it?

HyungJon:
Failures started happening before I merged. I suspect something went wrong with creating new panels (more specifically creating the elements on it) because ```NoNodesVisibleException``` happens right after panel creation in both tests. Fixing.

HyungJon:
Requesting review. I fixed the ```NoNodesVisibleException```.

src/main/java/ui/issuepanel/PanelControl.java
```diff
@@ -39,6 +45,8 @@ public PanelControl(UI ui, Preferences prefs) {
                 setCurrentlySelectedPanel(Optional.of(e.panelIndex)));
         ui.registerEvent((PanelClickedEventHandler) e ->
                 setCurrentlySelectedPanel(Optional.of(e.panelIndex)));
+        ui.registerEvent((ShowRenamePanelEventHandler) e -> Platform.runLater(() ->
+                ((FilterPanel) getPanel(e.panelId)).showRenameTextField()));
```
dariusf: I don't think there a need for `Platform.runLater`. `ShowRenamePanelEvent` is never triggered from a thread other than the UI thread, right?

src/main/java/ui/issuepanel/PanelControl.java
```diff
@@ -105,6 +113,7 @@ public FilterPanel addPanelAt(int index) {
 
         updatePanelIndices();
         setCurrentlySelectedPanel(Optional.of(index));
+        mainStage.show();
```
dariusf: Is this still needed? If not, `PanelControl` no longer needs to know about the stage.

src/main/java/ui/issuepanel/PanelControl.java
```diff
@@ -173,6 +182,28 @@ public void swapPanels(int panelIndex, int panelIndex2) {
...
+        String newName = result.orElse(panelName);
+        if (newName.equals("")) {
+            newName = panelName;
+        }
+        if (newName.length() > MAX_NAME_LENGTH) {
+            newName = newName.substring(0, MAX_NAME_LENGTH);
+        }
+        panel.renamePanel(newName);
+    }
+    **/
```
dariusf: This can be removed.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -67,6 +85,76 @@ public FilterPanel(UI ui, IModel model, PanelControl parentPanelControl, int pan
...
+        nameBox.getChildren().add(nameText);
+        nameBox.setMinWidth(330);
+        nameBox.setMaxWidth(330);
+        nameBox.setAlignment(Pos.CENTER_LEFT);
+        nameBox.setOnMouseClicked(new EventHandler<MouseEvent>() {
+            @Override
+            public void handle(MouseEvent mouseEvent) {
+                if (mouseEvent.getButton().equals(MouseButton.PRIMARY)) {
+                    if (mouseEvent.getClickCount() == 2) {
+                        mouseEvent.consume();
```
dariusf: This can be replaced with a lambda.

Also some nesting can be removed from the conditional:

```java
boolean isDoubleClick = ... && ...;
if (isDoubleClick) {
    ...
}
```

In future we can factor this out further maybe, since double clicks are fairly commonplace.

src/main/java/ui/components/PanelNameTextField.java
```diff
@@ -0,0 +1,55 @@
...
+import javafx.application.Platform;
+import javafx.scene.control.TextField;
+import javafx.scene.input.KeyCode;
+import javafx.beans.value.ObservableValue;
+import javafx.beans.value.ChangeListener;
+
+public class PanelNameTextField extends TextField {
+
+    private final FilterPanel panel;
+    private final int PANEL_MAX_NAME_LENGTH = 36;
```
dariusf: Should be static.

src/main/java/ui/components/PanelNameTextField.java
```diff
@@ -0,0 +1,55 @@
...
+        textProperty().addListener(new ChangeListener<String>() {
+            @Override
+            public void changed(ObservableValue<? extends String> observableValue, String s, String s2) {
+                // Prevent excessive characters
+                if (getText().length() > PANEL_MAX_NAME_LENGTH) {
+                    setText(previousText);
+                }
+                previousText = getText();
+            }
+        });
```
dariusf: Can be replaced with a lambda.

src/main/java/ui/issuepanel/PanelControl.java
HyungJon: This is to fix the ```NoNodesVisibleException``` caused when adding a new panel. I guess there is a way around ```PanelControl``` calling the stage, by creating a method in ```UI``` for showing mainstage.

dariusf: Alright, that's a good idea. We can document the `NoNodesVisibleException` quirkiness there.

src/main/java/ui/issuepanel/PanelControl.java
HyungJon: The tests do trigger it, to prevent ```NoNodesFoundException``` on Travis. I think with ```mainStage.show()``` moved to UI, the event handler can likely be removed again without causing the exception. Shall I do it?

dariusf: Yes, but `ShowRenamePanelEvent` is only triggered via clicks, from the UI thread. Removing `Platform.runLater` should be fine.

Which event handler are you referring to?

HyungJon: ```ShowRenamePanelEvent``` is called by ```UseGlobalConfigsTest``` and ```PanelRenameTest``` as well. Removing ```runLater``` resulted in those tests failing to show the rename text field.

dariusf: In that case, `PlatformEx.runAndWait` should be used at those places. We should try to keep application code agnostic to what is done in tests.

```java
// Use GlobalConfigsTest.java
PlatformEx.runAndWait(() -> UI.events.triggerEvent(new ShowRenamePanelEvent(0)));
type("Renamed panel");
```

HyungJon: Done.

dariusf:
Coverage seems to be nondeterministic...

Anyway, looks good. @lejolly

lejolly:
Looks fine to me too, @damithc?

src/test/java/guitests/PanelInfoEmptyTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.loadui.testfx.utils.FXTestUtils;
+
+import prefs.Preferences;
+import prefs.PanelInfo;
+
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+public class PanelInfoEmptyTest extends UITest {
```
damithc: Can use a more self-explanatory class name? If the name is not self-explanatory enough, add a header comment to the class.

src/test/java/guitests/PanelRenameTest.java
```diff
@@ -0,0 +1,120 @@
...
+    @Test
+    public void panelRenameTest() {
+        
+        // Test for saving panel information with corner cases
+        
+        // maximize
+        press(KeyCode.CONTROL).press(KeyCode.X).release(KeyCode.X).release(KeyCode.CONTROL);
+        sleep(EVENT_DELAY);
+
+        // Testing rename cancel
```
damithc: Comments should be "well written". Comments like this are a bit too cryptic for an outsider. Try to improve. e.g. `// Rename, but press ESC midway to cancel`

src/test/java/guitests/PanelRenameTest.java
```diff
@@ -0,0 +1,120 @@
...
+        PlatformEx.runAndWait(() -> UI.events.triggerEvent(new ShowRenamePanelEvent(0)));
+        push(KeyCode.ESCAPE);
+        sleep(EVENT_DELAY);
+        
+        press(KeyCode.CONTROL).press(KeyCode.P).release(KeyCode.P).release(KeyCode.CONTROL);
+
```
damithc: I'm not sure what verification is done here to confirm that the behavior is as expected.

src/test/java/guitests/PanelRenameTest.java
```diff
@@ -0,0 +1,120 @@
...
+        push(KeyCode.BACK_SPACE);
+        push(KeyCode.ENTER);
+        sleep(EVENT_DELAY);
+        
+        press(KeyCode.CONTROL).press(KeyCode.P).release(KeyCode.P).release(KeyCode.CONTROL);
+
+        // Testing typing excessive panel name
+        PlatformEx.runAndWait(() -> UI.events.triggerEvent(new ShowRenamePanelEvent(3)));
+        sleep(EVENT_DELAY);
+        type("1234567890123456789012345678901234567890");
```
damithc: How long is this and why was that value chosen? Is it a boundary value? If not, it should be. If the max length is 20, you should test 19,20,21. It may be useful to define a utility method that can give you a random string of a specified length. e.g. `StringHelper.getRandomString(PANEL_NAME_MAX_LENGTH+1)`

src/test/java/guitests/UseGlobalConfigsTest.java
```diff
@@ -56,6 +70,31 @@ public void globalConfigTest() {
...
         push(KeyCode.ENTER);
+
+        click("#dummy/dummy_col1_renameButton");
+        type("Dummy 2 panel");
+        PanelNameTextField renameTextField2 = find("#dummy/dummy_col1_renameTextField");
+        assertEquals("Dummy 2 panel", renameTextField2.getText());
+        push(KeyCode.ENTER);
+        
+        // Creating panel to the left
+        press(KeyCode.SHIFT).press(KeyCode.CONTROL).press(KeyCode.P);
```
damithc: It may be good to abstract actions like this to utility methods. e.g. `createPanelOnLeft()` which makes the code easier to read and maintain.

src/test/java/guitests/PanelRenameTest.java
```diff
@@ -0,0 +1,120 @@
...
+        assertEquals(7, openPanelNames.size());
+
+        assertEquals("Panel", openPanelNames.get(0));
+        assertEquals("Renamed panel", openPanelNames.get(1));
+        assertEquals("Panel", openPanelNames.get(2));
+        assertEquals("123456789012345678901234567890123456", openPanelNames.get(3));
+        assertEquals("Panel", openPanelNames.get(4));
+        assertEquals("123456789012345678901234567890123456", openPanelNames.get(5));
+        assertEquals("123456789012345678901234561234567890", openPanelNames.get(6));
+    }
```
damithc: Can the code handle json in the old format?

src/test/java/guitests/PanelRenameTest.java
```diff
@@ -0,0 +1,120 @@
...
+        renameTextField3.copy();
+        renameTextField3.positionCaret(20); // between second 0 and 1
+        type("1234567890");
+        push(KeyCode.ENTER);
+        sleep(EVENT_DELAY);
+        
+        // Quitting to update json
+        click("File");
+        push(KeyCode.DOWN).push(KeyCode.DOWN).push(KeyCode.ENTER);
+        sleep(EVENT_DELAY);
```
damithc: This can be abstracted into a `quit()` method.

damithc:
Some cosmetic suggestions added.

src/test/java/guitests/PanelRenameTest.java
HyungJon: There should be no problem, since saving to json is done at quit. The json will be changed to the new format before the checks. I decided to remove json checks in this test since the scope is different though.

HyungJon:
Is an updater that updates json files to new format required? The current version won't result in any errors, but it does ignore any previously saved data and has to start fresh. 

I have already prepared a basic updater method, but I think if an updater is added, it should be a standalone feature that stores and reads version number from json files, checks it at startup and update the json if outdated.

damithc:
Let's say a person already has the current version of HT and he auto-upgrades to this version (HT updates itself automatically). Can the user continue to use HT with minimum disruption? Note that we don't know who is using HT and we currently don't have a way to contact our users.

HyungJon:
No serious disruptions like failure to start, but previously saved panels and boards will be wiped, which can be a problem since users can't be warned.

damithc:
I guess that's good enough, as we don't have many users at the moment.

HyungJon:
Understood. I'll just fix this merge conflict and make sure the Travis build failure is random, then push again.

damithc:
Let's wrap this up quickly.

HyungJon:
Yes. Anything else to be done?

lejolly:
nope, except to merge with the latest master.

HyungJon:
Sorry, didn't realize it meant merging master into this PR and not other way round. Here's the new push.

PR #823, opened 2015-07-24
Add divider between label groups #786

### LINKED ISSUE #786: Add  divider between label groups
... to make it look more organized. A blank line or a subtle divider line?
image

lejolly:
Fixes #786

edit_labels_for_pr__823_in_hubturbo_hubturbo


damithc:
Make the group name bold? Or do some other thing to make it stand out a bit more.
Indent the labels of a group a bit? Currently group name and labels are indented at the same level.
How to show labels without a group?

damithc:
May be show the dot/dash as well, because that indicates nature of the group?

lejolly:
something like that?

edit_labels_for_issue__10_in_dummy_dummy

damithc:
Yes, but I prefer 'p.' rather than 'p (exclusive)'. Shorter and the meaning
of the dot is something the user knows already.

damithc:
Put ungrouped ones at the bottom rather than on the top? I feel the
ungrouped ones are the less important and less used ones.

damithc:
And a tiny bit more space between groups?

lejolly:
do we still need to indent the group labels?

edit_labels_for_issue__10_in_dummy_dummy

damithc:
Looks good. Yes, let's keep the indentation.

lejolly:
ok

damithc:
For toggle.png, can use a more representative string e.g. `hi d.b u`
It is unlikely that we can use just a sequence of single letters to select the labels we actually want.

acruis:
Great refactoring! The label picker is a lot clearer now. Good to merge.

Issue #821, opened 2015-07-24
json explosion false positive for large repos


damithc:
Tried to download rails/rails. It triggered json explosion detection.
```
69091573 24 Jul 2015 10:57:42 INFO UI [pool-4-thread-56] - Triggered event UpdateRateLimitsEvent
69092626 24 Jul 2015 10:57:43 INFO GitHubRepo [pool-4-thread-44] - rails/rails | Unable to get collaborators: Must have push access to view repository collaborators. (403)
69092626 24 Jul 2015 10:57:43 INFO DownloadRepoTask [pool-4-thread-44] - rails/rails | Downloaded 20989 issue(s), 30 label(s), 42 milestone(s), 0 user(s)
69092626 24 Jul 2015 10:57:43 INFO UpdateService [pool-4-thread-58] - Updating rails/rails/collaborators with ETag null
69092626 24 Jul 2015 10:57:43 INFO UpdateService [pool-4-thread-60] - Updating rails/rails/milestones with ETag null
69092627 24 Jul 2015 10:57:43 INFO UpdateService [pool-4-thread-44] - Updating rails/rails/issues with ETag null
69092627 24 Jul 2015 10:57:43 INFO UpdateService [pool-4-thread-62] - Updating rails/rails/labels with ETag null
69092912 24 Jul 2015 10:57:44 WARN UpdateService [pool-4-thread-58] - UserUpdateService: error getting updated items (403)
69092912 24 Jul 2015 10:57:44 INFO UpdateUsersTask [pool-4-thread-58] - rails/rails | 0 user(s)) changed
69092930 24 Jul 2015 10:57:44 INFO UpdateService [pool-4-thread-62] - rails/rails/labels: 200
69093025 24 Jul 2015 10:57:44 INFO UpdateService [pool-4-thread-60] - rails/rails/milestones: 200
69093190 24 Jul 2015 10:57:44 INFO UpdateService [pool-4-thread-44] - rails/rails/issues: 200
69093257 24 Jul 2015 10:57:44 INFO UpdateService [pool-4-thread-62] - rails/rails/labels | page 0: 30 items
69093257 24 Jul 2015 10:57:44 INFO UpdateService [pool-4-thread-62] - New ETag for rails/rails/labels: Optional[2899ed6a07e2f7beb371c398c3d64398]
[...]

PR #824, opened 2015-07-24
json explosion false positive for large repos #821

acruis:
Does not fully finish #821.


lejolly:
waiting for the download count limit to be added

acruis:
Fixes #821

damithc:
@lejolly can review this quickly?

lejolly:
Do we want to show an error message when we reach the download limit without being able to successfully download the repo?

damithc:
Yes, that would be better. Just a simple, 'could not download, it is unexpected, please let us know if you encounter this' kind of message.

lejolly:
ok, looks good.

PR #828, opened 2015-07-24
Add support for 64 bits Linux chrome driver #826

### LINKED ISSUE #826: Additional steps for developers on 64 bits Linux
The provided chromedriver_linux is 32 bits and doesn't work on a typical 64 bits Linux installation. So we either add a note in the developer/user guide for installing 32 bits support on 64 bits Linux (which is quite painful :P since in addition to installing base architecture support, a bunch of 32 bits libraries dependencies for chromedriver are also needed, most of which are not easily obtained and configuration steps differ between distributions).

Or we can provide support in HubTurbo using something like this to check for os architecture:
```java
Runtime.getRuntime().exec("uname -m");
```

I had this problem from the beginning and used the former method to resolve it but haven't mentioned it since I thought Linux is not supported. But now, it's mentioned in the developer's guide so well :P. Any opinion from the dev team?

ndt93:
Fixes #826 

Tested on both Windows and 64 bits Debian. Travis CI is also using the 64 bits linux chrome driver now so that's also another confirmation.

*Line 203* in `util.Utility` will screw up and hang on Ubuntu's Unity desktop but that's a java swing bug, not really our problem although we might figure out a way around it (if anyone wants to run on Unity then just comment the line out).


ndt93:
The freeze on Ubuntu's Unity is caused by a permission problem for getting the native gtk theme so I added a workaround. This PR should be ready for review.

lejolly:
looks good, do you think we should rename the chromedriver executables to include the version number, or include it somewhere else? (ideally it would be a gradle dependency)

ndt93:
Yup, I think it would be nice to have the version number with the file. But if we want to integrate with gradle, there is a complication since the file names are tied to the code at the moment so the program needs to know when gradle updates them to a new version. I can change the names here but if we want something more flexible (include it somewhere external) then I think this should be a new issue.

lejolly:
yup, let's leave the gradle integration to a new issue.

ndt93:
I renamed the files with the version number. @lejolly Can help me test if everything is working fine on Mac OS and also Eclipse? Seem like I have to clean the project for the files with the new names to be copied over to the *bin/* folder.

And also, I haven't gotten any CI test failed (Node not visible) after the Ubuntu Unity fix. Not sure if it helps :P (since Travis uses Ubuntu although it's a virtual X server, not  Unity)

lejolly:
@ndt93 yup it works for me on a fresh clone of the branch on OS X. if it works fine on windows too then please merge the latest master into the branch and i'll merge the pr.

ndt93:
Ok merged. I've tested on both windows and linux. No failed test on first build also. But I would need some more for tests for the coveralls thing.

ndt93:
Anyone who is familiar with *coveralls* can show me why public `determineChromeDriverBinaryName`'s branches in *BrowserComponent.java* are not considered to be covered by `testDetermineChromeDriverBinaryName` in `PlatformSpecificTests.java` and how to cover them? And also how to cover the *catch clauses* in `getLinuxKernelArchitecture` in *PlatformSpecific.java*?

Those are the only lines left I wrote for this PR that are considered failed by *coveralls*.

acruis:
For `BrowserComponent`, do you mean the `isOnMac`, `isOnWindows`, `isOn32BitsLinux` branches? They will never return true as they don't match with Travis' environment unless you use a `PlatformSpecific` stub to force the branches to execute.

As for `getLinuxKernelArchitecture`, the `catch` clause is only covered if an `IOException` occurs during the `try` block. I doubt you can force exceptions at this point without a stub as well.

I think it's fine if we merge this branch as is, but do pull from master and solve the merge conflicts first.

ndt93:
Ah ok, I thought coveralls does some kinds of static analysis only. Will merge later tonight.

ndt93:
Merged and tested on Windows and Linux.

lejolly:
ok, looks good.

PR #840, opened 2015-07-28
keyword search broken for large repos? #838

### LINKED ISSUE #838: keyword search broken for large repos?
When trying with homebrew, keyword search does not seem to narrow down search results. e.g. `is:open` and `is:open fix` seem to give the same result set.

acruis:
Fixes #838

### DIFF
--- a/src/main/java/backend/resource/TurboIssue.java
+++ b/src/main/java/backend/resource/TurboIssue.java
@@ -116,12 +116,16 @@ public class TurboIssue {
 
     public TurboIssue(String repoId, Issue issue) {
         this.id = issue.getNumber();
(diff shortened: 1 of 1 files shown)


dariusf:
Maybe it would be better to fix this at the model level? By the time we get to the level of TurboIssue objects, the null fields should have default values.

damithc:
Yup, a lower level fix is better. Rest of the code should be able to operate as if this inconsistency does not exist.

acruis:
Alright, got it.

dariusf:
Looks good. Would be great if you could squash this (the original commit is no longer relevant).

PR #847, opened 2015-07-30
F1 should go to UserGuide.md #809

### LINKED ISSUE #809: F1 should go to UserGuide.md
Currently it goes to GettingStarted.md. This was ok when it was in the wiki because every page had the navigation menu. But now, UserGuide.md makes more sense.

acruis:
Fixes #809


damithc:
Seems simple enough to merge without review. On a related note, is there a way to test these links?. These things can break without any of us noticing.

acruis:
We can do this based on the repo structure, so that when the userGuide file doesn't exist, a test will fail.

damithc:
test comment

PR #848, opened 2015-07-30
Chrome keeps blocking me #810

### LINKED ISSUE #810: Chrome keeps blocking me
The chrome window sometimes still stays in the foreground when I try to switch to HubTurbo.

acruis:
Fixes #810 

`user32.ShowWindow` doesn't do anything, apparently.

### DIFF
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -41,6 +41,7 @@ public class BrowserComponent {
 
     private static final int SWP_NOSIZE = 0x0001;
     private static final int SWP_NOMOVE = 0x0002;
+    private static final int SWP_NOACTIVATE = 0x0010;
     private static HWND browserWindowHandle;
     private static User32 user32;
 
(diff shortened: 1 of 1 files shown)


damithc:
Didn't break anything on my machine, so no regressions. Were you able to reproduce the bug and confirm this fix works?

acruis:
Yes, it's a valid issue. To reproduce:

- Open HubTurbo
- Manually expand the HubTurbo window (clicking on Maximize doesn't trigger this bug) like the following:

bugonchrome

- Navigate to a different window, such that HubTurbo loses focus
- Bring focus back to HubTurbo again (click on the icon or Alt Tab back to the window)

HubTurbo will be covered by, and lose focus to, the bView Chrome window a majority of the time. Prior to this PR, after bringing the Chrome window to the front, the program manually sets focus back to HubTurbo again, which once again tries to bring the Chrome window to the front, and so on, resulting in an infinite loop that probably gets terminated by Windows itself. Something like this:

Focus on HT -> HT brings Chrome window to front (loses focus) -> HT manually focuses back on itself -> repeat

The fix changes the second step such that HT brings Chrome window to front without losing focus on itself.

damithc:
Ok, go ahead and merge if the tests pass.

PR #849, opened 2015-07-31
Keep track of ETags for issues #778

### LINKED ISSUE #778: Keep track of ETags for issues
To use when getting issue comments and events. Helps with API usage (related to #765).

acruis:
Fixes #778.

Implementation is complete, but since this implementation deals with the GitHub API, testing stubs have to undergo an overhaul to properly reflect the changes.


src/main/java/backend/IssueMetadata.java
```diff
@@ -95,6 +102,10 @@ public boolean isUpdated() {
         return isUpdated;
     }
 
+    public String getIssueETag() {
```
dariusf: `getETag` might be sufficient as a name maybe? It's already a metadata method.

src/main/java/backend/Logic.java
```diff
@@ -127,9 +127,9 @@ public void refresh() {
         });
     }
 
-    public CompletableFuture<Boolean> getIssueMetadata(String repoId, List<Integer> issues) {
+    public CompletableFuture<Boolean> getIssueMetadata(String repoId, Map<Integer, String> issues) {
```
dariusf: This (issue id, ETag) map is used a lot throughout. Could we make this a class? Alternatively we could pass a collection of (copied) `TurboIssue`s around, but I'm not sure if the copy could still be affected by mutation of the originals.

src/main/java/github/IssueEventType.java
```diff
@@ -6,6 +6,7 @@
     Closed,
     Reopened,
     Subscribed,
+    Unsubscribed,
```
dariusf: This was missing the whole time?

acruis: I know, right?

src/main/java/backend/IssueMetadata.java
acruis: I wanted to make sure it won't be taken to mean one of the main 4 ETags in `UpdateSignature` (`issuesETag`, `milestonesETag`, `labelsETag`, `collaboratorsETag`). Is the method name a bit too redundant though?

src/main/java/backend/Logic.java
acruis: The copy would probably be affected if we're concerned about ETags (which we are), so if the `TurboIssue`s should be copied, then there should be a deeper copy mechanism. Why don't you want the map, though? Too verbose?

dariusf: It's more that it's not as clear as it could be, to someone just looking at the method signature in isolation.

The metadata that the issue points to can change, but the metadata object itself is immutable, so it wouldn't change in a shallow copy. I'm more concerned about the `labels` field, which is mutated (via `addLabel`) and not always copied.

The best approach seems to be to fix the mutability issues and reuse TurboIssues, rather than make another class. What do you think?

src/main/java/backend/IssueMetadata.java
dariusf: I feel it is, unless there can be more ETags in metadata in future. We'll need the metadata instance to call it, so it shouldn't be too confusing.
```java
issue.getMetadata().getETag()
model.getUpdateSignature().issuesETag
```

acruis: Good point. I'll change it to `getETag`.

src/main/java/backend/Logic.java
acruis: I'm all for immutables, but just to confirm, would mutating the labels field have an impact on metadata retrieval logic?

dariusf: In this case it doesn't, but being sure that nothing will change an object after it's copied is a good guarantee to have. Otherwise we'll need more ad-hoc ways to get around the inability to treat TurboIssues as values.

acruis: Ah, alright. I'll pass the TurboIssues around, but I think making TurboIssues immutable should be in a different PR as it touches on the labelpicker / undo functionality.

dariusf: Okay, I made an issue for this.

dariusf:
Needs a rebase, but otherwise you can merge this.

acruis:
Ah alright

acruis:
@dariusf: Can merge? Or do you think the merges I made are nontrivial?

dariusf:
Can merge. Somehow Coveralls isn't updating GitHub, but going over there shows that coverage has increased.

Issue #812, opened 2015-07-23
Repo field of login dialog is case sensitive


acruis:
1. Open `TEAMMATES/repo` in HubTurbo
2. Logout from HubTurbo (File -> Logout)
3. Start HubTurbo again.
4. Specify `teammates/repo` in the login dialog

In step 4, HubTurbo will attempt to redownload teammates/repo even though `TEAMMATES-repo.json` already exists in store.

dariusf:
This problem occurs anywhere a repo id is used as a key.

I guess the long-term solution would be a `RepoId` class with `hasCode` that normalises differences in case. We may want to make our own instead of using the GitHub version in case we want to add methods.

PR #851, opened 2015-08-05
Remove white spaces from input repo's id in repo selector and login dialog #843

### LINKED ISSUE #843: Extra space in repo dropdown breaks HT
image

Should be trimmed before using. People can easily add extra spaces without realizing and the mistake is hard to spot too.

ndt93:
Fixes #843


acruis:
Okay, can go ahead and merge.

On a side note though, as mentioned in #812, things related to the RepoID should probably be migrated into a class to encapsulate case checking, space removal etc.

PR #855, opened 2015-08-06
Specify charset when reading in local files #854

### LINKED ISSUE #854: Specify charset when reading in local files
We explicitly write data to file using UTF-8, as seen in `Utility.writeFile` which initializes the printwriter as `new PrintWriter(fileName, "UTF-8")`. However, we read data from files [using the platform's default charset](http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#String(byte[])) in `Utility.readFile` where we use `new String(Files.readAllBytes(new File(filename).toPath()))`.

Fixing this may solve #706.

acruis:
Fixes #854

* Before:
before

* After:
image

### DIFF
--- a/src/main/java/util/Utility.java
+++ b/src/main/java/util/Utility.java
@@ -52,7 +52,7 @@ public class Utility {
 
     public static Optional<String> readFile(String filename) {
         try {
-            return Optional.of(new String(Files.readAllBytes(new File(filename).toPath())));
+            return Optional.of(new String(Files.readAllBytes(new File(filename).toPath()), "UTF-8"));
         } catch (IOException e) {
             logger.error(e.getLocalizedMessage(), e);
         }


lejolly:
looks good, any way to reproduce the before situation?

acruis:
Does it display properly for you even without the `UTF-8` in the String constructor call?

lejolly:
yup, on OS X.

acruis:
Do you ever run into the json explosion problem?

dariusf:
`Charset.defaultCharset()` on Yosemite returns `UTF-8`. Maybe the defaults let us get away with not specifying it on OS X. In any case it's good to be explicit about this.

acruis:
Yeah, the default on OSX being `UTF-8` was what I expected. It's `windows-1252` for me.

Anyway, time to merge!

PR #857, opened 2015-08-07
ESC should remove focus from repo dropdown

### LINKED ISSUE #842: ESC should remove focus from repo dropdown
Currently, there is no way keyboard shortcut to remove focus from the repo dropdown. TAB can be used but it moves focus to the filter box.

rd1992:
Fixes #842

### DIFF
--- a/src/main/java/ui/RepositorySelector.java
+++ b/src/main/java/ui/RepositorySelector.java
@@ -3,7 +3,9 @@ package ui;
 import javafx.geometry.Insets;
 import javafx.geometry.Pos;
 import javafx.scene.control.ComboBox;
+import javafx.scene.input.KeyEvent;
 import javafx.scene.layout.HBox;
+import ui.components.KeyboardShortcuts;
 import util.Utility;
 
 import java.util.List;
@@ -47,6 +49,14 @@ public class RepositorySelector extends HBox {
 
             if (Utility.isWellFormedRepoId(repoId) && !changesDisabled) {
                 onValueChangeCallback.accept(repoId);
+            }
+        });
+
+        //remove focus from the repo selector
+        comboBox.addEventFilter(KeyEvent.KEY_RELEASED, event -> {
+            event.consume();
+            if (event.getCode() == KeyboardShortcuts.REMOVE_FOCUS) {
+                getParent().requestFocus();
             }
         });
     }
--- a/src/test/java/guitests/KeyboardShortcutsTest.java
+++ b/src/test/java/guitests/KeyboardShortcutsTest.java
@@ -1,7 +1,10 @@
 package guitests;
 
+import javafx.scene.control.ComboBox;
 import javafx.scene.input.KeyCode;
+
 import org.junit.Test;
+
 import ui.UI;
 import ui.components.KeyboardShortcuts;
 import ui.listpanel.ListPanel;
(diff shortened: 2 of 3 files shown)


damithc:
@lejolly do we have UI tests for keyboard shortcuts?

damithc:
Any update to docs required?

lejolly:
yup in `guitests.KeyboardShortcutsTest`

damithc:
Got it @lejolly. 
@rd1992 pls update tests and docs (if required).

rd1992:
@lejolly  could you review?

lejolly:
please remove error screenshots from commits.

lejolly:
could you clean up the commit history too?

src/main/java/ui/RepositorySelector.java
```diff
@@ -49,6 +52,17 @@ private void setupComboBox() {
                 onValueChangeCallback.accept(repoId);
             }
         });
+        
+        // remove focus from the repo selector
+        comboBox.addEventFilter(KeyEvent.KEY_RELEASED, event -> {
+            event.consume();
+            if (event.getCode() == KeyboardShortcuts.REMOVE_FOCUS) {
+                if (ui.isTestMode()) {
```
lejolly: why test mode?

rd1992:
@lejolly whoops. sorry this was done at 5 am so dint realize what i had committed. I'll clean up the history.

src/main/java/ui/RepositorySelector.java
rd1992: Because i don't want to trigger the event if its not in test mode.

lejolly:
please ensure that it also passes travis (at least most of the time).

src/main/java/ui/RepositorySelector.java
rd1992: plus we do the same thing [here] (https://github.com/HubTurbo/HubTurbo/blob/master/src/main/java/ui/listpanel/ListPanel.java#L328)

lejolly: hmm yeah although now that i think about it, could we check if the repo selector itself has focus instead of checking for the event?

rd1992: hmm the thing is we are _removing_ focus. Plus to keep consistent with other event related testing, shouldn't we stick to this approach? Or we switch event based testing to your suggestion

lejolly: if we're able to check for focus of the repo selector before and then no focus after the key has been pressed, i think that's a better way to do it. we can create a separate issue to change the existing tests.

rd1992:
@lejolly ok cleaned up the history. Also changed the tests. passes the CI build also. Could you review?

src/test/java/guitests/KeyboardShortcutsTest.java
```diff
@@ -96,6 +98,13 @@ public void keyboardShortcutsTest() {
         assertEquals(1, panelIndex);
         clearPanelIndex();
 
+        // remove focus from repo selector
+        ComboBox<String> comboBox = find("#repositorySelector");
+        doubleClick(comboBox);
```
lejolly: can we add another assert here to check that it is focused?

rd1992:
@lejolly done. you can merge it.

lejolly:
i haven't managed to get it to pass on travis, can you confirm?

rd1992:
Hmmm it builds fine locally. did you try it locally?

lejolly:
yeah it works fine locally.

lejolly:
seems to work now.

PR #858, opened 2015-08-07
Initialize label picker dialog's position

### LINKED ISSUE #814: Label picker goes off screen
image

This happened in a Win 8 laptop with only one screen.
resolution: 1920x1200

ndt93:
Fixes #814

Label picker's dialog is now positioned at the middle of the pView, aligning its left edge with pView's left edge.

image

### DIFF
--- a/src/main/java/ui/components/pickers/LabelPicker.java
+++ b/src/main/java/ui/components/pickers/LabelPicker.java
@@ -31,12 +31,6 @@ public class LabelPicker {
         List<TurboLabel> allLabels = ui.logic.getRepo(issue.getRepoId()).getLabels();
         // create new LabelPickerDialog
         LabelPickerDialog labelPickerDialog = new LabelPickerDialog(issue, allLabels, stage);
-
-        // Needs this since dialog's size is not available before the blocking showAndWait
-        Platform.runLater(() -> {
-            labelPickerDialog.positionDialog(stage);
-        });
-
         // show LabelPickerDialog and wait for result
         Optional<List<String>> result = labelPickerDialog.showAndWait();
         stage.show(); // ensures stage is showing after label picker is closed (mostly for tests)
--- a/src/main/java/ui/components/pickers/LabelPicker.java
+++ b/src/main/java/ui/components/pickers/LabelPicker.java
@@ -31,6 +31,12 @@ public class LabelPicker {
         List<TurboLabel> allLabels = ui.logic.getRepo(issue.getRepoId()).getLabels();
         // create new LabelPickerDialog
         LabelPickerDialog labelPickerDialog = new LabelPickerDialog(issue, allLabels, stage);
+
+        // Needs this since dialog's size is not available before the blocking showAndWait
+        Platform.runLater(() -> {
+            labelPickerDialog.positionDialog(stage);
+        });
+
         // show LabelPickerDialog and wait for result
         Optional<List<String>> result = labelPickerDialog.showAndWait();
         stage.show(); // ensures stage is showing after label picker is closed (mostly for tests)
--- a/src/main/java/ui/components/pickers/LabelPickerDialog.java
+++ b/src/main/java/ui/components/pickers/LabelPickerDialog.java
@@ -9,6 +9,7 @@ import javafx.scene.input.KeyCode;
 import javafx.scene.layout.FlowPane;
 import javafx.scene.layout.VBox;
 import javafx.stage.Modality;
+import javafx.stage.Screen;
 import javafx.stage.Stage;
 
 import java.util.Collections;
@@ -132,6 +133,18 @@ public class LabelPickerDialog extends Dialog<List<String>> {
         initModality(Modality.APPLICATION_MODAL); // TODO change to NONE for multiple dialogs
         setTitle("Edit Labels for " + (issue.isPullRequest() ? "PR #" : "Issue #") +
                 issue.getId() + " in " + issue.getRepoId());
+        positionDialog(stage);
+    }
+
+    private void positionDialog(Stage stage) {
+        double totalScreenWidth = Screen.getScreens().stream()
+                .mapToDouble(screen -> {
+                    return screen.getVisualBounds().getWidth();
+                }).sum();
+        double maxX = Math.max(0, totalScreenWidth - stage.getMinWidth());
+
+        setX(Math.min(maxX, stage.getX() + stage.getScene().getWidth()));
+        setY(stage.getY() + stage.getScene().getY());
     }
 
     private void createButtons() {
--- a/src/main/java/ui/components/pickers/LabelPickerDialog.java
+++ b/src/main/java/ui/components/pickers/LabelPickerDialog.java
@@ -42,13 +42,21 @@ public class LabelPickerDialog extends Dialog<List<String>> {
         textField = createTextField();
         bottomBox = createBottomBox();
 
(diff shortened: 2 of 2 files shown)


damithc:
Will it block any important part of the bView?

damithc:
Is there a way to position it relative to the position of the selected issue?

ndt93:
@damithc Yup, it will block 2/3 of the content area of an issue on Github in bView, we can make the *x-coordinate* of the dialog to overlap the pView also. In terms of y-coordinate, it's not straight forward to find the position of the selected cell in the layout but should be doable; the problem is that the current dialog will go off screen at the bottom.

damithc:
Yes, relative positioning is complicated. May be simpler to keep it over the pView as it is now but ensure the it doesn't go out of screen on the left by adjusting x coord?

ndt93:
Ok sure, so we just align the x-coord of the label picker with the pView's x-coord (with respect the the monitor)?

damithc:
Yes, let's try that.

damithc:
But we should not place it at the top (y position). That looks odd. 
x position: prevent dialog from going out of screen. Usually it is the left side that is of concern, but it can  be the right side too, if the panel is at that edge.
y position: place the window roughly in the center (that's how it is now).

ndt93:
For the y position, the dialog is quite long now due to the grouping of labels. So on my monitor it will go off screen at the bottom. What do you suggest as a solution for that?

lejolly:
is it possible to match the midpoint of the dialog's y position with the pView's?

ndt93:
The dialog's size is not available during initialization (before it was shown). Then when it's shown, due to the blocking `showAndWait`, we cannot set the size. So that's a challenge :P

lejolly:
hmm, maybe attaching a listener to the `heightProperty`?

ndt93:
Yah, there are some suggested solutions online. I'm trying them out.

damithc:
Even in my monitor it sometimes goes out of the bottom. So I guess it has to stay fairly high, but may be not at the top? May be place it to roughly after one issue in the list?

ndt93:
Yup I'm trying to get the size of the dialog so we can align the dialog's center with the pView's center. How do you think about that?

ndt93:
Ok, now it's positioned at the middle (in the screenshot). How does that look?

damithc:
Looks good. Go ahead.

lejolly:
should the width of the dialog be limited to the default width of the pView? i think i left it at `400px` or something.

ndt93:
I saw the 400, but changing the number doesn't seem to have any effect.

lejolly:
oh? ok then.

ndt93:
Yah, unless you want to check that part out. This PR should be ready.

lejolly:
for now the top of the label picker dialog is at the midpoint of the pView, but would it be better if both the midpoints were aligned?

lejolly:
whoops, my bad, cause `positionDialog` doesn't seem to get called when the dialog first pops up?

ndt93:
Hmm it got into the right position when it first pops up for me. Does it adjust to the correct position for you eventually after pop up? Otherwise, we will need to try another solution for getting the dialog's size.

lejolly:
nope, only if i alter the height of the dialog by selecting enough labels.

ndt93:
@lejolly I added a second solution using `Platform.runLater`, which has some elements of non-determinism in it so I don't particularly prefer it and still keep the listener for `heightProperty` along side. Can help me check if it works for you?

lejolly:
yup it does, but there's a little jump at the start. 

jump

ndt93:
Ok, I also suspect that the dialog's size is actually already available in the initialization in MacOS so the height property's listener is not called afterwards. You want to try calling `positionDialog` in initialiseDialog` too? I will add that in if it eliminates the jump for you since I cannot replicate the problem on my comp.

lejolly:
sure, go ahead, just add a comment for that line.

ndt93:
@lejolly I pushed the line already, can check if it's still jumping for you?

lejolly:
yup, it's still jumping.

ndt93:
Hmm, you want to attempt other solutions on your side? Since I couldn't observe the effect on my side.

lejolly:
ok, will try later.

ndt93:
I have another idea in mind, I will try it and let you know.

ndt93:
Nvm, doesn't work. Leave it to you to try on your side then.

src/main/java/ui/components/pickers/LabelPickerDialog.java
```diff
@@ -132,6 +145,17 @@ private void initialiseDialog(Stage stage, TurboIssue issue) {
         initModality(Modality.APPLICATION_MODAL); // TODO change to NONE for multiple dialogs
         setTitle("Edit Labels for " + (issue.isPullRequest() ? "PR #" : "Issue #") +
                 issue.getId() + " in " + issue.getRepoId());
+        // MacOS may need this call here to position the dialog properly
+        positionDialog(stage);
```
lejolly: putting it here is redundant for now. (still working on where's best to put it)

src/main/java/ui/components/pickers/LabelPicker.java
```diff
@@ -31,6 +31,12 @@ private void showLabelPicker(TurboIssue issue) {
         List<TurboLabel> allLabels = ui.logic.getRepo(issue.getRepoId()).getLabels();
         // create new LabelPickerDialog
         LabelPickerDialog labelPickerDialog = new LabelPickerDialog(issue, allLabels, stage);
+
+        // Needs this since dialog's size is not available before the blocking showAndWait
+        Platform.runLater(() -> {
+            labelPickerDialog.positionDialog(stage);
```
lejolly: this causes the initial jump on OS X, but without it the dialog is positioned correctly on launch.

src/main/java/ui/components/pickers/LabelPickerDialog.java
lejolly: @ndt93 actually i managed to minimize the problem by wrapping the line in a `Platform.runLater`, it helps slightly but i don't think it can be eliminated completely. does it work on windows if we keep this as the only instance of `postitionDialog`? i'm not exactly for tying it to the height property since if the dialog expands having it jump back to the center might be jarring for the user.

lejolly:
@ndt93 ok the best i can do for now is to listen on the `showingProperty` instead of `heightProperty`, i haven't tested it on windows yet, but to me it makes more sense to to just position the dialog once and not on every height change since the height doesn't change dramatically and positioning it again might be too jarring for the user.

ndt93:
I tried with `setOnShowing` and `setOnShown` last time but both didn't work on Windows. I will check if showingProperty's handler would work, then all other codes can be removed.

ndt93:
Yayy, that works on Windows also! Still a bit jumpy at first but not as bad as `Platform.runLater`, should be an acceptable compromise then.

lejolly:
was it jumping on windows from the start?

ndt93:
It just jerks a bit the first time it pops up after the application is launched, but I think it's not that annoying. Afterwards, it behaves quite nicely.

Between, the coverage failed on some code in `FilterPanel.java` which is unrelated to the changes in this PR. Anyone knows the reason or how to fix it?

lejolly:
if it's not related then no need to worry about it.

PR #865, opened 2015-08-14
Change 'comment' hotkey to R instead of C

### LINKED ISSUE #859: Change 'comment' hotkey to R instead of C
Gmail uses R.

rd1992:
Fixes #859

### DIFF
--- a/src/main/java/ui/components/KeyboardShortcuts.java
+++ b/src/main/java/ui/components/KeyboardShortcuts.java
@@ -82,7 +82,7 @@ public class KeyboardShortcuts {
     public static final KeyCode SHOW_CONTRIBUTORS = KeyCode.D;
 
     // TODO decouple manage/show labels/milestones?
-    public static final KeyCode NEW_COMMENT = KeyCode.C;
+    public static final KeyCode NEW_COMMENT = KeyCode.R;
     public static final KeyCode MANAGE_LABELS = KeyCode.L;
     public static final KeyCode MANAGE_ASSIGNEES = KeyCode.A;
     public static final KeyCode MANAGE_MILESTONE = KeyCode.M;
@@ -123,7 +123,7 @@ public class KeyboardShortcuts {
         assignedKeys.add(KeyCode.F5); //REFRESH
         assignedKeys.add(KeyCode.F1); //SHOW_DOCS
         assignedKeys.add(KeyCode.G); //GOTO_MODIFIER
-        assignedKeys.add(KeyCode.C); //NEW_COMMENT
+        assignedKeys.add(KeyCode.R); //NEW_COMMENT
         assignedKeys.add(KeyCode.A); //MANAGE_ASSIGNEES
         assignedKeys.add(KeyCode.SPACE); //DOUBLE_PRESS
     }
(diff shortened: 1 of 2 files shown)


src/main/java/ui/RepositorySelector.java
```diff
@@ -49,6 +52,12 @@ private void setupComboBox() {
                 onValueChangeCallback.accept(repoId);
             }
         });
+        comboBox.addEventFilter(KeyEvent.KEY_RELEASED, event -> {
+            event.consume();
+            if (event.getCode() == KeyboardShortcuts.REMOVE_FOCUS) {
+                getParent().requestFocus();
+            }
+        });
```
damithc: Perhaps add a comment to explain what this is doing?

src/test/java/guitests/ChromeDriverTest.java
```diff
@@ -145,7 +145,7 @@ public void chromeDriverStubTest() {
         clearKeyCode();
 
         // jump to comments
-        push(KeyCode.C);
+        push(KeyCode.R);
```
dariusf: Same here, should make use of the constant.

src/main/java/ui/components/KeyboardShortcuts.java
```diff
@@ -123,7 +126,7 @@ private static void addNonCustomizableShortcutKeys() {
         assignedKeys.add(KeyCode.F5); //REFRESH
         assignedKeys.add(KeyCode.F1); //SHOW_DOCS
         assignedKeys.add(KeyCode.G); //GOTO_MODIFIER
-        assignedKeys.add(KeyCode.C); //NEW_COMMENT
+        assignedKeys.add(KeyCode.R); //NEW_COMMENT
         assignedKeys.add(KeyCode.A); //MANAGE_ASSIGNEES
         assignedKeys.add(KeyCode.SPACE); //DOUBLE_PRESS
     }
```
dariusf: You could make use of the constants defined above here, then the comments wouldn't be needed.

dariusf:
This doesn't have to depend on the Esc PR. Could you rebase onto master? Should also squash, since it's just a single change.

dariusf:
Hmm, actually @damithc's comment should be addressed in #857. The squashing should be done there. This PR can just be 8aa3abb plus changes from feedback.

src/test/java/guitests/ChromeDriverTest.java
dariusf: Realised that the constants aren't used much in tests. Unless we can `push` a `KeyCombination`...

damithc:
remember to use labels to show current issue status.

damithc:
Shall we wrap this up quickly?

rd1992:
@dariusf i have squashed the commits in the other PR. You can merge this into master.

dariusf:
Yes, but this still depends on those old commits and needs to be rebased as well. Like I said earlier, this PR should have only one commit (which you can squash changes based on feedback into). The easiest way to do this is to `rebase -i origin/master` and pick only the last one.

dariusf:
Can we make use of the constants in `KeyboardShortcuts`? Can be merged otherwise.

rd1992:
@dariusf we should use constants throughout. Though that should be a different issue than this no?

dariusf:
Okay, sure.

HansNewbie:
The keyboard shortcut docs has not been updated to use `R` instead of `C` for commenting.

damithc:
Oops, should have been part of the same PR. @rd1992 please do as a separate issue ASAP. @darius, once it is done, we can simply merge to release branch. No need to do a new release because the documentation is online.

Issue #751, opened 2015-06-30
Try findbugs on HT


damithc:
We can give this a try http://findbugs.sourceforge.net/

dariusf:
I actually tried this around the time we were integrating CheckStyle, along with [PMD](http://pmd.sourceforge.net/). Both are very easy to integrate using Gradle, but some effort is needed to work out the violations.

PR #866, opened 2015-08-14
Try findbugs on HT #751

dariusf:
Integrates Findbugs and fixes existing violations. Addresses #751.


damithc:
Let's do peer reviews for all code from now on.
@lejolly can review this one?

dariusf:
The last commit is an interim fix for a Findbugs violation that will require some refactoring. Details in commit message. I'll make a separate issue for reverting the commit and fixing it in a better way when this is merged.

src/main/java/github/GitHubEventsResponse.java
```diff
@@ -42,7 +45,14 @@ private void parseEventParameters(InputStream jsonBody) {
...
             reader = new BufferedReader(new InputStreamReader(jsonBody, CHARSET_UTF8));
-            Map<String, Object>[] eventsWithParameters = gson.fromJson(reader, type);
+            Optional<Map<String, Object>[]> possibleParameters =
+                Optional.ofNullable(gson.fromJson(reader, type));
+
+            if (!possibleParameters.isPresent()) {
+                return;
+            }
+            Map<String, Object>[] eventsWithParameters = possibleParameters.get();
+
```
lejolly: is this the preferred way of null checking instead of `object == null`?

dariusf: Hmm, actually no, that's overkill :stuck_out_tongue:  I thought I'd be propagating the empty object but it didn't turn out to be needed.

src/main/java/ui/MenuControl.java
```diff
@@ -201,10 +201,11 @@ private void onBoardDelete(String boardName) {
...
             Map<String, List<String>> boards = prefs.getAllBoards();
 
-            for (final String boardName : boards.keySet()) {
-                final List<String> filterSet = boards.get(boardName);
+            for (Map.Entry<String, List<String>> entry : boards.entrySet()) {
+                final String boardName = entry.getKey();
+                final List<String> filterSet = entry.getValue();
 
-                MenuItem openItem = new MenuItem(boardName);
+                MenuItem openItem = new MenuItem();
```
lejolly: this is not intended right?

dariusf: Yes, good catch.

lejolly:
looks good to me.

PR #868, opened 2015-08-14
Hotkey for assignees ('a') doesn't work if the bView is showing commits

### LINKED ISSUE #860: Hotkey for assignees ('a') doesn't work if the bView is showing commits
Currently, it works only if the bView is showing the conversation view. It should load the 'conversation view' and then focus on the assignee property.

rd1992:
Fixes #860

### DIFF
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -486,4 +486,10 @@ public class BrowserComponent {
         return driver.getCurrentUrl();
     }
 
+	public void loadTab(String currentUrl) {
+		if (isCurrentUrlIssue() && (currentUrl.endsWith("/commits") || currentUrl.endsWith("/files"))) {
+            driver.findElement(By.xpath("//a[@data-container-id='discussion_bucket']")).click();
+		}
+	}
+
 }
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -486,10 +486,9 @@ public class BrowserComponent {
         return driver.getCurrentUrl();
     }
 
-	public void loadTab(String currentUrl) {
-		if (isCurrentUrlIssue() && (currentUrl.endsWith("/commits") || currentUrl.endsWith("/files"))) {
+    public void loadTab(String currentUrl) {
+        if (isCurrentUrlIssue() && (currentUrl.endsWith("/commits") || currentUrl.endsWith("/files"))) {
             driver.findElement(By.xpath("//a[@data-container-id='discussion_bucket']")).click();
-		}
-	}
-
+        }
+    }
 }
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -486,8 +486,8 @@ public class BrowserComponent {
         return driver.getCurrentUrl();
     }
 
-    public void loadTab(String currentUrl) {
-        if (isCurrentUrlIssue() && (currentUrl.endsWith("/commits") || currentUrl.endsWith("/files"))) {
+    public void switchToConversationTab() {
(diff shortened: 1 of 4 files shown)


src/main/java/browserview/BrowserComponent.java
```diff
@@ -486,4 +486,9 @@ public String getCurrentUrl() {
         return driver.getCurrentUrl();
     }
 
+    public void loadTab(String currentUrl) {
```
damithc: Method name more general than what the method actually does?

damithc:
If ready for review, label accordingly.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -486,4 +486,9 @@ public String getCurrentUrl() {
...
+        if (isCurrentUrlIssue() && (currentUrl.endsWith("/commits") || currentUrl.endsWith("/files"))) {
```
dariusf: The `endsWith` part can be factored out into `GitHubURL`.

Also I don't think we need to pass the URL as a parameter, given that all this method does is switch to the discussion tab. Conceptually it's not really parameterised by the current URL, and being a `BrowserComponent` method, can already access it.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -486,4 +486,9 @@ public String getCurrentUrl() {
...
+            driver.findElement(By.xpath("//a[@data-container-id='discussion_bucket']")).click();
```
lejolly: would it be better to just explicitly load the pr page instead of clicking on the tab?

rd1992: @lejolly no that would involve a page refresh for no reason. We already have visited that page before when user initially clicked on the PR issue in the p view

dariusf:
Looks good. @lejolly?

src/main/java/browserview/BrowserComponent.java
```diff
@@ -486,4 +486,9 @@ public String getCurrentUrl() {
         return driver.getCurrentUrl();
     }
 
+    public void switchToConversationTab() {
+        if (isCurrentUrlIssue() && GitHubURL.isPullRequestLoaded(getCurrentUrl())) {
```
lejolly: do we need the first check here? since it seems to be called too in all cases this method is called.

src/main/java/util/GitHubURL.java
```diff
@@ -47,4 +47,8 @@ public static String getPathForContributors(String repoId) {
     public static boolean isUrlIssue(String url) {
         return url.matches("https://github.com/([^/]+)/([^/]+)/(issues|pull)/([0-9]+)([/commits,/files]*)");
     }
+    
+    public static boolean isPullRequestLoaded(String url)  {
+        return (url.endsWith("/commits") || url.endsWith("/files"));
+    }
```
lejolly: cover this in `GitHubURLTest`?

rd1992: @lejolly have added the tests. can you check and merge?

src/main/java/browserview/BrowserComponent.java
lejolly: is `isCurrentUrlIssue` still needed here?

src/main/java/util/GitHubURL.java
lejolly: looks fine.

rd1992:
@dariusf this was approved my @lejolly. Could you take a look?

dariusf:
I think the only outstanding issue is whether `isCurrentUrlTest` is needed. Otherwise can be merged.

rd1992:
@dariusf you meant isCurrentUrlIssue() i think. I have removed that check

lejolly:
looks good, go ahead and merge it.

PR #871, opened 2015-08-15
pri.high should match priority.high #864

### LINKED ISSUE #864: pri.high should match priority.high
image
Even `priority.high` seems to match `Feature.TextHighlight`

lejolly:
Fixes #864


damithc:
Something like search/matching will probably need hundreds of test cases. It's unlikely we can do that level of testing via the GUI. Need to extract it as a separate function and test independently.

dariusf:
We could have a function for matching strings and make use of that in the label/issue matching methods. The GUI portions could then make use of both.

Could we refactor as a separate issue and merge this first?

damithc:
Can be done as a separate issue.

PR #872, opened 2015-08-15
Aggregate Labels #744

### LINKED ISSUE #744: Aggregate label updates to save space
Updates like this can be condensed into a single line. e.g. `user` : `added-label` ~~`removed-label`~~ 
image

ndt93:
Fixes #744

Current behavior: label update events are grouped into 1 line if they belong to the same author and their time stamps are within 1 minute of each other (which can be considered to be edited in 1 go in most cases).

The time stamp of the earliest event of the group is taken as the time stamp for the whole group. Groups are then sorted according to those representative time stamp.

Within a group, the labels are sorted according to their time stamps

image

image


src/test/java/tests/TurboIssueEventTests.java
```diff
@@ -0,0 +1,114 @@
...
+                new User().setLogin("C"), IssueEventType.Labeled,
+                new GregorianCalendar(2015, 1, 1, 2, 30, 15).getTime())
+            .setLabelName("C1"));
+
+
+        sampleEvents.add(
+            new TurboIssueEvent(
+                new User().setLogin("D"), IssueEventType.Unlabeled,
+                new GregorianCalendar(2015, 1, 1, 3, 20, 59).getTime())
+            .setLabelName("D1"));
```
damithc: Looks like a block of code is being repeated several times. Is it possible to abstract into a small utility method?

damithc: There is a similar scenario of repeating code block in a test method above too.

damithc:
Is it possible to separate the logic of combining updates from UI elements used to show updates? I would like to see a separate method that takes a bunch of updates and return the combined version. Then, I would like to see unit tests covering that method extensively. e.g. cases where two users modify something around the same time. Furthermore, tests should be well documented: it should be clear what scenario is being tested and what is the expected output.

ndt93:
@damithc At the moment, the logic of combining updates are sort of separated from the creation of UI elements used to show them:
* The static method `groupLabelUpdateEvents` takes in a list of TurboIssueEvents  and return a combined list of sub-lists of TurboIssueEvents
* The static method `createLabelUpdateEventNodes` calls upon `groupLabelUpdateEvents` and transform the result into a list of JavaFX nodes for display

So I guess we could have more extensive testings for `groupLabelUpdateEvents` in this case?

damithc:
I think the separation is good enough. I didn't notice it because the word `Event` threw me off (I thought it refers to events firing the app). 
Yes, we need more testing for that method.

ndt93:
Ok sure.

ndt93:
Updated tests: There are now 10 tests for `groupLabelUpdateEvents` with comments on the scenarios under tests and expected results.

Coveralls failed due to unrelated code in `UI.java`

src/main/java/ui/listpanel/ListPanelCard.java
```diff
@@ -118,14 +124,23 @@ private Node getEventDisplay(TurboIssue issue, final int withinHours) {
...
         VBox.setMargin(result, new Insets(3, 0, 0, 0));
 
-        // Events
+        // Label update events
+        TurboIssueEvent.createLabelUpdateEventNodes(
+                model,
+                events.stream()
+                      .filter(e -> e.isLabelUpdateEvent())
+                      .collect(Collectors.toList()))
+        .forEach(node -> result.getChildren().add(node));;
```
damithc: This line should be indented more? Right now it's aligned with it's own start.
Is the double semicolon intended?

src/main/java/ui/listpanel/ListPanelCard.java
```diff
@@ -118,14 +124,23 @@ private Node getEventDisplay(TurboIssue issue, final int withinHours) {
      * @param comments
      * @return
      */
-    private static Node layoutEvents(Model model, TurboIssue issue,
-                                     List<TurboIssueEvent> events, List<Comment> comments) {
+    public static Node layoutEvents(Model model, TurboIssue issue,
```
damithc: This need not be public just for the sake of testing. There is a way to test private methods using reflection.

src/test/java/tests/TurboIssueEventTests.java
```diff
@@ -0,0 +1,400 @@
...
+import java.util.List;
+
+import org.eclipse.egit.github.core.User;
+import org.junit.Test;
+
+public class TurboIssueEventTests {
+
+    public List<TurboIssueEvent> sampleEvents;
+
+    public static TurboIssueEvent createLabelUpdateEvent(
```
damithc: Does this need to be public?

src/test/java/tests/TurboIssueEventTests.java
```diff
@@ -0,0 +1,400 @@
...
+     * Test the events grouping when list of events is empty
+     */
+    @Test
+    public void testLabelUpdateEventsGroupingForNoEvent() {
+        assertEquals(new ArrayList<TurboIssueEvent>(),
+                     TurboIssueEvent.groupLabelUpdateEvents(new ArrayList<TurboIssueEvent>()));
+    }
+
+    /**
+     * Test the events grouping when there is only 1 event
```
damithc: Test**s** the events ... (it's a shorter way of saying 'This method tests the events ...).

src/test/java/tests/TurboIssueEventTests.java
```diff
@@ -0,0 +1,400 @@
...
+        events.add(sampleEvents.get(0));
+
+        List<TurboIssueEvent> expectedSubList1 = new ArrayList<>();
+        expectedSubList1.add(sampleEvents.get(0));
+
+        List<List<TurboIssueEvent>> expected = new ArrayList<>();
+        expected.add(expectedSubList1);
+
+        assertEquals(expected,
+                     TurboIssueEvent.groupLabelUpdateEvents(events));
```
damithc: Our length is 110-120 chars. This can probably fit in one line?

src/test/java/tests/TurboIssueEventTests.java
```diff
@@ -0,0 +1,400 @@
...
+                                          "L1"));
+        events.add(createLabelUpdateEvent("B", IssueEventType.Unlabeled,
+                                          new GregorianCalendar(2015, 1, 1, 1, 1, 25),
+                                          "L1"));
+        events.add(createLabelUpdateEvent("A", IssueEventType.Labeled,
+                                          new GregorianCalendar(2015, 1, 1, 1, 1, 40),
+                                          "L2"));
+        events.add(createLabelUpdateEvent("B", IssueEventType.Labeled,
+                                          new GregorianCalendar(2015, 1, 1, 1, 2, 11),
+                                          "L2"));
```
damithc: I didn't check. Do you incorporate boundary cases? i.e. events happening at 59,60,61 seconds apart.

damithc: Also, empty lists.

damithc:
Added some cosmetic comments. @lejolly can review?

damithc:
Also, now that @ndt93 is a project member, he should be able to initiate a travis run without closing and opening a PR right?

src/main/java/github/TurboIssueEvent.java
```diff
@@ -248,6 +261,123 @@ public Node display(Model model, TurboIssue issue) {
...
+     * Checks if this label update event in the same group
+     * as another label update event e
+     * @param e another label update event
+     * @return true if this event and e have same author and times within
+     *              MAX_TIME_DIFF from each other.
+     *         false otherwise
+     */
+    public boolean isInSameLabelUpdateEventGroup(TurboIssueEvent e) {
+        long timeDiffMs = Math.abs(getDate().getTime() - e.getDate().getTime());
+        long timeDiffMin = TimeUnit.MILLISECONDS.toSeconds(timeDiffMs);
```
lejolly: `toSeconds` returns the number of minutes?

ndt93: MAX_TIME_DIFF is 60s

src/test/java/tests/TurboIssueEventTests.java
ndt93: I have empty list, 59 and 60. Will add a case for 61

src/main/java/ui/listpanel/ListPanelCard.java
ndt93: The `forEach` is operatoring on each node returned by `TurboIssueEvent.createLabelUpdateEventNodes` so I was thinking that it should be at the same level. I will put the `forEach` right after the previous statement then.

The double semi-colon is not intended :-P

ndt93:
There is supposed to be a button for rebuild or something but I can't find it. Maybe because this PR is still on my fork.

src/main/java/ui/listpanel/ListPanelCard.java
```diff
@@ -124,8 +130,18 @@ private static Node layoutEvents(Model model, TurboIssue issue,
         result.setSpacing(3);
         VBox.setMargin(result, new Insets(3, 0, 0, 0));
 
-        // Events
+        // Label update events
+        List<TurboIssueEvent> labelUpdateEvents =
+                        events.stream()
+                        .filter(e -> e.isLabelUpdateEvent())
```
lejolly: replace with method reference: `.filter(TurboIssueEvent::isLabelUpdateEvent)`?

src/main/java/github/TurboIssueEvent.java
```diff
@@ -248,6 +261,123 @@ public Node display(Model model, TurboIssue issue) {
...
+        // Sort according to time
+        Collections.sort(events,
+                (e1, e2) -> {
+                    return e1.getDate().compareTo(e2.getDate());
+                });
+        // Sort according to author
+        Collections.sort(events,
+                (e1, e2) -> {
+                    return e1.getActor().getLogin().compareTo(e2.getActor().getLogin());
+                });
```
lejolly: can collapse:

```java
// Sort according to time
Collections.sort(events, (e1, e2) -> e1.getDate().compareTo(e2.getDate()));
// Sort according to author
Collections.sort(events, (e1, e2) -> e1.getActor().getLogin().compareTo(e2.getActor().getLogin()));
```

src/main/java/github/TurboIssueEvent.java
```diff
@@ -248,6 +261,123 @@ public Node display(Model model, TurboIssue issue) {
...
+        }
+        if (!currentSubList.isEmpty()) {
+            result.add(currentSubList);
+        }
+
+        // Sort according to time of the first event of each group
+        Collections.sort(result,
+                (l1, l2) -> {
+                    return l1.get(0).getDate().compareTo(l2.get(0).getDate());
+                });
```
lejolly: same for this.

src/main/java/ui/listpanel/ListPanelCard.java
ndt93: This one doesn't work since `e` is the object on which `isLabelUpdateEvent` is called, not technically its argument and also `isLabelUpdateEvent` is not static.

src/main/java/github/TurboIssueEvent.java
ndt93: Nice! I will fix this.

src/main/java/ui/listpanel/ListPanelCard.java
lejolly: i think it still works, i guess i would understand if you want to keep it this way to be consistent with the other `.filter` below it.

ndt93: I've tried it out. It gives a compile time error

lejolly: hmm that's strange, seems to work for me. ah well, small issue i guess.

ndt93: :/ Ok, could be because of my eclipse's compiler or something. Gradle builds fine. I will change it since this one looks nicer. Also, can help me rebuild it when the CI fail.

ndt93:
@lejolly I've merged this with master with Findbugs. You can check this again.

lejolly:
looks good.

PR #878, opened 2015-08-19
Refactor label matching #873

### LINKED ISSUE #873: Refactor label matching
See discussion in #864 and #871.

The logic for matching strings against label names should be factored out and shared.

lejolly:
Fixes #873

### DIFF
--- a/src/main/java/filter/expression/Qualifier.java
+++ b/src/main/java/filter/expression/Qualifier.java
@@ -623,20 +623,20 @@ public class Qualifier implements FilterExpression {
 
         String group = "";
         if (tokens.getGroup().isPresent()) {
-            group = tokens.getGroup().get().toLowerCase();
+            group = tokens.getGroup().get();
         }
-        String labelName = tokens.getName().toLowerCase();
+        String labelName = tokens.getName();
 
         for (TurboLabel label : model.getLabelsOfIssue(issue)) {
             if (label.getGroup().isPresent()) {
                 if (labelName.isEmpty()) {
                     // Check the group
-                    if (label.getGroup().get().toLowerCase().contains(group)) {
+                    if (Utility.containsIgnoreCase(label.getGroup().get(), group)) {
                        return true;
                     }
                 } else {
-                    if (label.getGroup().get().toLowerCase().contains(group)
-                        && label.getName().toLowerCase().contains(labelName)) {
+                    if (Utility.containsIgnoreCase(label.getGroup().get(), group)
+                        && Utility.containsIgnoreCase(label.getName(), labelName)) {
                        return true;
                     }
                 }
@@ -644,7 +644,7 @@ public class Qualifier implements FilterExpression {
                 // Check only the label name
                 if (!group.isEmpty()) {
                     return false;
-                } else if (!labelName.isEmpty() && label.getName().toLowerCase().contains(labelName)) {
+                } else if (!labelName.isEmpty() && Utility.containsIgnoreCase(label.getName(), labelName)) {
                     return true;
                 }
             }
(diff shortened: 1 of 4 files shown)


dariusf:
No problems with this, but I was hoping we could factor out [this method](https://github.com/HubTurbo/HubTurbo/blob/master/src/main/java/filter/expression/Qualifier.java#L618) as well, and have it shared across all the places where we match user input against label titles. It's quite similar to the label picker method.

The tests for it would then involve string inputs only.

The [existing tests](https://github.com/HubTurbo/HubTurbo/blob/master/src/test/java/tests/FilterEvalTests.java#L134) would only be concerned about the correctness of the model's methods for locating issues based on label, and no longer about the details of matching.

lejolly:
the only part of `labelsSatisfy` that i think are the same are the `label.getGroup().get().toLowerCase().contains(group)` parts which can be substituted with `Utility.containsIgnoreCase`. the rest of the code seems very specific to `Qualifier`. on a side note, why are we using `contains` and not `equals`?

dariusf:
I see, so the matching is done differently enough that things can't be shared? I'm aware of a few issues (complexity from interactivity, filter doesn't delimit with spaces), but there are other deep differences?

Just hoping we can share as much as possible so behaviour is consistent everywhere.

`contains` was just to increase allowance for typos ('fuzzier' than `equals`). `startsWith` could also be used if that's what we're doing in the label picker.

lejolly:
yeah i think they're pretty different as `labelsSatisfy` compares the filter text with the labels of a particular issue while the label picker compares the search query with all the labels within the repo itself.

dariusf:
True, but isn't the comparison of user input against label name similar? This part is agnostic to the context of matching -- either way we are comparing strings.

How we delimit the input, whether we consider group and/or name, whether we ignore case, whether we compare with `equals` or `contains`... These small things are what I'm concerned will gradually get out of sync if we have them in multiple places.

damithc:
yes, that's my feeling too. It's matching a string to another group of
strings. However, the return value may be different. In one case it is
boolean while in the other it is the best match?

lejolly:
yes we are comparing strings, but `LabelPickerUILogic` takes in a list of `PickerLabels` so as to be able to set certain flags when checking, such as setting it as faded (dependent on search results) or putting a tick next to it (not dependent on search results). 

i'm assuming you want to create a method that takes in a list of strings to be compared against a query and then return a boolean? that only works for `Qualifier` because the method in `LabelPickerUILogic` needs to return a list of `PickerLabels` (with the appropriate flags set), and also the code in `Qualifier` supports early termination which isn't what is needed for `LabelPickerUILogic`. 

how we delimit the input is part of `TurboLabel` right? whether we use `equals`, `contains` or `startsWith` is also dependent on the function of the implementation?

dariusf:
What I proposed is a function which takes two strings (user input and label name) and returns a boolean. It gets run against every label we want to select. It's a low-level predicate, so it's not concerned with early termination. The intuition was that `LabelPickerUILogic` would use this predicate to choose its list of `PickerLabel`s.

In any case, I assumed that the label picker and filter had the same matching semantics, but on closer inspection they differ quite a bit:

- Filter allows things like `.name` and `priority.` (wildcards), while label picker does not differentiate (group if valid, name otherwise)
- Label picker interprets input differently depending on if the group is valid, while filter doesn't

Unless we want to reconcile these differences (which I'm not sure is possible), we can't meaningfully factor things out right now.

damithc:
Yes, and in addition, label picker is supposed to looks for the best match. for example, it should favor full word match over partial match,  'startsWith' match over 'contains' and so on (not sure if it is done like that at the moment). So we may not be able to refactor the logic right now. Let's KIV this for now. We can reconsider at a later stage.

damithc:
I meant let's KIV further refactoring, but go ahead with the refactorings done so far.

PR #880, opened 2015-08-19
Redundant statement in keyboard shortcuts #835

### LINKED ISSUE #835: Redundant statement in keyboard shortcuts
The following should be removed:
`Use arrow keys to move from one issue to another. Pressing shift with arrow key will not load the issue in browser. These will work when an issue is selected in a panel.`

HansNewbie:
Fixes #835, removed a redundant statement from documentation of keyboard shortcut.


damithc:
Why so many commits? :-p

HansNewbie:
Is this better?

damithc:
Looks good.

PR #888, opened 2015-08-19
Some updates not received until restart #748

### LINKED ISSUE #748: Some updates not received until restart
I had left HT running overnight y'day and when I checked in the morning, some of the latest updates did not appear under my `updated` filter. I did a restart and the missing updated appeared. Not sure what's the cause or how to repro. I'm recording it anyway, for future reference.

acruis:
Fixes #748

Will put some regression tests in place.

### DIFF
--- a/src/main/java/backend/IssueMetadata.java
+++ b/src/main/java/backend/IssueMetadata.java
@@ -33,6 +33,15 @@ public class IssueMetadata {
         this.isUpdated = other.isUpdated;
     }
 
+    // Copy constructor used in reconciliation
+    public IssueMetadata(IssueMetadata other, boolean isUpdated) {
+        this.events = new ArrayList<>(other.events);
+        this.comments = new ArrayList<>(other.comments);
+        this.nonSelfUpdatedAt = other.nonSelfUpdatedAt;
+        this.nonSelfCommentCount  = other.nonSelfCommentCount;
+        this.isUpdated = isUpdated;
+    }
+
     // Constructor used in DownloadMetadataTask
     public IssueMetadata(List<TurboIssueEvent> events, List<Comment> comments) {
         this.events = events;
--- a/src/main/java/backend/IssueMetadata.java
+++ b/src/main/java/backend/IssueMetadata.java
@@ -13,6 +13,8 @@ public class IssueMetadata {
     private final List<Comment> comments;
     private final LocalDateTime nonSelfUpdatedAt;
     private final int nonSelfCommentCount;
+
+    // If isUpdated is true, nonSelfUpdatedAt will be used to sort/filter instead of updatedAt in TurboIssue
     private final boolean isUpdated;
 
     // Constructor for default use when initializing TurboIssue
--- a/src/main/java/util/events/testevents/UpdateDummyRepoEvent.java
+++ b/src/main/java/util/events/testevents/UpdateDummyRepoEvent.java
@@ -9,7 +9,7 @@ public class UpdateDummyRepoEvent extends Event {
     public final int itemId; // For issues and milestones
     public final String idString; // For labels and users
     public final String updateText;
-    public final String actor;
+    public final String actor; // Only for comments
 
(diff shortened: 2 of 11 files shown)


damithc:
@acruis go ahead and assign a reviewer.

src/main/java/util/events/testevents/UpdateDummyRepoEvent.java
```diff
@@ -25,43 +27,50 @@
...
     // e.g. Update label
     public UpdateDummyRepoEvent(UpdateType updateType, String repoId, String idString, String updateText) {
-        this(updateType, repoId, -1, idString, updateText);
+        this(updateType, repoId, -1, idString, updateText, "");
+    }
+
+    // e.g. Comment
+    public UpdateDummyRepoEvent(UpdateType updateType, String repoId, int itemId, String updateText, String actor) {
+        this(updateType, repoId, itemId, null, updateText, actor);
     }
```
dariusf: This can be replaced with factory methods, like how `Optional.of` and `Optional.empty` are used in lieu of constructors. They are more self-documenting and less error-prone, especially in cases like this, where we have many ways to construct the same thing.

How about something like `UpdateDummyRepoEvent.forLabel`, `UpdateDummyRepoEvent.forMilestone`, etc.? Can use a static import if this starts getting verbose.

`UpdateType` could even be an interval value, not exposed to users. If there's no need for a default constructor, it can be made private as well.

src/main/java/backend/IssueMetadata.java
```diff
@@ -33,6 +33,15 @@ public IssueMetadata(IssueMetadata other) {
         this.isUpdated = other.isUpdated;
     }
 
+    // Copy constructor used in reconciliation
+    public IssueMetadata(IssueMetadata other, boolean isUpdated) {
+        this.events = new ArrayList<>(other.events);
+        this.comments = new ArrayList<>(other.comments);
+        this.nonSelfUpdatedAt = other.nonSelfUpdatedAt;
+        this.nonSelfCommentCount  = other.nonSelfCommentCount;
+        this.isUpdated = isUpdated;
```
dariusf: We should document what `isUpdated` does (controls whether `nonSelfUpdatedAt` or `updatedAt` is used).

dariusf:
The double filter isn't implemented yet, right? This implements a way of interacting with and testing issue metadata.

src/main/java/backend/IssueMetadata.java
acruis: I was hoping the use of `isUpdated` would be obvious as it's only referenced in the `satisfiesUpdatedHours` call in `Qualifier`.

src/main/java/util/events/testevents/UpdateDummyRepoEvent.java
acruis: How would you differentiate between events to make & delete & update a label? Have a factory method for each of them?

acruis:
I think the double filter can be implemented in a separate PR.

src/main/java/backend/IssueMetadata.java
dariusf: Just thought it was a bit vague. I guess it's fine, since I don't have any ideas for a more precise or obvious name... So in other words, it tells us whether or not the metadata is valid, and if not, that it should be replaced by valid metadata?

dariusf:
Okay, sure.

src/main/java/backend/IssueMetadata.java
acruis: Yup. I think more comments wouldn't hurt, though, so I'll clarify its use in a comment.

src/main/java/util/events/testevents/UpdateDummyRepoEvent.java
dariusf: Yup. The point of this is to make the type of event we're making clear at each call site.

acruis: Alright.

PR #890, opened 2015-08-21
 Add 'Save' function to boards #781 

### LINKED ISSUE #781: Add 'Save' function to boards
Currently, the 'Save' function works as 'Save as'. Rename it and add a Save function to update the board under the current name.

HyungJon:
Fixes #781 

"Save" function now refers to saving the state of the last opened board under its current name. The old function has been renamed to "Save as".
Saving when there is no board open results in the same behavior as "Save as".


HyungJon:
Unexpected Travis failure again. This time I will squash noisy commits when it's fixed.

HyungJon:
All right, I think the last push has been squashed properly this time. At least they show up as squashed in Git log on my local repository.

src/main/java/prefs/GlobalConfig.java
```diff
@@ -28,6 +28,7 @@
     private String lastViewedRepository = "";
     private String lastLoginUsername = "";
     private byte[] lastLoginPassword = new byte[0];
+    private String lastOpenBoard = "";
```
dariusf: We should use `Optional<String>` here instead. Expresses the semantics of this field more clearly.

Also, the empty string is a valid board name... Probably shouldn't be, but its use here as an empty value could cause problems.

src/main/java/ui/MenuControl.java
```diff
@@ -32,12 +32,14 @@
     private final ScrollPane panelsScrollPane;
     private final UI ui;
     private final Preferences prefs;
+    private String openBoardName = "";
```
dariusf: Is there a need to keep track of this here as well, i.e. could this and the state in prefs ever go out of sync?

src/main/java/ui/MenuControl.java
```diff
@@ -129,12 +131,31 @@ public void changed(ObservableValue<? extends Number> arg0, Number arg1, Number
...
     }
+    
+    private void onBoardSave() {
+        logger.info("Menu: Boards > Save");
+        
+        List<PanelInfo> panels = getCurrentPanels();
+        
+        if (openBoardName.equals("")) {
+            onBoardSaveAs();
+        }
```
dariusf: Should there be a `return` here? Seems like after we go into `onBoardSaveAs` we'd either have saved the board or not, and we shouldn't have to come back here.

src/main/java/prefs/GlobalConfig.java
HyungJon: Hmm true, I should add some name validation here.

src/main/java/ui/MenuControl.java
dariusf: Also, if there should be a return, you could swap the conditional and the `getCurrentPanels()` line, to make it clear that the absence of a board name immediately delegates to `onBoardSaveAs` without needing to do anything else.

src/main/java/prefs/GlobalConfig.java
```diff
@@ -82,6 +84,14 @@ public void addBoard(String name, List<PanelInfo> panels) {
     public void removeBoard(String name) {
         savedBoards.remove(name);
     }
+    
+    public void setLastOpenBoard(String board) {
+        lastOpenBoard = Optional.of(board);
+    }
+    
+    public String getLastOpenBoard() {
+        return lastOpenBoard.get();
```
dariusf: The point of `Optional` is to express statically that a value may be absent, and force consumers of the value to check for this (unlike `null`, which doesn't make you check anything).

Calling `get` without checking `isPresent` will fail at runtime (similar to dereferencing a null pointer). You should instead propagate the `Optional` out to callers of this method.

src/main/java/prefs/GlobalConfig.java
```diff
@@ -82,6 +84,14 @@ public void addBoard(String name, List<PanelInfo> panels) {
     public void removeBoard(String name) {
         savedBoards.remove(name);
     }
+    
+    public void setLastOpenBoard(String board) {
```
dariusf: Similarly, there needs to be a way for callers to indicate that there is no last-opened board. You could either have it take an `Optional` or provide `clearLastOpenBoard`.

src/main/java/ui/MenuControl.java
```diff
@@ -129,12 +129,31 @@ public void changed(ObservableValue<? extends Number> arg0, Number arg1, Number
         cols.getItems().addAll(createRight, createLeft, closePanel);
         return cols;
     }
+    
+    private void onBoardSave() {
+        logger.info("Menu: Boards > Save");
+        
+        if (prefs.getLastOpenBoard().equals("")) {
```
dariusf: Now that callers get an `Optional`, they can check if it `isPresent`.

src/main/java/ui/MenuControl.java
```diff
@@ -182,6 +208,9 @@ private void onBoardDelete(String boardName) {
 
         if (response.isPresent() && response.get().getButtonData() == ButtonData.OK_DONE) {
             prefs.removeBoard(boardName);
+            if (prefs.getLastOpenBoard().equals(boardName)) {
+                prefs.setLastOpenBoard("");
```
dariusf: This is where that `clear` method would be useful.

We also have to handle the possibility of the last-opened board not being present.

src/main/java/ui/MenuControl.java
```diff
@@ -151,9 +170,15 @@ private void onBoardSave() {
         Optional<String> response = dlg.showAndWait();
 
         if (response.isPresent()) {
-            prefs.addBoard(response.get(), panels);
+            String boardName = response.get();
+            if (boardName.equals("")) {
+                logger.info("Did not save new board");
+                return;
+            }
```
dariusf: You might want to factor the validation out into another method, and log each case, for example if it's empty, if it consists only of whitespace, etc.

HyungJon:
Maybe I should look into the ```NoNodesException``` again.

src/main/java/prefs/GlobalConfig.java
```diff
@@ -28,6 +29,7 @@
     private String lastViewedRepository = "";
     private String lastLoginUsername = "";
     private byte[] lastLoginPassword = new byte[0];
+    private Optional<String> lastOpenBoard;
```
dariusf: Think you might have forgotten to initialise this (to `empty`). Got an NPE when I checked this out and tried to save. It wasn't caught in the test?

HyungJon:
Some ridiculous error causing Travis failure, somehow trying to fine "'Node should have label Save as'" instead of just "Save as". I'll look into squashing these commits as well after I resolve them.

dariusf:
No worries, just ping or change the label when you are done.

HyungJon:
I just changed the tests to select all menu items using arrow keys. At least temporarily.

dariusf:
Hmm, using arrow keys might be fragile if we add more menu items. @lejolly is there another way to avoid the problem?

Otherwise this looks good.

damithc:
@lejolly see darius' comment above.

lejolly:
Whoops, missed this, I don't think the ci works well with clicking on menus. The alternative would be to use the accelerator (keyboard shortcut).

lejolly:
I guess the upside of using arrow keys is that we know if the order of items is kept.

damithc:
That should be a separate issue/PR. Saving borads is not frequent, so we may not need a shortcut for that. But I would like to have a short cut for changing boards, similar to the one we have for changing repos.

HyungJon:
All right. Will Ctrl+B do, like Ctrl+R for changing repos?

damithc:
Create a new issue. We'll discuss it over there.

dariusf:
In future you could squash a bit more. Build failures, 'fixed bug'-type commits, review changes (i.e. things which should have been done to begin with) can be part of whichever commit they are most related to.

Test changes can be a single commit, unless there's some change in there that warrants more granular documentation.

HyungJon:
Noted, I'll make sure to squash commits as required before the pushing itself.

PR #894, opened 2015-08-23
Error messages on 'could not sync' and data corruption on empty repo #877

### LINKED ISSUE #877: Error messages on could not sync and data corruption on empty repo
How to reproduce:
1. Create a new empty repository
2. Use it for Hub Turbo
3. Error messages will appear (as below)

error

HansNewbie:
Fixes #877 - false error when repo has no issue

### DIFF
--- a/src/main/java/util/Utility.java
+++ b/src/main/java/util/Utility.java
@@ -85,7 +85,8 @@ public class Utility {
     private static boolean processFileGrowth(long sizeAfterWrite, int issueCount, String fileName) {
         // The average issue is about 0.75KB in size. If the total filesize is more than (2 * issueCount KB),
         // we consider the json to have exploded as the file is unusually large.
-        if (sizeAfterWrite > ((long) issueCount * 2000)) {
+        if ((issueCount > 0) &&
+                (sizeAfterWrite > ((long) issueCount * 2000))) {
             UI.events.triggerEvent(new ShowErrorDialogEvent("Possible data corruption detected",
                     fileName + " is unusually large.\n\n"
                             + "Now proceeding to delete the file and redownload the repository to prevent "
--- a/src/test/java/tests/ErrorJsonTests.java
+++ b/src/test/java/tests/ErrorJsonTests.java
@@ -2,6 +2,7 @@ package tests;
 
 import backend.interfaces.RepoStore;
 import guitests.UITest;
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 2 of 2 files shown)


src/main/java/util/Utility.java
```diff
@@ -72,7 +72,10 @@ public static boolean writeFile(String fileName, String content, int issueCount)
                 writer.println(content);
                 writer.close();
 
-                long sizeAfterWrite = Files.size(Paths.get(fileName));
+                long sizeAfterWrite = 0;
+                if (issueCount > 0) {
```
acruis: Is this check necessary if you only carry out the check in `processFileGrowth` when `issueCount` is nonzero?

HansNewbie: I am not sure if I understood you correctly, but the check is necessary because `Files.size` would throw an `IOException` if `issueCount` is 0.

Since we should account for 0 `issueCount`, I put it such that no exception would be thrown in such case. An alternative would be handling 0 `issueCount` in the `catch` such that the return value of `writeFile` is consistent (`false` when no error including 0 `issueCount`, `true` when error).

acruis: `Files.size` throws `IOException` if `issueCount` is 0? You mean if the file doesn't exist at the given path right? That doesn't really mean `issueCount` is 0, if that's what you mean.

You can put a check for whether the file exists in the first place (`Files.exists`) if that's the case, and then assign `sizeAfterWrite` accordingly.

acruis:
About `testJsonExplosionDetection`, this test is meant to fail when someone removes the json explosion detection code. 

The modifications in this PR result in the tests passing even when the json explosion detection code is modified. I'd say use a test case with a issueCount of 1 issue or more, and then instead of using `abcde` as the file content, generate it using the `toString` version of an actual model, or produce a >2000-character-long String so that the filesize is more than 2KB.

src/main/java/util/Utility.java
HansNewbie: Thank you for noticing that. I put the check because I altered `testJsonExplosionDetection` so that when `issueCount` is 0, `corruptedJson` will be `false`. It was always corrupted because the test case of 0 `issueCount` does not create the file. I will look at it again.

HansNewbie:
I will change the test case to have `issueCount` of 1.

acruis:
Alright, good to merge.

PR #899, opened 2015-08-25
Closing a panel should put focus on the previous panel

### LINKED ISSUE #876: Closing a panel should put focus on the previous panel
Currently focus is nowhere.

rd1992:
Fixes #876

### DIFF
--- a/src/main/java/ui/issuepanel/PanelControl.java
+++ b/src/main/java/ui/issuepanel/PanelControl.java
@@ -142,6 +142,7 @@ public class PanelControl extends HBox {
         Node child = getChildren().remove(index);
         updatePanelIndices();
         ((AbstractPanel) child).close();
+        setFocusToPreviousPanel(index);
     }
 
     private void updatePanelIndices() {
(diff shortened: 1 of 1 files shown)


src/main/java/ui/issuepanel/PanelControl.java
```diff
@@ -191,15 +192,20 @@ public void closeCurrentPanel() {
...
+        if (getChildren().size() == 0) {
+            setCurrentlySelectedPanel(Optional.empty());
+        } else if (closedPanelIndex != currentlySelectedPanel.get()) {
+            return;
+        } else {
+            int newPanelIndex = (closedPanelIndex > getChildren().size() - 1)
+                                 ? closedPanelIndex - 1
+                                 : closedPanelIndex;
+            setCurrentlySelectedPanel(Optional.of(newPanelIndex));
+            getPanel(currentlySelectedPanel.get()).requestFocus();
```
dariusf: This method should be named in a more general way, since it doesn't always shift focus to the previous panel. `updateFocus`?

src/main/java/ui/issuepanel/PanelControl.java
```diff
@@ -191,15 +192,20 @@ public void closeCurrentPanel() {
...
-                getPanel(currentlySelectedPanel.get()).requestFocus();
-            }
+        }
+    }
+    
+    public void setFocusToPreviousPanel(int closedPanelIndex) {
+        if (getChildren().size() == 0) {
+            setCurrentlySelectedPanel(Optional.empty());
+        } else if (closedPanelIndex != currentlySelectedPanel.get()) {
+            return;
```
dariusf: This check should come first, as a guard clause. That way it's clear that this method does nothing if it wasn't the currently-selected panel that was closed.

PR #901, opened 2015-08-25
There should be a way to jump to an issue directly

### LINKED ISSUE #844: There is should be a way to jump to an issue directly
In the situation below, the focus is not on any element that we can interact with. The user should be able to put focus on an issue without going through the filter box first.
image

Pressing `1` should jump to first issue, `2` to the 2nd issue and so on?

rd1992:
Fixes #844

### DIFF
--- a/src/main/java/ui/components/KeyboardShortcuts.java
+++ b/src/main/java/ui/components/KeyboardShortcuts.java
@@ -94,6 +94,7 @@ public class KeyboardShortcuts {
             new KeyCodeCombination(KeyCode.K);
     public static final KeyCodeCombination SHOW_CONTRIBUTORS =
             new KeyCodeCombination(KeyCode.D);
+    public static final Set<KeyCodeCombination> jumpToFirstIssueKeys = populateJumpToFirstIssueSet();
 
     // TODO decouple manage/show labels/milestones?
     public static final KeyCodeCombination NEW_COMMENT =
@@ -112,7 +113,6 @@ public class KeyboardShortcuts {
     public static final KeyCodeCombination REMOVE_FOCUS =
             new KeyCodeCombination(KeyCode.ESCAPE);
 
-    public static final Set<KeyCode> jumpToFirstIssueKeys = populateJumpToFirstIssueSet();
     // ui.MenuControl
     public static final KeyCodeCombination NEW_ISSUE =
             new KeyCodeCombination(KeyCode.I, KeyCombination.CONTROL_DOWN);
@@ -128,10 +128,10 @@ public class KeyboardShortcuts {
     public static final KeyCodeCombination CLOSE_PANEL =
             new KeyCodeCombination(KeyCode.W, KeyCombination.CONTROL_DOWN);
     
-    private static Set<KeyCode> populateJumpToFirstIssueSet() {
-        Set <KeyCode> result = new HashSet<KeyCode>();
-        result.add(KeyCode.DIGIT1);
-        result.add(KeyCode.ENTER);
+    private static Set<KeyCodeCombination> populateJumpToFirstIssueSet() {
+        Set <KeyCodeCombination> result = new HashSet<KeyCodeCombination>();
+        result.add(new KeyCodeCombination(KeyCode.DIGIT1));
+        result.add(new KeyCodeCombination(KeyCode.DOWN, KeyCombination.CONTROL_DOWN));
         return result;
     }
     
--- a/src/main/java/ui/components/KeyboardShortcuts.java
+++ b/src/main/java/ui/components/KeyboardShortcuts.java
@@ -112,6 +112,7 @@ public class KeyboardShortcuts {
     public static final KeyCodeCombination REMOVE_FOCUS =
             new KeyCodeCombination(KeyCode.ESCAPE);
 
+    public static final Set<KeyCode> jumpToFirstIssueKeys = populateJumpToFirstIssueSet();
     // ui.MenuControl
     public static final KeyCodeCombination NEW_ISSUE =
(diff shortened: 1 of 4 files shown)


rd1992:
@damithc okay sure. But i think we should only care about jumping to the __first__ issue. After that the user can use either T/V or the arrow keys to move through the issue list. What do you think?

damithc:
Yes, most of the time jumping to first is enough. Jumping to n'th issue may be useful in searching. e.g. after typing a few key words, the results is narrowed to 5 issues and you notice the 4th one is the one you are looking for.

dariusf:
Needs a rebase, and should probably take the changes in #909 into account.

@lejolly could you review?

lejolly:
sure, @rd1992 just comment here when you're done.

rd1992:
@lejolly done. But again i can pass this locally. Travis is failing though

rd1992:
ran it again and it passed this time

damithc:
Shall we make Ctrl+Down the key to jump to the issue list? It's consistent with the keyboard shortcut used to go from filter box to issue list.

rd1992:
We could. So you want <kbd>1</kbd> , <kbd>Enter</kbd> and <kbd>Ctrl</kbd> + <kbd>Down</kbd> as the three ways of jumping to first issue? 3 different shortcuts seems quite a lot.

damithc:
We don't need enter.
>
>

-- 
Sent from a mobile device.

src/main/java/ui/components/KeyboardShortcuts.java
```diff
@@ -112,6 +112,7 @@
     public static final KeyCodeCombination REMOVE_FOCUS =
             new KeyCodeCombination(KeyCode.ESCAPE);
 
+    public static final Set<KeyCode> jumpToFirstIssueKeys = populateJumpToFirstIssueSet();
```
lejolly: move this higher since it belongs under the `ListPanel` part?

src/test/java/guitests/KeyboardShortcutsTest.java
```diff
@@ -144,7 +159,9 @@ public void keyboardShortcutsTest() {
         assertEquals(5, selectedIssueId);
         
         // minimize window
-        press(KeyCode.CONTROL).press(KeyCode.N).release(KeyCode.N).release(KeyCode.CONTROL); // run this last
+        press(KeyCode.CONTROL).press(KeyCode.N).release(KeyCode.N).release(KeyCode.CONTROL); // run
```
lejolly: for emphasis?

src/test/java/guitests/KeyboardShortcutsTest.java
```diff
@@ -117,22 +130,24 @@ public void keyboardShortcutsTest() {
         int issueIdExpected = issueIdBeforeMark - 1;
         push(getKeyCode("MARK_AS_READ"));
         assertEquals(issueIdExpected, selectedIssueId);
-        push(getKeyCode("UP_ISSUE")); //required since focus has changed to next issue
+        push(getKeyCode("UP_ISSUE")); // required since focus has changed to
```
lejolly: why are all the comments getting chopped up?

lejolly:
ok, waiting for <kbd>Ctrl</kbd>+<kbd>Down</kbd> to be implemented.

src/test/java/guitests/KeyboardShortcutsTest.java
rd1992: okay fixed the chopping part

src/test/java/guitests/KeyboardShortcutsTest.java
```diff
@@ -59,7 +59,20 @@ public void keyboardShortcutsTest() {
         assertEquals(UIComponentFocusEvent.EventType.FILTER_BOX, uiComponentFocusEventType);
         clearUiComponentFocusEventType();
 
+        // jump to first issue using number key(1) or ENTER
+        push(KeyCode.ESCAPE);
+        push(KeyCode.DIGIT1);
```
lejolly: could we get the keycodes directly from `KeyboardShortcuts`?

src/test/java/guitests/KeyboardShortcutsTest.java
```diff
@@ -117,7 +130,7 @@ public void keyboardShortcutsTest() {
         int issueIdExpected = issueIdBeforeMark - 1;
         push(getKeyCode("MARK_AS_READ"));
         assertEquals(issueIdExpected, selectedIssueId);
-        push(getKeyCode("UP_ISSUE")); //required since focus has changed to next issue
+        push(getKeyCode("UP_ISSUE")); // required since focus has changed to next issue
```
lejolly: similar to what this is doing.

src/test/java/guitests/KeyboardShortcutsTest.java
rd1992: we could do it for single key codes. but then what do you suggest for key combinations? Like <kbd>Ctrl</kbd>+<kbd>Down</kbd>. Also since their are so many of such keys we can fix them in a separate issue no?

lejolly: Yeah let's just do it for single key codes first.

rd1992: I've created a separate issue #922 to update all shortcuts here. Shouldn't be part of this PR.

damithc:
Don't forget the documentation.

damithc:
Our usual practice is to assign milestones to issues, not PRs.

dariusf:
Couldn't get this to pass 4 times in a row. :fearful: Also, are reviews done? Can this be included in the release?

lejolly:
looks fine to me, apart from it failing on travis.

PR #902, opened 2015-08-26
Support sort by assignee #889

### LINKED ISSUE #889: Support 'sort:assignee'
As per title

ndt93:
Fixes #889


ndt93:
@dariusf Hey Darius, can help review this? Thanks.

dariusf:
Looks good. Small issue: we already have some tests for sorting [here](https://github.com/HubTurbo/HubTurbo/blob/master/src/test/java/tests/FilterEvalTests.java#L447-L628) and should combine them. Which file should we keep?

The reason for the name "FilterEvalTests" is that the tests in there involve all the filter syntax tree classes, but then "QualifierTests" also makes sense for unit tests.

ndt93:
Hmm, I was thinking that the test is more about the semantics of `getCompoundSortComparator` or `getSortComparator` rather then evaluating the filter expression itself. But there is good reasons for them to be in either files actually, so I'm fine with both. But if they are to move to `QualiferTests` then I think it's better to rename it to `SortingTests` or something along the line.

ndt93:
Ok, so I have moved all sort related tests in `FilterEvalTests` and tests in `QualifierTests` into a new class called `SortQualifierTests`

dariusf:
Okay, sure.

PR #904, opened 2015-08-26
Cater for 2 factor authentication #271

### LINKED ISSUE #271: Cater for 2 factor authentication
We need to look into this too... in case the user has enabled 2FA already

HansNewbie:
Fixes #271

### DIFF
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -8,6 +8,7 @@ import java.io.OutputStream;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
 
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringEscapeUtils;
 import org.apache.logging.log4j.LogManager;
@@ -48,6 +49,8 @@ public class BrowserComponent {
 
     private static final String CHROME_DRIVER_LOCATION = "browserview/";
     private static final String CHROME_DRIVER_BINARY_NAME = determineChromeDriverBinaryName();
+
+    private static final String CHROME_USER_DATA_DIR = System.getProperty("user.home") + File.separator + ".HubTurbo";
 
     private String pageContentOnLoad = "";
 
@@ -90,7 +93,13 @@ public class BrowserComponent {
             driver = createChromeDriver();
             logger.info("Successfully initialised browser component and ChromeDriver");
         });
-        login();
+
+        if (!isChromeCustomProfilePresent()) {
+            login();
+        }
+        else {
+            runBrowserOperation(() -> driver.get(GitHubURL.MAIN_PAGE, false));
+        }
     }
 
     /**
@@ -99,6 +108,34 @@ public class BrowserComponent {
     public void onAppQuit() {
         quit();
         removeChromeDriverIfNecessary();
+    }
+
+    private boolean isChromeCustomProfilePresent() {
+        File chromeCustomProfDir = new File(CHROME_USER_DATA_DIR);
+
+        if (chromeCustomProfDir.isDirectory()) {
+            String[] chromeCustomProfDirList = chromeCustomProfDir.list();
+            if (chromeCustomProfDirList != null && chromeCustomProfDirList.length > 0) {
+                return true;
+            }
+            else { // directory is empty
+                return false;
+            }
+        }
+        else { // directory does not exist yet
+            return false;
+        }
+    }
+
+    /**
+     * Reset Chrome Custom Profile directory so that all cookies are cleared
+     */
+    public void cleanChromeCustomProfile() {
+        try {
+            FileUtils.cleanDirectory(new File(CHROME_USER_DATA_DIR));
+        } catch (IOException e) {
+            logger.error(e.getLocalizedMessage(), e);
+        }
     }
 
     /**
@@ -123,6 +160,7 @@ public class BrowserComponent {
      */
     private ChromeDriverEx createChromeDriver() {
         ChromeOptions options = new ChromeOptions();
+        options.addArguments("user-data-dir=" + CHROME_USER_DATA_DIR);
         if (USE_MOBILE_USER_AGENT) {
             options.addArguments(String.format("user-agent=\"%s\"", MOBILE_USER_AGENT));
         }
--- a/src/main/java/ui/UI.java
+++ b/src/main/java/ui/UI.java
@@ -124,7 +124,7 @@ public class UI extends Application implements EventDispatcher {
                         showMainWindow(logic.loginController.getRepoId());
                         disableUI(false);
                     } else {
-                        quit();
+                        quit(false);
                     }
                     return true;
                 }).exceptionally(e -> {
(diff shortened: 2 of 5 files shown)


HansNewbie:
I need further discussion on this.

What I have done is create a Chrome custom profile stored in `~/.HubTurbo`. This allows Chrome to use the folder to store custom settings, extensions and **cookies**.

I have several things in mind:

1. Is it secure enough since cookies is involved?
2. Since cookies is stored, the session is recovered. This brings a benefit and a trouble.
   The benefit is that when user reopens HubTurbo, they won't be directed to the login page in browser view anymore.

   The trouble is that if user logs out of HubTurbo, the browser session is still logged in. E.g. I use my alternative GitHub account for HubTurbo, but I log out to use my main account. The browser will still be in session of my alternative account.

   I can think of 2 solutions:
  1. Log out of GitHub by crawling the log out button.
  2. Clear cookies or session preference if user log out. (sounds easier, need to dig some info)

Any suggestions? @dariusf @damithc

damithc:
We don't need to try too hard for this one though. I don't think many users have 2FA enabled for github.

damithc:
If there is an easy solution, go ahead. If no easy solution, we can KIV this for now.

src/main/java/ui/UI.java
```diff
@@ -307,7 +307,7 @@ private void setupMainStage(Scene scene) {
         mainStage.setTitle("HubTurbo " + Utility.version(VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH));
         mainStage.setScene(scene);
         mainStage.show();
-        mainStage.setOnCloseRequest(e -> quit());
+        mainStage.setOnCloseRequest(e -> quit(false));
```
HansNewbie: I think having a constant `USER_LOGOUT` which value is `true` will help readability as to why we need to pass a parameter to `quit`. I am not sure where can I put such constant

HansNewbie:
So basically I added a parameter `isLogout` to quitting HubTurbo. If user logs out, then this parameter will be true, which will make `browserComponent` to delete cookies from github.com before it quits.

src/main/java/ui/UI.java
dariusf: It may be better to make a new method for logging out instead, which does the cookie cleanup, then calls `quit`.

dariusf:
Hmm, so 2FA already works with HT, and this just persists the session so we don't have to keep keying in the 2FA code?

src/main/java/ui/UI.java
HansNewbie: Okay I will see what I can do.

HansNewbie:
@dariusf yes. The only caveat is that on logging in, the browser view will give error since HubTurbo will try to log in using the application token. So user will need to key in their password and 2FA code manually once.

dariusf:
Hmm, is that possible to fix? I know keying in the 2FA code is unavoidable. The error happens when HT goes to the login page and inputs the username/password?

HansNewbie:
I don't think it's possible to fix, unless we create another login dialog to cater to 2FA enabled user. Because for 2FA users, they need their application token for the application to call API, their password to login into the browser as well as their 2FA code. In this case, the login dialog should have one field for app token and one field for password.

dariusf:
I think this is useful and we can merge it, but maybe we shouldn't close the issue since there is room for future work to fully support 2FA. @damithc?

damithc:
Sure, we can merge. I guess it's neater to close this and open a new issue for the remaining work? We don't have a defined workflow for keeping issues open while merging the PR.

dariusf:
I checked out the branch to try it, and it works well. One small issue, in addition to my previous comments: user login is still carried out in the browser.
```
14061 31 Aug 2015 14:49:24 INFO BrowserComponent [pool-7-thread-1] - Unable to login, may already be logged in.
```
Visually nothing goes wrong, but we should have finer-grained error-handling there and log what is actually happening, so we can diagnose if something goes wrong in future.

HansNewbie:
@dariusf I can't solve the problem when the browser view is closed then user logs out. The browser view will be reset, but it will not quit, neither will HubTurbo. The last error message would be `ERROR UI [JavaFX Application Thread] - no such window: target window already closed`. Then it will open github.com to delete the cookies, but instead of that Logic will refresh the repo instead.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -101,6 +106,19 @@ public void onAppQuit() {
         removeChromeDriverIfNecessary();
     }
 
+    public void logout() {
```
HansNewbie: I made logout method in `BrowserComponent`

src/main/java/ui/UI.java
```diff
@@ -268,7 +268,10 @@ private boolean isCloseOnQuit() {
         return commandLineArgs.getOrDefault("closeonquit", "false").equalsIgnoreCase("true");
     }
 
-    public void quit() {
+    public void quit(boolean isLogout) {
```
HansNewbie: quit will still have boolean to denote logout until UI level. Is this fine? Or should logout be called by other component?

HansNewbie:
@dariusf I will look at the error message for user login.

HansNewbie:
@dariusf should not have anymore login error message due to saved session.

In summary:
- Create Chrome Custom Profile for HubTurbo which will save session.
- Upon logging out of HubTurbo, this Custom Profile is deleted completely (so as to clear the session).
- As such, on starting `browserComponent`, it will check if Custom Profile exists. If it does, first page would be github.com instead of github.com/login.

Given now login page would not be loaded at the start unless user is logged out, should the message still be changed? I can't think of any case where going to github.com/login would create exception unless the user is already logged in (in which the logger message is already correct).

src/main/java/browserview/BrowserComponent.java
```diff
@@ -90,7 +95,13 @@ public void initialise() {
             driver = createChromeDriver();
             logger.info("Successfully initialised browser component and ChromeDriver");
         });
-        login();
+
+        if (!isChromeCustomProfileExist()) {
```
dariusf: `isChromeCustomProfilePresent` may be a better name.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -31,6 +32,8 @@
 import com.sun.jna.platform.win32.WinDef.HWND;
 import com.sun.jna.platform.win32.WinUser;
 
+import static com.sun.jna.platform.FileUtils.*;
```
dariusf: Is this used? I think `cleanDirectory` is from Apache Commons.

src/main/java/ui/UI.java
dariusf: What I was thinking was

```java
class UI {
    void logout() {
        browserComponent.cleanChromeCustomProfile();
        quit();
    }
    void quit() {
        browserComponent.onAppQuit();
        // whatever else is here
    }
}
```
It's a minor thing though, just cosmetic. The UI mostly just interfaces directly with the browser component so there's not much else involved.

dariusf:
The log message is fine then, since there's now an explicit check.

src/main/java/ui/UI.java
HansNewbie: This will result in some code repetition because I need to call `browserComponent.onAppQuit()` before `browserComponent.cleanChromeCustomProfile()`, reason being if I clean the profile before the quit, some files would not be cleaned.

dariusf: Right, never mind then. We shouldn't needlessly complicate things.

PR #905, opened 2015-08-26
User can undo a change to the label #870

### LINKED ISSUE #870: User can undo a change to the label
... so that user can use the label picker fast in the safety of the knowledge that any mistake can be undone.
similar to Gmail 'undo send' feature?

lejolly:
Fixes #870

undo

Changes are pushed to GitHub immediately, notification pane auto-hides after 5 seconds, undo-ing changes pushes the original labels to GitHub.

### DIFF
--- a/src/main/java/ui/UI.java
+++ b/src/main/java/ui/UI.java
@@ -9,6 +9,7 @@ import com.sun.jna.platform.win32.User32;
 import com.sun.jna.platform.win32.WinDef.HWND;
 import javafx.application.Application;
 import javafx.application.Platform;
+import javafx.event.ActionEvent;
 import javafx.geometry.Pos;
 import javafx.scene.Node;
 import javafx.scene.Parent;
@@ -206,7 +207,7 @@ public class UI extends Application implements EventDispatcher {
             status::updateTimeToRefresh, logic::refresh, TimeUnit.SECONDS);
         refreshTimer.start();
         notificationPaneTimer = new TickingTimer("Notification Pane Timer", NOTIFICATION_PANE_VISIBLE_PERIOD,
-                integer -> {}, this::hideNotificationPane, TimeUnit.SECONDS);
+                integer -> {}, () -> Platform.runLater(this::hideNotificationPane), TimeUnit.SECONDS);
         notificationPaneTimer.start();
     }
 
@@ -552,8 +553,8 @@ public class UI extends Application implements EventDispatcher {
     }
 
     public void showNotificationPane(Node graphic, String text, Action action) {
-        hideNotificationPane();
         Platform.runLater(() -> {
+            hideNotificationPane();
             notificationPane.setGraphic(graphic);
             notificationPane.setText(text);
             notificationPane.getActions().clear();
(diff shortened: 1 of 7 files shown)


lejolly:
@acruis can help me check how it looks like on windows? does it cover the menu bar?

acruis:
Yup, it does.

image

lejolly:
that's ok right? as for the travis build, it fails because controlsfx depends on a javafx library that got shifted in 8u60, see: https://bitbucket.org/controlsfx/controlsfx/issues/526/comsunjavafxutils-changes-to

damithc:
Can bind to Ctrl+Z?
I'm not sure about the location of the message though. If we are going to
support undo more widely, may be we should find a more permanent location
to show the most recent action?
How long is the undo window?

acruis:
Yeah, it's ok. It looks great :)

So Travis is using 8u60?

lejolly:
travis downloads and installs `oracle-java8-installer` upon every run because by default they're still using `8u31` which doesn't support the new native dialogs (which were taken from controlsfx), so the latest release is `8u60` which is why this happens. two possible solutions, i find an updated version of controlsfx or somehow force travis to use `8u51`.

lejolly:
@damithc yes i can bind it to <kbd>Ctrl</kbd>+<kbd>Z</kbd>, the location is meant both to be obvious and not intrusive (as opposed to a pop-up) because it goes away after 5 seconds. the time available for the undo is the time when the notification pane is showing (which is 5 seconds by default).

damithc:
I see. Another label change during that 5 sec will make the previous change
no longer undoable right? is it like that?

lejolly:
yup.

lejolly:
@acuris ready for review!

acruis:
Okay @lejyllo!

acruis:
Good to merge! Can consider additional delay timings, the label picker window takes more than 1s on first run of the test for my com.

But then, that might just be my com being slow. I leave it up to you to decide whether to extend delay and then merge =D

lejolly:
ah ok, i'll change it back to 1.5s

dariusf:
Hmm, I think there's a slight concurrency issue. Could you wait a bit before merging? Trying to reason it out now.

lejolly:
sure thing.

src/main/java/ui/UI.java
```diff
@@ -544,4 +552,36 @@ public HWND getMainWindowHandle() {
         return mainWindowHandle;
     }
 
+    public void showNotificationPane(Node graphic, String text, Action action) {
+        hideNotificationPane();
+        Platform.runLater(() -> {
```
dariusf: What happens with this line and the previous is something like this:

```java
void showNotificationPanel(...) {
    Platform.runLater(() -> {
        // hide
    });
    Platform.runLater(() -> {
        // show
    });
}
```

The `show` block is guaranteed to follow the `hide` block in a sequential environment, but if `showNotificationPanel` is called from many threads concurrently, the `show` and `hide` blocks may interleave. Any permutation where `hideN` precedes `showN` is possible:

```
hide1 show1 hide2 show2 // normal
hide2 show2 hide1 show1 // also okay
hide1 hide2 show1 show2 // unexpected things may happen
hide2 hide1 show1 show2 // this too
```
I've been staring at it for a while and I don't *think* there is incorrect behaviour right now, but there might be in future. :grimacing: Just to be on the safe side we should ensure that the hiding, then showing is done as one atomic action.

Another issue is that if there are multiple concurrent calls, order between them is undefined and only the last will appear. We should buffer the notification requests somehow... but that's probably for another issue.

lejolly: good point, will change it. as for the buffer yeah, that'll be a separate issue =P

PR #908, opened 2015-08-28
'-label:p.' does not match issues without label group 'p' #903

### LINKED ISSUE #903: '-label:p.' does not match issues without label group 'p'
In TEAMMATES, priority labels are given as `p.*`. I was trying to use `-label:p.` to find issues without a priority, but it did not work.

dariusf:
Fixes #903.

### DIFF
--- a/src/test/java/tests/FilterEvalTests.java
+++ b/src/test/java/tests/FilterEvalTests.java
@@ -1,24 +1,22 @@
 package tests;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.fail;
+import static junit.framework.TestCase.assertFalse;
+import static org.junit.Assert.*;
 
 import java.time.LocalDateTime;
+import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


src/test/java/tests/FilterEvalTests.java
```diff
@@ -158,6 +156,22 @@ public void label() {
         } catch (ParseException ignored) {
         }
 
+        // Ensure that early termination does not occur
```
acruis: Unfortunately, without the commit message, it took me a while to figure out what the "early termination" refers to (I thought it was early termination of the `label()` test at first, but that didn't make sense).

Is it possible to clarify that this refers to the new `labelsSatisfy` loop?

dariusf: Sure, done. Also added a comment to `labelsSatisfy` to explain the approach.

acruis:
OK, good to merge.

PR #909, opened 2015-08-28
 Ctrl+R is triggering a 'reply' action #907 

### LINKED ISSUE #907: Ctrl+R is triggering a 'reply' action
When I use the Ctrl+R shortcut, it changes the repo, but also trigger a 'reply' action.

HyungJon:
Fixes #907

### DIFF
--- a/src/main/java/ui/MenuControl.java
+++ b/src/main/java/ui/MenuControl.java
@@ -7,7 +7,6 @@ import javafx.scene.Node;
 import javafx.scene.control.*;
 import javafx.scene.control.Alert.AlertType;
 import javafx.scene.control.ButtonBar.ButtonData;
-import javafx.scene.input.KeyCodeCombination;
 import javafx.stage.Modality;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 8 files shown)


dariusf:
I think a better solution here would be to change all `KeyCode`s into `KeyCodeCombination`s, and use `match` instead of `==` to check equality everywhere.

```java
public static final KeyCodeCombination NEW_COMMENT = new KeyCodeCombination(KeyCode.R);

if (KeyboardShortcuts.NEW_COMMENT.match(event) && ...) {
    // ...
}
```

Using ad-hoc conditional checks will get messy when we start using more chords -- <kbd>Shift</kbd>+<kbd>R</kbd> will require another exception, so will <kbd>Shift</kbd>+<kbd>Ctrl</kbd>+<kbd>R</kbd>, etc.

Also, we can static import the `KeyboardShortcuts` enums and `KeyCode` constants in files where we use them a lot.

HyungJon:
Done except importing the enums part.

dariusf:
Why not, explicitness? It seems awfully verbose.

I think the `KeyPress` interface should be changed as well, but we'll do that in the other issue.

HyungJon:
Just a status report, I'm working on it. I'm very pressed with overlapping assignments right now though. Can it be done in an another issue, maybe along with that ```KeyPress``` interface changing?

dariusf:
No worries. In that case there's nothing more to do here.

PR #911, opened 2015-08-31
Double space while in the middle of a filter leaves the spaces in the filter #897

### LINKED ISSUE #897: Double space while in the middle of a filter leaves the spaces in the filter
I could end up with something like `is:is  sue`

lejolly:
Fixes #897


acruis:
Good to merge.

Issue #912, opened 2015-08-31
`updated` filter missing some issues


damithc:
![image](https://cloud.githubusercontent.com/assets/1673303/9573105/db9049e2-4fea-11e5-8606-1eed39854688.png)

damithc:
Deleting the cache fixed the problem. There's something fishy going on w.r.t. the cache data. May be   related to the broken milestone problem? That too get fixed when I delete the cache.

damithc:
Here's the filter I had in use `(repo:teammates/repo OR repo:hubturbo/hubturbo OR repo:powerpointlabs/powerpointlabs) updated:12 is:unread sort:nonSelfUpdate`
I didn't check the cache at the time.
Simply restarting did not fix the problem. There were some issues missing even after a restart. One new PR that was created during the window was missing entirely.

ndt93:
Ok noted. I will troubleshoot the milestones bug first then come back to this unless some one wants to take on this in the mean time.

PR #924, opened 2015-09-01
Add a way to filter out self updates from the 'updated' filters

### LINKED ISSUE #898: Add a way to filter out self updates from the 'updated' filter
e.g. some special character to indicate that self updates should be ignored

codemanmeet:
Fixes #898

### DIFF
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -108,8 +108,8 @@ This will not work for ambiguous expressions (containing OR or NOT operators) an
 - [`is`](#is)
 - [`created`](#created)
 - [`updated`](#updated)
-- [`updatedByOthers`](#updatedByOthers)
-- [`updatedBySelf`](#updatedBySelf)
+- [`updated-others`](#updated-others)
+- [`updated-self`](#updated-self)
 - [`repo`](#repo)
 - [`sort`](#sort)
 
@@ -222,19 +222,19 @@ Matches issues which were created on a given date, or within a given date range.
 
 *Expects a number or  number range*
 
-Matches issues which were updated in the given number of hours. For example, `updated:<24` would match issues updated in the last day. If a number `n` is given, it is implicitly translated to `<n`.
+Matches issues which were updated in the given number of hours. For example, `updated:<24` would match issues updated in the last day. If a number `n` is given, it is implicitly translated to `<n`. Number ranges are written using a relational operator (.e.g `>5`, `<=10`).
 
-### updatedByOthers
-
-*Expects a number or  number range*
-
-Matches issues which were updated in the given number of hours by users other than the logged in user. For example, `updated:<24` would match issues updated in the last day by other users. If a number `n` is given, it is implicitly translated to `<n`.
-
-### updatedBySelf
+### updated-others
 
 *Expects a number or  number range*
 
-Matches issues which were updated in the given number of hours by the logged in user. For example, `updated:<24` would match issues updated in the last day by the logged in user. If a number `n` is given, it is implicitly translated to `<n`.
+Matches issues which were updated in the given number of hours by users other than the logged in user. For example, `updated:<24` would match issues updated in the last day by other users. Number ranges are written using a relational operator (.e.g `>5`, `<=10`). If a number `n` is given, it is implicitly translated to `<n`.
+
+### updated-self
+
+*Expects a number or  number range*
+
+Matches issues which were updated in the given number of hours by the logged in user. For example, `updated:<24` would match issues updated in the last day by the logged in user. Number ranges are written using a relational operator (.e.g `>5`, `<=10`). If a number `n` is given, it is implicitly translated to `<n`.
 
 ### repo
 
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -108,6 +108,8 @@ This will not work for ambiguous expressions (containing OR or NOT operators) an
 - [`is`](#is)
 - [`created`](#created)
 - [`updated`](#updated)
+- [`updatedByOthers`](#updatedByOthers)
+- [`updatedBySelf`](#updatedBySelf)
 - [`repo`](#repo)
 - [`sort`](#sort)
 
@@ -221,6 +223,18 @@ Matches issues which were created on a given date, or within a given date range.
 *Expects a number or  number range*
 
 Matches issues which were updated in the given number of hours. For example, `updated:<24` would match issues updated in the last day. If a number `n` is given, it is implicitly translated to `<n`.
+
+### updatedByOthers
+
+*Expects a number or  number range*
+
+Matches issues which were updated in the given number of hours by users other than the logged in user. For example, `updated:<24` would match issues updated in the last day by other users. If a number `n` is given, it is implicitly translated to `<n`.
+
+### updatedBySelf
+
+*Expects a number or  number range*
+
+Matches issues which were updated in the given number of hours by the logged in user. For example, `updated:<24` would match issues updated in the last day by the logged in user. If a number `n` is given, it is implicitly translated to `<n`.
 
 ### repo
 
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -228,13 +228,13 @@ Matches issues which were updated in the given number of hours. For example, `up
 
 *Expects a number or  number range*
 
-Matches issues which were updated in the given number of hours by users other than the logged in user. For example, `updated:<24` would match issues updated in the last day by other users. Number ranges are written using a relational operator (.e.g `>5`, `<=10`). If a number `n` is given, it is implicitly translated to `<n`.
+Like [`updated`](#updated), but considers only issues updated by others (not the logged in user).
 
 ### updated-self
 
 *Expects a number or  number range*
 
-Matches issues which were updated in the given number of hours by the logged in user. For example, `updated:<24` would match issues updated in the last day by the logged in user. Number ranges are written using a relational operator (.e.g `>5`, `<=10`). If a number `n` is given, it is implicitly translated to `<n`.
+Like [`updated`](#updated), but considers only issues updated by the logged in user.
 
 ### repo
 
--- a/src/main/java/backend/IssueMetadata.java
+++ b/src/main/java/backend/IssueMetadata.java
@@ -4,8 +4,10 @@ import github.TurboIssueEvent;
 import org.eclipse.egit.github.core.Comment;
 
 import java.time.LocalDateTime;
+import java.time.ZoneId;
 import java.time.ZoneOffset;
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.List;
 
 public class IssueMetadata {
@@ -16,10 +18,13 @@ public class IssueMetadata {
     private final int nonSelfCommentCount;
     private final int selfCommentCount;
 
-    // If isUpdated is true, nonSelfUpdatedAt will be used to sort/filter instead of updatedAt in TurboIssue
-    private final boolean isUpdated;
-    private final boolean isUpdatedBySelf;
-    private final boolean isUpdatedByOthers;
+    private final boolean isUpdated; //for any update
+    private final boolean isUpdatedBySelf; // for update by self
+    private final boolean isUpdatedByOthers; // for update by others
+
+    private enum UpdatedKind {
+        SELF_UPDATED, OTHER_UPDATED
+    }
 
     // Constructor for default use when initializing TurboIssue
     public IssueMetadata() {
@@ -74,18 +79,86 @@ public class IssueMetadata {
     }
 
     // Constructor used in Logic
-    public IssueMetadata(IssueMetadata other, LocalDateTime nonSelfUpdatedAt, LocalDateTime selfUpdatedAt,
-                         int nonSelfCommentCount, int selfCommentCount,
-                         boolean isUpdatedByOthers, boolean isUpdatedBySelf) {
-        this.events = new ArrayList<>(other.events);
-        this.comments = new ArrayList<>(other.comments);
-        this.nonSelfUpdatedAt = nonSelfUpdatedAt; // Calculated just prior to calling this constructor
-        this.selfUpdatedAt = selfUpdatedAt;
-        this.nonSelfCommentCount = nonSelfCommentCount;
-        this.selfCommentCount = selfCommentCount;
+    public IssueMetadata(IssueMetadata existingMetadata, String currentUser) {
+        Date lastNonSelfUpdate = new Date(0);
+        Date lastSelfUpdate = new Date(0);
+        boolean isUpdatedByOthers = false;
+        boolean isUpdatedBySelf = false;
+        for (TurboIssueEvent event : existingMetadata.getEvents()) {
+            if (isNewEventByOther(currentUser, event, lastNonSelfUpdate)) {
+                lastNonSelfUpdate = event.getDate();
+                isUpdatedByOthers = true;
+            }
+            if (isNewEventBySelf(currentUser, event, lastSelfUpdate)){
+                lastSelfUpdate = event.getDate();
+                isUpdatedBySelf = true;
+            }
+        }
+        for (Comment comment : existingMetadata.getComments()) {
+            if (isNewCommentByOther(currentUser, comment, lastNonSelfUpdate)) {
+                lastNonSelfUpdate = comment.getCreatedAt();
+                isUpdatedByOthers = true;
+            }
+            if (isNewCommentBySelf(currentUser, comment, lastSelfUpdate)){
+                lastSelfUpdate = comment.getCreatedAt();
+                isUpdatedBySelf = true;
+            }
+        }
+
+        this.events = new ArrayList<>(existingMetadata.events);
+        this.comments = new ArrayList<>(existingMetadata.comments);
+        this.nonSelfUpdatedAt = LocalDateTime.ofInstant(lastNonSelfUpdate.toInstant(), ZoneId.systemDefault());
+        this.selfUpdatedAt = LocalDateTime.ofInstant(lastSelfUpdate.toInstant(), ZoneId.systemDefault());
+        this.nonSelfCommentCount = calculateCommentCount(existingMetadata.getComments(), currentUser,
+                UpdatedKind.OTHER_UPDATED);
+        this.selfCommentCount = calculateCommentCount(existingMetadata.getComments(), currentUser,
+                UpdatedKind.SELF_UPDATED);
         this.isUpdated = true;
         this.isUpdatedByOthers = isUpdatedByOthers;
         this.isUpdatedBySelf = isUpdatedBySelf;
+
+    }
+
+    private static boolean isNewCommentByOther(String currentUser, Comment comment, Date lastNonSelfUpdate){
+        return !comment.getUser().getLogin().equalsIgnoreCase(currentUser)
+                && comment.getCreatedAt().after(lastNonSelfUpdate);
+    }
+
+    private static boolean isNewCommentBySelf(String currentUser, Comment comment, Date lastSelfUpdate){
+        return comment.getUser().getLogin().equalsIgnoreCase(currentUser)
+                && comment.getCreatedAt().after(lastSelfUpdate);
+    }
+
+    private static boolean isNewEventBySelf(String currentUser, TurboIssueEvent event, Date lastSelfUpdate){
+        return event.getActor().getLogin().equalsIgnoreCase(currentUser)
+                && event.getDate().after(lastSelfUpdate);
+    }
+
+    private static boolean isNewEventByOther(String currentUser, TurboIssueEvent event, Date lastNonSelfUpdate){
+        return !event.getActor().getLogin().equalsIgnoreCase(currentUser)
+                && event.getDate().after(lastNonSelfUpdate);
+    }
+
+    private static int calculateCommentCount(List<Comment> comments, String currentUser, UpdatedKind updatedKind) {
+        int result = 0;
+        if (updatedKind == UpdatedKind.OTHER_UPDATED) {
+            for (Comment comment : comments) {
+                if (!isCommentBySelf(currentUser, comment)) {
+                    result++;
+                }
+            }
+        } else if (updatedKind == UpdatedKind.SELF_UPDATED){
+            for (Comment comment : comments) {
+                if (isCommentBySelf(currentUser, comment)) {
+                    result++;
+                }
+            }
+        }
+        return result;
+    }
+
+    private static boolean isCommentBySelf(String currentUser, Comment comment){
+        return comment.getUser().getLogin().equalsIgnoreCase(currentUser);
     }
 
     // Constructor used in MultiModel
@@ -99,6 +172,21 @@ public class IssueMetadata {
         this.isUpdated = other.isUpdated;
         this.isUpdatedBySelf = other.isUpdatedBySelf;
         this.isUpdatedByOthers = other.isUpdatedByOthers;
+    }
+
+    //Constructor used in FilterEvalTests
+    public IssueMetadata(IssueMetadata other, LocalDateTime nonSelfUpdatedAt, LocalDateTime selfUpdatedAt,
+                         int nonSelfCommentCount, int selfCommentCount,
+                         boolean isUpdatedByOthers, boolean isUpdatedBySelf) {
+        this.events = new ArrayList<>(other.events);
+        this.comments = new ArrayList<>(other.comments);
+        this.nonSelfUpdatedAt = nonSelfUpdatedAt; // Calculated just prior to calling this constructor
+        this.selfUpdatedAt = selfUpdatedAt;
+        this.nonSelfCommentCount = nonSelfCommentCount;
+        this.selfCommentCount = selfCommentCount;
+        this.isUpdated = true;
+        this.isUpdatedBySelf = isUpdatedBySelf;
+        this.isUpdatedByOthers = isUpdatedByOthers;
     }
 
     public String summarise() {
--- a/src/main/java/backend/IssueMetadata.java
+++ b/src/main/java/backend/IssueMetadata.java
@@ -142,17 +142,9 @@ public class IssueMetadata {
     private static int calculateCommentCount(List<Comment> comments, String currentUser, UpdatedKind updatedKind) {
         int result = 0;
         if (updatedKind == UpdatedKind.OTHER_UPDATED) {
-            for (Comment comment : comments) {
-                if (!isCommentBySelf(currentUser, comment)) {
-                    result++;
-                }
-            }
+            return (int)(comments.stream().filter(c -> !isCommentBySelf(currentUser, c)).count());
         } else if (updatedKind == UpdatedKind.SELF_UPDATED){
-            for (Comment comment : comments) {
-                if (isCommentBySelf(currentUser, comment)) {
-                    result++;
-                }
-            }
+            return (int)(comments.stream().filter(c -> isCommentBySelf(currentUser, c)).count());
         }
         return result;
     }
--- a/src/main/java/backend/IssueMetadata.java
+++ b/src/main/java/backend/IssueMetadata.java
@@ -142,9 +142,9 @@ public class IssueMetadata {
     private static int calculateCommentCount(List<Comment> comments, String currentUser, UpdatedKind updatedKind) {
         int result = 0;
         if (updatedKind == UpdatedKind.OTHER_UPDATED) {
-            return (int)(comments.stream().filter(c -> !isCommentBySelf(currentUser, c)).count());
+            return (int) (comments.stream().filter(c -> !isCommentBySelf(currentUser, c)).count());
         } else if (updatedKind == UpdatedKind.SELF_UPDATED){
-            return (int)(comments.stream().filter(c -> isCommentBySelf(currentUser, c)).count());
+            return (int) (comments.stream().filter(c -> isCommentBySelf(currentUser, c)).count());
         }
         return result;
     }
--- a/src/main/java/backend/IssueMetadata.java
+++ b/src/main/java/backend/IssueMetadata.java
@@ -12,18 +12,26 @@ public class IssueMetadata {
     private final List<TurboIssueEvent> events;
     private final List<Comment> comments;
     private final LocalDateTime nonSelfUpdatedAt;
+    private final LocalDateTime selfUpdatedAt;
     private final int nonSelfCommentCount;
+    private final int selfCommentCount;
 
     // If isUpdated is true, nonSelfUpdatedAt will be used to sort/filter instead of updatedAt in TurboIssue
     private final boolean isUpdated;
+    private final boolean isUpdatedBySelf;
+    private final boolean isUpdatedByOthers;
 
     // Constructor for default use when initializing TurboIssue
     public IssueMetadata() {
         events = new ArrayList<>();
         comments = new ArrayList<>();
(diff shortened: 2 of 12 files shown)


src/main/java/backend/IssueMetadata.java
```diff
@@ -12,18 +12,26 @@
...
     private final List<Comment> comments;
     private final LocalDateTime nonSelfUpdatedAt;
+    private final LocalDateTime selfUpdatedAt;
     private final int nonSelfCommentCount;
+    private final int selfCommentCount;
 
     // If isUpdated is true, nonSelfUpdatedAt will be used to sort/filter instead of updatedAt in TurboIssue
     private final boolean isUpdated;
+    private final boolean isUpdatedBySelf;
+    private final boolean isUpdatedByOthers;
```
dariusf: Is there a difference between `isUpdatedByOthers` and `isUpdated`? See the comment above.

src/main/java/backend/IssueMetadata.java
```diff
@@ -49,17 +65,27 @@ public IssueMetadata(List<TurboIssueEvent> events, List<Comment> comments) {
...
         this.events = new ArrayList<>(other.events);
         this.comments = new ArrayList<>(other.comments);
         this.nonSelfUpdatedAt = nonSelfUpdatedAt; // Calculated just prior to calling this constructor
-        this.nonSelfCommentCount = nonSelfCommentCount; // Calculated just prior to calling
+        this.selfUpdatedAt = selfUpdatedAt;
+        this.nonSelfCommentCount = nonSelfCommentCount; // Calculated just prior to calling this constructor
+        this.selfCommentCount = selfCommentCount; //Calculated just prior to calling this constructor
         this.isUpdated = true;
+        this.isUpdatedByOthers = isUpdatedByOthers; //Calculated prior to calling this constructor
+        this.isUpdatedBySelf = isUpdatedBySelf; // Calculated prior to calling this constructor
```
dariusf: I don't think these comments are necessary. The constructor receives these values as parameters so it's kind of a given.

src/main/java/backend/Logic.java
```diff
@@ -157,32 +160,57 @@ public void refresh() {
...
+                    isUpdatedBySelf = true;
                 }
             }
 
             entry.setValue(new IssueMetadata(currentMetadata,
                     LocalDateTime.ofInstant(lastNonSelfUpdate.toInstant(), ZoneId.systemDefault()),
-                    calculateNonSelfCommentCount(currentMetadata.getComments(), currentUser)
+                    LocalDateTime.ofInstant(lastSelfUpdate.toInstant(), ZoneId.systemDefault()),
+                    calculateCommentCount(currentMetadata.getComments(), currentUser, IssueType.OTHER_UPDATED),
+                    calculateCommentCount(currentMetadata.getComments(), currentUser, IssueType.SELF_UPDATED),
```
dariusf: All this logic should be in the `IssueMetadata` constructor (or some factory method), and broken down further into different methods. This method is doing too much, plus `IssueMetadata` should handle its own construction.

src/main/java/backend/Logic.java
```diff
@@ -157,32 +160,57 @@ public void refresh() {
...
+            for (Comment comment : comments) {
+                if (!comment.getUser().getLogin().equalsIgnoreCase(currentUser)) {
+                    result++;
+                }
+            }
+        } else if (issueType == IssueType.SELF_UPDATED){
+            for (Comment comment : comments) {
+                if (comment.getUser().getLogin().equalsIgnoreCase(currentUser)) {
+                    result++;
+                }
```
dariusf: The conditional logic should be factored out into a predicate so we don't have to repeat ourselves. This may also be formulated more clearly in terms of a `filter`, then `count`, but we can try the predicate first.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -25,9 +25,14 @@
 public class Qualifier implements FilterExpression {
 
     public static final String UPDATED = "updated";
+    public static final String UPDATED_OTHERS = "updatedByOthers";
     public static final String REPO = "repo";
     public static final String SORT = "sort";
 
+    private enum IssueType{
```
dariusf: `IssueType` is too generic a name. How about `UpdatedKind`, to make it more consistent with how it's named below?

src/main/java/filter/expression/Qualifier.java
```diff
@@ -35,7 +40,8 @@
         "date", "nonSelfUpdate", "desc", "description", "has", "id", "in", "involves",
         "is", "issue", "keyword", "label", "labels", "merged", "milestone", "milestones",
         "no", "open", "pr", "pullrequest", "read", "repo", "sort", "state", "status",
-        "title", "type", "unmerged", "unread", "updated", "user"
+        "title", "type", "unmerged", "unread", "updated", "updatedByOthers",
+        "updatedBySelf", "user"
```
dariusf: It'd be good if we could put `nonSelfUpdate` in alphabetical position as well.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -527,17 +537,37 @@ private boolean satisfiesUpdatedHours(TurboIssue issue) {
...
+                    hoursSinceUpdate = Utility.safeLongToInt(issue.getMetadata().getNonSelfUpdatedAt()
+                            .until(getCurrentTime(), ChronoUnit.HOURS));
+                }
+                break;
+            case ALL_UPDATED:
+                if (issue.getMetadata().isUpdated()) {
+                    // Second time being filtered, we now have metadata from source, so we can use getNonSelfUpdatedAt.
+                    hoursSinceUpdate = Utility.safeLongToInt(issue.getUpdatedAt()
+                            .until(getCurrentTime(), ChronoUnit.HOURS));
+                }
```
dariusf: We should use the conditional to decide the date field to use, instead of the hours. The hours can be computed based on the date field after that.

Also, the comments are duplicated.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -527,17 +537,37 @@ private boolean satisfiesUpdatedHours(TurboIssue issue) {
...
         }
 
+        // First time being filtered (haven't gotten metadata from source yet).
+        if (hoursSinceUpdate == Integer.MIN_VALUE) {
+            hoursSinceUpdate = Utility.safeLongToInt(issue.getUpdatedAt().until(getCurrentTime(), ChronoUnit.HOURS));
```
dariusf: This case can be placed first, since it's shorter. Same issue -- we should decide the date instead of the hours.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -60,7 +60,9 @@ public ListPanel(UI ui, IModel model, PanelControl parentPanelControl, int panel
      * @return true if the issue has changed, false otherwise
      */
     private boolean issueHasNewComments(TurboIssue issue, boolean hasMetadata) {
-        if (currentFilterExpression.getQualifierNames().contains(Qualifier.UPDATED) && hasMetadata) {
+        List<String> filterQualifierNames = currentFilterExpression.getQualifierNames();
+        if ((filterQualifierNames.contains(Qualifier.UPDATED) ||
```
dariusf: Do we need to check for the self update qualifier as well?

`hasMetadata` could be put first the boolean expression. The presence of qualifiers can be checked into a static `Qualifier` helper method.

src/test/java/tests/FilterEvalTests.java
```diff
@@ -425,19 +425,39 @@ public void updated() {
         issue.setUpdatedAt(now.minusDays(2));
 
         assertEquals(false, matches("updated:<24", issue));
+        assertEquals(false, matches("updatedByOthers:<24", issue));
+        assertEquals(false, matches("updatedBySelf:<24", issue));
```
dariusf: These tests don't cover what was changed. You have to stub out each the test issue's metadata with events and comments, and ensure that the right qualifier matches in the right case. There are 2^3 cases to cover.

src/main/java/ui/listpanel/ListPanel.java
codemanmeet: Nope, we don't. We care about comments/events from other users in this context.

src/main/java/filter/expression/Qualifier.java
codemanmeet: Changed.

src/test/java/tests/FilterEvalTests.java
codemanmeet: I added tests to check for `updatedByOthers` and `updatedBySelf` separately with the `IssueMetadata` construction to indicate when the issue was last self updated or last updated by others.

src/main/java/filter/expression/Qualifier.java
codemanmeet: Changed.

src/main/java/ui/listpanel/ListPanel.java
dariusf: Okay. I guess you could add a comment explaining this, since you're purposefully omitting the self check.

src/main/java/backend/IssueMetadata.java
codemanmeet: Addressed!

src/main/java/backend/IssueMetadata.java
codemanmeet: Yup, I included a difference. The previous implementation of the updated filter was causing some updates to go missed. There's the issue of cache in doing the filtering with `nonSelfUpdatedAt` after metadata is available versus always doing with `updated` whether or not `metadata` is available. The issue has been addressed in #912

src/main/java/backend/Logic.java
codemanmeet: Addressed!

src/main/java/ui/listpanel/ListPanel.java
codemanmeet: Sure.

src/main/java/filter/expression/Qualifier.java
codemanmeet: Addressed!

src/main/java/filter/expression/Qualifier.java
codemanmeet: Addressed!

damithc:
Should there be some changes to documentation?

codemanmeet:
Yup, will add those as well.

src/main/java/backend/IssueMetadata.java
dariusf: So `isUpdated` tells us whether or not metadata is up to date in general, and the other two tell us whom the last update was by?

Maybe you could add a brief comment explaining the purpose of each field. Without context it can be difficult to figure out what they are for.

src/main/java/backend/Logic.java
```diff
@@ -157,37 +160,84 @@ public void refresh() {
...
+                    isUpdatedBySelf = true;
                 }
             }
 
             entry.setValue(new IssueMetadata(currentMetadata,
                     LocalDateTime.ofInstant(lastNonSelfUpdate.toInstant(), ZoneId.systemDefault()),
-                    calculateNonSelfCommentCount(currentMetadata.getComments(), currentUser)
+                    LocalDateTime.ofInstant(lastSelfUpdate.toInstant(), ZoneId.systemDefault()),
+                    calculateCommentCount(currentMetadata.getComments(), currentUser, UpdatedKind.OTHER_UPDATED),
+                    calculateCommentCount(currentMetadata.getComments(), currentUser, UpdatedKind.SELF_UPDATED),
```
dariusf: Same issue as before. The values of these two arguments can be derived from the others, so you can compute them in the `TurboIssue` constructor. We want to keep `Logic` as high-level as possible.

src/main/java/backend/Logic.java
```diff
@@ -157,37 +160,84 @@ public void refresh() {
...
             Date lastNonSelfUpdate = new Date(0);
+            Date lastSelfUpdate = new Date(0);
+            boolean isUpdatedByOthers = false;
+            boolean isUpdatedBySelf = false;
             for (TurboIssueEvent event : currentMetadata.getEvents()) {
-                if (!event.getActor().getLogin().equalsIgnoreCase(currentUser)
-                        && event.getDate().after(lastNonSelfUpdate)) {
+                if (isNewEventByOther(currentUser, event, lastNonSelfUpdate)) {
                     lastNonSelfUpdate = event.getDate();
+                    isUpdatedByOthers = true;
```
dariusf: All this logic is about the (re-)construction of `IssueMetadata` given the current user (which isn't available deeper in the back end). Could it be moved into a new `IssueMetadata` constructor? `new IssueMetadata(existingMetadata, currentUser)`

src/main/java/backend/Logic.java
```diff
@@ -157,37 +160,84 @@ public void refresh() {
         for (Map.Entry<Integer, IssueMetadata> entry : metadata.entrySet()) {
             IssueMetadata currentMetadata = entry.getValue();
             Date lastNonSelfUpdate = new Date(0);
+            Date lastSelfUpdate = new Date(0);
```
dariusf: You should use `LocalDateTime.now()` for these. `Date` is a legacy class and could only be used for compatibility.

src/main/java/backend/Logic.java
```diff
@@ -157,37 +160,84 @@ public void refresh() {
...
+                if (isNewCommentBySelf(currentUser, comment, lastSelfUpdate)){
+                    lastSelfUpdate = comment.getCreatedAt();
+                    isUpdatedBySelf = true;
                 }
             }
 
             entry.setValue(new IssueMetadata(currentMetadata,
                     LocalDateTime.ofInstant(lastNonSelfUpdate.toInstant(), ZoneId.systemDefault()),
-                    calculateNonSelfCommentCount(currentMetadata.getComments(), currentUser)
+                    LocalDateTime.ofInstant(lastSelfUpdate.toInstant(), ZoneId.systemDefault()),
```
dariusf: `LocalDateTime` objects are immutable, so you don't need to clone them.

src/main/java/backend/Logic.java
```diff
@@ -157,37 +160,84 @@ public void refresh() {
...
+                    calculateCommentCount(currentMetadata.getComments(), currentUser, UpdatedKind.SELF_UPDATED),
+                    isUpdatedByOthers,
+                    isUpdatedBySelf
             ));
         }
         return metadata;
     }
 
-    private int calculateNonSelfCommentCount(List<Comment> comments, String currentUser) {
+    private boolean isNewCommentByOther(String currentUser, Comment comment, Date lastNonSelfUpdate){
```
dariusf: All these helper methods can be static.

src/main/java/backend/Logic.java
```diff
@@ -157,37 +160,84 @@ public void refresh() {
...
         int result = 0;
-        for (Comment comment : comments) {
-            if (!comment.getUser().getLogin().equalsIgnoreCase(currentUser)) {
-                result++;
+        if (updatedKind == UpdatedKind.OTHER_UPDATED) {
+            for (Comment comment : comments) {
+                if (!isCommentBySelf(currentUser, comment)) {
+                    result++;
+                }
+            }
```
dariusf: `result` can be eliminated by filtering instead.
```java
return comments.stream()
    .filter(c -> !isCommentBySelf(currentUser, c))
    .count();
```

src/main/java/backend/Logic.java
```diff
@@ -157,37 +160,84 @@ public void refresh() {
...
+            for (Comment comment : comments) {
+                if (isCommentBySelf(currentUser, comment)) {
+                    result++;
+                }
             }
         }
         return result;
     }
 
+    private boolean isCommentBySelf(String currentUser, Comment comment){
```
dariusf: Can also be static.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -527,17 +538,33 @@ private boolean satisfiesUpdatedHours(TurboIssue issue) {
             return false;
         }
 
-        int hoursSinceUpdate;
+        int hoursSinceUpdate = Integer.MIN_VALUE;
+        LocalDateTime dateOfUpdate = LocalDateTime.ofEpochSecond(0, 0, ZoneOffset.ofHours(0));
+        // First time being filtered (haven't gotten metadata from source yet).
+        if (hoursSinceUpdate == Integer.MIN_VALUE) {
```
dariusf: Moving this up might have caused a bug -- this is always true. Also you might have intended to return early.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -71,6 +72,12 @@ private boolean issueHasNewComments(TurboIssue issue, boolean hasMetadata) {
         }
     }
 
+    private static boolean containsUpdatedQualifier(List<String> filterQualifierNames){
```
dariusf: This method and the fields above could go in `Qualifier`. The name is a bit too generic for what it does also.

src/test/java/tests/FilterEvalTests.java
```diff
@@ -433,6 +434,38 @@ public void created() {
     }
 
     @Test
+    public void updatedByOther(){
+        LocalDateTime now = LocalDateTime.now();
```
dariusf: We can probably merge these two tests, since they're testing very similar things. You don't have to test ranges, since they are covered elsewhere. Something more important is which qualifiers match in which scenarios: given an issue, there are 4 possibilities for whether the two qualifiers match, so that would be more useful to focus on.

```java
TurboIssue issue = ...;
assertEquals(false, matches("updatedByOthers:24", issue));
assertEquals(false, matches("updatedBySelf:24", issue));
// repeat for (true, true), (false, true), (true, false)
```
There's also the cases where issues can be updated by **both** others and self, or **none**. A test to demonstrate which takes precedence would be useful.

src/main/java/backend/IssueMetadata.java
codemanmeet: Yup, exactly.

src/main/java/backend/Logic.java
codemanmeet: I have to use `Date` because `event.getDate()` and `comment.getCreatedAt()` return `Date` instead of `LocalDateTime` objects. This would require a lot of refactoring of old code again. We can add that refactoring part in a separate issue.

src/main/java/backend/Logic.java
codemanmeet: Same as above. The above code basically converts the `Date` object to `LocalDateTime` object.

src/main/java/backend/Logic.java
codemanmeet: Changed.

src/main/java/backend/Logic.java
codemanmeet: Done!

src/main/java/backend/Logic.java
codemanmeet: Done!

codemanmeet: I found this really useful!

src/main/java/filter/expression/Qualifier.java
codemanmeet: Fixed.

src/main/java/ui/listpanel/ListPanel.java
codemanmeet: Fixed.

src/test/java/tests/FilterEvalTests.java
codemanmeet: Fixed. Added all the possible test cases.

docs/filters.md
```diff
@@ -108,6 +108,8 @@ This will not work for ambiguous expressions (containing OR or NOT operators) an
 - [`is`](#is)
 - [`created`](#created)
 - [`updated`](#updated)
+- [`updatedByOthers`](#updatedByOthers)
+- [`updatedBySelf`](#updatedBySelf)
```
damithc: Didn't we decide to use `-word` format instead, which is similar to git command format?

docs/filters.md
```diff
@@ -222,6 +224,18 @@ Matches issues which were created on a given date, or within a given date range.
 
 Matches issues which were updated in the given number of hours. For example, `updated:<24` would match issues updated in the last day. If a number `n` is given, it is implicitly translated to `<n`.
 
+### updatedByOthers
+
+*Expects a number or  number range*
+
+Matches issues which were updated in the given number of hours by users other than the logged in user. For example, `updated:<24` would match issues updated in the last day by other users. If a number `n` is given, it is implicitly translated to `<n`.
+
```
damithc: What's meant by 'number range'? Something like `2-4`? In that case, give an example.

src/main/java/backend/IssueMetadata.java
```diff
@@ -4,26 +4,39 @@
...
+    private final int selfCommentCount;
 
-    // If isUpdated is true, nonSelfUpdatedAt will be used to sort/filter instead of updatedAt in TurboIssue
-    private final boolean isUpdated;
+    private final boolean isUpdated; //for any update
+    private final boolean isUpdatedBySelf; // for update by self
+    private final boolean isUpdatedByOthers; // for update by others
+
+    private enum UpdatedKind {
+        SELF_UPDATED, OTHER_UPDATED
```
damithc: `UPDATED_BY_SELF, UPDATED_BY_OTHERS` ?

docs/filters.md
damithc: Or we can refer to some other place where more examples of date ranges are given. No need to repeat everywhere.

docs/filters.md
dariusf: Also these links don't work. Seems to be case-sensitive.

docs/filters.md
dariusf: The format for number ranges is given at the top. Maybe in future we can add links pointing back up.

I agree about the duplication. Since these come right after `updated`, it should be enough to just point out the differences in behaviour.

src/main/java/backend/IssueMetadata.java
```diff
@@ -49,17 +70,87 @@ public IssueMetadata(List<TurboIssueEvent> events, List<Comment> comments) {
...
+            if (isNewCommentBySelf(currentUser, comment, lastSelfUpdate)){
+                lastSelfUpdate = comment.getCreatedAt();
+                isUpdatedBySelf = true;
+            }
+        }
+
+        this.events = new ArrayList<>(existingMetadata.events);
+        this.comments = new ArrayList<>(existingMetadata.comments);
+        this.nonSelfUpdatedAt = LocalDateTime.ofInstant(lastNonSelfUpdate.toInstant(), ZoneId.systemDefault());
+        this.selfUpdatedAt = LocalDateTime.ofInstant(lastSelfUpdate.toInstant(), ZoneId.systemDefault());
```
dariusf: Use `Utility.dateToLocalDateTime` instead.

src/main/java/backend/IssueMetadata.java
```diff
@@ -49,17 +70,87 @@ public IssueMetadata(List<TurboIssueEvent> events, List<Comment> comments) {
...
+        return !event.getActor().getLogin().equalsIgnoreCase(currentUser)
+                && event.getDate().after(lastNonSelfUpdate);
+    }
+
+    private static int calculateCommentCount(List<Comment> comments, String currentUser, UpdatedKind updatedKind) {
+        int result = 0;
+        if (updatedKind == UpdatedKind.OTHER_UPDATED) {
+            return (int) (comments.stream().filter(c -> !isCommentBySelf(currentUser, c)).count());
+        } else if (updatedKind == UpdatedKind.SELF_UPDATED){
+            return (int) (comments.stream().filter(c -> isCommentBySelf(currentUser, c)).count());
```
dariusf: Use `Utility.safeLongToInt` instead.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -527,18 +538,34 @@ private boolean satisfiesUpdatedHours(TurboIssue issue) {
...
+                }
+                break;
+            case OTHER_UPDATED:
+                if (issue.getMetadata().isUpdatedByOthers()) {
+                    dateOfUpdate = issue.getMetadata().getNonSelfUpdatedAt();
+                }
+                break;
+            case ALL_UPDATED:
+                // The or condition for dateOfUpdate is for first time check
+                if (issue.getMetadata().isUpdated() || dateOfUpdate == null) {
```
dariusf: This condition is again always true. Potential bug.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -527,18 +538,34 @@ private boolean satisfiesUpdatedHours(TurboIssue issue) {
...
+                }
         }
 
-        return updatedRange.encloses(hoursSinceUpdate);
+        if (dateOfUpdate != null){
+            int hoursSinceUpdate = Utility.safeLongToInt(dateOfUpdate.until(getCurrentTime(), ChronoUnit.HOURS));
+            return updatedRange.encloses(hoursSinceUpdate);
+        } else {
+            return false;
+        }
```
dariusf: It'd be good to switch the order of these branches, so when reading the easier condition comes first.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -60,8 +60,16 @@ public ListPanel(UI ui, IModel model, PanelControl parentPanelControl, int panel
      * @param issue
      * @return true if the issue has changed, false otherwise
      */
+
+    private static boolean qualifierNamesHaveUpdatedQualifier(List<String> filterQualifierNames){
+        // omitting self-updated issues since we only care about updated comments from other users.
+        return (filterQualifierNames.contains(Qualifier.UPDATED) ||
+                filterQualifierNames.contains(Qualifier.UPDATED_OTHERS));
+    }
```
dariusf: This should be moved to `Qualifier`, along with the code that is calling it. The whole thing can take in a `FilterExpression` and return a boolean value.

src/main/java/filter/expression/Qualifier.java
damithc: Yup, as a rule of thumb, 'pick the lower hanging fruits first' to improve readability.

src/test/java/tests/FilterEvalTests.java
```diff
@@ -433,6 +434,59 @@ public void created() {
...
 
     @Test
+    public void updatedBySelfAndOther(){
+        LocalDateTime now = LocalDateTime.now();
+        Qualifier.setCurrentTime(now);
+        TurboIssue issueUpdated = new TurboIssue(REPO, 1, "");
+        IssueMetadata updatedByOtherMetadata = new IssueMetadata(issueUpdated.getMetadata(), now.minusHours(4),
+                now.minusHours(4), 2, 2, true, false); //UpdatedByOther true with updated time 4 hours ago.
+        issueUpdated.setMetadata(updatedByOtherMetadata);
+        assertEquals(false, matches("updatedByOthers:<2", issueUpdated));
```
damithc: Why 2? is that a boundary value?

src/main/java/filter/expression/Qualifier.java
codemanmeet: Yup figured the `if` condition is not even required. Just obtain the `updatedDate` and check whether it encloses the range.

src/test/java/tests/FilterEvalTests.java
codemanmeet: Wanted to check for a value less than 4. I also changed tests to include 4 as the boundary.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -20,17 +21,23 @@
 public class Qualifier implements FilterExpression {
 
     public static final String UPDATED = "updated";
+    public static final String UPDATED_OTHERS = "updatedByOthers";
```
dariusf: This isn't updated to reflect the new convention.

docs/filters.md
```diff
@@ -220,7 +222,19 @@ Matches issues which were created on a given date, or within a given date range.
...
+*Expects a number or  number range*
+
+Matches issues which were updated in the given number of hours by users other than the logged in user. For example, `updated:<24` would match issues updated in the last day by other users. Number ranges are written using a relational operator (.e.g `>5`, `<=10`). If a number `n` is given, it is implicitly translated to `<n`.
+
+### updated-self
+
+*Expects a number or  number range*
+
+Matches issues which were updated in the given number of hours by the logged in user. For example, `updated:<24` would match issues updated in the last day by the logged in user. Number ranges are written using a relational operator (.e.g `>5`, `<=10`). If a number `n` is given, it is implicitly translated to `<n`.
```
dariusf: The descriptions for `updated-others` and `updated-self` still refer to `updated`.

Also, no need to duplicate the examples, just say something like, 'Like `updated`, but only matches issues updated by other users/the current user'.

src/main/java/backend/IssueMetadata.java
```diff
@@ -3,27 +3,40 @@
...
     private final LocalDateTime nonSelfUpdatedAt;
+    private final LocalDateTime selfUpdatedAt;
     private final int nonSelfCommentCount;
+    private final int selfCommentCount;
 
-    // If isUpdated is true, nonSelfUpdatedAt will be used to sort/filter instead of updatedAt in TurboIssue
-    private final boolean isUpdated;
+    private final boolean isUpdated; //for any update
+    private final boolean isUpdatedBySelf; // for update by self
+    private final boolean isUpdatedByOthers; // for update by others
```
dariusf: Sorry, this again... is `isUpdated` equal to `isUpdatedBySelf || isUpdatedByOthers`? Like, is there a case where the values of these can differ? If not, we can get rid of this field and implement `isUpdated()` in terms of the other ones.

src/main/java/ui/listpanel/ListPanel.java
dariusf: This bit is still relevant I think. We want to minimise operations on `FilterExpression`s outside `Qualifier` and the subclasses themselves.

src/test/java/tests/FilterLexerTests.java
```diff
@@ -37,8 +37,8 @@ public void lexer() {
             new Lexer("a,b").lex();
         } catch (ParseException ignored) {}
 
-        assertEquals(new Lexer("sort: a, b ").lex(), Arrays.asList(
-            new Token(TokenType.QUALIFIER, "sort:"),
+        assertEquals(new Lexer("sort-self: a, b ").lex(), Arrays.asList(
+            new Token(TokenType.QUALIFIER, "sort-self:"),
```
dariusf: Don't change this test, just add another. You can also cover more pathological cases, like `-aaa-bbb`, `---aaa-bbb`, `aaa-bbb-`, etc.

dariusf:
Last round. I have a feeling that uncovered line is a false positive, so will merge once this batch is done and CI passes.

damithc:
Wasn't there some change to sort command? Was the documentation updated for it?

dariusf:
The tests just happen to contain the word 'sort'. No changes to functionality.

damithc:
Oh? We didn't change `sort:nonSelfUpdate`?

dariusf:
No, it still uses camel case.

damithc:
I see. Let's change it to be consistent. Can be done separately.

acruis:
@codemanmeet Is `updated-others` supposed to be used in conjunction with the original `updated` filter?

codemanmeet:
Nope it doesn't have to be used in conjunction.

Issue #739, opened 2015-06-26
Use runAndWait in tests instead of sleep


acruis:
As mentioned by @dariusf, `PlatformEx.runAndWait` has a stronger guarantee that what comes after executes after compared to `sleep`, which makes integration tests more deterministic.

lejolly:
may not work for every test case, but i'd love to be proven wrong for this.

dariusf:
If it doesn't work, the use of explicit latches and barriers may be required. This part won't be as beginner-friendly and can be a separate issue.

dariusf:
Subsumed by new testing guidelines (always synchronise explicitly!).

PR #932, opened 2015-09-02
 Add keyboard shortcut for changing the board #919 

### LINKED ISSUE #919: Add keyboard shortcut for changing the board
Suggestion: Ctrl+B

HyungJon:
Fixes #919 

Changes board when Ctrl+B is pushed.
Only happens if 1) there are at least 2 boards saved, 2) one of them is currently open

2 might be debatable. I chose that because I wanted to prevent changing the board and losing the current board panels. I'll modify if there are other opinions.

### DIFF
--- a/src/main/java/ui/MenuControl.java
+++ b/src/main/java/ui/MenuControl.java
@@ -17,6 +17,7 @@ import ui.issuepanel.FilterPanel;
 import ui.issuepanel.PanelControl;
 import util.events.*;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -270,6 +271,20 @@ public class MenuControl extends MenuBar {
                 return Stream.of();
             }
         }).collect(Collectors.toList());
+    }
+    
+    public void switchBoard() {
+        if (prefs.getLastOpenBoard().isPresent()) {
+            List<String> boardNames = new ArrayList<>(prefs.getAllBoards().keySet());
+            int index = boardNames.indexOf(prefs.getLastOpenBoard().get());
+            if (index == boardNames.size()-1) {
+                index = 0;
+            } else {
+                index++;
+            }
+            
+            onBoardOpen(boardNames.get(index), prefs.getBoardPanels(boardNames.get(index)));
+        }
     }
 
     private MenuItem createDocumentationMenuItem() {
--- a/src/main/java/ui/MenuControl.java
+++ b/src/main/java/ui/MenuControl.java
@@ -17,7 +17,6 @@ import ui.issuepanel.FilterPanel;
 import ui.issuepanel.PanelControl;
 import util.events.*;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -274,12 +273,9 @@ public class MenuControl extends MenuBar {
     }
     
     public void switchBoard() {
-        if (prefs.getLastOpenBoard().isPresent() && prefs.getAllBoards().size() > 1) {
-            List<String> boardNames = new ArrayList<>(prefs.getAllBoards().keySet());
-            int lastBoard = boardNames.indexOf(prefs.getLastOpenBoard().get());
-            int index = (lastBoard + 1) % boardNames.size();
-            
-            onBoardOpen(boardNames.get(index), prefs.getBoardPanels(boardNames.get(index)));
+        Optional<String> name = prefs.switchBoard();
+        if (name.isPresent()) {
+            onBoardOpen(name.get(), prefs.getBoardPanels(name.get()));
         }
     }
 
--- a/src/main/java/ui/MenuControl.java
+++ b/src/main/java/ui/MenuControl.java
@@ -274,10 +274,10 @@ public class MenuControl extends MenuBar {
     }
     
     public void switchBoard() {
-        if (prefs.getLastOpenBoard().isPresent()) {
+        if (prefs.getLastOpenBoard().isPresent() && prefs.getAllBoards().size() > 1) {
             List<String> boardNames = new ArrayList<>(prefs.getAllBoards().keySet());
             int index = boardNames.indexOf(prefs.getLastOpenBoard().get());
-            if (index == boardNames.size()-1) {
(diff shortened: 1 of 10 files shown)


damithc:
Showing the current board name should be done soon too. Without it, this feature is a big hard to use.

src/test/java/guitests/MenuControlTest.java
```diff
@@ -21,6 +22,8 @@ public void menuControlTest() {
         modelUpdatedEventTriggered = false;
         UI.events.registerEvent((ModelUpdatedEventHandler) e -> modelUpdatedEventTriggered = true);
         PanelControl panelControl = (PanelControl) find("#dummy/dummy_col0").getParent();
+        Preferences testPref = panelControl.getPreferences();
```
dariusf: Can we use `UI.prefs` instead?

src/main/java/ui/issuepanel/PanelControl.java
```diff
@@ -273,4 +273,8 @@ public int getNumberOfPanels() {
     public int getNumberOfSavedBoards() {
         return prefs.getAllBoards().size();
     }
+    
+    public Preferences getPreferences() {
```
dariusf: We wouldn't need this getter in that case.

src/main/java/ui/MenuControl.java
```diff
@@ -271,6 +272,20 @@ private void onBoardDelete(String boardName) {
             }
         }).collect(Collectors.toList());
     }
+    
+    public void switchBoard() {
+        if (prefs.getLastOpenBoard().isPresent() && prefs.getAllBoards().size() > 1) {
+            List<String> boardNames = new ArrayList<>(prefs.getAllBoards().keySet());
```
dariusf: This is backed by a `HashMap`, which doesn't guarantee key ordering: the boards may be cycled in a different order [over time](http://stackoverflow.com/a/2144822). You might want to change it to a `LinkedHashMap`.

src/main/java/ui/MenuControl.java
```diff
@@ -271,6 +272,20 @@ private void onBoardDelete(String boardName) {
...
+            int index = boardNames.indexOf(prefs.getLastOpenBoard().get());
+            if (index == boardNames.size() - 1) {
+                index = 0;
+            } else {
+                index++;
+            }
```
dariusf: Use `%` instead?
```java
int index = (lastBoard + 1) % boardNames.size();
```

dariusf:
Could you also see if the `sleep`s in tests can be replaced with `PlatformEx.waitOnFxThread`?

src/test/java/guitests/MenuControlTest.java
HyungJon: That will result in non-static reference error. Can I move the ```getPrefs()``` from ```PanelControl``` to ```UI``` for this purpose?

src/main/java/ui/MenuControl.java
HyungJon: Do I just have to change ```savedBoards``` in GlobalConfig to LinkedHashMap, or should every access to it be changed accordingly to ensure the order will be retained?

HyungJon:
Confirmed that ```MenuControlTest``` runs okay with ```PlatformEx.waitOnFxThread```. Shall I create a new issue for changing all ```sleep``` with ```PlatformEx.waitOnFxThread```? #739 has a related discussion though.

src/test/java/guitests/MenuControlTest.java
dariusf: Right, `prefs` isn't static. Moving it to `UI` would still require a reference to the current `UI` instance though. How about making `prefs` static, then doing that?

src/main/java/ui/MenuControl.java
dariusf: Just the former should do it. What do you mean by changing every access?

It'd be good to add a non-GUI test for this too. Save and delete a few boards, then test that cycling returns stuff in the expected order.

dariusf:
Great! We should get rid of `sleep` eventually, but I'm not sure yet if in general it's possible to mechanically replace it with `waitOnFxThread`. I guess we should do it test by test for now.

That issue can be repurposed. `waitOnFxThread` is generally what we want when we use `sleep` in tests, not `runAndWait`.

src/test/java/guitests/MenuControlTest.java
dariusf: You can exclude it [here](https://github.com/HubTurbo/HubTurbo/blob/master/config/findbugs/excludeFilter.xml#L23-L30).

HyungJon:
Oops, forgot to commit the last fix to findbugsTest. Squashed and pushed.

I'm really unable to figure out what this status check thing is. Looks like I should be able to fix and push it from my end, how can I do it?

dariusf:
Status check? You mean the drop in coverage?

HyungJon:
The branch became unmergeable because "Required status must pass before merging". Probably because I edited the excludeFilter I think?

dariusf:
That would cause the Travis build to fail if it were an issue. The only remaining check I see is the coveralls one. How were you trying to merge?

damithc:
Merging should be done locally, not using GitHub.

HyungJon:
cap 2015-09-05 10-59-08-132

This is displayed. It's visible only to the branch owner then?

HyungJon:
Oh wait, just branch out of date I think.

dariusf:
That's a new GitHub feature that enforces that all pre-merge checks have passed before allowing a merge.

Isn't the problem that coverage has dropped?

HyungJon:
Ah, it makes sense now. I think the coverage decrease might be because I added a simplified version of ```switchBoard``` to the non-GUI test as well. I think I should totally move ```switchBoard``` implementation to ```Preferences``` and make ```MenuControl``` only call that function.

HyungJon:
Bingo. But I'm really really not a fan of the change I had to make (calling ```switchBoard``` normally with Ctrl+B will result in one redundant ```setLastOpenBoard```), so if there's any way to achieve higher coverage without doing this, I'll immediately do that.

dariusf:
I think it's fine. Can't think of a clearer way to write it.

PR #935, opened 2015-09-02
 Display currently open board on title bar #929 

### LINKED ISSUE #929: Display currently open board on title bar
Format: "HubTurbo V3.3.0 (Board 1)"

If there is no board open, indicate "No board open"

HyungJon:
Fixes #929

### DIFF
--- a/src/main/java/ui/UI.java
+++ b/src/main/java/ui/UI.java
@@ -268,7 +268,7 @@ public class UI extends Application implements EventDispatcher {
     }
 
     private void setupMainStage(Scene scene) {
-        mainStage.setTitle("HubTurbo " + Utility.version(VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH));
+        updateTitle();
         mainStage.setScene(scene);
         mainStage.show();
         mainStage.setOnCloseRequest(e -> quit());
@@ -515,5 +515,12 @@ public class UI extends Application implements EventDispatcher {
     public void triggerNotificationAction() {
         notificationController.triggerNotificationAction();
     }
-
+    
+    public void updateTitle() {
+        String openBoard = prefs.getLastOpenBoard().orElse("none");
+        String title = String.format("HubTurbo " + Utility.version(VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH)
+                + " (%s)", openBoard);
+        mainStage.setTitle(title);
+    }
+    
 }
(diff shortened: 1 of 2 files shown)


src/main/java/ui/UI.java
```diff
@@ -585,5 +585,18 @@ public void triggerNotificationPaneAction() {
             }
         });
     }
-
+    
+    public void setTitle() {
```
dariusf: `updateTitle` might be a clearer name.

src/main/java/ui/UI.java
```diff
@@ -585,5 +585,18 @@ public void triggerNotificationPaneAction() {
...
+        String openBoard = "";
+        if (prefs.getLastOpenBoard().isPresent()) {
+            openBoard = prefs.getLastOpenBoard().get();
+        } else {
+            openBoard = "No board open";
+        }
```
dariusf: Can be replaced with
```java
String openBoard = prefs.getLastOpenBoard().orElse("No board open");
```
Also I'm not sure about `HubTurbo V1.2.3 (No board open)`. Maybe `HubTurbo V1.2.3 (none)` (terser), or hide the parentheses entirely?

dariusf:
Looks good. Final thing, on the title text: we can't differentiate a board named 'none' and having no board open now. :stuck_out_tongue: Is that a problem? @damithc

HyungJon:
I've been thinking about that too, which is why I chose a name very unlikely for a board at first. Wasn't the best choice, but I guess we should use something along that line. I thought "none" is unlikely enough, but it'd be good to have a way to prevent the confusion altogether. Maybe use some special characters and add an board name input validation that forbids those characters?

dariusf:
Yup, or hide the parentheses.

dariusf:
@damithc do you have any opinions on this?

damithc:
It's ok to use 'none'. It's unlikely that a user will name a board 'none'. We can also block that name during naming boards, but it can be done as a separate issue.

damithc:
No tests?

HyungJon:
When #951 is merged, allowing tests to access the UI instance. Then I'll be able to add ```getTitle``` so that the test can check whether the name was updated properly.

damithc:
I see. May be merge this without tests and create a new issue for adding tests?

HyungJon:
All right. Shall I create an issue for board name validation and test everything there?

damithc:
Sure.

dariusf:
Can this be in today as well?

HyungJon:
Okay, Travis build passed. Coverage will fall since I'm creating tests in #975 .Can I have a toMerge?

dariusf:
I'm merging this. It's passed before and won't cause any issues.

PR #936, opened 2015-09-03
Label picker only submits label changes to GitHub only after undo window passes #923

### LINKED ISSUE #923: Label picker only submits label changes to GitHub only after undo window passes
Noticed some unusual behavior when applying labels.

1. Applied label.
2. The labels showed up on bView almost immediately, although the undo window is still open.
3. The labels did not show up on the pView until the next sync.

May be a periodic sync happened just as I was applying the labels?

lejolly:
Fixes #923

### DIFF
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -224,21 +224,25 @@ public class Logic {
      *
      * @param issue The issue whose labels are to be replaced
      * @param labels The labels to be applied to the given issue
-     * @return The list of labels on the issue after the request is received by GitHub
+     * @param originalLabels The original labels to be applied to the UI in case of failure
+     * @return A boolean indicating the result of the label replacement from GitHub
      */
-    public CompletableFuture<Boolean> replaceIssueLabels(TurboIssue issue, List<String> labels) {
-        logger.info(HTLog.format(issue.getRepoId(), "Applying labels " + labels + " to " + issue));
-
-        return repoIO.replaceIssueLabels(issue, labels).handle((resultLabels, ex) -> {
-            if (ex == null) {
-                issue.setLabels(labels);
-                updateUIAndShow();
-                return true;
-            } else {
-                return false;
-            }
+    public CompletableFuture<Boolean> replaceIssueLabelsRepo
+            (TurboIssue issue, List<String> labels, List<String> originalLabels) {
+        logger.info(HTLog.format(issue.getRepoId(), "Sending labels " + labels + " for " + issue + " to GitHub"));
+        return repoIO.replaceIssueLabels(issue, labels)
+                .thenApply(e -> true)
+                .exceptionally(e -> {
+                    replaceIssueLabelsUI(issue, originalLabels);
+                    logger.error(e.getLocalizedMessage(), e);
+                    return false;
         });
     }
 
-}
+    public void replaceIssueLabelsUI(TurboIssue issue, List<String> labels) {
+        logger.info(HTLog.format(issue.getRepoId(), "Applying labels " + labels + " to " + issue + " in UI"));
+        issue.setLabels(labels);
+        updateUIAndShow();
+    }
 
+}
--- a/src/main/java/ui/UI.java
+++ b/src/main/java/ui/UI.java
@@ -9,9 +9,7 @@ import com.sun.jna.platform.win32.User32;
 import com.sun.jna.platform.win32.WinDef.HWND;
 import javafx.application.Application;
 import javafx.application.Platform;
-import javafx.event.ActionEvent;
 import javafx.geometry.Pos;
-import javafx.scene.Node;
 import javafx.scene.Parent;
 import javafx.scene.Scene;
 import javafx.scene.control.Label;
@@ -27,10 +25,10 @@ import javafx.stage.Stage;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.controlsfx.control.NotificationPane;
-import org.controlsfx.control.action.Action;
 import prefs.Preferences;
 import ui.components.HTStatusBar;
 import ui.components.KeyboardShortcuts;
+import ui.components.Notification;
 import ui.components.StatusUI;
 import ui.components.pickers.LabelPicker;
 import ui.issuepanel.PanelControl;
@@ -62,7 +60,6 @@ public class UI extends Application implements EventDispatcher {
     private static HWND mainWindowHandle;
 
     private static final int REFRESH_PERIOD = 60;
-    private static final int NOTIFICATION_PANE_VISIBLE_PERIOD = 5;
 
     // Application-level state
 
@@ -74,8 +71,9 @@ public class UI extends Application implements EventDispatcher {
     public EventBus eventBus;
     private HashMap<String, String> commandLineArgs;
     private TickingTimer refreshTimer;
-    private TickingTimer notificationPaneTimer;
     public GUIController guiController;
+    private NotificationController notificationController;
+
 
     // Main UI elements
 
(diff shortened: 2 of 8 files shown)


src/main/java/backend/Logic.java
```diff
@@ -252,19 +250,31 @@ public Model getRepo(String repoId) {
      * @param labels The labels to be applied to the given issue
      * @return The list of labels on the issue after the request is received by GitHub
      */
-    public CompletableFuture<Boolean> replaceIssueLabels(TurboIssue issue, List<String> labels) {
-        logger.info(HTLog.format(issue.getRepoId(), "Applying labels " + labels + " to " + issue));
-
+    public CompletableFuture<Boolean> replaceIssueLabelsRepo(TurboIssue issue, List<String> labels) {
```
dariusf: The return javadocs here don't seem to be accurate?

src/main/java/backend/Logic.java
```diff
@@ -252,19 +250,31 @@ public Model getRepo(String repoId) {
...
+        logger.info(HTLog.format(issue.getRepoId(), "Sending labels " + labels + " for " + issue + " to GitHub"));
         return repoIO.replaceIssueLabels(issue, labels).handle((resultLabels, ex) -> {
             if (ex == null) {
-                issue.setLabels(labels);
-                updateUIAndShow();
                 return true;
             } else {
+                undoUILabelChange(issue);
```
dariusf: Can be formulated more clearly as:
```java
return repoIO.replaceIssueLabels(issue, labels).thenApply(e -> {
    undoUILabelChange(issue);
    return true;
}).exceptionally(withResult(false));
```

src/main/java/ui/components/pickers/LabelPicker.java
```diff
@@ -47,22 +41,24 @@ private void showLabelPicker(TurboIssue issue) {
         stage.show(); // ensures stage is showing after label picker is closed (mostly for tests)
         // if result is present (user did not cancel) then replace issue labels with result
         if (result.isPresent()) {
-            ui.logic.replaceIssueLabels(issue, result.get().stream().sorted().collect(Collectors.toList()))
-                    .thenApply(success -> postReplaceLabelActions(success, issue, originalLabels));
+            replaceLabels(issue, result.get().stream().sorted().collect(Collectors.toList()));
+//            ui.logic.replaceIssueLabels(issue, result.get().stream().sorted().collect(Collectors.toList()))
+//                    .thenApply(success -> postReplaceLabelActions(success, issue, originalLabels));
```
dariusf: Unused code.

src/main/java/backend/Logic.java
```diff
@@ -36,6 +33,7 @@
 
     private RepoIO repoIO;
     public LoginController loginController;
+    private Optional<List<String>> originalLabels = Optional.empty();
```
dariusf: Maybe now isn't the time to implement a fully-fledged undo feature, but we should be very selective of the state we maintain here, because practically every application-level feature will go through this class.

Ideally the state required to undo each action should be encapsulated in some kind of command object. 'Smart actions', basically, which can do and undo themselves.

src/main/java/ui/UI.java
```diff
@@ -555,13 +556,19 @@ public HWND getMainWindowHandle() {
         return mainWindowHandle;
     }
 
-    public void showNotificationPane(Node graphic, String text, Action action) {
+    public void showNotificationPane(Node graphic, String text, Runnable runnable, Runnable undoRunnable) {
```
dariusf: This also seems rather ad-hoc, because this method is for showing a notification in general, which may not have undo semantics. Again, I'm now sure how much of the architecture we should flesh out now.

src/main/java/ui/UI.java
```diff
@@ -555,13 +556,19 @@ public HWND getMainWindowHandle() {
...
         Platform.runLater(() -> {
             hideNotificationPane();
+            notificationRunnable = Optional.of(runnable);
+            notificationUndoRunnable = Optional.of(undoRunnable);
             notificationPane.setGraphic(graphic);
             notificationPane.setText(text);
             notificationPane.getActions().clear();
-            notificationPane.getActions().add(action);
+            notificationPane.getActions().add(new Action("Undo",
+                    actionEvent -> Platform.runLater(() -> {
```
dariusf: This may not need to be wrapped in `Platform.runLater`? `Action` callbacks should originate from the UI thread.

src/main/java/ui/UI.java
```diff
@@ -555,13 +556,19 @@ public HWND getMainWindowHandle() {
...
+                        notificationRunnable = Optional.of(undoRunnable);
```
dariusf: We should try to keep state local. Maybe a class in charge of controlling the notification panel?

src/main/java/backend/Logic.java
lejolly: i think the `undoUILabelChange` should be in the `exceptionally`?

dariusf: Oh, right, yes. It should be reversed.
```java
return repoIO.replaceIssueLabels(issue, labels)
    .thenApply(newLabels -> true)
    .exceptionally(e -> {
        undoUILabelChange(issue);
        return false;
    });
```
I guess we don't need a `withResult`-like function for `thenApply`.

lejolly:
@dariusf i'm not too sure how else to implement the undo besides storing state in the `NotificationController`, not sure if creating another class to encapsulate the actions/runnables would be a better solution. any thoughts?

dariusf:
Making another class was what I was thinking of. `notificationRunnable` corresponds to what you would do on timeout, `notificationUndoRunnable` to what would happen when clicking an action. This class (say it's a `Notification`) would contain those, plus a duration and a message maybe.

This way the notifications are generic and not coupled to the undo functionality.

lejolly:
If i were to create a class for notifications, then i'll still have to cater to varying numbers of runnables tied to each notification, but for the scope for this pr, i'll just cater for two runnables. (Other possible numbers are 0: when you just want to show a notification, or 1: when the runnable is tied to the button and not the timeout)

dariusf:
I guess you could have a bunch of overloaded constructors/factory methods, and fill in the missing runnables with `() -> {}`s. Doesn't need to be complicated as we don't know all the use cases for notifications yet.

lejolly:
i'm using optionals for now, but i've only written one constructor so far, we can fill in the rest if we ever need them.

src/main/java/ui/components/Notification.java
```diff
@@ -0,0 +1,60 @@
...
+    public static final int DEFAULT_NOTIFICATION_PANE_VISIBLE_PERIOD = 10;
+
+    public enum NotificationType { NOACTION, ACTIONONBUTTON, ACTIONONTIMEOUT, ACTIONONBUTTONANDTIMEOUT }
+
+    private final NotificationType notificationType;
+    private final Node icon;
+    private final String message;
+    private final Optional<String> buttonLabel;
+    private final Optional<Runnable> timeoutRunnable;
+    private final Optional<Runnable> buttonRunnable;
```
dariusf: These don't have to be optional. Runnables have an 'empty' value: the no-op `() -> {}`. We could default to no-ops when runnables are absent. The interface could then be slightly higher-level, with `timeout` and `action` (for example) methods which just `run` the corresponding actions.

Is it possible to have a button without a button label? If not, it probably doesn't have to be optional also.

src/test/java/guitests/LabelPickerTests.java
```diff
@@ -56,6 +57,9 @@ public void addAndRemoveLabelTest() {
         sleep(EVENT_DELAY);
         assertEquals(0, listPanelCell.getIssueLabels().size());
 
+        // sleep long enough for label changes to be made permanent (i.e. longer than the undo window period)
+        sleep(Notification.DEFAULT_NOTIFICATION_PANE_VISIBLE_PERIOD * 1000);
```
dariusf: Could this be synchronised using `UITest.waitUntilNodeDisappears`? As in waiting for the notification UI to go away.

Also, could you see if the tests work with `waitOnFxThread`?

dariusf:
Looks good. Is everything covered? If so we can merge.

lejolly:
yup, ok.

PR #938, opened 2015-09-05
New milestones cannot be used in filters unless the cache is refreshed #906

### LINKED ISSUE #906: New milestones cannot be used in filters unless the cache is refreshed
It happened again. 
`milestone:V5.51` does not capture issues assigned to it unless I refresh the cache. Note that the milestone V5.51 was created after the previous refresh of cache.

ndt93:
Fix #906 

This seem to fix a number of other weird behaviors also (missing assignee's name and profile picture in issues etc.). The effect is widespread to all core resources (issues, labels, milestones, users) and likely need users to refresh all old caches to clear up past update problems that stick in the cache and cause regression due to this fix.

Several direct interactions with Github API are implemented and are not tested in unit tests so would be nice if the reviewer use this and check for any regression. Also, future maintenance should include checking for any changes in Github API.


ndt93:

src/main/java/backend/resource/TurboUser.java
```diff
@@ -90,9 +92,9 @@ public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         TurboUser that = (TurboUser) o;
-        return loginName.equals(that.loginName) &&
-                realName.equals(that.realName) &&
-                avatarURL.equals(that.avatarURL);
+        return (loginName == that.loginName || (loginName != null && loginName.equals(that.loginName))) &&
```
dariusf: Could this be fixed by moving the null checks to the `User` constructor instead (and defaulting to `""`)? I was hoping to avoid `null` entirely in the model classes.

src/main/java/github/update/PageHeaderIterator.java
```diff
@@ -0,0 +1,137 @@
...
+        final String param = UrlUtils.getParam(parsed, PARAM_PAGE);
+        if (param == null || param.length() == 0) {
+            return -1;
+        }
+
+        try {
+            return Integer.parseInt(param);
+        } catch (NumberFormatException nfe) {
+            return -1;
+        }
```
dariusf: This would be a great place to use the Maybe monad. Unfortunately lambdas that throw exceptions need a whole new set of functional interfaces... I guess it isn't worth it? :disappointed:

src/test/java/tests/UpdateServiceTests.java
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.Test;
+
+public class UpdateServiceTests {
+    /**
+     * Tests that head request to nonexistent repo throws an exception
+     * @throws IOException
+     */
+    @Test(expected = FileNotFoundException.class)
+    public void testHeadRequest() throws IOException {
+        GitHubClientExtended client = new GitHubClientExtended();
```
dariusf: These tests aren't authenticated, so could there be a problem with quota use? Or do HEAD requests not consume quota?

dariusf:
We'll probably need a way to clear the cache upon update. More conservative than necessary but we don't have a way to do schema changes yet.

What do you mean by checking for GitHub API regressions?

src/main/java/backend/resource/TurboUser.java
ndt93: Ok, good idea.

src/main/java/github/update/PageHeaderIterator.java
ndt93: Yah, might not worth the trouble. This method I copy directly from egit's repo.

src/test/java/tests/UpdateServiceTests.java
ndt93: This tests the exception only, so it already failed from opening the `HttpURLConnection` due to the bogus URL.

ndt93:
Oh, it's for the case when Github API changes in the future (their resource paths or parameters), so it's a good idea to keep a note to check for such thing since there are parts we can't rely on egit's updates anymore.

dariusf:
Keep a note meaning in a comment?

ndt93:
Maybe just a note in the developer guide saying that we are using Github API v3 so that if someone notices an API's update then they could know that `egit` needs to be updated and stuffs in `GithubClientExtended` need to be rechecked.

ndt93:
@dariusf I fixed the TurboUser null check, you can check this again.

dariusf:
`replaceNull` seems like something we should have added a long time ago. Maybe you could static import it in `TurboUser`? It's clear enough what it refers to.

Apart from that, looks good.

ndt93:
Ok, I imported it. I'm not entirely sure if #912 is related to this so may be best to leave it open for a while for any reports.

damithc:
@ndt93 when merging pull requests, do it locally, not using GitHub interface. That's because we want the merge commit message to be in a specific format.

ndt93:
Oh right. Sure. I completely forgot about that :/

PR #940, opened 2015-09-06
User can remove a repo #914

### LINKED ISSUE #914: User can remove a repo
... so that it doesn't appear in the repo list anymore.

This requires adding a new menu called `Repos` (similar to `Panels` and `Boards`)

HansNewbie:
Fixes #914

### DIFF
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -208,6 +208,11 @@ public class Logic {
         return models.getDefaultRepo();
     }
 
+    public CompletableFuture<Boolean> removeRepository(String repoId) {
+        models.removeRepoModelById(repoId);
+        return repoIO.removeRepository(repoId);
+    }
+
     /**
      * Carries the current model in Logic to the GUI and triggers metadata updates if panels require
      * metadata to display their issues, in which case the changes in the model are not presented to the user.
--- a/src/main/java/backend/json/JSONStore.java
+++ b/src/main/java/backend/json/JSONStore.java
@@ -12,6 +12,7 @@ import util.Utility;
 
 import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
@@ -54,6 +55,12 @@ public class JSONStore extends RepoStore {
             logger.error("Unable to open stored repository directory. ");
             return new ArrayList<>();
         }
+    }
+
+    public CompletableFuture<Boolean> removeStoredRepo(String repoId) {
+        CompletableFuture<Boolean> response = new CompletableFuture<>();
+        addTask(new DeleteTask(repoId, response));
+        return response;
     }
 
     private static Optional<String> getRepositoryIdFromJson(Path p) {
--- a/src/main/java/backend/resource/MultiModel.java
+++ b/src/main/java/backend/resource/MultiModel.java
@@ -2,6 +2,8 @@ package backend.resource;
 
 import backend.IssueMetadata;
 import backend.interfaces.IModel;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import prefs.Preferences;
 
 import java.time.LocalDateTime;
@@ -26,6 +28,8 @@ public class MultiModel implements IModel {
     // Guaranteed to have a value throughout
     private String defaultRepo = null;
 
+    private static final Logger logger = LogManager.getLogger(MultiModel.class.getName());
+
     public MultiModel(Preferences prefs) {
         this.models = new HashMap<>();
         this.pendingRepositories = new HashSet<>();
@@ -46,6 +50,16 @@ public class MultiModel implements IModel {
 
     private synchronized MultiModel add(Model model) {
         this.models.put(model.getRepoId(), model);
+        return this;
+    }
+
+    public synchronized MultiModel removeRepoModelById(String repoId) {
+        Optional<Model> repoModelToBeDeleted = getModelById(repoId);
+        if (repoModelToBeDeleted.isPresent()) {
+            this.models.remove(repoModelToBeDeleted.get().getRepoId());
+        } else {
+            logger.error("RepoModel to be deleted does not exist.");
+        }
         return this;
     }
 
--- a/src/main/java/ui/MenuControl.java
+++ b/src/main/java/ui/MenuControl.java
@@ -1,5 +1,6 @@
 package ui;
 
+import javafx.application.Platform;
 import javafx.beans.value.ChangeListener;
 import javafx.beans.value.ObservableValue;
 import javafx.geometry.Orientation;
@@ -17,9 +18,11 @@ import ui.issuepanel.FilterPanel;
 import ui.issuepanel.PanelControl;
 import util.events.*;
 
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -51,12 +54,15 @@ public class MenuControl extends MenuBar {
         Menu boards = new Menu("Boards");
         boards.getItems().addAll(createBoardsMenu());
 
+        Menu repos = new Menu("Repos");
+        repos.getItems().addAll(createReposMenu());
+
         Menu view = new Menu("View");
         view.getItems().addAll(
                 createRefreshMenuItem(),
                 createDocumentationMenuItem());
 
-        getMenus().addAll(file, newMenu, panels, boards, view);
+        getMenus().addAll(file, newMenu, panels, boards, repos, view);
     }
 
     private Menu createFileMenu() {
@@ -322,6 +328,43 @@ public class MenuControl extends MenuBar {
         newMilestoneMenuItem.setAccelerator(KeyboardShortcuts.NEW_MILESTONE);
 
         return new MenuItem[] { newIssueMenuItem, newLabelMenuItem, newMilestoneMenuItem };
+    }
+
+    private MenuItem[] createReposMenu() {
+        Menu remove = new Menu("Remove");
+
+        ui.registerEvent((OpenReposChangedEventHandler) e -> {
+            Platform.runLater(() -> updateRepoRemoveList(remove));
+        });
+
+        return new MenuItem[] { remove };
+    }
+
+    private void updateRepoRemoveList(Menu remove) {
+        remove.getItems().clear();
+
+        HashSet<String> currentlyUsedRepo = getCurrentlyUsedRepo();
+
+        for (String repoId : ui.logic.getStoredRepos()) {
+            if (!currentlyUsedRepo.contains(repoId)) {
+                MenuItem removeItem = new MenuItem(repoId);
+                removeItem.setOnAction(e1 -> onRepoRemove(repoId));
+                remove.getItems().add(removeItem);
+            }
+        }
+    }
+
+    private HashSet<String> getCurrentlyUsedRepo() {
+        HashSet<String> currentlyUsedRepos = new HashSet<>();
+        String defaultRepo = ui.logic.getDefaultRepo();
+        currentlyUsedRepos.add(defaultRepo);
+        currentlyUsedRepos.addAll(panels.getRepositoriesReferencedOnAllPanels());
+
+        return currentlyUsedRepos;
+    }
+
+    private void onRepoRemove(String repoId) {
+        ui.logic.removeRepository(repoId).thenRun(() -> ui.triggerEvent(new OpenReposChangedEvent()));
     }
 
     public void scrollTo(int panelIndex, int numOfPanels){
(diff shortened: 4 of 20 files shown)


src/main/java/backend/Logic.java
```diff
@@ -119,6 +120,7 @@ public void refresh() {
                         .thenApply(models::addPending)
                         .thenRun(this::updateUI)
                         .thenRun(() -> UI.events.triggerEvent(new RepoOpenedEvent(repoId)))
+                        .thenRun(() -> UI.events.triggerEvent(new RepoListUpdatedEvent()))
```
HansNewbie: This keeps giving me
`Exception in thread "pool-4-thread-2" java.lang.IllegalStateException: This operation is permitted on the event thread only; currentThread = pool-4-thread-2`

Anything I can do about it?

HansNewbie:
Some comments:
- I don't list default repo under remove.
- I am yet to clear a filter if it contains a repo that's just removed. Should I clear the filter or should the repo be re-downloaded?
- As with the previous point, if a user loads a board which contains a removed repo, should the repo be re-downloaded or should the board be cleared?  
I am looking at this scenario:  
> The board was used for a project  
> User saves the board and remove the repo once the project is closed  
> After 3 months or so, user needs to use the board again as the project needs some update.  
In this scenario, it makes much more sense to re-download the repo instead of making the board invalid.

src/main/java/backend/Logic.java
HansNewbie: Let me debug a bit more on this.

damithc:
I'm fine with redownloading missing repos. What happens when a repo is deleted while it is being used in the current board?

HansNewbie:
What I would like to do when a repo that's being used is deleted is to re-download the repo. But right now it is not doing that. This would consequently increases the API call as well.

Would it be better not to list repos that are being used in the current board? Hence no re-download.

dariusf:
About the `IllegalStateException`: the future `openRepository` returns is completed from an arbitrary thread, and anything that modifies the UI must run on the JavaFX Application Thread. You can use `Platform.runLater` to synchronise it. See what `RepoOpenedEvent` does.

Also, instead of making a standalone event, you can make it a superclass of `RepoOpenedEvent` (and change the name to reflect this, for example `OpenReposChangedEvent`). That way when `RepoOpenedEvent`s are triggered, `OpenReposChangedEventHandler`s will also be run.

The benefits of this approach are that you won't have to trigger it from so many different places, or worry about missing some out (`repo:` filter currently doesn't work with this).

damithc:
It is fine not to list repos in use.

HansNewbie:
@dariusf Yep, I have changed the code to use `Platform.runLater` instead. Will do superclassing.

@damithc Noted.

src/main/java/backend/Logic.java
```diff
@@ -208,6 +208,11 @@ public String getDefaultRepo() {
         return models.getDefaultRepo();
     }
 
+    public void removeRepository(String repoId) {
+        repoIO.removeRepository(repoId);
+        models.removeRepoModelById(repoId);
```
HansNewbie: @dariusf How do I update `models`? Here it doesn't seem that the changes is applied to the `models`.

src/main/java/backend/resource/MultiModel.java
```diff
@@ -49,6 +49,12 @@ private synchronized MultiModel add(Model model) {
         return this;
     }
 
+    public synchronized MultiModel removeRepoModelById(String RepoId) {
+        Optional<Model> repoModelToBeDeleted = getModelById(RepoId);
+        this.models.remove(repoModelToBeDeleted.get().getRepoId());
```
dariusf: You need to check if the value is present here, otherwise this can fail at runtime (like dereferencing a null pointer).

I got an email with a comment where you said this wasn't working. Is that resolved?

HansNewbie: Yeah, it's resolved. I used the `Model` instead of the repoId. Okay, will do a check.

HansNewbie:
@dariusf what tests do I need to write? I plan to add on to `MenuControlTest`. Any other tests I should consider?

src/main/java/backend/json/JSONStore.java
```diff
@@ -56,6 +57,18 @@
...
+        ensureDirectoryExists();
+        try {
+            Path repoJsonFile = Paths.get(RepoStore.directory).resolve(escapeRepoName(repoName));
+            Files.delete(repoJsonFile);
+        } catch (NoSuchFileException e) {
+            logger.error("JSON file does not exist. Repo list is incorrect. ");
+        } catch (IOException e) {
+            logger.error("Unable to delete repo JSON file. ");
+        }
+    }
```
dariusf: Rather than do the I/O in here, you should create a task that does it asynchronously, in case someday we switch to a data store where the operation takes a while. See how `saveRepository` is implemented.

dariusf:
That's integration tests. I guess you could add unit tests for back end additions if possible -- see how it's all currently tested. Also, a test that the new `Event` superclass handler fires when triggering a subclass event would be good.

src/main/java/ui/MenuControl.java
```diff
@@ -317,6 +325,59 @@ private MenuItem createRefreshMenuItem() {
...
+            if (!currentlyUsedRepo.contains(repoId)) {
+                MenuItem removeItem = new MenuItem(repoId);
+                removeItem.setOnAction(e1 -> onRepoRemove(repoId));
+                remove.getItems().add(removeItem);
+            }
+        }
+    }
+
+    private HashSet<String> getCurrentlyUsedRepo() {
+        HashSet<String> currentlyUsedRepos = new HashSet<String>();
```
dariusf: I don't think we need to traverse the panels and filters to figure out which repos are open. This should be available from `Logic`.

HansNewbie: From what I understand, a repo will never be closed. In this case, say I used a repo for a filter, then I clear the filter, the repo is still marked as open. As such, I need to traverse the panels and filters again.

dariusf: So if a repo isn't referenced in some filter, it's deleted? From a feature perspective, I'm not sure about this. A repo may not be open in the current session but I may want to keep its cached data.

HansNewbie: No, it's not deleted when it isn't referenced. If a repo is ever being used in the current session, it's marked as open throughout the whole session, even when nothing is using it anymore (say panel is closed).

As such, if a user clears a `repo` filter, the repo status is still open. There is no way of telling what `repo` filter was just cleared either. As such, I need to traverse panels to check what repos are currently in use.

dariusf: Okay, sure. I guess by 'in use' you mean 'referenced from some filter'?

My original question was whether you could use `logic.getOpenRepositories()`, because it makes sense that you should be able to close any repo that is currently open, even if it isn't referenced from some filter.

HansNewbie: Yes, I meant referenced by some filter. I can't use
`logic.getOpenRepositories()` since repo is only so-called closed when
HubTurbo quits. I guess it's for caching reason. By open, it means there is
a model for the repo (deleting the json would not close the repo either).
Right now there is no such close method.

dariusf: Sorry, I edited my post in the meantime.

src/main/java/backend/resource/MultiModel.java
```diff
@@ -49,6 +49,14 @@ private synchronized MultiModel add(Model model) {
         return this;
     }
 
+    public synchronized MultiModel removeRepoModelById(String repoId) {
+        Optional<Model> repoModelToBeDeleted = getModelById(repoId);
+        if (repoModelToBeDeleted.isPresent()) {
+            this.models.remove(repoModelToBeDeleted.get().getRepoId());
+        }
```
dariusf: I guess you could log if the repo doesn't exist, since that's probably indicative of a bug.

src/main/java/util/Utility.java
```diff
@@ -90,6 +90,20 @@ public static boolean writeFile(String fileName, String content, int issueCount)
...
+        if (validPath) {
+            try {
+                Path repoFile = Paths.get(fileName);
+                Files.delete(repoFile);
+            } catch (IOException e) {
+                logger.error(e.getLocalizedMessage(), e);
+                return true;
+            }
+        }
+        return false;
```
dariusf: What does the return value of this signify? Might need javadocs to explain. If it's whether the file ultimately *does not* exist, there might be a missing return (in the `try` block).

src/test/java/tests/EventTests.java
```diff
@@ -60,4 +60,15 @@ public void basics() {
...
+    @Test
+    public void testSuperclassHandlerOnSubclassEvent() {
+        events.register(superclassHandler);
+
+        RepoOpenedEvent subclassEvent = new RepoOpenedEvent("");
+
+        events.post(subclassEvent);
+    }
+
+    private final EventHandler superclassHandler = (OpenReposChangedEventHandler) e -> assertTrue(true);
```
dariusf: These fields should be locals, since they aren't being shared. Also the `EventBus` instance probably shouldn't be shared, as it's possible for events to interfere.

We want to ensure that the subclass handler also fires, so I guess we should have both handlers:

- Dispatch superclass event, ensure subclass doesn't fire.
- Dispatch subclass event, sure both fire.

src/test/java/tests/MultiModelTest.java
```diff
@@ -25,4 +39,21 @@ public void multiModelTest() {
...
 
+    @Test
+    public void testRemoveModel() throws ExecutionException, InterruptedException {
+        final String repoId = "dummy1/dummy1";
+        MultiModel models = new MultiModel(new Preferences(true));
+        models.queuePendingRepository(repoId);
+        RepoIO testIO = new RepoIO(true, false);
+        testIO.openRepository(repoId).thenApply(models::addPending);
+
+        TestUtils.delay(3);
```
dariusf: `sleep` in tests is bad, as it may cause randomness and timing issues. Is it possible to replace this with `PlatformEx.waitOnFxThread()`?

src/test/java/tests/StoreTests.java
```diff
@@ -121,9 +123,24 @@ public void testLoadNonExistentRepo() throws InterruptedException, ExecutionExce
...
-        UITest.clearTestFolder();
+    @Test
+    public void testRemoveRepo() throws InterruptedException, ExecutionException {
+        RepoIO testIO = new RepoIO(true, true);
+
+        Model dummy1 = testIO.openRepository("dummy1/dummy1").get();
+        UI.events.triggerEvent(UpdateDummyRepoEvent.newIssue("dummy1/dummy1"));
+        testIO.updateModel(dummy1).get();
+
+        TestUtils.delay(3);
```
dariusf: Same here.

dariusf:
Good work! Just a few minor things left.

src/main/java/backend/resource/MultiModel.java
HansNewbie: As there is no logger in `MultiModel`, do I need to create one inside?

src/main/java/util/Utility.java
HansNewbie: Will add a javadoc. It's whether there was any error on deleting.

HansNewbie:
@dariusf `PlatformEx.waitOnFxThread()` doesn't work on both tests you mentioned since both of them has no JavaFX component. Should I leave it with delay or should I use something else? The wait is necessary due to asynchronous behaviour.

src/main/java/backend/resource/MultiModel.java
dariusf: Yes, just follow how other loggers are created.

src/test/java/guitests/RepositorySelectorTest.java
```diff
@@ -97,6 +98,15 @@ public void repositorySelectorTest() {
         assertEquals(4, comboBox.getItems().size());
         assertEquals("dummy4/dummy4", primaryRepo);
 
+        // we check if delete repo works
+        click("Repos");
+        push(KeyCode.DOWN);
+        push(KeyCode.RIGHT);
+        push(KeyCode.DOWN);
+        push(KeyCode.ENTER);
+        sleep(5000);
```
dariusf: This can be replaced with a combination of `PlatformEx.waitOnFxThread` and `UITest.waitForValue`/`UITest.waitUntilNodeAppears`.

dariusf:
Right. In that case you could block on the future returned by `thenApply` with `get` until it's done. `removeRepoModelById` isn't async so you don't have to sleep.

You should be able to do the same thing for `StoreTests`, but if you can't there is a problem with how `updateModel` is written.

HansNewbie:
I may need to push this to the next milestone because I have another deadline today.

dariusf:
Nice work!

HansNewbie:
Sorry, I named the merge commit wrongly. Should I change the commit message?

dariusf:
It's fine. Force pushing to master can cause worse problems.

HansNewbie:
Alright sure. Thanks.

PR #944, opened 2015-09-07
Issues are marked as unread after a self-update

### LINKED ISSUE #750: Issues are marked as unread after a self-update
Followed from discussions in #679.

The issue passes the `updated:XX` filter because there is indeed a non-self update in the specified window. It passes the `is:unread` filter as it is marked as read before the model in the GUI is updated (which causes it to be marked as unread again).

rd1992:
Fixes #750

### DIFF
--- a/src/main/java/backend/resource/MultiModel.java
+++ b/src/main/java/backend/resource/MultiModel.java
@@ -207,13 +207,32 @@ public class MultiModel implements IModel {
                     // TODO O(n^2), optimise by preprocessing into a map or sorting
                     if (!existingModel.getIssueById(i).equals(model.getIssueById(i))) {
                         assert model.getIssueById(i).isPresent();
-                        // It's no longer currently read, but it retains its updated time.
-                        // No changes to preferences.
-                        model.getIssueById(i).get().setIsCurrentlyRead(false);
+                        /** If the issue was marked read after the last update, but update came in later
+                         * due to network latency, ensure the issue retains its read state
+                        **/
+                        if (updateReadAtState(model.getIssueById(i))) {
+                            model.getIssueById(i).get().setIsCurrentlyRead(false);
+                        }
                     }
                 }
             }
         }
+    }
+
+    private boolean updateReadAtState(Optional<TurboIssue> issue) {
+        if (issue.get().isCurrentlyRead()) {
+            LocalDateTime markedReadAt = issue.get().getMarkedReadAt().get();
+            LocalDateTime updatedAt = issue.get().getUpdatedAt();
+            if (updatedAt.isAfter(markedReadAt)) {
+                return true;
+            } else if (updatedAt.isBefore(markedReadAt) &&
+                       (markedReadAt.getSecond() - updatedAt.getSecond() < 3)) {
+                return true;
+            } else {
+                return false;
+            }
+        }
+        return false;
     }
 
     private void ______BOILERPLATE______() {
--- a/src/main/java/backend/resource/MultiModel.java
+++ b/src/main/java/backend/resource/MultiModel.java
@@ -5,6 +5,7 @@ import backend.interfaces.IModel;
 import prefs.Preferences;
 
 import java.time.LocalDateTime;
+import java.time.temporal.ChronoUnit;
 import java.util.*;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -201,15 +202,15 @@ public class MultiModel implements IModel {
         for (Model model : newModels) {
             assert models.containsKey(model.getRepoId());
             Model existingModel = models.get(model.getRepoId());
-            if (existingModel.getIssues().equals(model.getIssues())) return;
+            if (existingModel.getIssues().equals(model.getIssues())) continue;
             for (int i = 1; i <= model.getIssues().size(); i++) {
-                if (existingModel.getIssueById(i).equals(model.getIssueById(i))) return;
+                if (existingModel.getIssueById(i).equals(model.getIssueById(i)) || 
+                    !wasNotJustMarkedAsRead(model.getIssueById(i).get())) continue;
                 // TODO O(n^2), optimise by preprocessing into a map or sorting
                 assert model.getIssueById(i).isPresent();
                 /** If the issue was marked read after the last update, but update came in later
                  * due to network latency, ensure the issue retains its read state
                 **/
-                if (!wasNotJustMarkedAsRead(model.getIssueById(i).get())) return;
                 model.getIssueById(i).get().setIsCurrentlyRead(false);
             }
         }
@@ -219,10 +220,9 @@ public class MultiModel implements IModel {
         if (!issue.isCurrentlyRead() || !issue.getMarkedReadAt().isPresent()) return false;
         LocalDateTime markedReadAt = issue.getMarkedReadAt().get();
         LocalDateTime updatedAt = issue.getUpdatedAt();
-        if (updatedAt.isAfter(markedReadAt)) {
-            return true;
-        } else if (updatedAt.isBefore(markedReadAt) &&
-                   (markedReadAt.toEpochSecond(null) - updatedAt.toEpochSecond(null) < 3L)) {
+        if (updatedAt.isBefore(markedReadAt) && (updatedAt.until(markedReadAt, ChronoUnit.SECONDS) < 3L)) {
+            return false;
+        } else if (updatedAt.isAfter(markedReadAt)) {
             return true;
         } else {
             return false;
--- a/src/main/java/backend/resource/MultiModel.java
+++ b/src/main/java/backend/resource/MultiModel.java
@@ -201,38 +201,32 @@ public class MultiModel implements IModel {
         for (Model model : newModels) {
             assert models.containsKey(model.getRepoId());
             Model existingModel = models.get(model.getRepoId());
-            if (!existingModel.getIssues().equals(model.getIssues())) {
-                // Find issues that have changed and update preferences with them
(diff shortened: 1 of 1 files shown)


src/main/java/backend/resource/MultiModel.java
```diff
@@ -207,9 +207,12 @@ private void preprocessUpdatedIssues(List<Model> newModels) {
...
                         assert model.getIssueById(i).isPresent();
-                        // It's no longer currently read, but it retains its updated time.
-                        // No changes to preferences.
-                        model.getIssueById(i).get().setIsCurrentlyRead(false);
+                        /** If the issue was marked read after the last update, but update came in later
+                         * due to network latency, ensure the issue retains its read state
+                        **/
+                        if (model.getIssueById(i).get().getUpdatedAt().isAfter(model.getIssueById(i).get().getMarkedReadAt().get())) {
+                            model.getIssueById(i).get().setIsCurrentlyRead(false);
+                        }
```
damithc: Shouldn't there be some time check? The situation in concern happens because the update arrives a few seconds after marking as read. If it arrived much later, it should be recognized as a legitimate update.

rd1992: yes there should be. i forgot to sync my local branch. i have kept the interval at 3 seconds for now. or is that too long in your opinion?

damithc: Only dog fooding can tell us if 3s is enough.

src/main/java/backend/resource/MultiModel.java
```diff
@@ -207,9 +207,14 @@ private void preprocessUpdatedIssues(List<Model> newModels) {
                     // TODO O(n^2), optimise by preprocessing into a map or sorting
                     if (!existingModel.getIssueById(i).equals(model.getIssueById(i))) {
                         assert model.getIssueById(i).isPresent();
-                        // It's no longer currently read, but it retains its updated time.
-                        // No changes to preferences.
-                        model.getIssueById(i).get().setIsCurrentlyRead(false);
+                        /** If the issue was marked read after the last update, but update came in later
+                         * due to network latency, ensure the issue retains its read state
+                        **/
+                        if (model.getIssueById(i).get().getUpdatedAt().isAfter(model.getIssueById(i).get().getMarkedReadAt().get())
```
dariusf: That's a monster conditional... you should factor it out into a predicate. Binding `model.getIssueById(i)` will probably help.

Also you need to handle the case where `markedReadAt` is not present, otherwise it can fail at runtime like a dereferenced null pointer.

dariusf:
There are some unrelated changes in this commit. You should probably split them out.

rd1992:
@dariusf i went with a method instead of predicate. Seemed cleaner. you can review it now

src/main/java/backend/resource/MultiModel.java
```diff
@@ -207,15 +207,34 @@ private void preprocessUpdatedIssues(List<Model> newModels) {
...
+                         * due to network latency, ensure the issue retains its read state
+                        **/
+                        if (updateReadAtState(model.getIssueById(i))) {
+                            model.getIssueById(i).get().setIsCurrentlyRead(false);
+                        }
                     }
                 }
             }
         }
     }
```
damithc: This method has too much nesting. Nesting can be reduced by reversing `if` conditions.
Instead of 
```
if(condition){
   //do stuff
}
```
use,
```
if(!condition) return;
//do stuff
```

src/main/java/backend/resource/MultiModel.java
```diff
@@ -207,15 +207,34 @@ private void preprocessUpdatedIssues(List<Model> newModels) {
...
+            LocalDateTime updatedAt = issue.get().getUpdatedAt();
+            if (updatedAt.isAfter(markedReadAt)) {
+                return true;
+            } else if (updatedAt.isBefore(markedReadAt) &&
+                       (markedReadAt.getSecond() - updatedAt.getSecond() < 3)) {
+                return true;
+            } else {
+                return false;
+            }
+        }
```
damithc: Same here. Instead of a very tall `if` block, reverse the condition and convert it to an early return.

dariusf:
- The method name for `updateReadAtState`. Maybe change it to `wasNotJustMarkedAsRead`? It's very nuanced... That's the clearest name I can think of for it. Or maybe its result could be inverted, then you could check `if (!wasJustMarkedAsRead(issue)) { ... }`?
- The method signature: `updateReadAtState` can just take a `TurboIssue`. We already assert its presence outside.
- The case where `issue.get().getMarkedReadAt()` is not present must be handled. `if (issue.getMarkedReadAt().isPresent()) { ... } else { ... }`
- If `updatedAt` is at `0:09:59` and `markedReadAt` is at `0:10:03`, the comparison will return the wrong result, i.e. it's not enough to compare seconds-of-a-minute. You should compare instant (and deal with `long`s), or there might be a `LocalDateTime` method for it.

src/main/java/backend/resource/MultiModel.java
```diff
@@ -201,21 +201,34 @@ private void preprocessUpdatedIssues(List<Model> newModels) {
...
-                }
+            if (existingModel.getIssues().equals(model.getIssues())) return;
+            for (int i = 1; i <= model.getIssues().size(); i++) {
+                if (existingModel.getIssueById(i).equals(model.getIssueById(i))) return;
+                // TODO O(n^2), optimise by preprocessing into a map or sorting
+                assert model.getIssueById(i).isPresent();
+                /** If the issue was marked read after the last update, but update came in later
+                 * due to network latency, ensure the issue retains its read state
+                **/
+                if (!wasNotJustMarkedAsRead(model.getIssueById(i).get())) return;
```
dariusf: Returning here doesn't look right. Did you mean to continue instead?

src/main/java/backend/resource/MultiModel.java
```diff
@@ -201,21 +201,34 @@ private void preprocessUpdatedIssues(List<Model> newModels) {
...
     }
 
+    private boolean wasNotJustMarkedAsRead(TurboIssue issue) {
+        if (!issue.isCurrentlyRead() || !issue.getMarkedReadAt().isPresent()) return false;
+        LocalDateTime markedReadAt = issue.getMarkedReadAt().get();
+        LocalDateTime updatedAt = issue.getUpdatedAt();
+        if (updatedAt.isAfter(markedReadAt)) {
+            return true;
+        } else if (updatedAt.isBefore(markedReadAt) &&
+                   (markedReadAt.toEpochSecond(null) - updatedAt.toEpochSecond(null) < 3L)) {
```
dariusf: The zone offset [must be non-null](http://cr.openjdk.java.net/~sherman/jdk8_threeten/defaultM2/raw_files/new/src/share/classes/javax/time/chrono/ChronoLocalDateTime.java).

There is `Utility.localDateTimeToLong`, which I guess you could use. An alternative is the [`until` method](https://github.com/HubTurbo/HubTurbo/blob/master/src/main/java/filter/expression/Qualifier.java#L534-L535). That might be better actually.

src/main/java/backend/resource/MultiModel.java
```diff
@@ -201,21 +202,33 @@ private void preprocessUpdatedIssues(List<Model> newModels) {
...
-                        assert model.getIssueById(i).isPresent();
-                        // It's no longer currently read, but it retains its updated time.
-                        // No changes to preferences.
-                        model.getIssueById(i).get().setIsCurrentlyRead(false);
-                    }
-                }
+            if (existingModel.getIssues().equals(model.getIssues())) continue;
+            for (int i = 1; i <= model.getIssues().size(); i++) {
+                if (existingModel.getIssueById(i).equals(model.getIssueById(i)) || 
+                    !wasNotJustMarkedAsRead(model.getIssueById(i).get())) continue;
```
dariusf: The double negation is quite confusing. Could we negate the method name instead?

src/main/java/backend/resource/MultiModel.java
```diff
@@ -201,21 +202,33 @@ private void preprocessUpdatedIssues(List<Model> newModels) {
...
+                /** If the issue was marked read after the last update, but update came in later
+                 * due to network latency, ensure the issue retains its read state
+                **/
+                model.getIssueById(i).get().setIsCurrentlyRead(false);
             }
         }
     }
 
+    private boolean wasNotJustMarkedAsRead(TurboIssue issue) {
+        if (!issue.isCurrentlyRead() || !issue.getMarkedReadAt().isPresent()) return false;
```
dariusf: Maybe it's just me, but I find it quite difficult to reason about this function's correctness still. We should break it up into smaller functions. That would make it easier to test as well.

Something like this? The logic should be equivalent.

```java
/**
 * Determines if an issue was recently marked as read. This happens if
 * it is marked as read a short time before an update occurs.
 */
private boolean wasJustMarkedAsRead() {
    if (!isCurrentlyRead()) return false;
    return wasReadWithinInterval(3L, ChronoUnit.SECONDS);
}

/**
 * Determines if an issue was read in a short interval before its last update.
 */
private boolean wasReadWithinInterval(long seconds, ChronoUnit unit) {
    if (!getMarkedReadAt().isPresent()) return false;
    return getUpdatedAt().isBefore(getMarkedReadAt().get())
        && getUpdatedAt().until(getMarkedReadAt().get(), unit) < seconds;
}
```

I put it inside TurboIssue. Seems to fit better because it's entirely a function of the issue's state.

dariusf:
Sorry, couple more things.

Might also need tests. Waiting to see the coverage now.

rd1992:
Closing this as this is handled in #955 now. Closing this.

Issue #953, opened 2015-09-09
Only consider self updates when deciding not to mark issues as unread


dariusf:
Follow-up to #750. For now (#944) it works for all updates, not just self ones.

PR #955, opened 2015-09-09
Unread updates appearing as read #939

### LINKED ISSUE #939: Unread updates appearing as read
This appears as read although it was updated in the last 9 hours and I did not use HT during that 9 hours. I noticed it for two HT issues so far.
image

ndt93:
Fix #939 

This PR also helps with the same problem as in #944, so @rd1992 might want to take a look.

Also, I think the logic of fixes or extensions to read/unread state should be kept within TurboIssue and expose through a unifying interface such as `TurboIssue.isCurrentlyRead`, since they are inherently tied to an issue. For example, #953 should be ideally done by extending `isCurrentlyRead` and no complex logic is necessary in the Model level (like what's being done in `MultiModel.preprocessUpdatedIssues`)

### DIFF
--- a/src/main/java/backend/resource/TurboIssue.java
+++ b/src/main/java/backend/resource/TurboIssue.java
@@ -4,6 +4,7 @@ import backend.IssueMetadata;
 import backend.resource.serialization.SerializableIssue;
 import org.eclipse.egit.github.core.Issue;
 import org.eclipse.egit.github.core.Label;
+import prefs.Preferences;
 import util.Utility;
 
 import java.time.LocalDateTime;
@@ -346,6 +347,17 @@ public class TurboIssue {
         }
 
         return getMarkedReadAt().get().isAfter(getUpdatedAt());
+    }
+
+    public void markAsRead(Preferences prefs) {
+        LocalDateTime now = LocalDateTime.now();
+        setMarkedReadAt(Optional.of(now));
+        prefs.setMarkedReadAt(getRepoId(), getId(), getMarkedReadAt().get());
+    }
+
+    public void markAsUnread(Preferences prefs) {
+        setMarkedReadAt(Optional.empty());
+        prefs.clearMarkedReadAt(getRepoId(), getId());
     }
 
     /**
--- a/src/main/java/backend/resource/TurboIssue.java
+++ b/src/main/java/backend/resource/TurboIssue.java
@@ -6,6 +6,7 @@ import org.eclipse.egit.github.core.Issue;
 import org.eclipse.egit.github.core.Label;
 import prefs.Preferences;
 import util.Utility;
+import static util.Utility.replaceNull;
 
 import java.time.LocalDateTime;
 import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


ndt93:
Coverage failed due to unrelated code. @dariusf Can help to review this? Thanks.

src/main/java/backend/resource/TurboIssue.java
```diff
@@ -334,17 +329,35 @@ public IssueMetadata getMetadata() {
...
     public void setMarkedReadAt(Optional<LocalDateTime> markedReadAt) {
         this.markedReadAt = markedReadAt;
     }
+
     public boolean isCurrentlyRead() {
-        return isCurrentlyRead;
+        if (!getMarkedReadAt().isPresent()) {
+            return false;
+        }
+        if (getUpdatedAt() == null) {
```
dariusf: Is the update time ever null? If this is to account for GitHub Issues that are just created, maybe we could set it to the creation time in the TurboIssue constructor.

dariusf:
I like the idea of removing the preprocessing step, and letting issues decide whether or not to toggle their unread status, but I'm not sure how it works at the moment. Is it done via the TurboIssue constructor, so the check is done every time we get an updated version of the same issue and construct a new one?

src/main/java/backend/resource/TurboIssue.java
ndt93: Not likely since Github will set `updatedAt` to `createdAt` the first time. But I will assign it to the `createdAt` in the constructor then, for safety.

ndt93:
The preprocessing step is already not doing anything in this PR (only has an assert statement in the for loop) since there isn't static a read state to be toggled anymore (`markedReadAt` is checked against `updatedAt` everytime `isCurrentlyRead()` is called) but I didn't remove it since it can clash with what @rd1992 is doing in #944

dariusf:
I see, very nice. The only thing left is to reconcile these two PRs then, and remove the processing step in one of them. With this, I'm not sure how much of the other one is still needed, since the read state will never be wrong and never have to be updated. @rd1992?

#953 would then involve one more check using the new metadata fields from #924.

rd1992:
Good work @ndt93 . @dariusf yeah so the #944 is not needed as this PR and the #953 will encompass whatever #944 is doing.

ndt93:
All right, I will remove the preprocessing step for this PR then.

ndt93:
@dariusf I put the `updatedAt` null replacement in constructors and setter and removed preprocessing step MultiModel update. coveralls website is down but the change only involves deleting code so coverage shouldn't decrease. You can check again.

src/main/java/backend/resource/TurboIssue.java
```diff
@@ -126,7 +126,7 @@ public TurboIssue(String repoId, Issue issue) {
         this.description = issue.getBody() == null
             ? ""
             : issue.getBody();
-        this.updatedAt = Utility.dateToLocalDateTime(issue.getUpdatedAt());
+        this.updatedAt = replaceNull(Utility.dateToLocalDateTime(issue.getUpdatedAt()), this.createdAt);
```
dariusf: `dateToLocalDateTime` will fail if it gets a null value. Might need a manual check here.

ndt93:
@dariusf All right, fixed.

dariusf:
Nice work!

PR #959, opened 2015-09-12
Import KeyCodeCombinations enums to related classes #916

### LINKED ISSUE #916: Import KeyCodeCombinations enums to related classes
Discussed in #909 

In ```ListPanel```, ```PanelControl```, ```NavigableListView```, ```MenuControl```, ```RepositorySelector``` and a few tests.

HyungJon:
Fixes #916

### DIFF
--- a/src/main/java/ui/listpanel/ListPanel.java
+++ b/src/main/java/ui/listpanel/ListPanel.java
@@ -1,10 +1,9 @@
 package ui.listpanel;
 
-import java.time.LocalDateTime;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Optional;
+
 import backend.interfaces.IModel;
 import backend.resource.TurboIssue;
 import filter.expression.Qualifier;
@@ -14,8 +13,8 @@ import javafx.scene.input.KeyCodeCombination;
 import javafx.scene.input.KeyEvent;
 import javafx.scene.layout.Priority;
 import ui.UI;
-import ui.components.IssueListView;
 import ui.components.KeyboardShortcuts;
+import ui.components.IssueListView;
 import ui.issuepanel.FilterPanel;
 import ui.issuepanel.PanelControl;
 import util.KeyPress;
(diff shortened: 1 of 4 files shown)


src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -145,138 +172,137 @@ private void setupListView() {
...
-            if (KeyboardShortcuts.SWITCH_DEFAULT_REPO.match(event)) {
+            if (SWITCH_DEFAULT_REPO.match(event)) {
                 ui.switchDefaultRepo();
             }
-            if (KeyboardShortcuts.UNDO_LABEL_CHANGES.match(event)) {
+            if (UNDO_LABEL_CHANGES.match(event)) {
                 ui.triggerNotificationAction();
             }
-            for (KeyCodeCombination key:KeyboardShortcuts.JUMP_TO_FIRST_ISSUE_KEYS) {
+            for (KeyCodeCombination key:JUMP_TO_FIRST_ISSUE_KEYS) {
```
dariusf: This should be spaced out properly.

dariusf:
Hmm, maybe we shouldn't static import those that are not NAMED_LIKE_CONSTANTS... May be confusing in the contexts in which they are used.

HyungJon:
It does make sense since those aren't immutable, we might not want to import them like that. Making the change.

dariusf:
Going to assume the coverage drop is a false positive. Could you squash? I'll merge then.

HyungJon:
Okay. Can advice me on how to squash when there's a merge (with all involved commits) in between?

dariusf:
`git rebase -i origin/master` should do it.

Hmm, it's more involved than I thought. I managed to get it on my end, but if you 
git checkout Issue916


```sh
git rebase -i origin/master
# Fix merge conflicts
git add .
git rebase --continue

# If everything fails,

```

HyungJon:
Right. That should be squashed now.

dariusf:
Nice.

PR #960, opened 2015-09-12
user can sort issues by status #913

### LINKED ISSUE #913: user can sort issues by status
so that the user can group closed issues and open issues separately

rd1992:
Fixes #913

### DIFF
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -244,7 +244,7 @@ Matches issues of the given repository. If omitted, will match issues of the def
 
 ### sort
 
-*Expects a comma-separated list of sorting keys. For example, `repo, ~updated, -comments`. Any negation operator can be used to invert them. Keys can be any of `comments`, `repo`, `updated`, `date` (an alias for `updated`), `id`, `assignee` or a label group. Label groups can be disambiguated by appending a `.`.*
+*Expects a comma-separated list of sorting keys. For example, `repo, ~updated, -comments`. Any negation operator can be used to invert them. Keys can be any of `comments`, `repo`, `updated`, `date` (an alias for `updated`), `id`, `assignee`, `status`(`open` or `closed`) or a label group. Label groups can be disambiguated by appending a `.`.*
 
 Sorts a repository by the list of keys, going from left to right. Negated keys will reverse the ordering that they describe.
 
--- a/src/test/java/guitests/SortTest.java
+++ b/src/test/java/guitests/SortTest.java
@@ -16,20 +16,17 @@ public class SortTest extends UITest {
         type("id");
         push(KeyCode.ENTER);
         doubleClick("#dummy/dummy_col0_filterTextField");
-        doubleClick("#dummy/dummy_col0_filterTextField");
         // Descending ID (default)
         type("sort");
         press(KeyCode.SHIFT).press(KeyCode.SEMICOLON).release(KeyCode.SEMICOLON).release(KeyCode.SHIFT);
         type("-id");
         push(KeyCode.ENTER);
         doubleClick("#dummy/dummy_col0_filterTextField");
-        doubleClick("#dummy/dummy_col0_filterTextField");
         // Comment count
         type("sort");
         press(KeyCode.SHIFT).press(KeyCode.SEMICOLON).release(KeyCode.SEMICOLON).release(KeyCode.SHIFT);
         type("comments");
         push(KeyCode.ENTER);
-        doubleClick("#dummy/dummy_col0_filterTextField");
         doubleClick("#dummy/dummy_col0_filterTextField");
         // Repo ID (e.g. dummy2/dummy)
         type("sort");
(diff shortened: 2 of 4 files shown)


rd1992:
@dariusf dont know why coverage went down. Have added the relevant tests though.

dariusf:
Looks good, but these are integration tests -- we should have [tests for just the qualifiers](https://github.com/HubTurbo/HubTurbo/blob/master/src/test/java/tests/SortQualifierTests.java) as well.

Maybe we can support `sort:status` also? Just an alias for `sort:open`.

rd1992:
Added the tests. But i dont know about the `sort:status` thing. Seems a bit ambiguous to me. @damithc  what do you think?

damithc:
Actually, what I wanted was `sort:status`. `sort:open` is a bit weird because it's like saying `sort:red` when we actually mean `sort:color`.

rd1992:
So what would sort:status do? Open issues will be shown first? And what if user wants to show closed first?

damithc:
Yes, `sort:status` will show open issues first and `sort:-status` will show closed issues first. This is based on the assumption that people are interested in open issues more often than not.

rd1992:
@dariusf i have made the relevant changes. Travis is being weird yet again though.

src/test/java/guitests/SortTest.java
```diff
@@ -50,5 +50,19 @@ public void sortTest () {
         press(KeyCode.SHIFT).press(KeyCode.SEMICOLON).release(KeyCode.SEMICOLON).release(KeyCode.SHIFT);
         type("-date");
         push(KeyCode.ENTER);
+        // Issue state (open)
+        doubleClick("#dummy/dummy_col0_filterTextField");
+        doubleClick("#dummy/dummy_col0_filterTextField");
```
dariusf: Is double-clicking twice needed here?

dariusf:
Apart from that, this looks good. The build usually needs to be run a few times.

docs/filters.md
```diff
@@ -244,7 +244,7 @@ Matches issues of the given repository. If omitted, will match issues of the def
 
 ### sort
 
-*Expects a comma-separated list of sorting keys. For example, `repo, ~updated, -comments`. Any negation operator can be used to invert them. Keys can be any of `comments`, `repo`, `updated`, `date` (an alias for `updated`), `id`, `assignee` or a label group. Label groups can be disambiguated by appending a `.`.*
+*Expects a comma-separated list of sorting keys. For example, `repo, ~updated, -comments`. Any negation operator can be used to invert them. Keys can be any of `comments`, `repo`, `updated`, `date` (an alias for `updated`), `id`, `assignee`, `status`(`open` or `closed`) or a label group. Label groups can be disambiguated by appending a `.`.*
```
damithc: Not strictly related to this issue but can we add some examples here? I didn't even know we can sort by label groups and not quite sure how to do that anyway.

damithc: Rephrase as:
... `status`(shows `open` issues first followed by `closed` issues)

src/test/java/guitests/SortTest.java
```diff
@@ -50,5 +50,19 @@ public void sortTest () {
...
+        type("sort");
+        press(KeyCode.SHIFT).press(KeyCode.SEMICOLON).release(KeyCode.SEMICOLON).release(KeyCode.SHIFT);
+        type("status");
```
damithc: So it's not possible to type "sort:status" in one shot? If cannot, perhaps we should write our own wrapper method to take a string and convert it to the correct key strokes?

docs/filters.md
dariusf: `sort:priority` will arrange issues within the group by name (high, low, etc.) in alphabetical order, and put issues without that label group at the bottom.

If we're going to describe each input to `sort`, we should change the formatting and make it a list.

damithc: Yes, a list would  be better.

src/test/java/guitests/SortTest.java
rd1992: special chars such as `:` or `_` wont work for `type()`. I'll create a separate issue for this. The whole class needs to be changed.

docs/filters.md
rd1992: So are we going with the list suggestion? I personally like it since it would explain the use of each qualifier.

src/test/java/guitests/SortTest.java
damithc: Yes, it can be a separate issue.

src/test/java/guitests/SortTest.java
dariusf: So is this needed?

rd1992: Hmm no. But this file contains this double doubleclick code all over the place. Lets fix it in #966?

damithc: Can be a separate issue, but doesn't sound like it is related to #966

rd1992: ok lets fix it here. seems too minor to create an issue for deleting 8 lines

rd1992:
@dariusf removed the redundant code.

PR #961, opened 2015-09-13
Ctrl+1,2,3,...9 keys should jump to nth issue

### LINKED ISSUE #937: Ctrl+1,2,3,...9 keys should jump to nth issue
Currently we can use `1` to jump to first issue if the focus is not in a filter box.
Instead of that, may be we allow to use `Ctrl+number` to jump to `n`th issue? That can work from anywhere.

codemanmeet:
Fixes #937


src/main/java/ui/components/KeyboardShortcuts.java
```diff
@@ -130,10 +130,17 @@
...
+        Map <Integer, KeyCodeCombination> result = new HashMap<>();
+        result.put(1, new KeyCodeCombination(KeyCode.DIGIT1, KeyCombination.CONTROL_DOWN));
+        result.put(2, new KeyCodeCombination(KeyCode.DIGIT2, KeyCombination.CONTROL_DOWN));
+        result.put(3, new KeyCodeCombination(KeyCode.DIGIT3, KeyCombination.CONTROL_DOWN));
+        result.put(4, new KeyCodeCombination(KeyCode.DIGIT4, KeyCombination.CONTROL_DOWN));
+        result.put(5, new KeyCodeCombination(KeyCode.DIGIT5, KeyCombination.CONTROL_DOWN));
+        result.put(6, new KeyCodeCombination(KeyCode.DIGIT6, KeyCombination.CONTROL_DOWN));
+        result.put(7, new KeyCodeCombination(KeyCode.DIGIT7, KeyCombination.CONTROL_DOWN));
+        result.put(8, new KeyCodeCombination(KeyCode.DIGIT8, KeyCombination.CONTROL_DOWN));
+        result.put(9, new KeyCodeCombination(KeyCode.DIGIT9, KeyCombination.CONTROL_DOWN));
```
dariusf: You can return an immutable view of the map using `Collections.unmodifiableMap`. Given the name, it's expected to be constant/immutable.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -158,11 +158,11 @@ protected void setupKeyEvents() {
             }
             if (FIRST_ISSUE.match(e)) {
                 e.consume();
-                selectFirstItem();
+                selectnthItem(1);
```
dariusf: I think this should be named 'selectNthItem', to be more consistent with how we capitalise in names. Same for the other occurrences.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -158,11 +158,11 @@ protected void setupKeyEvents() {
...
             }
             if (LAST_ISSUE.match(e)) {
                 e.consume();
-                selectLastItem();
+                selectnthItem(getItems().size() - 1);
```
dariusf: This doesn't look right, given that the method uses 1-based indexing.

Is there a test for this hotkey?

src/main/java/ui/components/NavigableListView.java
```diff
@@ -188,21 +188,12 @@ public void setOnItemSelected(IntConsumer callback) {
         onItemSelected = callback;
     }
 
-    public void selectFirstItem() {
+    public void selectnthItem(int n) {
```
dariusf: You can add a javadoc here saying the indexing is 1-based.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -295,10 +294,10 @@ private void setupKeyboardShortcuts() {
             if (UNDO_LABEL_CHANGES.match(event)) {
                 ui.triggerNotificationAction();
             }
-            for (KeyCodeCombination key : JUMP_TO_FIRST_ISSUE_KEYS) {
-                if (key.match(event)) {
+            for (Integer nKey:JUMP_TO_NTH_ISSUE_KEYS.keySet()) {
```
dariusf: This should be spaced out properly.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -295,10 +294,10 @@ private void setupKeyboardShortcuts() {
...
+                if (JUMP_TO_NTH_ISSUE_KEYS.get(nKey).match(event)){
                     event.consume();
-                    listView.selectFirstItem();
+                    listView.selectnthItem(nKey);
```
dariusf: You might want to break out of the loop after this, because the key events are mutually exclusive.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -188,21 +188,12 @@ public void setOnItemSelected(IntConsumer callback) {
...
-        onItemSelected.accept(selectedIndex.get());
-    }
-
-    public void selectLastItem() {
-        requestFocus();
-        if (getItems().size() == 0) return;
-        getSelectionModel().clearAndSelect(getItems().size() - 1);
-        scrollAndShow(getItems().size() - 1);
-        selectedIndex = Optional.of(getItems().size() - 1);
+        if (getItems().size() < n) return;
```
dariusf: This can be changed into an assertion, since off-by-one errors here are programmer errors. It should also assert the lower bound (`> 0`).

src/main/java/ui/listpanel/ListPanel.java
dariusf: Also you might want to iterate through the `entrySet` instead, since you need both key and value in the loop.

src/test/java/guitests/KeyboardShortcutsTest.java
```diff
@@ -69,9 +69,44 @@ public void keyboardShortcutsTest() {
...
+        assertEquals(9, selectedIssueId);
+        clearSelectedIssueId();
+        push(KeyCode.ESCAPE);
+        press(KeyCode.CONTROL).press(KeyCode.DIGIT3).release(KeyCode.CONTROL).release(KeyCode.DIGIT3);
+        sleep(1000);
+        assertEquals(8, selectedIssueId);
+        clearSelectedIssueId();
+        push(KeyCode.ESCAPE);
+        press(KeyCode.CONTROL).press(KeyCode.DIGIT4).release(KeyCode.CONTROL).release(KeyCode.DIGIT4);
+        sleep(1000);
```
dariusf: Could you see if the sleeps can be removed, or replaced with `PlatformEx.waitOnFxThread`?

src/main/java/ui/components/KeyboardShortcuts.java
codemanmeet: changed to `Collections.unmodifiableMap`

src/main/java/ui/components/KeyboardShortcuts.java
```diff
@@ -130,10 +131,17 @@
     public static final KeyCodeCombination CLOSE_PANEL =
             new KeyCodeCombination(KeyCode.W, KeyCombination.CONTROL_DOWN);
     
-    private static Set<KeyCodeCombination> populateJumpToFirstIssueSet() {
-        Set <KeyCodeCombination> result = new HashSet<KeyCodeCombination>();
-        result.add(new KeyCodeCombination(KeyCode.DIGIT1));
-        result.add(new KeyCodeCombination(KeyCode.DOWN, KeyCombination.CONTROL_DOWN));
+    private static Map<Integer, KeyCodeCombination> populateJumpToNthIssueMap() {
+        Map<Integer, KeyCodeCombination> result = Collections.unmodifiableMap(new HashMap<>());
```
dariusf: This won't work -- the point of an unmodifiable map is that it's immutable. You have to make the map unmodifiable after putting the key combinations in, before returning.

codemanmeet: Just realised that. fixed it.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -188,21 +188,17 @@ public void setOnItemSelected(IntConsumer callback) {
...
+     */
 
-    public void selectLastItem() {
+    public void selectNthItem(int n) {
         requestFocus();
-        if (getItems().size() == 0) return;
-        getSelectionModel().clearAndSelect(getItems().size() - 1);
-        scrollAndShow(getItems().size() - 1);
-        selectedIndex = Optional.of(getItems().size() - 1);
+        assert(getItems().size() < n);
```
dariusf: Needs assertion for the lower bound

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -295,10 +296,11 @@ private void setupKeyboardShortcuts() {
             if (UNDO_LABEL_CHANGES.match(event)) {
                 ui.triggerNotificationAction();
             }
-            for (KeyCodeCombination key : JUMP_TO_FIRST_ISSUE_KEYS) {
-                if (key.match(event)) {
+            for (Map.Entry<Integer, KeyCodeCombination> entry:JUMP_TO_NTH_ISSUE_KEYS.entrySet()) {
```
dariusf: There should be spaces around the `:`

src/main/java/ui/components/NavigableListView.java
```diff
@@ -188,21 +188,17 @@ public void setOnItemSelected(IntConsumer callback) {
...
+     */
 
-    public void selectLastItem() {
+    public void selectNthItem(int n) {
         requestFocus();
-        if (getItems().size() == 0) return;
-        getSelectionModel().clearAndSelect(getItems().size() - 1);
-        scrollAndShow(getItems().size() - 1);
-        selectedIndex = Optional.of(getItems().size() - 1);
+        assert(getItems().size() >= n);
```
dariusf: What I meant was `assert n > 0 && n <= size`. It's good to put `n` on the left because we read from left to right, and `n` is the variable whose range we are constraining. Also `assert` isn't a keyword so we don't need to parenthesise.

PR #971, opened 2015-09-15
Adding labels rapidly interferes with undo #964

### LINKED ISSUE #964: Adding labels rapidly interferes with undo
e.g. 
1. add label1 to issue 1
2. before the undo window is over, add label2 to issue 2
expected: label 1 is applied to issue 1 and label2 is applied to issue 2
actual: HT shows label 1 applied to issue 1, but it is not updated in the server

lejolly:
Fixes #964


src/main/java/ui/NotificationController.java
```diff
@@ -34,6 +32,12 @@ void showNotification(Notification notification) {
                     }));
             notificationPaneTimer = new TickingTimer("Notification Timer", notification.getTimeoutDuration(),
                     integer -> {}, () -> Platform.runLater(this::triggerTimeoutAction), TimeUnit.SECONDS);
+            notificationPane.setOnHiding(event -> Platform.runLater(() -> {
```
dariusf: The hiding event seems to always originate from the UI thread, so the Platform.runLater isn't needed.

Can we factor the callback out into its method? It looks kind of similar to triggerTimeoutAction so maybe it could be merged.

damithc:
How about adding a test case to cover this behavior.

damithc:
This PR doesn't have any status label.

lejolly:
added test, ready for review.

dariusf:
No problems except for the sleeps in tests. Are there better ways to synchronise? For example if there is `sleep`, then `find`, we can use `waitUntilNodeAppears` instead.

In the general case (`resetDummyRepo`), maybe we can pass the event a CountdownLatch/Semaphore and block on that? Something like this (untested):

```java
void resetDummyRepo() {
    UILogicRefreshEvent event = new UILogicRefreshEvent();
    UI.events.triggerEvent(event);
    event.latch.await();
    // At this point, the refresh is done
}

// UI.java

protected void registerTestEvents() {
    registerEvent((UILogicRefreshEventHandler) e ->
        Platform.runLater(() ->
            logic.refresh().thenRun(() -> e.latch.countDown())));
}

// Logic.java

public CompletableFuture<Void> refresh() {
    // ...
    return Futures.sequence(models.toModels().stream()
        // ...
        .exceptionally(Futures::log)
        .thenRun(() -> {});
}
```

lejolly:
can't use `waitUntilNodeAppears` because the node is always present, because what i'm testing for is if the content of the node changes.

lejolly:
i used a `CountDownLatch` in `UILogicRefreshEvent` but it wasn't late enough in the update chain, still needed a `sleep(100)` just to run reliably locally. is that good enough or should i find someplace lower, the trouble is we don't call the other events within the test, so we'll have to pass the `CountDownLatch` around which will make it messier.

dariusf:
The change that you're detecting is issues appearing in panels, right? Could we wait for that (maybe the appearance of some issue card, or taking note of the content before and waiting until it changes)?

Sleeping is unreliable because it's effectively unsynchronised: there's no ordering between events. We can only specify a constant amount of time to wait, while the actual event could always take longer.

lejolly:
ok, made it compare `updateAt` times to check for a repo reset.

dariusf:
Looks good.

Issue #664, opened 2015-06-01
Only the last 30 issue events per issue are retrieved


dariusf:
The API likely has some paging functinality for issue events which we are not using

damithc:
How does this affect functionality?

dariusf:
I encountered this today. Events *are* paged, and as #938 showed, ETags differ across pages.

Since we store ETags for events now and only for the first page, issues won't receive any event updates after their 30th event (the default page size).

This issue should probably be addressed soon.

PR #980, opened 2015-09-16
'updated' filter only shows comments #976

### LINKED ISSUE #976: 'updated' filter only shows comments
image

image

acruis:
Fixes #976


damithc:
Any chance of test cases to cover this?

acruis:
Good idea, I'll put some tests in to specify the `updated` filter's general behavior.

dariusf:
Nice work!

On a side note, I encountered #664 when testing. Looks like events are paged, and we'll eventually need to roll all the ETags into one like in #938. The ETag we're storing here is for the first page only.

Issue #950, opened 2015-09-09
Create TestController


lejolly:
Mostly to allow calling of the `UI` instance during tests. 

Should help with #933.

PR #982, opened 2015-09-16
Focus goes to wrong filter after start up #969

### LINKED ISSUE #969: Focus goes to wrong filter after start up
Actual: the first panel is selected, but the cursor is in the filter box of the last panel
Expected: the first panel is selected, cursor is not in any filter box. **At this point user should be able to use F/D to navigate panels and Ctrl+down to enter a panel**

HansNewbie:
Fixes #969

### DIFF
--- a/src/main/java/ui/issuepanel/FilterPanel.java
+++ b/src/main/java/ui/issuepanel/FilterPanel.java
@@ -60,10 +60,7 @@ public abstract class FilterPanel extends AbstractPanel {
         this.ui = ui;
         
         getChildren().addAll(createPanelMenuBar(), createFilterBox());
-        this.setOnMouseClicked(e-> {
-            ui.triggerEvent(new PanelClickedEvent(panelIndex));
-            requestFocus();
-        });
+        setUpEventHandler();
         focusedProperty().addListener((unused, wasFocused, isFocused) -> {
             if (isFocused) {
                 getStyleClass().add("panel-focused");
@@ -72,6 +69,13 @@ public abstract class FilterPanel extends AbstractPanel {
             }
         });
         
+    }
+
+    private void setUpEventHandler() {
+        this.setOnMouseClicked(e-> {
+            ui.triggerEvent(new PanelClickedEvent(this.panelIndex));
+            requestFocus();
+        });
     }
 
     private final ModelUpdatedEventHandler onModelUpdate = e -> {
@@ -290,5 +294,4 @@ public abstract class FilterPanel extends AbstractPanel {
     public void close() {
         ui.unregisterEvent(onModelUpdate);
     }
-
 }
--- a/src/main/java/ui/issuepanel/PanelControl.java
+++ b/src/main/java/ui/issuepanel/PanelControl.java
@@ -5,6 +5,7 @@ import filter.expression.FilterExpression;
 import filter.expression.Qualifier;
 import javafx.geometry.Insets;
 import javafx.scene.Node;
+import javafx.scene.control.ScrollPane;
 import javafx.scene.input.KeyEvent;
 import javafx.scene.layout.HBox;
 import prefs.Preferences;
@@ -29,6 +30,7 @@ public class PanelControl extends HBox {
 
     private final UI ui;
     private final Preferences prefs;
+    private ScrollPane panelsScrollPane;
     private IModel model;
     private GUIController guiController;
     private Optional<Integer> currentlySelectedPanel = Optional.empty();
@@ -53,8 +55,9 @@ public class PanelControl extends HBox {
     /**
      * Called on login.
      */
-    public void init(GUIController guiController) {
+    public void init(GUIController guiController, ScrollPane panelsScrollPane) {
         this.guiController = guiController;
+        this.panelsScrollPane = panelsScrollPane;
         restorePanels();
         selectFirstPanel();
     }
@@ -119,6 +122,8 @@ public class PanelControl extends HBox {
 
     public void selectFirstPanel() {
         setCurrentlySelectedPanel(Optional.of(0));
+        scrollToPanel(0);
+        getPanel(0).requestFocus();
     }
 
     private void setCurrentlySelectedPanel(Optional<Integer> selectedPanel) {
@@ -166,11 +171,11 @@ public class PanelControl extends HBox {
     }
 
     public void createNewPanelAtStart() {
-        addPanelAt(0);
+        addPanelAt(0).filterTextField.requestFocus();
     }
 
     public void createNewPanelAtEnd() {
-        addPanel();
+        addPanel().filterTextField.requestFocus();
     }
 
     public void swapPanels(int panelIndex, int panelIndex2) {
@@ -263,11 +268,11 @@ public class PanelControl extends HBox {
             }
         }
         ui.triggerEvent(new PanelClickedEvent(currentlySelectedPanel.get()));
-        scrollandShowPanel(currentlySelectedPanel.get(), getChildren().size());
+        scrollToCurrentlySelectedPanel();
     }
 
-    private void scrollandShowPanel(int selectedPanelIndex, int numOfPanels) {
-        ui.getMenuControl().scrollTo(selectedPanelIndex, numOfPanels);
+    public void scrollToCurrentlySelectedPanel() {
+        scrollToPanel(currentlySelectedPanel.get());
     }
 
     public GUIController getGUIController() {
@@ -296,5 +301,13 @@ public class PanelControl extends HBox {
 
     public int getNumberOfSavedBoards() {
         return prefs.getAllBoards().size();
+    }
+
+    public void scrollToPanel(int panelIndex){
+        setHvalue(panelIndex * (panelsScrollPane.getHmax()) / (getNumberOfPanels() - 1));
+    }
+
+    private void setHvalue(double val) {
+        panelsScrollPane.setHvalue(val);
     }
 }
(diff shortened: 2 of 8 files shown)


HansNewbie:
Will create the test(s)

HansNewbie:
Current problem: must press `Ctrl` + :arrow_up: before going down to issues. Will see what's wrong.

damithc:
Can add test cases to confirm this behavior?

HansNewbie:
@damithc yes I will create test(s) to confirm the behavior.

src/test/java/guitests/MenuControlTest.java
```diff
@@ -117,6 +117,9 @@ public void menuControlTest() {
 
         // Testing First Panel selected
         assertEquals(0, (int) panelControl.getCurrentlySelectedPanel().get());
+        press(KeyCode.CONTROL).press(KeyCode.DIGIT1).release(KeyCode.DIGIT1).release(KeyCode.CONTROL);
+        assertEquals(0, (int) panelControl.getCurrentlySelectedPanel().get());
+        // TODO (currently not possible) test that panelsScrollPane.getHValue() is 0
```
HansNewbie: @dariusf I added this as todo because when I implemented the function, the panel that is focused was not in view, i.e. the pane doesn't scroll to the one in focus.
However, there is no current functionality to check the caret/pane position right now unless I make a public method to do so, which breaks encapsulation. Is the comment sufficient?

dariusf: A public method in ScrollableListView wouldn't be breaking encapsulation. Is that what you're referring to?

Alternatively, what if we assert that the first issue of that panel is selected? After ensuring that it isn't selected before pressing Ctrl + 1.

HansNewbie: Not ScrollableListView, but on `MenuControl`>`panelsScrollPane`. It's a JavaFX ScrollPane object. If I want to see its pane position, I need to create a public method in `MenuControl` to see the scroll position, but that method's only purpose is for my test and nothing else, which I think is not a good design.

I will see if I can do the first issue selected test. The purpose would be different from the TODO I wrote, though. The TODO is because the focused panel was outside of view (say the focused panel is the first panel but the view is stuck on last panel, hence user is doing things without looking at the panel he is working on). I thought a test on this would be good, but I need to access the ScrollPane value.

dariusf: Okay, I see. Agreed that the test is good to have. The ScrollPane is passed to MenuControl, but it doesn't originate there -- the UI actually manages it. If you can get a UI reference, you could add some way of accessing it.

(The MenuControl probably doesn't need to have a direct reference to the ScrollPane... it should communicate with it either via the UI reference or event.)

On a side note, this test might be named wrongly. It should be PanelControlTest?

HansNewbie: ScrollPane originates from UI, but it doesn't have a variable that stores it. It stores one that is the grandparent of the ScrollPane, hence accessing it from UI will be quite convoluted. So in this case, MenuControl actually have direct reference to ScrollPane (under the variable `panelsScrollPane`.

This test is added to MenuControlTest where they do panel menu testing and board menu testing. I added it to OpenBoard testing. The test for panel focus is on another file (PanelFocusTest). This is because I test the panel focus for 2 cases: on startup and on opening a board.

dariusf: I know, but I think it is better in the long run that we have a reference to the ScrollPane in UI/PanelControl, and eventually remove the one in MenuControl. The former is really just an implementation detail of the panels, and it isn't conceptually part of the menu.

The name seems weird to me because MenuControl is just a subclass of JavaFX's MenuBar... it can't really have any meaningful tests without going into the functionality of its items.

HansNewbie: Alright, I will add a reference for the ScrollPane in UI or PanelControl.

The MenuControlTest tests the items in menu control, like Add Panel, Save Board, Open Board, Remove Repo. You want to rename it to MenuControlItemTest?

HansNewbie:
I may need more time. There is a funny bug where 2 panels can be focused when creating new panel (and this exists in the main branch as well). Will see if I can solve this at the same time.

Steps to reproduce:
1. Have about 3 or 4 panels open.
2. Click on the 2nd or 3rd panel.
3. Create a new panel (whether at the end or at the left is fine)
4. Click all the panels one by one and observe that sometimes there are multiple panels selected.

HansNewbie:
Found out why. Apparently panel index update method is not working. This bug can be easily observed upon creating panel on the left and upon closing a panel. Will try to solve it.

HansNewbie:
@dariusf Here are some things that I have done:
- make `panelsScrollPane` in `UI` and move all scrolling to UI
  - `panelsScrollPane` still exists in MenuControl as we need a listener to it on adding a panel at the end
- make more thorough panel focus test on startup
  - I still can't check the scrolling bar position in  `MenuControlTest` on opening board as I can't access the ui instance
  - As such, I didn't add the public method to get scrolling bar position as it will only be used in my test.

One big question here is that I may not be able to solve the bug where two panels are being "selected". Should I create a separate issue for that and work on it separately?

HansNewbie:
Just additional comment:
I managed to solve one part of the bug, that is two panels being coloured blue. This is due to calling requestFocus in eventHandler of panelOnClick.
The problem right now is that 'PanelControl' `updatePanelIndices` does not work, as in panel index is immutable.

HansNewbie:
Ok, found out more about the bug:
so the `panelIndex` is mutable, but the event triggered will use its original `panelIndex` instead of its current `panelIndex`. Will look into it more.

HansNewbie:
@dariusf I managed to solve the bug. Any other comments on my commits?

HansNewbie:
@damithc with this commit, filter box is no longer in focus on startup. However, this is also the case when we create a new panel (the focus will be on the panel). Is that ok?

HansNewbie:
@dariusf  Just found a bug. Will look at it more.

damithc:
At startup or change of board, the focus should be in the panel rather than filter box.
However, when creating a new panel the focus should go to filter box because that's the most likely place the user is going to go next anyway.

HansNewbie:
Noted. Will make the focus on the filter box upon creating a new panel.

src/main/java/ui/issuepanel/PanelControl.java
```diff
@@ -267,7 +269,7 @@ private void handleKeys(boolean isForwardKey) {
     }
 
     private void scrollandShowPanel(int selectedPanelIndex, int numOfPanels) {
-        ui.getMenuControl().scrollTo(selectedPanelIndex, numOfPanels);
+        ui.scrollTo(selectedPanelIndex, numOfPanels);
```
dariusf: Could you help to fix the name of this method?

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -61,7 +61,7 @@ public FilterPanel(UI ui, IModel model, PanelControl parentPanelControl, int pan
         
         getChildren().addAll(createPanelMenuBar(), createFilterBox());
         this.setOnMouseClicked(e-> {
-            ui.triggerEvent(new PanelClickedEvent(panelIndex));
+            ui.triggerEvent(new PanelClickedEvent(this.panelIndex));
```
dariusf: Nice catch! To prevent this from happening again we should factor the event handler out into its own method. That way it won't have a lexical reference to the local `panelIndex`.

dariusf:
Hmm, seeing as we need the maximum number of panels every time we call `ui.scrollTo`, it should be in PanelControl. Wrong decision for me to have suggested that it be in UI, sorry. It's a good abstraction though.

We should be able to access the UI instance now, since we have #950?

Also, if you don't want to add a getter, maybe you could use reflection to get hold of the ScrollPane instance (FieldUtils.readField), then query it directly?

damithc:
BTW, lookout for conflicts. Rajan is changing the keyboard shortcut for entering the issue list to `Ctrl+enter`

src/main/java/ui/issuepanel/PanelControl.java
HansNewbie: @dariusf from `ui.scrollTo` to `ui.scrollAndShowPanel`? Also, taking note of below (moving from ui to panelControl) so I will rename and move it to panel control.

HansNewbie: As I moved it to `PanelControl`, I combine `scrollTo` with `scrollAndShowPanel` as `scrollToPanel` and I have changed all the parameters so that now only 1 parameter, `panelIndex` since we always use number of panels as the second parameter.

dariusf: What I meant was the non-camel-cased name, but yup. :stuck_out_tongue: Those methods do the same thing.

HansNewbie:
What I just changed:
- Create a method for event handler for `FilterPanel`
- Move `panelsScrollPane` methods to `PanelControl`, also clearing up the methods pertaining to scrolling.
- Create reflection to test scrollbar set at the left most upon opening a board
  - For this, I left a basic focus test on `MenuControl` upon opening a board just to be sure that the focus is correct
  - I moved the test for scroll bar to my `PanelFocusTest` and I am using reflection

I will lookout for conflicts when merging.

dariusf:
Nice work! Just have to update to latest master and get CI to pass.

damithc:
I think #1000 is related to this issue.

PR #983, opened 2015-09-16
Nothing shown in the panel on mac #978

### LINKED ISSUE #978: Nothing shown in the panel on mac
This is what my panel looks like

This is my java version.

dariusf:
Fixes #978.

Coverage drop seems like a false positive.

### DIFF
--- a/src/main/java/util/Utility.java
+++ b/src/main/java/util/Utility.java
@@ -17,10 +17,7 @@ import java.text.SimpleDateFormat;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
-import java.util.Date;
-import java.util.List;
-import java.util.Optional;
-import java.util.TimeZone;
+import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


src/test/java/tests/UtilityTest.java
```diff
@@ -55,13 +55,26 @@ public void stripQuotesTest() {
...
         testDate.set(1994, Calendar.NOVEMBER, 15, 12, 45, 26);
         testDate.setTimeZone(TimeZone.getTimeZone("GMT"));
 
         assertEquals(testDate.getTime().toString(),
-                parseHTTPLastModifiedDate("Tue, 15 Nov 1994 12:45:26 GMT").toString());
+            parseHTTPLastModifiedDate("Tue, 15 Nov 1994 12:45:26 GMT").toString());
+
+        try {
+            assertEquals(testDate.getTime().toString(),
+                parseHTTPLastModifiedDate("some non-date string").toString());
```
lejolly: include some chinese? =P

src/main/java/util/Utility.java
```diff
@@ -190,17 +187,16 @@ public static int safeLongToInt(long l) {
     public static Date parseHTTPLastModifiedDate(String dateString) {
         assert dateString != null;
         try {
-            return new SimpleDateFormat("EEE, d MMM yyyy HH:mm:ss Z").parse(dateString);
+            return new SimpleDateFormat("EEE, d MMM yyyy HH:mm:ss Z", Locale.US).parse(dateString);
         } catch (ParseException e) {
-            assert false : "Error in date format string!";
+            logger.error(e.getLocalizedMessage(), e);
+            throw new IllegalArgumentException("Could not parse date " + dateString);
```
HansNewbie: This is not caught in `UpdateService.UpdateCheckTime()`.

dariusf: It can't be handled there (without defaulting to some date, which I'm not sure we want). Instead we log the exception, then let it propagate back up to RepoIO, where it will be handled [here](https://github.com/HubTurbo/HubTurbo/blob/master/src/main/java/backend/RepoIO.java#L160).

src/test/java/tests/UtilityTest.java
dariusf: The input date is always given in English -- the problem was SimpleDateFormat trying to interpret the English input as Chinese, due to the default locale. :stuck_out_tongue: I think any input which causes a parse failure should work.

src/main/java/util/Utility.java
HansNewbie: That's true. This sounds good, then.

HansNewbie:
@dariusf I am not sure if this should be in this issue/pr, but would it be better to make the time in logger in English, too? Right now the month is still in Chinese.

dariusf:
The simplest solution I can think of for that is to render months as numbers instead of some locale-dependent term. The latter seems to a bit more involved.

PR #990, opened 2015-09-17
updated-others and updated-self not working #968

### LINKED ISSUE #968: updated-others and updated-self not working
The filters `updated-others` and `updated-self` do not show issues after filtering.

codemanmeet:
Fixes #968.

### DIFF
--- a/src/test/java/guitests/UpdateIssuesTest.java
+++ b/src/test/java/guitests/UpdateIssuesTest.java
@@ -1,6 +1,7 @@
 package guitests;
 
 import javafx.scene.control.Label;
+import javafx.scene.control.TextField;
 import javafx.scene.input.KeyCode;
 import org.junit.Test;
 import ui.UI;
@@ -20,8 +21,9 @@ public class UpdateIssuesTest extends UITest {
     public void updateIssues() throws InterruptedException, ExecutionException {
         Label apiBox = find("#apiBox");
         resetRepo();
+        TextField filterField = find("#dummy/dummy_col0_filterTextField");
 
-        click("#dummy/dummy_col0_filterTextField");
+        click(filterField);
         type("updated");
         press(KeyCode.SHIFT).press(KeyCode.SEMICOLON).release(KeyCode.SEMICOLON).release(KeyCode.SHIFT);
         type("24");
@@ -33,7 +35,7 @@ public class UpdateIssuesTest extends UITest {
 
         // After updating, issue with ID 5 should have title Issue 5.1
         updateIssue(5, "Issue 5.1");
-        click("#dummy/dummy_col0_filterTextField");
+        click(filterField);
         push(KeyCode.ENTER);
         PlatformEx.waitOnFxThread();
 
@@ -44,7 +46,7 @@ public class UpdateIssuesTest extends UITest {
         // Then have a non-self comment for Issue 9.
         UI.events.triggerEvent(UpdateDummyRepoEvent.addComment("dummy/dummy", 9, "Test comment", "test-nonself"));
         UI.events.triggerEvent(new UILogicRefreshEvent());
-        click("#dummy/dummy_col0_filterTextField");
+        click(filterField);
         push(KeyCode.ENTER);
         PlatformEx.waitOnFxThread();
         assertEquals(3492, getApiCount(apiBox.getText())); // 2 calls for Issue 9
@@ -54,6 +56,27 @@ public class UpdateIssuesTest extends UITest {
         push(KeyCode.ENTER);
         PlatformEx.waitOnFxThread();
         assertEquals(3492, getApiCount(apiBox.getText())); // No change to issues, so no additional API quota spent
+
+        updateIssue(8, "Issue 8.1");
+        filterField.selectAll();
+        push(KeyCode.BACK_SPACE);
+        type("updated-others");
+        press(KeyCode.SHIFT).press(KeyCode.SEMICOLON).release(KeyCode.SEMICOLON).release(KeyCode.SHIFT);
+        type("24");
+        push(KeyCode.ENTER);
+        PlatformEx.waitOnFxThread();
+
+        assertEquals(4, countIssuesShown());
+
+        filterField.selectAll();
+        push(KeyCode.BACK_SPACE);
+        type("updated-self");
+        press(KeyCode.SHIFT).press(KeyCode.SEMICOLON).release(KeyCode.SEMICOLON).release(KeyCode.SHIFT);
+        type("24");
+        push(KeyCode.ENTER);
+        PlatformEx.waitOnFxThread();
+
+        assertEquals(1, countIssuesShown());
     }
 
     public void resetRepo() {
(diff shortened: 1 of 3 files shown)


damithc:
No tests?

dariusf:
You can add an integration test in UpdateIssuesTest, and maybe create QualifierTests and just assert that the method behaves as it should in there.

codemanmeet:
I added GUI tests for updated-others and updated-self.

src/test/java/guitests/UpdateIssuesTest.java
```diff
@@ -44,16 +46,50 @@ public void updateIssues() throws InterruptedException, ExecutionException {
         // Then have a non-self comment for Issue 9.
         UI.events.triggerEvent(UpdateDummyRepoEvent.addComment("dummy/dummy", 9, "Test comment", "test-nonself"));
         UI.events.triggerEvent(new UILogicRefreshEvent());
-        click("#dummy/dummy_col0_filterTextField");
+        click(filterField);
         push(KeyCode.ENTER);
         PlatformEx.waitOnFxThread();
-        assertEquals(3492, getApiCount(apiBox.getText())); // 2 calls for Issue 9
```
acruis: Why remove the `// 2 calls for Issue 9` comment?

codemanmeet: removed it by mistake. added back in

dariusf:
I checked out the branch and tested this. Also had a long discussion with @acruis.

The `updated` qualifier actually has 4 effects:

1. Filters issues by updated time
2. Sorts issues by non-self update time
3. Triggers a metadata download
4. Changes the appearance of `ListPanelCard`

This isn't documented (and it should be). `updated-self` and `updated-others` should behave as similarly as possible.

Currently `updated-self` and `updated-others` only support 1 (filtering by the appropriate issue field), and possibly 2 as well, but that's difficult to verify right now.

The problem this PR addresses is actually two-fold:

- The new qualifiers don't trigger metadata downloads
- The new qualifiers don't change the appearance of `ListPanelCard`

Neither is actually fixed. The first one doesn't happen despite the changes here because the new qualifiers are [not registered as meta-qualifiers](https://github.com/HubTurbo/HubTurbo/blob/968-updated-others-fix/src/main/java/filter/expression/Qualifier.java#L412), therefore their presence doesn't trigger the metadata download. The second [has not been done](https://github.com/HubTurbo/HubTurbo/blob/968-updated-others-fix/src/main/java/ui/listpanel/ListPanelCard.java#L88).

These can only be tested via integration tests, which need to be strengthened (clearly). In particular we have to verify that *the qualifiers actually trigger the downloads*, and that the UI reflects this.

We also have to verify that 2 is supported. This can be unit-tested.

Finally, the new qualifiers have a quirky semantics: issues need to have metadata to pass their filter. As metadata updates are only triggered for issues which pass a filter, `updated-self` and `updated-others` won't work on the first pass; issues must have passed an `updated` filter before they will work with the new qualifiers. *This should be taken into account when writing integration tests.*

damithc:
item 4 can be done in a separate issue.
Do we still need 2? We already have a separate sort filter now.

dariusf:
We can get rid of 2 and move it into `sort`, but there isn't a criterion specifically for non-self update time yet, so we'll have to implement it.

acruis:
Are you, by any chance, referring to `sort:nonSelfUpdate`?

dariusf:
Right, forgot about that.

codemanmeet:
Yup, we should separate out issues 2 and 4. We don't want pull requests to be massive. I have changed the meta qualifier list to include `updated-self` and `updated-others`. The integration tests in UpdateIssues [GUI test](https://github.com/HubTurbo/HubTurbo/blob/968-updated-others-fix/src/test/java/guitests/UpdateIssuesTest.java#L72) do test metadata by adding comments which are part of the issue metadata. The only way to check whether *real* metadata is obtained would be through making actual HTTP requests which requires another system altogether.

@dariusf, I changed the [meta qualifier list](https://github.com/HubTurbo/HubTurbo/blob/968-updated-others-fix/src/test/java/guitests/UpdateIssuesTest.java#L72) but I'm still facing a problem where the `updated-others` and `updated-self` only update after the `updated` filter has been run first. I tried to dig in to find the problem and delve into various functions in [GUIController.java](https://github.com/HubTurbo/HubTurbo/blob/968-updated-others-fix/src/main/java/ui/GUIController.java) and [Logic.java](https://github.com/HubTurbo/HubTurbo/blob/968-updated-others-fix/src/main/java/backend/Logic.java) but the code seems alright to me. @acruis and @dariusf, I need your help in helping me find out where the problem is because as seen in `GUIController.java`, if the filter expression contains any meta-qualifier it should download metadata. I don't know why the metadata doesn't get immediately downloaded upon entering `updated-others` and `updated-self`.

dariusf:
I have already explained many times why the tests are insufficient. Just because metadata is involved doesn't mean they assert the behaviour under test. HTTP requests are not necessary to ensure this.

The second point is addressed in the last paragraph of my long post.

codemanmeet:
I don't understand the quirky semantics part. The meta qualifier list has been updated. Why aren't `updated-others` and `updated-self` pulling in data?

codemanmeet:
Actually, my tests for `updated-others` in the integration tests don't pass if `updated` is not included first which is a good indicator of the problem.

acruis:
@codemanmeet The point about qualifiers actually triggering the downloads doesn't mean you have to download metadata from the actual server. You just have to give an indication that you actually received metadata from the backend, which is where it ties in to 4) which demands you to reflect the presence of metadata through the UI. But counting issues, like what you're doing now, works since you're not implementing 4) here anyway.

However, `updateIssuesTest` is not downloading metadata with `updated-others` and `updated-self`. It's doing it through `updated`, and then filtering with `updated-others` and `updated-self` afterwards. Clever, but we want `updated-others` and `updated-self` to work without `updated`. That's @dariusf's point when he demands tests that test whether the qualifiers actually download metadata.

Do read the flow of execution in `GUIController` more carefully. To answer your problem, think about the scenario when the program does not have metadata and the user enters a filter string with `updated-others`: We don't download _all_ issues' metadata with the `updated` filter, so where in the flow of execution do we decide which issues to download metadata for, and what is the problem there?

codemanmeet:
Yup, I thought of the scenario and read `GUIController.java`. The `processPanel` has a boolean value [hasUpdatedQualifier](https://github.com/HubTurbo/HubTurbo/blob/968-updated-others-fix/src/main/java/ui/GUIController.java#L150) which checks if any meta qualifier (I have added `updated-others` and `updated-self` to this list) has been updated. If it has it populates the update list. Once this has been processed, `modelUpdated` calls `processPanel` and subsequently also calls `dispatchMetadataRequest` (which does the actual job getting metadata). I don't know what else I missed when it comes to including `updated-self` and `updated-others` to lists that trigger metadata download. Do you guys have any idea?

acruis:
What does `populateUpdateList` do? Is that method where the program _decides_ which issues to dispatch metadata for, or is that done somewhere else?

Do you have problems understanding the lambda in `populateUpdateList`?

PR #991, opened 2015-09-18
Support ReviewComment and CommitComment #970

### LINKED ISSUE #970: Support ReviewComment and CommitComment in GithubClientEx
Subtask of #654 

*egit* does not support Review Comments, so we need to interact directly with Github API for this.

ndt93:
Fixes #970

The support is implemented in a new class `PullRequestServiceEx` (together with a new `ReviewComment` class) instead of `GitHubClientEx` for consistency with `egit`. All classes that extend `egit` are also renamed with `Ex` suffice. This PR should not contain any apparent functional change.

### DIFF
--- a/src/main/java/github/PullRequestServiceEx.java
+++ b/src/main/java/github/PullRequestServiceEx.java
@@ -1,9 +1,11 @@
 package github;
 
 import com.google.gson.reflect.TypeToken;
+import org.apache.logging.log4j.Logger;
 import org.eclipse.egit.github.core.IRepositoryIdProvider;
 import org.eclipse.egit.github.core.client.PagedRequest;
 import org.eclipse.egit.github.core.service.PullRequestService;
+import util.HTLog;
 
 import java.io.IOException;
 import java.util.List;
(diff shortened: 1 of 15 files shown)


ndt93:

src/main/java/github/PullRequestServiceEx.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Gets a pull request's review comments
+     *
+     * @param repository
+     * @param pullRequestNumber
+     * @return list of review comments
+     * @throws IOException
+     */
+    public List<ReviewComment> getReviewComments(IRepositoryIdProvider repository,
```
dariusf: Is this method needed? It doesn't ensure the PR number is an int. Maybe we should have the int constructor as the only interface for this functionality. This will allow us to skip some checks below too (only need one for `> 0`).

ndt93: I was mirroring the structure of `egit`'s `getComments` interface in their service classes. The method with `int` is the commonly used one. The one with string is a fallback in case that some time in the future issue's id is not a number anymore (which is weird but still) since GitHub never guarantees that issue's id is a number.

It's the same case with page number. Although we are using numbers for getting pages from GitHub API, in rare cases, the page's number can be a hash string.

dariusf: I see. Okay then, looks good!

PR #925, opened 2015-09-02
Ensure that globalConfigTest passes #920

dariusf:
Fixes #920.

This needs opinions on the names of utility methods and where to put them. This is just a start; eventually we can use these new APIs in other tests.

Issue #900, opened 2015-08-25
Make use of KeyboardShortcuts constants


dariusf:
See discussion in #865. We should use the constants defined in `KeyboardShortcuts` as much as possible. Even in tests, if there is a way to `push` a `KeyCombination`.

HansNewbie:
I am thinking it would be cool if we can link the `KeyboardShortcuts` to automatically update the `keyboardShortcuts.md` docs, so that whenever we update the shortcut, the docs will be updated accordingly.

dariusf:
Meaning, generate docs from source code?

HansNewbie:
@dariusf only for the keyboard shortcut page. So that whenever we change anything in `KeyboardShortcuts` constant, we will not forget to update the keyboard shortcut doc page.

I am thinking where we create a custom parser for the MD only for keyboard page. Basically, we will use the constants in `KeyboardShortcuts` to be written in a template keyboard shortcut MD (which is not put in the folder docs). The parser will replace the constants with the relevant `KeyCode` button upon pushing and the generated MD is put in the docs folder.

dariusf:
Feasible, sure, but it shouldn't be done in an ad-hoc way, i.e. it should be a part of the build system, developers shouldn't be able to forget to build it or that the output in `docs` should not be modified by hand, etc.

PR #992, opened 2015-09-18
Nondeterministic tests #952

dariusf:
Addresses #952 and #900. General fixes for test nondeterminism.

- Sometimes key press commands in tests get delayed, causing hundreds of key press events to be registered (happened in #925, detailed in TestFx/TestFx#43). This adds a few methods for pressing keys (and `KeyCodeCombination`s!) without delays, which seems to fix this problem. Ran `UseGlobalConfigsTest` 20 times and it failed only once (JVM crash). Method names need feedback.
- Removes `--continue` flag for CI tests, so they fail as soon as one fails
- Various test fixes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ env:
     - TERM=dumb
 
 script:
-    - "./gradlew clean check -i --continue"
+    - "./gradlew clean check -i"
 
 after_success:
     - "./gradlew jacocoRootReport coveralls"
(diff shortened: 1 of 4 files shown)


lejolly:
Looks good, would going to testfx 4 help matters?

dariusf:
Eventually, yes, but 4 seems to have changed APIs quite a bit, and is still in alpha now (so APIs will continue to change). It's not a drop-in replacement.

lejolly:
ah ok 👌

PR #993, opened 2015-09-19
updated-others and updated-self not working #968

dariusf:
Fixes #968. Approach outlined [here](https://github.com/HubTurbo/HubTurbo/pull/990#issuecomment-141165501).

PR #999, opened 2015-09-21
updated-others missing issues #997

### LINKED ISSUE #997: updated-others missing issues
image

dariusf:
Fixes #997. Disables the new qualifiers.

I'm reverting #924 and #993 here, but keeping the useful parts. This should bring us back to before both were implemented. We can then focus on redesigning how `updated` is done, then selectively revert parts of *this* PR and put the functionality back in.

Also taking the chance to redesign IssueMetadata while things aren't so messy.

### DIFF
--- a/src/main/java/backend/IssueMetadata.java
+++ b/src/main/java/backend/IssueMetadata.java
@@ -24,144 +24,147 @@ public class IssueMetadata {
     // Properties computed from events and comments on instantiation, so we
     // don't have to recompute on querying.
 
+    private final String user;
     private final LocalDateTime nonSelfUpdatedAt;
     private final int nonSelfCommentCount;
 
     private final String eventsETag; // Only modified in the DownloadMetadataTask constructor
     private final String commentsETag;
 
-    // Constructor for default use when initializing TurboIssue
-    public IssueMetadata() {
-        events = new ArrayList<>();
-        comments = new ArrayList<>();
-
-        nonSelfUpdatedAt = LocalDateTime.ofEpochSecond(0, 0, ZoneOffset.ofHours(0));
-        nonSelfCommentCount = 0;
-
-        eventsETag = "";
-        commentsETag = "";
-        isLatest = false;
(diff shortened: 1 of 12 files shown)


src/main/java/backend/IssueMetadata.java
```diff
@@ -180,32 +157,11 @@ public IssueMetadata(IssueMetadata other, LocalDateTime nonSelfUpdatedAt,
...
 
         this.nonSelfUpdatedAt = nonSelfUpdatedAt; // After creation date reconciliation
-        this.selfUpdatedAt = other.selfUpdatedAt;
         this.nonSelfCommentCount  = other.nonSelfCommentCount;
-        this.selfCommentCount = other.selfCommentCount;
-        this.isUpdatedBySelf = other.isUpdatedBySelf;
-        this.isUpdatedByOthers = other.isUpdatedByOthers;
 
-        this.eventsETag = other.eventsETag;
+        this.eventsETag = currEventsETag;
```
dariusf: @acruis the ETag never changes here. Is this a bug?

acruis: The ETag never changes after or before this diff was applied?

dariusf: Before. I'm wondering if this fix is correct.

dariusf: Never mind, it's not.

dariusf:
The isUpdated flag is back, but it isn't currently used for anything because the `updated` qualifier now has very simple semantics (no more filtering by non-self update time, since that should be done by a separate qualifier). I'm not sure if we'll still need it in future so I left it in.

acruis:
Thanks for putting in the factory methods, I was thinking of doing that too. Anything else to add to this PR, btw?

dariusf:
Nothing for now, unless you have suggestions.

acruis:
OK, good to merge.

PR #1003, opened 2015-09-23
Add a windows script to update to latest master #973

### LINKED ISSUE #973: Add a windows script to update to latest master
This for devs to use (at least I need this, because I update my HT whenever there is an update to master).

0. pull latest master
1. shut down the running instance of HT
2. build without tests (e.g. `gradlew build -x test`)
3. launch newly-build HT

acruis:
Fixes #973

The script is `rbar.bat`.


damithc:
Why is it named rbar.bat? Update developer guide regarding this?

damithc:
OK, I read the commit message that explains the name.
The script itself didn't work. For one thing, the build failed. For another, when it launched HT, it asked me for login/password although I have previously logged in.
Here's the output.
Closing all java and javaw processes with window title format: HubTurbo V3.7.1*.
SUCCESS: Sent termination signal to the process "javaw.exe" with PID 16140.

INFO: No tasks running with the specified criteria.

Starting Gradle build. This will take a while...

To honour the JVM settings for this build a new JVM will be forked. Please consider using the daemon: http://gradle.org/docs/2.4/userguide/gradle_daemon.html.
:clean FAILED

BUILD FAILED

Total time: 5.397 secs

Starting HubTurbo V3.7.1. You can close this cmd window once HubTurbo begins launching.

acruis:
Did you have `IntelliJ` open at the time? `clean` doesn't fail unless it's trying to delete some files that are currently open.

damithc:
I tried again in my other comp and it worked.
And it wiped out all my settings. That's definitely a 'no no'. Now I have to set up all my filters/boards again :-(
There is no need to clean.

damithc:
When it failed in my other comp, I might have had eclipse open, I'm not sure. I'll try again later after you've removed the clean action.

acruis:
I'll still run `clean` (see [this thread](http://stackoverflow.com/questions/29028748/why-run-gradle-clean) for why `clean` is necessary, we run `clean` when releasing as well), and the problem of settings and cache being wiped out can be solved by simply backing up the data in a different folder before running `clean`.

damithc:
Sure, as long as I don't have to set up everything everytime I upgrade.

damithc:
Observations after running a few times: 
At times the `clean` task fails because it can't delete `chromedriver_2-18.exe` because it is still in use. 
Also note that xcopy prompts for overwriting if the file exists.

acruis:
What files in particular were to be overwritten?

damithc:
I think the overwrite issue happens if the previous run failed in the middle. For example, the back up folder does not get cleaned if the script failed and causes an overwrite alert next time.

acruis:
I see. Thanks a lot for dogfooding this issue! I'll see what I can do to make `clean` work properly.

damithc:
@acruis Can finish this one?

acruis:
Yep, I'll finish this first.

acruis:
@damithc Ready for review again! The script now waits for chromedriver to finish closing down before proceeding.

If a rogue chromedriver exists (let's say if HubTurbo closes down but somehow leaves a chromedriver running), then the script will stall. The only option then would be to force-kill all chromedriver processes (as we do not know which belongs to the parent HT process), which is drastic, so I didn't put that in. Do tell me if that is OK, or if the script should forcefully close chromedriver processes.

damithc:
Looks good. Go ahead and merge.

PR #1009, opened 2015-09-24
Ctrl+R doesn't work if the focus is on the repo dropdown #962

### LINKED ISSUE #846: Document 'change repo' keyboard shortcut
Can be documented under the 'Control panel view' section.
The shortcut is Ctrl+R

### LINKED ISSUE #962: Ctrl+R doesn't work if the focus is on the repo dropdown
As per title

HansNewbie:
Fixes #962, fixes #846

### DIFF
--- a/src/main/java/ui/UI.java
+++ b/src/main/java/ui/UI.java
@@ -16,6 +16,7 @@ import javafx.scene.control.Label;
 import javafx.scene.control.ScrollPane;
 import javafx.scene.control.ScrollPane.ScrollBarPolicy;
 import javafx.scene.control.Tooltip;
+import javafx.scene.input.KeyEvent;
 import javafx.scene.layout.BorderPane;
 import javafx.scene.layout.HBox;
 import javafx.scene.layout.Priority;
@@ -45,6 +46,9 @@ import java.awt.*;
 import java.util.HashMap;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
+
+import static ui.components.KeyboardShortcuts.SWITCH_DEFAULT_REPO;
+
 
 public class UI extends Application implements EventDispatcher {
 
@@ -180,7 +184,7 @@ public class UI extends Application implements EventDispatcher {
         UI.events = this;
 
         Thread.currentThread().setUncaughtExceptionHandler((thread, throwable) ->
-            logger.error(throwable.getMessage(), throwable));
+                logger.error(throwable.getMessage(), throwable));
 
         TestController.setUI(this, getParameters());
         prefs = new Preferences(TestController.isTestMode());
@@ -219,6 +223,7 @@ public class UI extends Application implements EventDispatcher {
 
         Scene scene = new Scene(createRootNode());
         setupMainStage(scene);
+        setupGlobalKeyboardShortcuts(scene);
         notificationController = new NotificationController(notificationPane);
         notificationPane.setId("notificationPane");
 
@@ -292,6 +297,14 @@ public class UI extends Application implements EventDispatcher {
             });
         });
         mainStage.hide();
+    }
+
+    private void setupGlobalKeyboardShortcuts(Scene scene) {
+        scene.addEventHandler(KeyEvent.KEY_PRESSED, event -> {
+            if (SWITCH_DEFAULT_REPO.match(event)) {
+                switchDefaultRepo();
+            }
+        });
     }
 
     private static void getMainWindowHandle(String windowTitle) {
--- a/src/test/java/guitests/KeyboardShortcutsTest.java
+++ b/src/test/java/guitests/KeyboardShortcutsTest.java
@@ -13,7 +13,10 @@ import util.events.IssueSelectedEventHandler;
 import util.events.PanelClickedEventHandler;
 import util.events.testevents.UIComponentFocusEvent;
 import util.events.testevents.UIComponentFocusEventHandler;
+
 import static org.junit.Assert.assertEquals;
+
+import static ui.components.KeyboardShortcuts.SWITCH_DEFAULT_REPO;
 
 public class KeyboardShortcutsTest extends UITest {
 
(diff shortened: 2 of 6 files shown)


HansNewbie:
@damithc this PR is regarding Ctrl+R on repo dropdown. I have a few clarifications needed:
- Currently if you press `Esc` on repo dropdown, the focus is on the bar above the panel, hence all panel shortcut will not work. Would you like to keep it this way or if `Esc` is pressed the focus should return to panel?
- On pressing `Ctrl+R`, I will just put the cursor at the end of the repo name. Is that alright, or would you like focus to be transferred back to panel?

I will update the keyboard shortcut documentation about `Ctrl+R` and create a test case for `Ctrl+R` on repo dropdown soon.

dariusf:
I added the documentation issue to the description.

dariusf:
Actually I'm not sure we should duplicate the hotkey binding here. Would it be possible to bind Ctrl+R globally, by registering it [on the scene](http://stackoverflow.com/a/25421376)?

src/main/java/ui/UI.java
```diff
@@ -294,6 +299,14 @@ private void setupMainStage(Scene scene) {
         mainStage.hide();
     }
 
+    private void setupGlobalKeyboardShortcut(Scene scene) {
```
HansNewbie: @dariusf Something like this?

dariusf: Yup!

src/test/java/guitests/KeyboardShortcutsTest.java
```diff
@@ -144,15 +144,48 @@ public void keyboardShortcutsTest() {
...
+        PlatformEx.waitOnFxThread();
+        assertEquals(2, repoSelectorComboBox.getItems().size());
+        assertEquals(repoSelectorComboBox.getValue(), "dummy1/dummy1");
+        // test shortcut on repo dropdown
+        doubleClick(repoSelectorComboBox);
+        press(KeyCode.CONTROL).press(KeyCode.R).release(KeyCode.R).release(KeyCode.CONTROL);
+        PlatformEx.waitOnFxThread();
+        push(KeyCode.ESCAPE);
+        PlatformEx.waitOnFxThread();
+        assertEquals(repoSelectorComboBox.getValue(), "dummy/dummy");
```
HansNewbie: @dariusf I got a problem here. Altho it passes locally and also on manual testing everything was good, this case always fails in Travis. In Travis, `Ctrl+R` does not work on repo dropdown but it works on panels. Any guess on why and what I could do about it?

I tried several things:
  - `PlatformEx.waitOnFxThread()` - you can say I spam it since even without any of those the test would pass locally
  - `sleep(1000)` - didn't work
  - press `Esc` after `Ctrl+R` on repo dropdown (in the hope that it will update) - didn't work

dariusf: Possible causes... the event handler could not be bound, but that's unlikely.

Try using the new `pushKeys` method on the KeyCodeCombination directly. Another possibility is to wait until isFocused() returns true (UITest.waitUntil).

HansNewbie: @dariusf I tried to check if the key press works on Travis (by printing out some message on the keypress event) and it works.

I tried `pushKeys`, same result, does not work on repo dropdown. I am not sure what you mean with isFocused() but let me try it.

dariusf: Something like [this](https://github.com/HubTurbo/HubTurbo/blob/master/src/test/java/guitests/UITest.java#L226)? I don't know whether the problem is that the repo dropdown is not in focus when the key press event happens though.

Anyway pushKeys is preferred to press/push. More convenient and there is less chance of problems with timing issues.

HansNewbie: Since the keyboard shortcut is global, where the focus is should not matter, right? Then do I still need wait until is focused?

On pushKeys, apparently platformEx.waitOnThread did not work. It will fail locally. I tried using sleep for now just to see if it works on Travis with pushKeys.

HansNewbie:
@dariusf should I just revert back to putting the shortcut in the panels and selector box instead of making it global? Or should I not put the failing test case (bad idea, but I don't think there is any workaround)?

dariusf:
Revert I guess, and leave a comment there saying why it's duplicated.

HansNewbie:
Noted

dariusf:
Thanks for exploring it! So this PR will now only bind the key in the repo dropdown, and update the docs, right?

HansNewbie:
And also test cases. Originally `Ctrl+R` has no test cases at all.

And I think I figure out why Travis keeps failing at the test. It seems `Ctrl+R` on repo dropdown is non-deterministic. I tried changing the test case to be neater and now I get failing test case on my local machine as well on repo dropdown.

dariusf:
Nondeterministic in that it only triggers sometimes?

By the way, did you try `comboBox.addEventHandler`? `setOnKeyPressed`, `setOnKeyReleased`, `setOnKeyTyped` should work, failing that?

HansNewbie:
I just added on to current code under `comboBox.addEventFilter(KeyEvent.KEY_RELEASED, event -> {`.

It's kind of negative nondeterministic. Most of the time it fails, but sometimes it passes. I managed to make it work by having a while loop for `sleep` until the value changes. I noticed sometimes it needs 1x `sleep(100)` sometimes it does not, as in the behaviour is weird. The next problem is that the panel names are not updating. After that sleep cycle and all, I keep getting error that no node is found for `"#dummy/dummy_col1_1"` although the default repo is now `"dummy/dummy"`.

HansNewbie:
Just realized it's still `KEY_RELEASED`. Will  change to `KEY_PRESSED`.

damithc:
Do a check for other instances of `KEY_RELEASED` and create issues if they need to be changed.

HansNewbie:
@damithc Noted. I will create issues if I find them, except for the one in listPanel and filterPanel since Rajan is fixing it.

PR #1010, opened 2015-09-24
Refreshing the bView at the end of undo-period is disruptive #985

### LINKED ISSUE #985: Refreshing the bView at the end of undo-period is disruptive
I'm referring to the refresh we do after a label change is effected after the undo-window. Several times I was adding an explanatory comment in the bView when it got reloaded automatically. I don't think we need to refresh the entire page. GitHub pages can pull label changes from the server periodically.
May be we can trigger that behavior ourselves? If not, we can let it run its own course?

lejolly:
Fixes #985


dariusf:
Hmm, this is straightforward enough, but is the coverage drop legit?

lejolly:
ok, removed the unused code

dariusf:
Okay, go ahead and merge.

PR #1011, opened 2015-09-26
 Add name validation for boards #975 

### LINKED ISSUE #975: Add name validation for boards
1. "none" not allowed
2. Maximum length? (preferably based on width rather than char count, which will link back to #892 )
3. Duplicate board name allowed, but user will be warned that saving will overwrite previous board with that name.

Also going to include tests for #935

HyungJon:
Fixes #975 

Board name length is calculated as char count rather than pixel width since 1) unlike panels, displaying board name is not bound to a fixed space, 2) it would be unintuitive for a user to be know that "board name cannot exceed XX pixels".

Problematic board names will prompt text warnings to be displayed below the input ```TextField```. If name is invalid, then the submit button will be disabled to forbid submission rather than accepting the name and then (not) notifying the user the board wasn't saved.

May still need some tweaks to dialog layout, please point out if so.


src/main/java/ui/BoardNameDialog.java
```diff
@@ -0,0 +1,150 @@
...
+import javafx.stage.Modality;
+import javafx.stage.Stage;
+
+public class BoardNameDialog extends Dialog<String> {
+    
+    Preferences prefs;
+    TextField nameField;
+    Text prompt;
+    Text errorText;
+    Button submitButton;
```
dariusf: These should be private.

src/main/java/ui/BoardNameDialog.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private ArrayList<String> invalidNames = new ArrayList<>();
```
dariusf: A better data structure for this is a Set? Also it can be static, and you can use Collections.unmodifiableSet to freeze it.

src/main/java/ui/BoardNameDialog.java
```diff
@@ -0,0 +1,150 @@
...
+    private void createButtons() {
+        ButtonType submitButtonType = new ButtonType("Save", ButtonBar.ButtonData.OK_DONE);
+        getDialogPane().getButtonTypes().addAll(submitButtonType, ButtonType.CANCEL);
+        submitButton = (Button) getDialogPane().lookupButton(submitButtonType);
+        
+        setResultConverter(submit -> {
+            if (submit == submitButtonType) {
+                return nameField.getText();
+            }
+            return null;
```
dariusf: Does the null here translate into an empty Optional result?

src/main/java/ui/BoardNameDialog.java
```diff
@@ -0,0 +1,150 @@
...
+        });
+        
+    }
+    
+    private boolean isBoardNameInvalid(String newName) {
+        return invalidNames.contains(newName);
+    }
+    
+    private boolean isBoardNameEmpty(String newName) {
+        if (newName.equals("")) {
```
dariusf: This can be replaced with
```
return newName.isEmpty();
```

src/main/java/ui/BoardNameDialog.java
```diff
@@ -0,0 +1,150 @@
...
+            return true;
+        }
+        return false;
+    }
+    
+    private boolean isBoardNameDuplicate(String newName) {
+        if (prefs.getAllBoards().containsKey(newName)) {
+            return true;
```
dariusf: Same here, we can just return the result of the predicate.

src/main/java/ui/UI.java
```diff
@@ -522,4 +526,20 @@ public void updateTitle() {
         mainStage.setTitle(title);
     }
     
+    public String getTitle() {
+        return mainStage.getTitle();
+    }
+    
+    public int getVersionMajor() {
+        return VERSION_MAJOR;
```
dariusf: Do we really need these getters? They're constants, so we can just make them public I guess.

dariusf:
You can use [ValidationSupport](http://controlsfx.bitbucket.org/org/controlsfx/validation/ValidationSupport.html) from ControlsFX to show validation errors. See how it's used in FilterTextField.

HyungJon:
Hmm, I don't know if it's just me, but I don't see the validation failure messages, even in ```FilterTextField``` so I can't be sure whether I'm doing this right. I'm very sure I did it like in ```FilterTextField``` though.

src/main/java/ui/BoardNameDialog.java
HyungJon: It does as far as I know. I remember returning null because I can't return an empty string or "" and I had to search online.

dariusf: Okay, just confirming.

dariusf:
They only appear when you mouse over the 'x' on the bottom left. It's kind of unintuitive. Maybe we can leave the text display in, but use validation support to show the 'x'?

HyungJon:
Hmm... in that case I think the submit buttons being disabled and grayed out is sufficient as an error indication compared to the red 'x', if the message isn't displayed unless the cursor is over it.

dariusf:
I suppose that's fine. My thinking was that errors should be presented in a consistent way across the application. I don't know of a good way to do that yet though, as we don't have that many forms. We can defer this for now I guess.

HyungJon:
That makes sense. I'll use it for panel names later. It would fit well in panel rename textfiels.

HyungJon:
Hmm I can't get that Travis failure to pass....

dariusf:
What's causing the failures?

HyungJon:
I resolved it, Travis was failing to control the board name input properly. I got a second problem now probably because of which stage has the focus. I'm expecting this push to resolve it. Will have to look into it more if not.

HyungJon:
Seems to have sent Travis to unusually long build. Restarting hasn't fixed yet.

dariusf:
If you update to the latest master, the CI won't be as painful to deal with (and you can run the unstable tests offline).

HyungJon:
All unstable tests passed locally, except ```RepositorySelectorTest``` due to ```NullPointerException``` at line 100. I'm not sure whether it's because I'm running it from Eclipse. Still checking.

HyungJon:
All right, unstable tests passed.

src/test/java/unstable/MenuControlTest.java
```diff
@@ -22,10 +25,17 @@
     @Test
     public void menuControlTest() {
         
+        UI ui = TestController.getUI();
+        PanelControl panelControl = ui.getPanelControl();
+        Preferences testPref = UI.prefs;
+        
+        int vMajor = ui.VERSION_MAJOR;
+        int vMinor = ui.VERSION_MINOR;
+        int vPatch = ui.VERSION_PATCH;
```
dariusf: These can be accessed through `UI`. We shouldn't access static fields through instances. Also these extra variables aren't really needed I think.

src/test/java/unstable/UpdateIssuesTest.java
```diff
@@ -73,7 +75,8 @@ public void updateIssue(int issueId, String newIssueTitle) {
 
     @SuppressWarnings("unchecked")
     public int countIssuesShown() throws InterruptedException, ExecutionException {
-        FutureTask countIssues = new FutureTask(((ListPanel) find("#dummy/dummy_col0"))::getIssueCount);
+        FutureTask countIssues = new FutureTask(((ListPanel)
```
dariusf: The FutureTask should should be parameterised by its result type (Integer?).

dariusf:
The NPE in the unstable tests is being fixed. In the meantime you can test by copying `src/test/resources/guitests` to `src/test/resources/unstable`.

Also, you could clean up history a bit, since there are a few commits which are refinements of earlier ones (just squash these) and merges.

HyungJon:
Most of those noisy commits should have been already squashed since I've been squashing as I push.

dariusf:
I'm mostly referring to the merges, but I suppose resolving the conflicts again would take unnecessary effort.

I also think the level of abstraction could be higher: one commit adding validation and tests, maybe another two to document/revert the attempts at using ValidationSupport. That's likely all a future reader would need. Not super important though, so it shouldn't hold this back.

HyungJon:
Will keep that in mind. That said, changes to reflect what you pointed out wil have low abstraction level since they're just simple fixes. It's pretty hard to squash when there are merge commits involved though, how do you recommend I commit in cases like this?

dariusf:
It's okay, no need to do it for this one. No point expending effort to redo merges when everything is already working.

PR #1014, opened 2015-09-27
Ctrl+key combination is hard to register #1001

### LINKED ISSUE #1001: Ctrl+key combination is hard to register
I have to be very deliberate when using `Ctrl+key` combinations. Most of the time only the `key` is registered.

rd1992:
Fixes #1001

### DIFF
--- a/src/main/java/ui/issuepanel/FilterPanel.java
+++ b/src/main/java/ui/issuepanel/FilterPanel.java
@@ -1,5 +1,13 @@
 package ui.issuepanel;
 
+import static ui.components.KeyboardShortcuts.DEFAULT_SIZE_WINDOW;
+import static ui.components.KeyboardShortcuts.JUMP_TO_FILTER_BOX;
+import static ui.components.KeyboardShortcuts.JUMP_TO_FIRST_ISSUE;
+import static ui.components.KeyboardShortcuts.JUMP_TO_NTH_ISSUE_KEYS;
+import static ui.components.KeyboardShortcuts.MAXIMIZE_WINDOW;
+import static ui.components.KeyboardShortcuts.MINIMIZE_WINDOW;
+import static ui.components.KeyboardShortcuts.SWITCH_BOARD;
+import static ui.components.KeyboardShortcuts.SWITCH_DEFAULT_REPO;
 import backend.interfaces.IModel;
 import backend.resource.TurboIssue;
 import backend.resource.TurboUser;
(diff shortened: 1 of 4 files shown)


rd1992:
@dariusf is this what you had in mind?

dariusf:
Yup, this is a step in the right direction! The filter field being a child of the panel means we won't have to bind most things globally.

I don't think we need to move everything to FilterPanel though; binding those keys to the ListPanel should have the same effect. Plus if we do this, we can no longer have panels which are polymorphic in the presentation of issues but can still benefit from filtering.

I'm not sure if there are any keybindings left that still need to be (effectively) global. Does the repo dropdown share any with panels? Not sure if this will conflict with what @HansNewbie is doing in #1009.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -68,7 +83,55 @@ public FilterPanel(UI ui, IModel model, PanelControl parentPanelControl, int pan
...
+    
+    private void setFocusToFilterBox() {
+        if (TestController.isTestMode()) {
+            ui.triggerEvent(new UIComponentFocusEvent(UIComponentFocusEvent.EventType.FILTER_BOX));
+        }
+        filterTextField.requestFocus();
+        filterTextField.setText(filterTextField.getText().trim());
+        filterTextField.positionCaret(filterTextField.getLength());
+
+        addEventHandler(KeyEvent.KEY_PRESSED, event -> {
```
dariusf: This looks like a bug (from a while back it seems). Every time we jump to the filter box this handler will be bound.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -68,7 +83,55 @@ public FilterPanel(UI ui, IModel model, PanelControl parentPanelControl, int pan
...
-        
+        setupKeyboardShortcuts();
+    }
+
+    private void setupKeyboardShortcuts() {
+        addEventHandler(KeyEvent.KEY_PRESSED, event -> {
+            if (MAXIMIZE_WINDOW.match(event)) {
+                ui.maximizeWindow();
+            }
+            if (MINIMIZE_WINDOW.match(event)) {
```
dariusf: Could we use `else if` here instead, so we guarantee that the bindings are mutually exclusive?

HansNewbie:
Pretty sure will have conflict with mine as I made changes to `ListPanel`.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -68,7 +83,55 @@ public FilterPanel(UI ui, IModel model, PanelControl parentPanelControl, int pan
...
+                ui.minimizeWindow();
+            }
+            if (DEFAULT_SIZE_WINDOW.match(event)) {
+                ui.setDefaultWidth();
+            }
+            if (SWITCH_DEFAULT_REPO.match(event)) {
```
HansNewbie: This can be removed as I made it a global keyboard shortcut (i.e. binded to Scene) in #1009.

rd1992:
@dariusf havent moved everything to the Filterpanel. Only those which were being repeated as part of filtertextfield and the issue list. I'll take a look at @HansNewbie's PR.

src/main/java/ui/issuepanel/FilterPanel.java
HansNewbie: @rd1992 Sorry, you can ignore this. I am going to change my PR since the global shortcut is messing with the test. I will keep the switch repo shortcut in the panel.

PR #1017, opened 2015-10-03
The 'updated' filter does not consider comments on the diff of PRs

### LINKED ISSUE #654: The 'updated' filter does not consider comments on the diff of PRs
Looks like it does not consider the time stamp of comments added to the diff of PRs

ndt93:
Fix #654 

Depending on how large a user's `updated` filter window is, the correctness of this fix may take some times to stabilize since only PRs updated after the fix will be combined with the existing Issues.

Another issue is with commit comments. Commenting on a commit does not affect the `updatedAt` time of both PR and Issue. So we need to retrieve a list of commits for a PR and inject their latest `updatedAt` into the PR. This has potential implementation difficulties due to API usage, comments and PRs being retrieved on 2 separate update cycles and also is not as critical to be in this PR since we don't comment on commit that often. So I will open a separate issue for this.

### DIFF
--- a/src/main/java/backend/github/UpdateIssuesTask.java
+++ b/src/main/java/backend/github/UpdateIssuesTask.java
@@ -28,7 +28,7 @@ public class UpdateIssuesTask extends GitHubRepoTask<GitHubRepoTask.Result<Turbo
         ImmutableTriple<List<TurboIssue>, String, Date> changes = repo.getUpdatedIssues(model.getRepoId(),
             model.getUpdateSignature().issuesETag, model.getUpdateSignature().lastCheckTime);
         List<PullRequest> updatedPullRequests = repo.getUpdatedPullRequests(
-                model.getRepoId(), "", model.getUpdateSignature().lastCheckTime);
+                model.getRepoId(), model.getUpdateSignature().lastCheckTime);
 
         List<TurboIssue> existing = model.getIssues();
         List<TurboIssue> updatedIssues = changes.left;
--- a/src/main/java/backend/github/UpdateIssuesTask.java
+++ b/src/main/java/backend/github/UpdateIssuesTask.java
@@ -6,6 +6,7 @@ import backend.resource.Model;
 import backend.resource.TurboIssue;
 import org.apache.commons.lang3.tuple.ImmutableTriple;
 import org.apache.logging.log4j.Logger;
+import org.eclipse.egit.github.core.PullRequest;
 import util.HTLog;
 
 import java.util.Date;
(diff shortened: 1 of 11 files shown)


ndt93:

src/main/java/backend/github/UpdateIssuesTask.java
```diff
@@ -26,16 +27,21 @@ public UpdateIssuesTask(TaskRunner taskRunner, Repo repo, Model model) {
     public void run() {
         ImmutableTriple<List<TurboIssue>, String, Date> changes = repo.getUpdatedIssues(model.getRepoId(),
             model.getUpdateSignature().issuesETag, model.getUpdateSignature().lastCheckTime);
+        List<PullRequest> updatedPullRequests = repo.getUpdatedPullRequests(
+                model.getRepoId(), "", model.getUpdateSignature().lastCheckTime);
```
dariusf: Are we not using ETags for PRs at all? Given that nothing is passed in here, and PullRequestUpdateService.getUpdatedItems is overridden so it doesn't check them.

dariusf:
Looks good. Would it be possible to set up an integration test? Can just be a simple one like MetadataUpdateTests, where we cover the whole workflow of getting an issue, detecting updates to its PR, then filtering and having it (dis)appear.

Also, just to clarify, when issues are updated we iterate through all PRs, sorted by date, and get everything since the last issue update, kind of like our own implementation of a `since` parameter?

About commit comments, would it be as straightforward as doing the PRs, then getting relevant commits and injecting the new times into issues the same way?

damithc:
What's the cost on the API usage? For example, if there are no commit comments or diff comments since the previous refresh, how many calls are used on trying to find possible updates. What does it depend on? the number of PRs in the repo?

ndt93:
Yup, so we kind of implement our own `since` parameter by asking GitHub to sort the list of PRs by its `updated_at` field. The pages downloading stops as soon as I detect an PR that is updated before the `lastIssueCheckTime`. I also reduces number of items per page to 30 so we can detect this with smaller overhead.

@damithc So according to the mechanism describe above, there should be only 1 extra API call more than optimal to get all updated PRs in each refresh e.g. if there is no PR updated since the last refresh, we would stop after getting the first page. The extra APIs consumption now is mostly due to getting the review/commit comments for the `updated` filter.

I haven't thought too much about the commit comments problem but it might be more difficult. Matching commits with prs is harder as b/w issues and prs (e.g. the commits returned by GitHub API don't point to their containing PRs), and they are also obtained in 2 separate update cycles. Also, commits are not persistent as compared to PRs. Commits API supports `since` parameter though so that's a plus.

I will check if I can put up an integration test for this.

src/main/java/backend/github/UpdateIssuesTask.java
ndt93: Yup, it's there just to conform to the interface. Since we are implementing our own `since` (which cancel the page downloading early, the ETags don't really help much but can make things confusing to reason about and require us to maintain and store them.

ndt93:
I was looking at some of the integration tests for the updating procedure and we are using stub classes like `DummyRepo` and `DummySource` which are substituted when the flag `isTestMode` is on. They contain a different set of code to mimic the same behavior as the actual classes. I can add in some dummy PRs update code to them but it doesn't seem to add more testing value to the backend changes in this PR.

@dariusf Any thoughts?

dariusf:
I was hoping that interactions between some of these changed classes could be tested. Are there none at all? If all the behaviour that's been changed here is going to be stubbed out, then yes, there's no point.

src/main/java/backend/github/UpdateIssuesTask.java
dariusf: Could we remove it as a parameter then? Or is that possible that ETags may be introduced in future?

ndt93: This is because I inherits `UpdateService` and call the superclass constructor which always requires and ETag. Unless I overload another constructor for `UpdateService` that has a default ETag?

ndt93:
To summarize the changes:
* `PullRequestServiceEx` -> to download review comments from github (need to stub out)
* `ReviewComment` -> I make it inherits Comment so existing tests would test them as normal comments. Structures specific to ReviewComment are covered in unit tests.
* `PullRequestUpdateService` -> download updated pull requests from github (need to stub out)
* Merging issues and pull requests (consists of some utility methods in `TurboIssue` which are covered in unit tests)

Those parts that would benefit the most from an integration test are stubbed out while those left are utility methods or new class structures which I think are more suitably covered by unit tests.

src/main/java/backend/github/UpdateIssuesTask.java
ndt93: Oh wait, I can just pass `""` to the super constructor all the time. I will change this then.

ndt93:
@dariusf I've removed parts related to ETags, you can check this again.

dariusf:
Okay, no more issues then.

It would be good to squash (commits should describe features, so no need for 'added tests'- and 'created class'-type commits, or things done in retrospect like coverage fixes), but maybe when CI is less finicky.

ndt93:
Yeah, I usually try to squash but the CI is giving me phobia recently :P

PR #1018, opened 2015-10-03
Make Keyboard Shortcuts work in Mac #957

### LINKED ISSUE #957: Make Keyboard Shortcuts work in Mac
Currently not all `Ctrl` keyboard shortcut works in Mac due to Mac OS wide shortcut (e.g. `Ctrl` + :arrow_down: is App Expose shortcut, hence will not jump to issue list). Consider using Mac `Cmd` button instead (e.g. copy in Mac is `Cmd` + `C` instead of `Ctrl` + `C` by default).

JavaFX has `KeyCode.META` for this, but I tested and it didn't work (might be version problem or JavaFX bug). It was suggested to use `KeyCode.SHORTCUT` to make OS agnostic key (it will be mapped to `KeyCode.CONTROL` in Windows and `KeyCode.META` in Mac, but no mention of Linux).

Info: http://stackoverflow.com/questions/12710468/using-javafx-2-2-mnemonic

dariusf:
Fixes #957. Tests broken as a result of this change have been updated to use the KeyboardShortcuts constants so things are more DRY.

### DIFF
--- a/docs/keyboardShortcuts.md
+++ b/docs/keyboardShortcuts.md
@@ -2,12 +2,11 @@
 
 ## Go to page ...
 
-These key sequences (one key followed by the other) are used to load (in the browser view)
- various pages of the default repository
+These key sequences (one key followed by the other) are used to load various pages of the default repository in the browser view.
 
-| **Page to go**| **Shortcut**|**Explanation**|
+| **Page to go to**| **Shortcut**|**Explanation**|
 | ------------- |:-------------:| ------------|
-| **G**o to **I**ssues page|<kbd>G</kbd> then <kbd>I</kbd>|Loads the *issues* page in the browser view|
+| **G**o to **I**ssues page|<kbd>G</kbd> <kbd>I</kbd>|Loads the *issues* page in the browser view|
 | **L**abels page|<kbd>G</kbd> <kbd>L</kbd>|Loads the *labels* page |
 | **M**ilestones page|<kbd>G</kbd> <kbd>M</kbd>|Loads the *milestones* page |
 | **P**ull requests page|<kbd>G</kbd> <kbd>P</kbd>|Loads the *pull requests* page |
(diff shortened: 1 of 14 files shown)


dariusf:
@damithc @rd1992 there's one conflict with default OS X hotkeys: Cmd+M minimises windows, but we use Cmd+N for that, and Cmd+M for changing milestones. This might be a bit dissonant. How should we resolve this?

damithc:
Are there any other letter we can use for milestones?

dariusf:
Cmd+E? I and L are for issues and labels.

damithc:
Previously, we used L for labelling and Ctrl+L for creating a new label.
We can't use E and Ctrl+E the same way because E is used to mark an issue
as read.
Upgrade all 'create new ...' shortcuts to use Ctrl+Shift+... ?

dariusf:
Sure, that will work.

docs/keyboardShortcuts.md
```diff
@@ -17,11 +16,13 @@ These key sequences (one key followed by the other) are used to load (in the bro
...
 | **Item to create**| **Shortcut**|**Explanation**|
 | ------------- |:-------------:| --------- |
-| Issue|<kbd>Ctrl</kbd> + <kbd>I</kbd>|Loads the 'New Issue' page in the browser view|
-| Label|<kbd>Ctrl</kbd> + <kbd>L</kbd>|Loads the 'Labels' page in the browser view|
-| Milestone|<kbd>Ctrl</kbd> + <kbd>M</kbd>|Loads the 'New Milestone' page in the browser view|
+| Issue|<kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>I</kbd>|Loads the 'New Issue' page in the browser view|
+| Label|<kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>L</kbd>|Loads the 'Labels' page in the browser view|
+| Milestone|<kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>M</kbd>|Loads the 'New Milestone' page in the browser view|
 | Panel (on right)|<kbd>Ctrl</kbd> + <kbd>P</kbd>| Create a new panel to the right of current panel|
 | Panel (on left)|<kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>P</kbd>|Create a new panel to the left of current panel|
```
damithc: The default behavior should be `Ctrl+Shift+p`?

dariusf: Default behaviour?

damithc: The default behavior (that is, the more common behavior) is 'create panel on right' which should be bound to `ctrl+shift+p`? May be it doesn't matter much because we have both `ctrl+p` and `ctrl+shift+p`. The current one is less work for the user (the common behavior requires only 2 keys), but not consistent with other shortcuts.

dariusf: The current one feels fine to me. It's consistent with how browsers work, where the tab opens on the right.

damithc: Yes, the default should be on the right.
An HT user is expected to translate 'create new' to Ctrl+Shift (to be
consistent with other similar shortcuts) and 'panel' to P. then he will be
surprised to see the panel opening on left. That's the thing I'm worried
about. But it's ok, we can keep it as it is.

dariusf:
This works and passes offline, so I guess we could start reviewing.

src/test/java/tests/SortQualifierTests.java
```diff
@@ -397,4 +399,34 @@ public void statusOrdering() {
                 .map(TurboIssue::getId)
                 .collect(Collectors.toList()));
     }
+
+    @Test
+    public void stability() {
```
lejolly: is this meant to be part of the pr?

dariusf: It was to address a coverage drop. Not exactly sure why it happens, but the build has only passed twice over the past couple of days, and both times showed some line in the sorting code as not being covered.

lejolly: can we move this out to another pr?

dariusf: New issue?

lejolly: yup.

dariusf: Okay, done. Rebased this and it passes now.

lejolly:
ok, good to go then.

PR #1020, opened 2015-10-04
There should be a way to navigate between tabs of a PR #869

### LINKED ISSUE #869: There should be a way to navigate between tabs of a PR
as per title

rd1992:
Fixes #869

### DIFF
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -12,6 +12,7 @@ import org.openqa.selenium.chrome.ChromeOptions;
 import org.openqa.selenium.os.Kernel32;
 import ui.UI;
 import util.GitHubURL;
+import util.GithubURLPageElements;
 import util.PlatformSpecific;
 import util.events.testevents.JumpToCommentEvent;
 import util.events.testevents.SendKeysToBrowserEvent;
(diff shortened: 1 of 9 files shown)


src/main/java/browserview/BrowserComponent.java
```diff
@@ -487,9 +487,9 @@ public String getCurrentUrl() {
         return driver.getCurrentUrl();
     }
 
-    public void switchToConversationTab() {
+    public void switchToTab(String tabName) {
```
dariusf: Could this string be an enum instead? We don't want callers of this method to have to know how GitHub's DOM elements are named. The enum can be in GitHubURL and have a toString method which generates the string.

src/main/java/util/GitHubURL.java
```diff
@@ -49,7 +49,7 @@ public static boolean isUrlIssue(String url) {
     }
     
     public static boolean isPullRequestLoaded(String url)  {
-        return (url.endsWith("/commits") || url.endsWith("/files"));
+        return url.matches("https://github.com/([^/]+)/([^/]+)/pull/([0-9]+)([/commits,/files]*)");
```
dariusf: The last part doesn't look right. `[/commits,/files]*` is a character class, so it will match things like `cc`, `,,` -- 0 or more of any character in the brackets.

Also we don't have to parenthesise groups that we're not going to refer to. That usually makes the regex a tiny bit faster.

This should probably have tests, given how mistakes in regexes can sometimes be very subtle.

damithc: When using regular expressions, use a comment to explain the intended meaning of the expression (e.g. 8 character strings ending with a number) and write test cases with a few examples of matching strings and boundary cases.

rd1992:
@dariusf final review before i merge it? (took ages to pass the travis test)

dariusf:
Looks good. Nice work!

PR #1021, opened 2015-10-04
Scroll to bottom when loading an existing issue in bView

### LINKED ISSUE #1006: Scroll to bottom when loading an existing issue in bView
In most cases I find myself pressing 'N' immediately after loading an issue because I'm interested in what's happened most recently to that issue. May be we should do that by default?

codemanmeet:
Fixes #1006

### DIFF
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -204,12 +204,11 @@ public class BrowserComponent {
         if (isPullRequest) {
             logger.info("Showing pull request #" + id);
             runBrowserOperation(() -> driver.get(GitHubURL.getPathForPullRequest(repoId, id), isForceRefresh));
-            runBrowserOperation(() -> driver.scrollToBottom());
         } else {
             logger.info("Showing issue #" + id);
             runBrowserOperation(() -> driver.get(GitHubURL.getPathForIssue(repoId, id), isForceRefresh));
-            runBrowserOperation(() -> driver.scrollToBottom());
         }
+        runBrowserOperation(() -> scrollToBottom());
     }
 
     public void jumpToComment(){
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -204,9 +204,11 @@ public class BrowserComponent {
         if (isPullRequest) {
             logger.info("Showing pull request #" + id);
             runBrowserOperation(() -> driver.get(GitHubURL.getPathForPullRequest(repoId, id), isForceRefresh));
+            runBrowserOperation(() -> driver.scrollToBottom());
         } else {
             logger.info("Showing issue #" + id);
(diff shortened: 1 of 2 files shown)


src/main/java/browserview/ChromeDriverEx.java
```diff
@@ -109,6 +109,16 @@ public WebElement findElement(By by) throws NoSuchElementException {
...
+    public void scrollToTop() {
+        String script = "window.scrollTo(0, 0)";
+        executeScript(script);
+    }
+
+    public void scrollToBottom() {
+        String script = "window.scrollTo(0, document.body.scrollHeight)";
+        executeScript(script);
+    }
+
```
damithc: We already had hot keys to scroll to top and bottom. So why the need to add these methods now? We should use the same methods in both places right?

src/main/java/browserview/BrowserComponent.java
```diff
@@ -204,9 +204,11 @@ public void showIssue(String repoId, int id, boolean isPullRequest, boolean isFo
         if (isPullRequest) {
             logger.info("Showing pull request #" + id);
             runBrowserOperation(() -> driver.get(GitHubURL.getPathForPullRequest(repoId, id), isForceRefresh));
+            runBrowserOperation(() -> driver.scrollToBottom());
         } else {
             logger.info("Showing issue #" + id);
             runBrowserOperation(() -> driver.get(GitHubURL.getPathForIssue(repoId, id), isForceRefresh));
+            runBrowserOperation(() -> driver.scrollToBottom());
```
dariusf: Since they occur in both cases, these should be outside the conditional. Also, a direct method reference can be used.

src/main/java/browserview/ChromeDriverEx.java
```diff
@@ -109,6 +109,16 @@ public WebElement findElement(By by) throws NoSuchElementException {
         throw new NoSuchElementException();
     }
 
+    public void scrollToTop() {
```
dariusf: There are already functions for doing this in BrowserComponent.

src/main/java/browserview/ChromeDriverEx.java
codemanmeet: Yup, will remove it!

src/main/java/browserview/BrowserComponent.java
codemanmeet: Done!

src/main/java/browserview/ChromeDriverEx.java
codemanmeet: Done!

dariusf:
Please squash. Also needs to pass on Travis.

dariusf:
No squash? In future please wait for the green light (the `toMerge` label) before merging.

codemanmeet:
My bad, I just pressed the "Merge from Master" button on Github. Yup will do.

PR #1023, opened 2015-10-04
Auto-complete suggests 'assignees' when it should suggest 'assignee' #1022

### LINKED ISSUE #1022: Auto-complete suggests 'assignees' when it should suggest 'assignee'
When I type 'a' in filter box, I get the suggestion 'assignees' when I'm expecting to see 'assignee'

lejolly:
Fixes #1022

### DIFF
--- a/src/test/java/guitests/UIComponentsTest.java
+++ b/src/test/java/guitests/UIComponentsTest.java
@@ -3,14 +3,30 @@ package guitests;
 import javafx.scene.input.KeyCode;
 import org.junit.Test;
 import ui.UI;
+import ui.components.FilterTextField;
 import util.events.UpdateProgressEvent;
+
+import static org.junit.Assert.assertEquals;
 
 public class UIComponentsTest extends UITest {
 
     // TODO check that filter text field does indeed do autocomplete correctly, etc
     @Test
-    public void filterTextFieldTest() {
+    public void keywordCompletionTest() {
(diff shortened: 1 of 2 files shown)


src/test/java/guitests/UIComponentsTest.java
```diff
@@ -3,14 +3,30 @@
...
+        type("ass");
+        push(KeyCode.TAB);
+        FilterTextField filterTextField = find("#dummy/dummy_col0_filterTextField");
+        assertEquals("assignee", filterTextField.getText());
+    }
+
+    @Test
     public void filterTextFieldTest() {
+        doubleClick("#dummy/dummy_col0_filterTextField");
         click("#dummy/dummy_col0_filterTextField");
```
acruis: Doesn't this `click` remove the select-all on the filter (that was caused by the previous `doubleClick`)?

lejolly: doubleclick has never been a select all, it's select word. triple click is select all.

acruis: Ah alright

acruis:
Good to merge.

src/test/java/guitests/UIComponentsTest.java
dariusf: I know this is used in quite a few places, and seems to be supported across the platforms we test on, but we should eventually factor it out into some kind of select-all or triple-click method.

Issue #947, opened 2015-09-08
Add a 'Reviewer checklist' to dev docs


dariusf:
This seems like a good idea with the growing number of things to keep track of. The code ones can cover common scenarios only, otherwise the list is potentially never-ending.

**General**

- Branch name, PR name, "fixes" description
- Tests for relevant features are added/updated
- Documentation is updated
- CI passes with no drop in coverage
- Unstable tests pass offline
- No CheckStyle warnings (these don't cause the CI build to fail, maybe they should?)

**Code**

- No `sleep` in tests, find a way to synchronise. `PlatformEx.waitOnFxThread` is usually a drop-in replacement.
- `Optional`s must be checked before use.
- Overall SLAP and putting methods in the right places
- Static imports of methods from the `util` package are generally fine, but otherwise shouldn't be used.

Add more suggestions here. After a while we can add this to the docs.

Issue #757, opened 2015-07-02
Document development workflow


dariusf:
We should add a few things to either the Workflow or Development Environment documents, mostly so new developers can get started more easily:

- [x] How to run tests
- [x] How to run unstable tests
- [x] How to view CheckStyle violations locally
- [x] How to view code coverage locally
- [x] Commonly-used Gradle tasks

acruis:
Agreed.

damithc:
Not sure a beginner can document the workflow that he/she is new to. :-)

HansNewbie:
- [ ] How to merge PR
  - follow command line instruction on GitHub PR merge, use git merge --no-ff -m "message" origin/your-branch-name.

PR #1025, opened 2015-10-05
Run unstable tests separately #1024

### LINKED ISSUE #1024: Run unstable tests separately
Devs/reviewers will manually check these. The Travis build should be as deterministic as possible. Related to #952.

dariusf:
Fixes #1024.

Docs will be updated separately in #947 and #757.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,6 @@
  * user guide available at http://gradle.org/docs/2.2.1/userguide/tutorial_java_projects.html
  */
 
-// Plugins
 plugins {
     id "com.github.kt3k.coveralls" version "2.4.0"
     id "com.github.johnrengelman.shadow" version "1.2.1"
@@ -76,28 +75,31 @@ allprojects {
 
     test {
         forkEvery = 1
+        if (System.env.'CI') {
+            exclude 'unstable'
+        }
+    }
 
-        // uncomment to disable gui tests
-//        exclude 'guitests'
+    sourceSets {
+        unstableTests {
+            java.srcDir 'src/test/java/unstable'
+            compileClasspath += main.output + test.output
+            runtimeClasspath += main.output + test.output
+        }
+    }
+
+    configurations {
+        unstableTestsCompile.extendsFrom testCompile
+        unstableTestsRuntime.extendsFrom testRuntime
+    }
+
+    task unstableTests(type: Test) {
+        testClassesDir = sourceSets.unstableTests.output.classesDir
+        classpath = sourceSets.unstableTests.runtimeClasspath
     }
 
     jacoco {
         toolVersion = "0.7.5.201505241946"
-    }
-
-    jacocoTestReport {
-        dependsOn test
-        sourceDirectories = project.files(sourceSets.main.allSource.srcDirs)
-        classDirectories =  project.files(sourceSets.main.output)
-        afterEvaluate {
-            classDirectories = files(classDirectories.files.collect {
-                fileTree(dir: it, exclude: 'tests/**')
-            })
-        }
-        reports {
-            xml.enabled = true
-            html.enabled = true
-        }
     }
 
     jar {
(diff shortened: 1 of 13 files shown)


dariusf:
An environment variable is used to control whether the unstable tests run. This is the simplest solution I could get to work. Tried splitting the tests up into source sets, but that caused problems with coverage (the task doesn't support computing coverage from multiple source sets by default). I'm still not sure how to define a task that operates on a subset of a default task's inputs.

@lejolly do you know if jacocoTestReport is needed? AFAIK it's not used. The parameter it supplies to jacocoRootReport has the default value.

dariusf:
Forgot to rename the task. I'm thinking of calling it either `testUnstable` or `unstableTests`. The first follows conventions, the second is more descriptive. Opinions welcome.

lejolly:
not sure if `jacocoTestReport` was needed, possibly to generate both html and xml reports locally?

i would go with `unstableTests`.

dariusf:
Renamed. In that case, I think it isn't needed. jacocoRootReport does the same thing.

Added a couple more unstable tests.

dariusf:
Added MenuControlTest. I think that should be all for now, unless there are other suggestions. We don't have too many samples, so I've only added those I've seen fail 2-3 times in the recent PRs.

lejolly:
Got this on the second restart of the build. 

```
guitests.MetadataUpdateTest > testUpdatedTriggersMetadata FAILED
    java.lang.RuntimeException: java.lang.NullPointerException
        at org.loadui.testfx.utils.TestUtils.awaitCondition(TestUtils.java:49)
        at org.loadui.testfx.utils.TestUtils.awaitCondition(TestUtils.java:30)
        at guitests.MetadataUpdateTest.testUpdatedTriggersMetadata(MetadataUpdateTest.java:24)
        Caused by:
        java.lang.NullPointerException
            at org.loadui.testfx.utils.FXTestUtils.isNodeWithinSceneBounds(FXTestUtils.java:249)
            at org.loadui.testfx.utils.FXTestUtils.isNodeVisible(FXTestUtils.java:242)
            at org.loadui.testfx.utils.FXTestUtils$6.apply(FXTestUtils.java:233)
            at org.loadui.testfx.utils.FXTestUtils$6.apply(FXTestUtils.java:230)
            at com.google.common.collect.Iterators$7.computeNext(Iterators.java:652)
            at com.google.common.collect.AbstractIterator.tryToComputeNext(AbstractIterator.java:143)
            at com.google.common.collect.AbstractIterator.hasNext(AbstractIterator.java:138)
            at com.google.common.collect.Iterators.getNext(Iterators.java:865)
            at com.google.common.collect.Iterables.getFirst(Iterables.java:775)
            at org.loadui.testfx.GuiTest.find(GuiTest.java:466)
            at org.loadui.testfx.GuiTest.find(GuiTest.java:306)
            at guitests.UITest.findQuiet(UITest.java:211)
            at guitests.MetadataUpdateTest.lambda$testUpdatedTriggersMetadata$20(MetadataUpdateTest.java:25)
            at org.loadui.testfx.utils.TestUtils.awaitCondition(TestUtils.java:38)
            ... 2 more
```

dariusf:
Added that. It's already in the nondeterministic test issue.

lejolly:
i think you forgot to update the findbugs exception file?

```
Write to static field ui.UI.status from instance method unstable.SavedLoginTest.setupMethod()
```

dariusf:
Yeah, just realised.

dariusf:
Passed another 5 times. Another failure from RepositorySelectorTest so I removed it as well.

So when do we decide if the CI build is stable enough?

lejolly:
i guess 100% passing rate is kinda unattainable? 90%? 80%?

dariusf:
Passed 9 more times. Not the most formal experiment, but we're at 20/22 (90.9%) now.

lejolly:
Ok, good to merge then. We can always include more tests in the future if they're giving us problems.

PR #1026, opened 2015-10-06
Delete Model of Unused Repositories #1028

### LINKED ISSUE #1028: Delete Model of Unused Repositories
So that its reference is not kept and it can be collected by garbage collector.

Part of effort to #481.

HansNewbie:
Fixes #1028

### DIFF
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -183,9 +183,14 @@ public class Logic {
         return models.getDefaultRepo();
     }
 
-    public CompletableFuture<Boolean> removeRepository(String repoId) {
-        models.removeRepoModelById(repoId);
+    public CompletableFuture<Boolean> removeStoredRepository(String repoId) {
         return repoIO.removeRepository(repoId);
+    }
+
+    public void removeUnusedModels(Set<String> reposInUse) {
+        models.toModels().stream().map(Model::getRepoId).
+                filter(repoId -> !reposInUse.contains(repoId)).
+                forEach(repoIdNotInUse -> models.removeRepoModelById(repoIdNotInUse));
     }
 
     /**
--- a/src/test/java/guitests/KeyboardShortcutsTest.java
+++ b/src/test/java/guitests/KeyboardShortcutsTest.java
@@ -5,6 +5,8 @@ import javafx.scene.input.KeyCode;
 
 import org.junit.Test;
 
+import org.loadui.testfx.utils.TestUtils;
+import ui.TestController;
 import ui.UI;
 import ui.components.KeyboardShortcuts;
 import ui.listpanel.ListPanel;
@@ -169,10 +171,9 @@ public class KeyboardShortcutsTest extends UITest {
         // test shortcut on repo dropdown
         doubleClick(repoSelectorComboBox);
         pushKeys(SWITCH_DEFAULT_REPO);
-        PlatformEx.waitOnFxThread();
+        waitUntilNodeAppears("#dummy/dummy_col1_1");
         assertEquals(repoSelectorComboBox.getValue(), "dummy/dummy");
         // test shortcut when focus is on panel
-        PlatformEx.waitOnFxThread();
         click("#dummy/dummy_col1_1");
         press(SWITCH_DEFAULT_REPO);
         PlatformEx.waitOnFxThread();
--- a/src/main/java/ui/GUIController.java
+++ b/src/main/java/ui/GUIController.java
@@ -113,7 +113,7 @@ public class GUIController {
 
         dispatchMetadataRequests(toUpdate);
 
-        UI.events.triggerEvent(new OpenReposChangedEvent());
+        UI.events.triggerEvent(new UsedReposChangedEvent());
     }
 
     /**
--- a/src/main/java/ui/MenuControl.java
+++ b/src/main/java/ui/MenuControl.java
@@ -17,10 +17,10 @@ import ui.issuepanel.FilterPanel;
 import ui.issuepanel.PanelControl;
 import util.events.*;
 
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -205,6 +205,8 @@ public class MenuControl extends MenuBar {
         panels.selectFirstPanel();
         prefs.setLastOpenBoard(boardName);
         ui.updateTitle();
+
+        ui.triggerEvent(new UsedReposChangedEvent());
     }
 
     /**
@@ -341,7 +343,7 @@ public class MenuControl extends MenuBar {
     private MenuItem[] createReposMenu() {
         Menu remove = new Menu("Remove");
 
-        ui.registerEvent((OpenReposChangedEventHandler) e -> {
+        ui.registerEvent((UnusedStoredReposChangedEventHandler) e -> {
             Platform.runLater(() -> updateRepoRemoveList(remove));
         });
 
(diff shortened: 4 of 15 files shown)


HansNewbie:
@dariusf I didn't put as fix because I am not sure if this is going to fix.

What I did:
- Remove repo model if repo is no longer in use (i.e. not in any panel and not default repo)
- Then I reworked removing repo to just removing JSON since the model will not exist anymore

So what happens now is that if a repo is no longer used, its model will be deleted but its JSON still exists. On preliminary test, the loading from JSON is not slow so performance should be fine. I didn't see any changes to memory use but I am going to take a deeper look.

I don't think this can be done in current milestone.

HansNewbie:
Did this memory test:
- prepare 2 boards:
  - One board has 1 panel with no filter - let's call this empty board
  - Another board has 3 panels opened, each having repo filter. 2 of the panels' filters are repo:HubTurbo/HubTurbo and repo:TEAMMATES/repo - let's call this full board
- Open the empty board
- Switch to full board
- After all panels are loaded, switch back to empty board
- Using YourKit memory profiler, force GC. Take snapshot of memory after forced GC.

The result is as expected: 3 Model-s are deleted, but this only freed about 120 bytes of data. Everything else is virtually the same. I don't think this fixes the issue.

HansNewbie:
@damithc to follow up on my comment above, the memory that is actually being used is actually quite small, in the sense that project data is released.

This problem seems to be more related to http://stackoverflow.com/questions/4952568/is-there-a-way-to-lower-java-heap-when-not-in-use. What happens is that once memory is allocated to JVM, it will never be released to the OS again. In HubTurbo, the memory that is being used in the long run is quite small, but the allocated memory does not go down - which is why it seems to be eating lots of memory.

Should this delete-model PR be continued/merged? Practically it doesn't change the total memory consumption and hence the issue in question still exist.

damithc:
I think the right thing to do is to let go of the model that we no longer use. If we keep a reference to it, it can never be collected by the garbage collector anyway. Actually, my observation is the memory allocated to JVM is eventually released when not needed, but this takes a very long time. e.g. leave it running overnight.

HansNewbie:
@damithc Sure, but should this PR fixes the issue or the issue should be kept open? Indeed deleting the Model will release the project data but since the issue is about memory, I am not sure if this PR should close the issue.

damithc:
You can create a separate issue and change the title of this PR to match the new issue.

HansNewbie:
Noted.

src/test/java/guitests/KeyboardShortcutsTest.java
```diff
@@ -171,6 +171,7 @@ public void keyboardShortcutsTest() {
         assertEquals(repoSelectorComboBox.getValue(), "dummy/dummy");
         // test shortcut when focus is on panel
         PlatformEx.waitOnFxThread();
+        sleep(1000);
```
HansNewbie: This one is just to make sure it is reliable on local. Seems to be very unreliable without sleep.

src/main/java/backend/Logic.java
```diff
@@ -178,11 +178,25 @@ public String getDefaultRepo() {
...
 
+    private void removeRepoModel(String repoId) {
+        models.removeRepoModelById(repoId);
+    }
+
+    public void removeUnusedModels(Set<String> reposInUse) {
+        // Not using getOpenRepositories since it returns all in lower case
+        // in which we can't feed to removeRepoModel(repoId)
+        Set<String> openRepos = models.toModels().stream().map(Model::getRepoId).collect(Collectors.toSet());
+        openRepos.removeAll(reposInUse);
```
dariusf: This can be written as a map, then filter, then forEach, since the set supports a fast `contains` operation.

src/test/java/guitests/KeyboardShortcutsTest.java
dariusf: Is there another way to synchronise? How about waiting for col1 to be visible, or for an issue card to be selected?

dariusf:
The logic for this is starting to get convoluted, with all the events going around... I can still understand it, but we might need something like GUIController soon, to manage the opening and closing of repos.

This is fine in general, but I don't think it will have much of an effect yet. We need to comprehensively evaluate how model objects go from the back end to the UI if we want to hunt down leaks, and even then model objects don't dominate memory usage so we won't see a huge drop.

src/test/java/guitests/KeyboardShortcutsTest.java
HansNewbie: The problem here is not col1 but rather dummy/dummy. The repo change makes the node traversal very unpredictable. I tried dummy/dummy_blah and dummy1/dummy1_blah and both gives Node not found. I can make it wait until dummy/dummy_col1 can be found, but it will make the time non-deterministic. Would that be ok?

src/main/java/backend/Logic.java
HansNewbie: Ok, will do.

src/test/java/guitests/KeyboardShortcutsTest.java
dariusf: The time waited being nondeterministic is fine, as long as it always eventually finds the node. Is that possible?

HansNewbie: I think it is. I can't remember clearly but I tried that before and it works, except that that time I used a while loop waiting for the node, and if the node is not available yet, I will sleep. I will try it out.

src/test/java/guitests/KeyboardShortcutsTest.java
```diff
@@ -170,7 +172,7 @@ public void keyboardShortcutsTest() {
         PlatformEx.waitOnFxThread();
         assertEquals(repoSelectorComboBox.getValue(), "dummy/dummy");
         // test shortcut when focus is on panel
-        PlatformEx.waitOnFxThread();
+        findOrWaitFor("#dummy/dummy_col1_1");
```
HansNewbie: Using this to make waiting more reliable. It passes on all my tries locally.

src/main/java/backend/Logic.java
```diff
@@ -178,11 +178,16 @@ public String getDefaultRepo() {
...
-    public CompletableFuture<Boolean> removeRepository(String repoId) {
-        models.removeRepoModelById(repoId);
+    public CompletableFuture<Boolean> removeStoredRepository(String repoId) {
         return repoIO.removeRepository(repoId);
     }
 
+    public void removeUnusedModels(Set<String> reposInUse) {
+        models.toModels().stream().map(Model::getRepoId).
+                filter(repoId -> !reposInUse.contains(repoId)).
+                forEach(repoIdNotInUse -> models.removeRepoModelById(repoIdNotInUse));
```
HansNewbie: Using this

src/test/java/guitests/KeyboardShortcutsTest.java
HansNewbie: Sorry, changing to another method.

src/test/java/guitests/KeyboardShortcutsTest.java
```diff
@@ -170,7 +172,7 @@ public void keyboardShortcutsTest() {
         PlatformEx.waitOnFxThread();
         assertEquals(repoSelectorComboBox.getValue(), "dummy/dummy");
         // test shortcut when focus is on panel
-        PlatformEx.waitOnFxThread();
+        waitUntilNodeAppears("#dummy/dummy_col1_1");
```
HansNewbie: using this for now

dariusf: If this works reliably, good to merge.

HansNewbie:
@dariusf It passes tests locally, except 1; RepositorySelectorTest. But it seems the problem is not because of this PR because the error is like this:

```
Exception in Application start method
Exception in thread "Thread-5" java.lang.RuntimeException: Exception in Application start method
	at com.sun.javafx.application.LauncherImpl.launchApplication1(LauncherImpl.java:917)
	at com.sun.javafx.application.LauncherImpl.lambda$launchApplication$152(LauncherImpl.java:182)
	at com.sun.javafx.application.LauncherImpl$$Lambda$15/899668028.run(Unknown Source)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.NullPointerException
	at ui.UI.initCSS(UI.java:267)
	at ui.UI.initUI(UI.java:234)
	at ui.UI.start(UI.java:102)
	at unstable.RepositorySelectorTest$RepositorySelectorTestUI.start(RepositorySelectorTest.java:35)
```

It can't set up the test. Other than that, everything is fine.

dariusf:
That's a problem with the runtime classpath, probably introduced when I tweaked the build configuration. I'll fix it. In the meantime you can test by copying `src/test/resources/guitests` to `src/test/resources/unstable`.

HansNewbie:
Passed all tests more than once.

HansNewbie:
@dariusf is this good to merge?

dariusf:
Yeah, good to go.

PR #1033, opened 2015-10-08
Keep the label positions in the label list stable while typing labels #789

### LINKED ISSUE #789: Keep the label positions in the label list stable while typing labels
Currently, ticks that are dynamically inserted (to mark selected labels) can change the position of other labels. May be we should reserve space for the tick in every label?
image

lejolly:
Fixes #789


lejolly:
good to go?

acruis:
Yup, looks good to merge.

Issue #1008, opened 2015-09-24
Show 'pending' label changes differently


damithc:
After we do a change to labels, the changes are shown in pView during the 'undo period' although they are not yet applied. Perhaps we should show those labels differently? e.g. bold
Otherwise the user has to remember what he did in case he wants to undo.

PR #1034, opened 2015-10-08
Added label not appearing in pView #974

### LINKED ISSUE #974: Added label not appearing in pView
I added a label using label picker. It appeared in bView after the undo window, but it took much longer for it appear in the pView. Shouldn't it appear immediately after I hit 'enter' on the label picker? 

image

lejolly:
Fixes #974

### DIFF
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -68,17 +68,22 @@ public class Logic {
         return repoIO.isRepositoryValid(repoId);
     }
 
-    public void refresh() {
+    public void refresh(boolean isNotificationPaneShowing) {
+        // TODO fix refresh to take into account the possible pending actions associated with the notification pane
+        if (isNotificationPaneShowing) {
+            logger.info("Notification Pane is currently showing, not going to refresh. ");
+            return;
+        }
         String message = "Refreshing " + models.toModels().stream()
-            .map(Model::getRepoId)
-            .collect(Collectors.joining(", "));
+                .map(Model::getRepoId)
+                .collect(Collectors.joining(", "));
 
         logger.info(message);
         UI.status.displayMessage(message);
 
         Futures.sequence(models.toModels().stream()
-            .map(repoIO::updateModel)
-            .collect(Collectors.toList()))
+                .map(repoIO::updateModel)
+                .collect(Collectors.toList()))
                 .thenApply(models::replace)
                 .thenRun(this::updateUI)
                 .thenCompose(n -> getRateLimitResetTime())
--- a/src/main/java/ui/MenuControl.java
+++ b/src/main/java/ui/MenuControl.java
@@ -11,20 +11,11 @@ import javafx.scene.control.ButtonBar.ButtonData;
 import javafx.stage.Modality;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import prefs.Preferences;
 import prefs.PanelInfo;
+import prefs.Preferences;
 import ui.issuepanel.FilterPanel;
 import ui.issuepanel.PanelControl;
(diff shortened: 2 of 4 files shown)


lejolly:
@damithc does this fix the issue for you?

dariusf:
Why does this fix the problem? Or rather, what's the problem with refreshing in the undo window?

lejolly:
Here's an example: 

Issue starts with one label: `label 1`. 
1. I add `label 2` to the issue (UI shows 2 labels)
2. I remove `label 1` from the same issue (at this point, `label 2` is sent to github). (UI shows 1 label)
3. While the undo notification is showing (that means the removal of `label 1` has not been sent yet), if a refresh is triggered (either by the ticking timer or by a browser change (which is what happens most of the time)), the new etag points to a change in the issue (because of the addition of `label 2`) and thus the issue is updated with the state of the issue at step 1 (where it had 2 labels). So the UI ends up being updated and shows 2 labels, when it's still supposed to be showing just 1.

dariusf:
Hmm, in that case I think the problem goes a lot deeper... Mutating the model for eager updates to the GUI (which may not have happened yet) forces us to reason about the order in which changes happen, and that's very tough when we can refresh at any time.

Making refresh a no-op sometimes won't solve the general case. We have to track the eager changes to the model separately and invalidate them when they actually happen. This is a large change that requires some design though, so I'm not sure if we should proceed with this first.

lejolly:
Yup, this is a short-term fix since the model isn't built to account for pending updates. I guess it would make more sense for the model/logic to be involved with the ability to undo rather than the `NotificationController`, that way we can properly keep track of actions and their inverses. Such a change would also solve other issues like #1008. 

I would push for putting this fix out first as it's not that hard to reproduce the bug?

damithc:
I'm fine with a temporary fix but we should also work toward a more robust undo mechanism.

dariusf:
Okay then.

By the way, the coverage drop seems to be from that line I was talking about [here](https://github.com/HubTurbo/HubTurbo/pull/1018#discussion-diff-41219954). Somehow it isn't always covered.

src/main/java/backend/Logic.java
```diff
@@ -68,22 +68,26 @@ public Logic(UIManager uiManager, Preferences prefs, boolean isTestMode, boolean
...
-        Futures.sequence(models.toModels().stream()
-            .map(repoIO::updateModel)
-            .collect(Collectors.toList()))
-                .thenApply(models::replace)
-                .thenRun(this::updateUI)
-                .thenCompose(n -> getRateLimitResetTime())
-                .thenApply(this::updateRemainingRate)
-                .exceptionally(Futures::log);
+    public void refresh(boolean isNotificationPaneShowing) {
+        if (!isNotificationPaneShowing) {
```
dariusf: We could add a TODO comment here, so we know to revert it in future. Also the simpler branch can come first with a guard clause.
```java
if (isNotificationPaneShowing) {
    // log
    return;
}
// rest
```

src/main/java/ui/MenuControl.java
```diff
@@ -310,7 +303,15 @@ private MenuItem createRefreshMenuItem() {
         MenuItem refreshMenuItem = new MenuItem("Refresh");
         refreshMenuItem.setOnAction((e) -> {
             logger.info("Menu: View > Refresh");
-            ui.logic.refresh();
+            if (ui.isNotificationPaneShowing()) {
+                // we trigger the notification timeout action first before refreshing
+                Platform.runLater(() -> {
```
dariusf: Is this needed? The menu item will always be clicked/navigated to from the UI thread.

dariusf:
Okay, looks good.

PR #1040, opened 2015-10-09
Some unstable tests don't work #1039

### LINKED ISSUE #1039: Some unstable tests don't work
Tests which require login don't work because the necessary resources aren't on the runtime classpath.

dariusf:
Fixes #1039

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -85,6 +85,9 @@ allprojects {
             java.srcDir 'src/test/java/unstable'
             compileClasspath += main.output + test.output
             runtimeClasspath += main.output + test.output
+
+            ext.guiTestResources = "${sourceSets.test.resources.getSrcDirs()[0]}/guitests"
+            ext.unstableTestResources = "${sourceSets.test.output.resourcesDir}/unstable"
         }
     }
 
@@ -94,8 +97,8 @@ allprojects {
     }
 
     task generateUnstableTestResources(type: Copy) {
-        from 'src/test/resources/guitests'
-        into 'build/resources/test/unstable'
+        from sourceSets.unstableTests.ext.guiTestResources
+        into sourceSets.unstableTests.ext.unstableTestResources
     }
 
     task unstableTests(type: Test) {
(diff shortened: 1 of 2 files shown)


dariusf:
The problem this addresses is that the unstable tests need all the resources in `src/test/resources/guitest` when they run. Now the resources are copied to `build`; the less DRY alternative I could get to work is having a copy of all the resources in`src/test/resources/unstable`.

@lejolly do you know if there Is a more abstract way of specifying the paths, or an overall cleaner approach?

lejolly:
maybe defining `sourceSets.unstableTests.output.resourcesDir`?

on a somewhat related note, are we able to share resources between the main classes and test classes (for launching via the IDE and not through gradle)

dariusf:
Like this?

The right way to do things would be to put the main resource directories on the test classpath, but I haven't figured out how to do it without breaking other tasks yet. An interim solution would be to copy like we do here.

build.gradle
```diff
@@ -85,6 +85,9 @@ allprojects {
             java.srcDir 'src/test/java/unstable'
             compileClasspath += main.output + test.output
             runtimeClasspath += main.output + test.output
+
+            ext.guiTestResources = "${sourceSets.test.resources.getSrcDirs()[0]}/guitests"
+            ext.unstableTestResources = "${sourceSets.test.output.resourcesDir}/unstable"
```
lejolly: yeah something like this, although i wonder if something like this would work?

```
sourceSets {
        unstableTests {
            ...
            output.resourcesDir = "${sourceSets.test.output.resourcesDir}"
        }
    }
```

dariusf: Source sets are assumed to be at the top level, so `output.resourcesDir` refers to `build/resources/unstableTests`, which we aren't using. We have this odd nested source set layout, where `unstableTests` is a subset of `test`.

I tried changing it to use one source set per type of test, but then that interfered with the coverage task (we'd now have to merge multiple .exec files and would have to merge them).

lejolly:
ok looks good, can finally successfully run them locally.

PR #1044, opened 2015-10-10
Repo list is not updated when a repo is added #967

### LINKED ISSUE #967: Repo list is not updated when a repo is added
It seems the repo list (given under `Repos`) menu is not updated when a new repo is downloaded by typing the repo in a filter. The same goes for the repo dropdown.

HansNewbie:
Fixes #967

### DIFF
--- a/src/main/java/ui/MenuControl.java
+++ b/src/main/java/ui/MenuControl.java
@@ -353,6 +353,22 @@ public class MenuControl extends MenuBar {
             removeItem.setOnAction(e1 -> onRepoRemove(repoId));
             remove.getItems().add(removeItem);
         }
+
+        remove.getItems().add(new SeparatorMenuItem());
+        MenuItem nonRemovableMsg = new MenuItem("[Non-Removable - In Use]");
+        nonRemovableMsg.setDisable(true);
+        remove.getItems().add(nonRemovableMsg);
+
+        // Supposedly, we would like the menu not to close when the disabled MenuItem-s
+        // below are clicked. But this is a JDK bug; we can use CustomMenuItem.setHideOnClick(false)
+        // if we want to. The bug is that it only works for ContextMenu and not Menu (which
+        // we are using).
+        for (String usedRepoId : currentlyUsedRepos) {
+            MenuItem disabledRemoveItem = new MenuItem(usedRepoId);
+            disabledRemoveItem.setDisable(true);
+            remove.getItems().add(disabledRemoveItem);
+        }
+
     }
 
     private void onRepoRemove(String repoId) {
--- a/src/test/java/unstable/RepositorySelectorTest.java
+++ b/src/test/java/unstable/RepositorySelectorTest.java
@@ -100,16 +100,39 @@ public class RepositorySelectorTest extends UITest {
         assertEquals(4, comboBox.getItems().size());
         assertEquals("dummy4/dummy4", primaryRepo);
 
-        // we check if delete repo works
+        // we check if deleting used repo does not remove it
         click("Repos");
         push(KeyCode.DOWN);
         push(KeyCode.RIGHT);
         push(KeyCode.DOWN);
+        push(KeyCode.DOWN);
+        push(KeyCode.DOWN); // first used repo
+        push(KeyCode.ENTER);
+        PlatformEx.waitOnFxThread();
+        assertEquals(4, comboBox.getItems().size());
+
+        // we check if delete repo works
+        click("Boards"); // trigger clicking "Repos"
+        click("Repos");
+        push(KeyCode.DOWN);
+        push(KeyCode.RIGHT);
+        push(KeyCode.ENTER);
+        PlatformEx.waitOnFxThread();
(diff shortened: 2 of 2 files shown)


src/main/java/ui/MenuControl.java
```diff
@@ -353,6 +353,17 @@ private void updateRepoRemoveList(Menu remove) {
             removeItem.setOnAction(e1 -> onRepoRemove(repoId));
             remove.getItems().add(removeItem);
         }
+
+        // Supposedly, we would like the menu not to close when the disabled MenuItem-s
+        // below are clicked. But this is a JDK bug; we can use CustomMenuItem.setHideOnClick(false)
+        // if we want to. The bug is that it only works for ContextMenu and not Menu (which
+        // we are using).
+        for (String usedRepoId : currentlyUsedRepos) {
+            MenuItem disabledRemoveItem = new MenuItem(usedRepoId + " (in use; non-removable)");
```
lejolly: I don't think the semicolon there looks nice =P

Maybe something shorter like just `(In Use)`?

HansNewbie: I agree, but do you think it will be obvious to user that it is not removable? I mean the visual cue is there but I am not sure if people get it.

lejolly: I should think so? One other option would be to add the `(In Use)` but not disable it, so that when the user clicks on it, it popups up a dialog telling the user why it can't be removed? But that seems even more convoluted. 

Ah, what if, at the very top, you have a disabled menu item that says that repos in use cannot be deleted?

HansNewbie: Would it be nice if I add a separator, which just below it says `(Non-removable)`, and each items behind has `(In Use)`?

lejolly: Ah that would work too.

src/test/java/unstable/RepositorySelectorTest.java
```diff
@@ -100,16 +100,39 @@ public void repositorySelectorTest() {
...
+        push(KeyCode.RIGHT);
+        push(KeyCode.DOWN);
+        push(KeyCode.DOWN);
+        push(KeyCode.DOWN); // first used repo
+        push(KeyCode.ENTER);
+        PlatformEx.waitOnFxThread();
+        assertEquals(4, comboBox.getItems().size());
+
         // we check if delete repo works
+        click("Boards"); // trigger clicking "Repos"
```
HansNewbie: This and below is to trigger the next simulations. Without this, some of them are just skipped. It went undetected since it passes the `assert`. Seems that the framework has a problem with using the same node again (this is also evident in `KeyboardShortcutTests`).

lejolly: So does it work locally?

HansNewbie: With this yes, it works locally. Previously, the test will skip some
commands except the assertion.

HansNewbie:
This has passed all tests locally for more than 3 times.

src/test/java/unstable/RepositorySelectorTest.java
lejolly: Ah ok.

src/main/java/ui/MenuControl.java
```diff
@@ -353,6 +354,22 @@ private void updateRepoRemoveList(Menu remove) {
             removeItem.setOnAction(e1 -> onRepoRemove(repoId));
             remove.getItems().add(removeItem);
         }
+
+        remove.getItems().add(new SeparatorMenuItem());
+        MenuItem nonRemovableMsg = new MenuItem("[Non-Removable]");
```
lejolly: I was wondering if we should just put the `In Use` here so that we don't have to repeat it below?

HansNewbie: so put like `[In User - Non-Removable]` at the top? That's fine with me, too.

src/main/java/ui/MenuControl.java
```diff
@@ -353,6 +354,22 @@ private void updateRepoRemoveList(Menu remove) {
             removeItem.setOnAction(e1 -> onRepoRemove(repoId));
             remove.getItems().add(removeItem);
         }
+
+        remove.getItems().add(new SeparatorMenuItem());
+        MenuItem nonRemovableMsg = new MenuItem("[Non-Removable - In Use]");
```
HansNewbie: This good?

lejolly: Yup.

src/main/java/ui/MenuControl.java
```diff
@@ -18,6 +18,7 @@
 import ui.issuepanel.PanelControl;
 import util.events.*;
 
+import javax.swing.*;
```
lejolly: is this needed?

HansNewbie: Oops. Will remove.

lejolly:
ok if it passes both locally and on travis, it's good to go.

Issue #933, opened 2015-09-02
Improve or change panel element selection in tests


HyungJon:
Node IDs are unreliable and immutable.

Alternative solution proposed today: add ```get``` method for each element that allows tests to select elements through ```PanelControl``` -> ```FilterPanel``` at index -> element.

This will need some way for tests to access ```PanelControl``` or ```UI``` directly.

dariusf:
[This](https://github.com/HubTurbo/HubTurbo/issues/685#issuecomment-111802527) might help, though the approach is rather low-level and should be discussed further.

HyungJon:
Or perhaps setting an ID for the UI itself?

dariusf:
What do you mean?

My earlier post was a way to get a reference to a node by traversing the UI tree directly. It's low-level and breaks abstraction (i.e. it's fragile), but it sidesteps the problem of having to assign and maintain a dynamic id.

HyungJon:
Working on this now. Can I clarify:
# How should the ```PanelControl``` be passed as a parameter? Also because If we can access it from tests, we could just use ```getPanel``` in ```PanelControl``` to get the panel directly.
# Would it be okay to add getters in ```FilterPanel``` to return elements (i.e. filterTextField, closeButton, etc)?

dariusf:
1. As a parameter to?
2. I guess that would be fine.

HyungJon:
For the ```UITraversal``` class. I'm going to pass ```PanelControl``` since this one is about panels.

PR #1047, opened 2015-10-11
Filter and sort issues in Logic #1004

### LINKED ISSUE #1004: Filter and sort issues in Logic
Currently, upon a program refresh, `GUIController` requests a refresh in `Logic`, which retrieves data from GitHub and passes it to the `GUIController` without metadata. `GUIController` then filters and sorts issues to retrieve metadata for, which it then sends to `Logic` again, which retrieves more data from GitHub and passes it to the `GUIController`, upon which it will filter and sort the issues again, and then finally display the issues to the user.

Filtering and sorting should be moved to `Logic` instead. In this manner, `GUIController` does not have to worry about managing the flow of execution of the program.

There are multiple implementation design concerns with migrating the filtering and sorting of issues in `Logic`, which are followed by how they are to be dealt with below:
* As `GUIController` does not do filtering and sorting any more, `Logic` should pass the issues to be displayed, for each panel, in the order in which they are to be displayed.
 * Solution: `Logic` attaches issues to be displayed with a filter text string. This works under the assumption that panels with the same filter text will be displayed exactly the same (which is true for the forseeable future)
* Each `ListPanelCard` needs to know if its panel's filter contains an `updated` qualifier so that it can decide when to display metadata for: 
 * Solution: `Logic` attaches a flag signifying whether a filter text string has an `updated` qualifier together with the filter text string, too.
* Currently, <kbd>F5</kbd> / `TickingTimer` / bView refresh event call `ui.logic.refresh()` directly. However, `Logic` needs to get the filter text from the panels.
 * Solution: ~~<kbd>F5</kbd> / `TickingTimer` / bView refresh event will enter the program from `GUIController` instead, which will pass the filter Strings to `Logic` as method call arguments.~~ Expose a method in `GUIController` for the `uiManager` to retrieve filters from panels.

acruis:
Fixes #1004 

My apologies for the huge initial commit, could not think of a way of breaking it up into small pieces.

The next step would be to remove the dependency on `isSortableByNonSelfUpdates` from Qualifier, which is now straightforward as we no longer sort for the first filter.


src/main/java/backend/Logic.java
```diff
@@ -247,13 +209,204 @@ public Model getRepo(String repoId) {
...
+        Map<String, List<TurboIssue>> toUpdate = tallyMetadataUpdate(filterTexts);
+
+        if (toUpdate.size() > 0) {
+            // If there are issues requiring metadata update, we dispatch the metadata requests...
+            ArrayList<CompletableFuture<Boolean>> metadataRetrievalTasks = new ArrayList<>();
+            toUpdate.forEach((repoId, issues) -> {
+                metadataRetrievalTasks.add(getIssueMetadata(repoId, issues));
+            });
+            // ...and then wait for all of them to complete.
+            CompletableFuture<Void> allCompletables = CompletableFuture.allOf(
```
dariusf: This can be replaced with `Futures.sequence`. It solves the more general problem of turning a `List<Future<T>>` into a `Future<List<T>>`.

Each metadata retrieval task returns a boolean, which we probably shouldn't ignore also. Can probably fold it and do something based on whether any failed?

An alternative is to change the return type, which might give us more info to act on, such as which failed.

src/main/java/backend/Logic.java
```diff
@@ -247,13 +209,204 @@ public Model getRepo(String repoId) {
...
+            updateUI(filterAndSort(filterTexts));
+        }
+    }
+
+    /**
+     * Given a list of filter expressions, open all repositories necessary for processing the filter expressions.
+     *
+     * @param filterTexts Filter expressions to process.
+     */
+    private void openRepositoriesInFilters(List<FilterExpression> filterTexts) {
```
dariusf: The variable name might give the impression that the method expects a string. `filterExprs` maybe?

src/main/java/backend/Logic.java
```diff
@@ -247,13 +209,204 @@ public Model getRepo(String repoId) {
...
+
+        filterTexts.forEach(filterText -> {
+            filterText.find(Qualifier::isMetaQualifier).stream()
+                    .filter(metaQualifier ->
+                            metaQualifier.getName().equals(Qualifier.REPO) && metaQualifier.getContent().isPresent())
+                    .forEach(repoQualifier ->
+                            reposToOpen.add(repoQualifier.getContent().get()));
+        });
+
+        reposToOpen.forEach(this::openRepositoryFromFilter);
```
dariusf: Should be equivalent to

```java
filterTexts.stream()
    .flatMap(f -> Qualifier.getContentOfMetaQualifier(f, Qualifier.REPO).stream())
    .distinct()
    .forEach(this::openRepositoryFromFilter);
```
Maybe even define Qualifier::isRepoQualifier, since we use this so much.

src/main/java/backend/Logic.java
```diff
@@ -247,13 +209,204 @@ public Model getRepo(String repoId) {
...
+     *
+     * @param filterTexts Filter expressions to process for metadata requests.
+     * @return Repo IDs and the corresponding issues in the repo requiring a metadata update.
+     */
+    private Map<String, List<TurboIssue>> tallyMetadataUpdate(List<FilterExpression> filterTexts) {
+        HashMap<String, List<TurboIssue>> toUpdate = new HashMap<>();
+
+        List<TurboIssue> allModelIssues = models.getIssues();
+
+        filterTexts.stream().filter(Qualifier::hasUpdatedQualifier).forEach(filterText -> {
```
dariusf: Do we need to preprocess here to convert `updated-self` (and any other qualifier which requires metadata to work correctly) into `update`?

src/main/java/backend/Logic.java
```diff
@@ -247,13 +209,204 @@ public Model getRepo(String repoId) {
...
+                    // If eixsts, just add the issue
+                    issuesInRepo.add(issue);
+                } else {
+                    // If not exists, create first then add
+                    issuesInRepo = new ArrayList<>();
+                    issuesInRepo.add(issue);
+                    toUpdate.put(issue.getRepoId(), issuesInRepo);
+                }
+            });
+        });
```
dariusf: Can be written as

```java
return filterTexts.stream()
    .filter(Qualifier::hasUpdatedQualifier)
    .flatMap(updatedQ -> allModelIssues.stream()
        .filter(issue -> Qualifier.process(models, updatedQ, issue)))
    .collect(Collectors.groupingBy(TurboIssue::getRepoId));
```

src/main/java/backend/Logic.java
```diff
@@ -247,13 +209,204 @@ public Model getRepo(String repoId) {
...
+     * update timings.
+     *
+     * @param repoId The repository containing issues to retrieve metadata for.
+     * @param issues Issues sharing the same repository requiring a metadata update.
+     * @return True if metadata retrieval was a success, false otherwise.
+     */
+    public CompletableFuture<Boolean> getIssueMetadata(String repoId, List<TurboIssue> issues) {
+        String message = "Getting metadata for " + repoId + "...";
+        logger.info("Getting metadata for issues "
+                + issues.stream().map(TurboIssue::getId).map(Object::toString).collect(Collectors.joining(", ")));
```
dariusf: Is this needed? I think List::toString invokes toString on its contents.

src/main/java/backend/Logic.java
```diff
@@ -247,13 +209,204 @@ public Model getRepo(String repoId) {
...
+        UI.status.displayMessage(message);
+
+        String currentUser = prefs.getLastLoginUsername();
+
+        return repoIO.getIssueMetadata(repoId, issues).thenApply(this::processUpdates)
+                .thenApply(metadata -> {
+                    String updatedMessage = "Received metadata from " + repoId + "!";
+                    UI.status.displayMessage(updatedMessage);
+                    models.insertMetadata(repoId, metadata, currentUser);
+                    return true;
```
dariusf: We should pull this out into its own method (SLAP).

src/main/java/backend/Logic.java
```diff
@@ -247,13 +209,204 @@ public Model getRepo(String repoId) {
...
+                .exceptionally(withResult(false));
+    }
+
+    /**
+     * Filters and sorts issues within the model according to the given filter expressions.
+     *
+     * @param filterTexts Filter expressions to process.
+     * @return Filter expressions and their corresponding issues after filtering and sorting.
+     */
+    private Map<FilterExpression, ImmutablePair<List<TurboIssue>, Boolean>>
```
dariusf: Noticed that this boolean value is propagated quite a long way to ListPanel to determine if an issue should show non-self comments. Is this required? Could we check in ListPanel to see if the filter has an updated qualifier, then decide there?

If the boolean value is needed, probably should factor this out into a class. I'm starting to think that all the ImmutablePairs and ImmutableTriples were a mistake... :worried:

Can probably refactor further with groupingBy depending on above.

src/main/java/backend/Logic.java
```diff
@@ -247,13 +209,204 @@ public Model getRepo(String repoId) {
...
+        Map<FilterExpression, ImmutablePair<List<TurboIssue>, Boolean>> filteredAndSorted = new HashMap<>();
+
+        List<TurboIssue> allModelIssues = models.getIssues();
+
+        filterTexts.forEach(filterText -> {
+            ImmutablePair<List<TurboIssue>, Boolean> filterAndSortedExpression = filteredAndSorted.get(filterText);
+
+            if (filterAndSortedExpression == null) { // If it already exists, no need to filter anymore
+                List<Qualifier> metaQualifiers = filterText.find(Qualifier::isMetaQualifier);
+                boolean hasUpdatedQualifier = Qualifier.hasUpdatedQualifier(metaQualifiers);
```
dariusf: hasUpdatedQualifier has an overload for dealing with FilterExpressions, so we don't need the `find`.

dariusf:
Sorry, took me a while to get through this. Should we put all the new stuff in Logic into another class, like UpdateController or whatever we discussed?

src/main/java/backend/Logic.java
acruis: Alright, I was thinking of a short name for it. `filterExprs` is great.

src/main/java/backend/Logic.java
acruis: Yup, probably is. The difference is that now there's no square bracket around the issues? lol

src/main/java/backend/Logic.java
acruis: Yes, we should.

src/main/java/backend/Logic.java
acruis: We'd be doing filtering in both the GUI and the Logic then. But I'm OK with that, actually. Are you OK with that?

src/main/java/backend/Logic.java
acruis: `metaQualifiers` is used in the `determineComparator` call. Maybe I should just pass the filter expression itself to `determineComparator`?

src/main/java/backend/Logic.java
acruis: Ah yup, I didn't think it'd be necessary to use `Futures.sequence` since we aren't currently using the return values from `getIssueMetadata`. But it makes sense to prep it for a future PR.

src/main/java/backend/Logic.java
acruis: `distinct` and `flatMap` look like what I was looking for, yup.

src/main/java/backend/Logic.java
acruis: Yup, that's the idea. In a PR after this, though?

src/main/java/backend/Logic.java
acruis: I was looking for this all my life.

acruis:
Would the new class hold `Logic` or `RepoIO` as a field? Since the new class would have to invoke `getIssueMetadata`.

src/main/java/backend/Logic.java
dariusf: It's a bit more SLAP also. Could we do the simplest thing possible, like put something in the log?

src/main/java/backend/Logic.java
dariusf: Okay, probably should be after #1050 since some useful methods are being implemented there.

src/main/java/backend/Logic.java
dariusf: Yes. Filters are still changing, so I'm not sure about our specific needs yet. There's currently quite a bit of processing them in the GUI (by calling Qualifier methods).

src/main/java/backend/Logic.java
dariusf: That could work also.

dariusf:
Hmm, so Logic would have to do `new UpdateController(this)`? I think having everything in Logic is still okay, just depends on whether you think there's meaningful distinction between what you've implemented here and whatever else is in there already.

src/main/java/backend/Logic.java
acruis: Yup, I'll put that in.

acruis:
I think this is quite a bit of code to add to Logic, and I'm not sure whether it's good to place everything API-related in one place. I think it's still fine now, but it would be nice to have more organization to Logic.

Is there a problem with calling `new UpdateController(this)` actually, aside from aesthetic concerns?

dariusf:
Nope, no problem.

src/main/java/backend/Logic.java
acruis: I think the reason why this code was here in the first place is that `TurboIssue::toString` prints the issue ID, as well as the issue title (which can be long). In a big array, let's say in an `updated:24` filter, we get a long text like:

`[#1061  Improve or change panel element selection in tests #933 , #1060 Model is deleted when repo letter casing is wrong, #1055 milestone missing from some issues #1054, #1054 milestone missing from some issues]`

Which might clutter the log without adding useful information, compared to just printing the issueIDs like the original code.

dariusf: Right. We can leave it then, if the list can get long. Or should we make TurboIssue::toString more succinct?

acruis: I think it might be a good idea to make it a combination of `[issueId, repoId]` instead of `[issueId, issueTitle`. What do you think?

dariusf: Sounds good.

dariusf:
Hmm, could you update to the latest master? In case there are upstream changes which haven't been accounted for.

dariusf:
Seems like this is consistently failing due to panel creation being triggered many times. Would pushKey help?

acruis:
Ready for review :)

dariusf:
Looks good. :+1:

PR #1048, opened 2015-10-11
Implement sort by milestone

### LINKED ISSUE #1027: User can sort issues by milestone
`sort:milestone`

ndt93:
Fixes #1027

### DIFF
--- a/src/main/java/filter/expression/Qualifier.java
+++ b/src/main/java/filter/expression/Qualifier.java
@@ -1,14 +1,5 @@
 package filter.expression;
 
-import java.time.LocalDate;
-import java.time.LocalDateTime;
-import java.time.temporal.ChronoUnit;
-import java.util.*;
-import java.util.function.BiConsumer;
-import java.util.function.Predicate;
-import java.util.stream.Collectors;
-
-import util.Utility;
 import backend.interfaces.IModel;
 import backend.resource.TurboIssue;
 import backend.resource.TurboLabel;
@@ -16,6 +7,15 @@ import backend.resource.TurboMilestone;
 import backend.resource.TurboUser;
 import filter.MetaQualifierInfo;
 import filter.QualifierApplicationException;
+import util.Utility;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.temporal.ChronoUnit;
+import java.util.*;
+import java.util.function.BiConsumer;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 public class Qualifier implements FilterExpression {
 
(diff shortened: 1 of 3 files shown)


ndt93:
@dariusf Can help review this?

damithc:
ongoing or toReview?

ndt93:
Oops, forget to change label.

src/test/java/tests/SortQualifierTests.java
```diff
@@ -1,27 +1,74 @@
...
+            }
+
+            Integer id = issue.getMilestone().get();
+
+            for (TurboMilestone milestone : getMilestones()) {
+                if (milestone.getId() == id) {
+                    return Optional.of(milestone);
+                }
+            }
+            return Optional.empty();
```
dariusf: I agree that we need a less ad-hoc way of setting up test data, but this method feels a bit like reinvention. Why not add it to an existing MultiModel instance? We can come up with something better to replace it (factories?) later.

src/test/java/tests/SortQualifierTests.java
```diff
@@ -177,6 +242,48 @@ public void testSortQualifer5() {
...
+        issues.add(issue6);
+        issues.add(issue7);
+
+        Comparator<TurboIssue> comparator = getComparatorForSortQualifier("sort:milestone,id");
+
+        Collections.shuffle(issues);
+        Collections.sort(issues, comparator);
+
+        assertEquals("[#3 Issue3, #2 Issue2, #1 Issue1, #4 Issue4, #7 Issue7, #5 Issue5, #6 Issue6]",
+                      issues.toString());
```
dariusf: Could we compare ids here instead (`issues.map(getId)` against `Arrays.asList(1, 2, 3)`)? So we don't rely on TurboIssue::toString.

src/test/java/tests/SortQualifierTests.java
ndt93: Not sure if I understood correctly but here I'm overriding an existing `getMilestoneOfIssue` method of `MultiModel` which `getSortComparator` makes use of. The existing method is more complicated and involve dependencies on Models, TurboIssues, matching repo id etc. which we don't want to simulate since we are not concerned with testing the method itself.

dariusf: Hmm, okay. I guess we should do this for the other comparators as well, eventually. We could maybe also use Mockito or something to reduce the boilerplate.

ndt93: I just take a look at Mockito, looks pretty neat and could be exactly what we need.

ndt93:
@dariusf I've changed all the string comparisons to integers list comparisons. You can check this again.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -460,6 +460,33 @@ public static boolean hasUpdatedQualifier(FilterExpression expr) {
...
+                        } else if (!aDueDate.isPresent()) {
+                            return 1;
+                        } else if (!bDueDate.isPresent()) {
+                            return -1;
+                        } else {
+                            return -aDueDate.get().compareTo(bDueDate.get());
+                        }
+                    }
+                };
+                break;
```
damithc: Generally, I don't like long methods with deep nesting. Not sure if there is a way to make this more SLAP though.

ndt93: The last outer else block can be flattened with @HansNewbie's method in #1050 so either of us can do it depending on which one gets merged first.

dariusf:
Okay, I think this looks fine. So are we going to do anything about the long comparator method here? Maybe factor it out?

HansNewbie:
I can replace it with my comparator if this PR is merged.

dariusf:
Okay, sure.

PR #1050, opened 2015-10-12
User can use aliases to refer to milestones #1019

### LINKED ISSUE #1019: User can use aliases to refer to milestones
... so that filters containing milestones do not have to change after each milestone.

e.g. `current` `previous` `next` `next+1` and so on. These aliases should be dynamically resolved based on the list of milestones.

HansNewbie:
Fixes #1019

### DIFF
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -148,7 +148,7 @@ Matches all issues with a body (or description) containing the given string.
 
 *Expects a string*
 
-Matches all issues associated with any milestones whose names contain the given string.
+Matches all issues associated with any milestones whose names contain the given string. `current` or `curr` can be used to refer to an open milestone with earliest due date, or an overdue open milestone with open issues if any. `curr-2`, `curr-1`, `curr+1`, `curr+2`, etc. (no space before or after `-`/`+`) can then be used to refer to milestones before or after `current`, sorted by due date.
 
 ### label
 
--- a/src/main/java/filter/expression/Qualifier.java
+++ b/src/main/java/filter/expression/Qualifier.java
@@ -1,21 +1,22 @@
 package filter.expression;
 
-import backend.interfaces.IModel;
-import backend.resource.TurboIssue;
-import backend.resource.TurboLabel;
-import backend.resource.TurboMilestone;
-import backend.resource.TurboUser;
-import filter.MetaQualifierInfo;
-import filter.QualifierApplicationException;
-import util.Utility;
-
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.temporal.ChronoUnit;
 import java.util.*;
 import java.util.function.BiConsumer;
+import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+
+import backend.resource.*;
+import util.Utility;
+import backend.interfaces.IModel;
+import filter.MetaQualifierInfo;
+import filter.QualifierApplicationException;
+
 
 public class Qualifier implements FilterExpression {
 
@@ -93,6 +94,47 @@ public class Qualifier implements FilterExpression {
         this.sortKeys = new ArrayList<>(keys);
     }
 
+    public static FilterExpression replaceMilestoneAliases(IModel model, FilterExpression expr) {
+        List<String> repoIds = getContentOfMetaQualifier(expr, REPO).stream()
+                .map(repoId -> repoId.toLowerCase())
+                .collect(Collectors.toList());
+
+        if (repoIds.isEmpty()) {
+            repoIds.add(model.getDefaultRepo().toLowerCase());
+        }
+
+        List<TurboMilestone> allMilestones = model.getMilestones().stream()
+                .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
+                .sorted((a, b) -> {
+                    if (!a.getDueDate().isPresent() && !b.getDueDate().isPresent()) {
+                        return 0;
+                    } else if (!a.getDueDate().isPresent()) {
+                        return 1;
+                    } else if (!b.getDueDate().isPresent()) {
+                        return -1;
+                    } else {
+                        return getMilestoneDueDateComparator().compare(a, b);
+                    }
+                })
+                .collect(Collectors.toList());
+
+        Optional<Integer> currentMilestoneIndex = getCurrentMilestoneIndex(allMilestones);
+
+        if (!currentMilestoneIndex.isPresent()) {
+            return expr;
+        }
+
+        expr = expr.map(q -> {
+            if (Qualifier.isMilestoneQualifier(q)) {
+                return q.convertMilestoneAliasQualifier(allMilestones, currentMilestoneIndex.get());
+            } else {
+                return q;
+            }
+        });
+
+        return expr;
+    }
+
     /**
      * Helper function for testing a filter expression against an issue.
      * Ensures that meta-qualifiers are taken care of.
@@ -106,7 +148,7 @@ public class Qualifier implements FilterExpression {
         boolean containsRepoQualifier = metaQualifiers.stream()
                 .map(Qualifier::getName)
                 .collect(Collectors.toList())
-            .contains("repo");
+                .contains("repo");
 
         if (!containsRepoQualifier) {
             exprWithNormalQualifiers = new Conjunction(
@@ -115,6 +157,29 @@ public class Qualifier implements FilterExpression {
         }
 
         return exprWithNormalQualifiers.isSatisfiedBy(model, issue, new MetaQualifierInfo(metaQualifiers));
+    }
+
+    private static Optional<Integer> getCurrentMilestoneIndex(List<TurboMilestone> allMilestones) {
+        if (allMilestones.isEmpty()) {
+            return Optional.empty();
+        }
+
+        int currentIndex = 0;
+
+        for (TurboMilestone checker : allMilestones) {
+            boolean overdue = checker.getDueDate().isPresent() &&
+                    checker.getDueDate().get().isBefore(LocalDate.now());
+            boolean relevant = !(overdue && (checker.getOpenIssues() == 0));
+
+            if (checker.isOpen() && relevant) {
+                return Optional.of(currentIndex);
+            }
+
+            currentIndex++;
+        }
+
+        // if no open milestone, set current as last milestone
+        return Optional.of(allMilestones.size() - 1);
     }
 
     public static void processMetaQualifierEffects(FilterExpression expr,
@@ -287,6 +352,11 @@ public class Qualifier implements FilterExpression {
         }
     }
 
+    @Override
+    public FilterExpression map(Function<Qualifier, Qualifier> func) {
+        return func.apply(this);
+    }
+
     /**
      * This method is used to serialise qualifiers. Thus whatever form returned
      * should be syntactically valid.
@@ -378,6 +448,15 @@ public class Qualifier implements FilterExpression {
             return true;
         default:
             return isUpdatedQualifier(q);
+        }
+    }
+
+    public static boolean isMilestoneQualifier(Qualifier q) {
+        switch (q.getName()) {
+            case "milestone":
+                return true;
+            default:
+                return false;
         }
     }
 
@@ -482,7 +561,8 @@ public class Qualifier implements FilterExpression {
                         } else if (!bDueDate.isPresent()) {
                             return -1;
                         } else {
-                            return -aDueDate.get().compareTo(bDueDate.get());
+                            return -(getMilestoneDueDateComparator()
+                                    .compare(aMilestone.get(), bMilestone.get()));
                         }
                     }
                 };
@@ -796,7 +876,7 @@ public class Qualifier implements FilterExpression {
 
         // Find milestones containing partial title
         List<TurboMilestone> milestones = model.getMilestones().stream()
-            .filter(m -> m.getTitle().toLowerCase().contains(content.get().toLowerCase()))
+                .filter(m -> m.getTitle().toLowerCase().contains(content.get().toLowerCase()))
             .collect(Collectors.toList());
 
         if (milestones.isEmpty()) {
@@ -829,7 +909,7 @@ public class Qualifier implements FilterExpression {
 
         // Find labels containing the label name
         List<TurboLabel> labels = model.getLabels().stream()
-           .filter(l -> l.getActualName().toLowerCase().contains(content.get().toLowerCase()))
+                .filter(l -> l.getActualName().toLowerCase().contains(content.get().toLowerCase()))
             .collect(Collectors.toList());
 
         if (labels.isEmpty()) {
@@ -924,5 +1004,56 @@ public class Qualifier implements FilterExpression {
 
     public String getName() {
         return name;
+    }
+
+    private Qualifier convertMilestoneAliasQualifier(List<TurboMilestone> allMilestones, int currentIndex) {
+        if (!content.isPresent()) {
+            return Qualifier.EMPTY;
+        }
+
+        String contents = content.get();
+        String alias = contents;
+
+        Pattern aliasPattern = Pattern.compile("(curr(?:ent)?)(?:(\\+|-)(\\d+))?$");
+        Matcher aliasMatcher = aliasPattern.matcher(alias);
+        int offset = 0;
+
+        if (!aliasMatcher.find()) {
+            return this;
+        }
+
+        if (aliasMatcher.group(2) != null && aliasMatcher.group(3) != null) {
+            offset = Integer.parseInt(aliasMatcher.group(3));
+            if (aliasMatcher.group(2).equals("+")) {
+                currentIndex += offset;
+            } else {
+                currentIndex -= offset;
+            }
+        }
+
+        // clamp
+        if (currentIndex >= allMilestones.size()) {
+            currentIndex = allMilestones.size() - 1;
+        } else if (currentIndex < 0) {
+            currentIndex = 0;
+        }
+
+        contents = allMilestones.get(currentIndex).getTitle().toLowerCase();
+
+        return new Qualifier(name, contents);
+    }
+
+    /**
+     * Condition: milestone must have due dates
+     */
+    public static Comparator<TurboMilestone> getMilestoneDueDateComparator() {
+        return (a, b) -> {
+            assert(a.getDueDate().isPresent());
+            assert(b.getDueDate().isPresent());
+            LocalDate aDueDate = a.getDueDate().get();
+            LocalDate bDueDate = b.getDueDate().get();
+
+            return (aDueDate.compareTo(bDueDate));
+        };
     }
 }
--- a/src/test/java/tests/FilterEvalTests.java
+++ b/src/test/java/tests/FilterEvalTests.java
@@ -3,11 +3,13 @@ package tests;
 import static junit.framework.TestCase.assertFalse;
 import static org.junit.Assert.*;
 
+import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Optional;
 
+import filter.expression.FilterExpression;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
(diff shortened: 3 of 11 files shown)


docs/filters.md
```diff
@@ -148,7 +148,7 @@ Matches all issues with a body (or description) containing the given string.
 
 *Expects a string*
 
-Matches all issues associated with any milestones whose names contain the given string.
+Matches all issues associated with any milestones whose names contain the given string. `current` or `curr` can be used to refer to an open milestone with earliest due date, or an overdue open milestone with open issues if any. `prev`, `prev-1`, `prev-2`, `next`, `next+1, `next+2`, etc. can be used to refer to milestones before or after `current`, sorted by due date.
```
damithc: Should we use a special symbol to indicate variables? e.g. `$current`?

damithc: BTW, you missed the closing backquote for next+1

src/main/java/backend/resource/TurboMilestoneDueDateComparator.java
```diff
@@ -0,0 +1,27 @@
...
+        } else if (!m2.getDueDate().isPresent()) {
+            return -1;
+        } else {
+            LocalDate m1DueDate = m1.getDueDate().get();
+            LocalDate m2DueDate = m2.getDueDate().get();
+
+            return (m1DueDate.isBefore(m2DueDate)) ? -1 :
+                    (m1DueDate.isEqual(m2DueDate)) ? 0 : 1;
+        }
+    }
```
damithc: Any overlap between this and 'sort by milestone' feature? @ndt93

src/main/java/filter/expression/Qualifier.java
```diff
@@ -717,6 +714,61 @@ private boolean milestoneSatisfies(IModel model, TurboIssue issue) {
...
+        } else if (alias.startsWith("next+")) {
+            int offsetPos = alias.indexOf('+') + 1;
+            int offset = Integer.parseInt(alias.substring(offsetPos).trim());
+            if (currentIndex + 1 + offset < allMilestones.size()) {
+                contents = allMilestones.get(currentIndex + 1 + offset).getTitle();
+            }
+        }
+
+        contents = contents.toLowerCase();
+
```
damithc: Looks like this method is doing more than advertised.

docs/filters.md
HansNewbie: For documentation purpose? I think it may make it confusing to user, like
am I supposed to type `current` or `$current`.

If you are thinking of using it in the filter, I think it's a good idea in
case user has a milestone named `V0.1 previously` and he wants to search
for that `prev` instead of previous milestone.

src/main/java/filter/expression/Qualifier.java
damithc: Use more SLAP. Try to write cohesive (does only one thing) method that are independently testable.

src/main/java/backend/resource/TurboMilestoneDueDateComparator.java
HansNewbie: Yes. @ndt93 I think this is exactly the same comparator, but I extracted it out as a file.

docs/filters.md
damithc: I meant in the filter it self. Currently there is no differentiation between a normal string literal (e.g. 'V0.1' and a variable string (e.g 'current'). @dariusf

src/main/java/filter/expression/Qualifier.java
HansNewbie: Will abstract it into its own method.

docs/filters.md
dariusf: Hmm, it depends on if we want users to the think of the contents of a qualifier as a actual arithmetic expression rather than just a string literal, i.e. if we may have arithmetic operations in future. Otherwise I think just 'current' is fine, since there's less syntax to remember.

src/main/java/backend/resource/TurboMilestoneDueDateComparator.java
dariusf: A method would be fine, which you can refer to with `this::comparator`. But yeah, we can only resolve in favour of one of your implementations.

HansNewbie: Which one is better? As a new file or method? I think method is cleaner
since I put the file in the same folder with `TurboMilestone` right now but
it seems to be out of place.

dariusf: Method, so it's closer to where it's used. By "one of your implementations" I meant yours or @ndt93's.

HansNewbie: Yep, I will follow @ndt93's.

docs/filters.md
damithc: Perhaps variable is the wrong term. I meant it is a keyword and we have no way to show the difference. Still, may be simpler to keep as it is. Certainly easier to type.
@HansNewbie do we allow spaces around `+`?

HansNewbie: I don't allow, to make it consistent with `prev-` since I think the current
regex will read ` - ` as negation.

damithc: That's fine. May be mention that in the documentation?

HansNewbie: Sure.

src/main/java/backend/resource/TurboMilestoneDueDateComparator.java
ndt93: Sorry didn't see this earlier. If yours is merged first then I will replace the similar part with the method. Otherwise, you can help me replace the part.

HansNewbie: @dariusf now that I look at it closer, @ndt93's and my comparator has a difference. His comparator has to check that a milestone exists since it is `Optional` while my comparator deals directly with `TurboMilestone` and not `Optional`. I have extracted mine into a method (instead of a `Comparator` class, so would it be fine to keep both?

docs/filters.md
HansNewbie: @damithc I am adding `current+1`, `current-1` as well for the alias. Is that okay with you? I will update the document accordingly.

src/main/java/backend/resource/TurboMilestoneDueDateComparator.java
dariusf: Unless they have very different semantics, we should keep things as DRY as possible. Maybe his comparator could call yours for the case where both milestones are present?

src/main/java/filter/expression/Qualifier.java
dariusf: You should probably use a regex for this. It would be a lot easier to reason about than all the explicit string processing.

Also, this method will be invoked for every issue, so we probably shouldn't sort over and over in here. One alternative I can think of is to preprocess the filter expression and replace the alias with its concrete equivalent, so it's only done once.

HansNewbie: But even if I make regex, it would still have that many `if` cases. And since each alias has its own kind of 'signature', the regex would just be as many, wouldn't it?

I noticed that this method will be invoked for every issue, but where should preprocessing of the milestones be done? In `Model`/`MultiModel`? That's the only place I can think of since putting it in `Qualifier` would not make it invoked once. In that case, should the milestone comparator be extracted into its own class so that it can be used by sort milestone as well?

On a side note, I might need to change parts of this since right now, it will refer to milestones of all open repos; so if there is a panel using `repo:sth/sth`, that repo's milestones will be listed together instead of only the relevant repo to the filter. I am not sure how to do it yet, though.

src/main/java/backend/resource/TurboMilestoneDueDateComparator.java
HansNewbie: Yes, I think so. Putting link here if needed to be changed later https://github.com/HubTurbo/HubTurbo/pull/1048/files#diff-ee9e6db7c94d86bc0cf7eba36b712932R474

docs/filters.md
damithc: Yes, it is fine to have a alias.
Actually, why not just use `current` only (i.e. not use previous/next)? That will make it simpler.

src/main/java/filter/expression/Qualifier.java
dariusf: How about something like `(curr(?:ent)?|prev(?:ious)?)(?:(\+|-)1)?`? `\1` will tell you the first word, and `\2` will tell you whether the offset is positive or negative, or zero if it's not present.

I guess you would need another method to be used alongside Qualifier::process. Some of it should also be moved out there. The new method could return a MetaQualifierInfo and/or a FilterExpression... or maybe we need two methods.

Find contents of repo qualifiers, then use that to filter the list of milestones? If we have something like `repo:a || repo:b` we will have to get milestones across repos.

On a side note we should clean up the helper methods in Qualifier... Make them higher-level and better at doing things like this.

docs/filters.md
HansNewbie: Sounds good to me.

src/main/java/filter/expression/Qualifier.java
HansNewbie: Okay, will do the regex. I will changed everyhig to just current/curr +/-
so no need for prev or next.

Just to confirm, so should I do the sorting and getting all milestones of a
repo/repos outside Qualifier?

About cleaning up helper methods, which ones are you talking about? Like
the `milestoneAliasConverter`?

HansNewbie: @dariusf in my current implementation, I just replaced the alias with the milestone name. However, this pose a problem when say I replace `current-10

HansNewbie: @dariusf sorry, pressed wrongly. So in my current implementation, I replace the alias with a milestone name and just let it go to normal milestone search. It poses a problem since the milestone search is by `contains`, e.g. `current-10` now returns `V3.1` but it will include issues with milestone `V3.10` as well. Should I ignore this and post this as another issue (or searching milestone) or should alias use be strict match instead of `contains`?

dariusf: Ignore and make an issue I think. We could either get rid of the `contains` matching or add another flag to Qualifier to control strictness.

HansNewbie:
@dariusf I can't think of any way of not sorting the milestones per issue unless we store the list of milestones relevant to a filter into it (say in FilterExpression). Let's say one panel has no `repo` qualifier but the next panel has 2 `repo` qualifier, then each of them will have different sets of milestones to be used.

What I can think of is make it more efficient by storing current milestone of each `Model`, so getting current milestone is comparing the current milestone of each relevant repo `Model`.

HansNewbie:
@dariusf ignore my previous comment. I just understood what you were saying. I'll see what I can do.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -103,20 +102,58 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+        List<TurboMilestone> allMilestones = model.getMilestones().stream()
+                .filter(ms -> repoIds.contains(ms.getRepoId()))
+                .collect(Collectors.toList());
+        Collections.sort(allMilestones, Qualifier::milestoneDueDateComparator);
+
+        int currentMilestoneIndex = getCurrentMilestoneIndex(allMilestones);
+
+        List<Qualifier> milestoneQualifiers = expr.find(Qualifier::isMilestoneQualifier);
+        milestoneQualifiers.stream()
+                .forEach(msQualifier -> msQualifier.replaceMilestoneAlias(allMilestones, currentMilestoneIndex));
```
HansNewbie: @dariusf is this what you mean? Should I abstract more?

HansNewbie:
I haven't updated the tests, but does the implementation look good?

src/test/java/tests/FilterEvalTests.java
```diff
@@ -130,6 +132,192 @@ public void milestone() {
...
+
+        model = TestUtils.singletonModel(new Model(REPO,
+                new ArrayList<>(Arrays.asList(iCurrMin3, iCurrMin2, iCurrMin1,
+                        iCurr, iCurrPlus1, iCurrPlus2, iCurrPlus3)),
+                new ArrayList<>(),
+                new ArrayList<>(Arrays.asList(msCurrMin3, msCurrMin2, msCurrMin1,
+                        msCurr, msCurrPlus1, msCurrPlus2, msCurrPlus3)),
+                new ArrayList<>()));
+
+        assertEquals(true, Qualifier.process(model, Parser.parse("milestone:curr-3"), iCurrMin3));
```
HansNewbie: @dariusf I have trouble with testing. Even though the alias works in running the program, it will not in test. The problem is that the milestone `Qualifier` content seems to be immutable in testing. In this case, the content was replaced correctly in the method that replaces the alias, but during the comparison in `milestoneSatisfies`, the content is back to `curr-3` (instead of `V0.1` in this test case).

Would my implementation create a problem in certain environment or is this just test framework quirk? Then should I just test the methods that do the conversion of aliases in another file using reflection and don't test the `Qualifier.process()`?

HansNewbie:
@dariusf I refactored the milestone into its own method in `Qualifier`, not only done once per panel before applying the `FilterExpression` to each issue. I will write more tests for the new methods I created.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -93,6 +92,47 @@ public Qualifier(String name, List<SortKey> keys) {
         this.sortKeys = new ArrayList<>(keys);
     }
 
+    public static FilterExpression replaceMilestoneAliases(IModel model, FilterExpression expr) {
+        List<Qualifier> metaQualifiers = expr.find(Qualifier::isMetaQualifier);
+        List<String> repoIds = getReposInMetaQualifiers(metaQualifiers);
```
dariusf: Use `getContentOfMetaQualifier(expr, REPO)` here? Can also rename it to `getMetaQualifierContent`.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -93,6 +92,47 @@ public Qualifier(String name, List<SortKey> keys) {
...
+
+        if (repoIds.isEmpty()) {
+            repoIds.add(model.getDefaultRepo().toLowerCase());
+        }
+
+        List<TurboMilestone> allMilestones = model.getMilestones().stream()
+                .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
+                .collect(Collectors.toList());
+        Collections.sort(allMilestones, Qualifier::milestoneDueDateComparator);
```
dariusf: You can use `.filter(...).sorted(comparator).collect(...)`.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -93,6 +92,47 @@ public Qualifier(String name, List<SortKey> keys) {
...
+
+        int currentMilestoneIndex = getCurrentMilestoneIndex(allMilestones);
```
dariusf: This function should probably return an Optional, for the case where there are no milestones. We should check that before continuing.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -93,6 +92,47 @@ public Qualifier(String name, List<SortKey> keys) {
...
+        milestoneQualifiers.stream()
+                .filter(msQualifier -> msQualifier.getContent().isPresent())
+                .forEach(msQualifier -> {
+                    convertedMilestoneQualifiers.add(
+                            msQualifier.replaceMilestoneAliasQualifier(allMilestones, currentMilestoneIndex)
+                    );
+                });
+
+        for (Qualifier msQualifier : convertedMilestoneQualifiers) {
+            exprWithoutMilestones = new Conjunction(msQualifier, exprWithoutMilestones);
```
dariusf: This can change semantics: `(milestone:curr+1 && is:pr) || (milestone:curr && is:issue)` will become `(is:pr || is:issue) && milestone:a && milestone:b`, which isn't equivalent.

I think a better approach would be implement `map` for FilterExpressions.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -117,6 +154,31 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+            // current milestone must be open
+            if (checker.isOpen()) {
+                // if milestone is overdue but have no open issue, ignore
+                if (checker.getDueDate().isPresent() &&
+                        checker.getDueDate().get().isBefore(LocalDate.now()) &&
+                        checker.getOpenIssues() == 0) {
+                    continue;
+                }
+                currentIndex = allMilestones.indexOf(checker);
+                break;
```
dariusf: It's a bit strange to have both break and continue in here. How about

```java
if (allMilestones.isEmpty()) {
    return Optional.empty();
}

int index = 0;

for (TurboMilestone milestone : allMilestones) {
    boolean eligible = ...;
    if (eligible) {
        return index;
    }
    index++;
}

return Optional.of(allMilestones.size() - 1);
```

src/main/java/filter/expression/Qualifier.java
```diff
@@ -898,4 +969,53 @@ private void applyState(TurboIssue issue) throws QualifierApplicationException {
     public String getName() {
         return name;
     }
+
+    private Qualifier replaceMilestoneAliasQualifier(List<TurboMilestone> allMilestones, int currentIndex) {
+        String contents = content.get();
```
dariusf: We could move the check for presence here. That will ensure that callers of this method never forget to check.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -898,4 +969,53 @@ private void applyState(TurboIssue issue) throws QualifierApplicationException {
...
+
+        while (aliasMatcher.find()) {
+            if (aliasMatcher.group(2) != null && aliasMatcher.group(3) != null) {
+                offset = Integer.parseInt(aliasMatcher.group(3));
+                if (aliasMatcher.group(2).equals("+")) {
+                    currentIndex += offset;
+                } else {
+                    currentIndex -= offset;
+                }
+            } else if (!aliasMatcher.group(0).equalsIgnoreCase(alias)) {
```
dariusf: The regex could be changed to `current((\+|-)\d+)?$` (some details left out) to preclude this case and make `find` return false if it happens.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -898,4 +969,53 @@ private void applyState(TurboIssue issue) throws QualifierApplicationException {
...
+                // in this case found curr/current but there are something behind it
+                // e.g. curr+asdf
+                break;
+            }
+
+            if (currentIndex < allMilestones.size() &&
+                    currentIndex >= 0) {
+                contents = allMilestones.get(currentIndex).getTitle().toLowerCase();
```
dariusf: Probably better to clamp this to the required range rather than reject it entirely.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -898,4 +969,53 @@ private void applyState(TurboIssue issue) throws QualifierApplicationException {
...
+            }
+
+            break;
```
dariusf: Is this right? The loop will only execute once because of this. If it's intended the loop can be a guard clause instead, i.e `if (!matcher.find()) return this;`.

HansNewbie: Yes, intended. Okay, will change it.

src/test/java/tests/FilterEvalTests.java
dariusf: Is this still a problem? I'm not sure why it would be immutable. Generally unit tests aren't as quirky as UI tests.

Tests for both would be good, with more emphasis on the former I guess.

src/main/java/filter/expression/Qualifier.java
HansNewbie: :+1:

src/test/java/tests/FilterEvalTests.java
HansNewbie: No longer a problem as I refactored out things :) Both tests will be implemented.

src/main/java/filter/expression/Qualifier.java
HansNewbie: Just noticed it. Will change accordingly.

src/main/java/filter/expression/Qualifier.java
HansNewbie: I will return `Qualifier.EMPTY` if no presence.

src/main/java/filter/expression/Qualifier.java
HansNewbie: I will still need the continue as otherwise the eligible boolean will be very long (listing 1 negative vs listing 3 positives). I will refactor it to like what you suggested to be cleaner.

src/main/java/filter/expression/Qualifier.java
HansNewbie: I am not familiar with `map`; I will try it first. I show you tomorrow and see if it's good.

HansNewbie:
I did the mapping. I need time to update my tests, though.

src/test/java/tests/FilterEvalTests.java
HansNewbie: After changing everything, this similar problem comes back. Maybe a bug in my code. I will see to it.

HansNewbie:
Is this good now? Are the tests enough?

src/main/java/filter/expression/Qualifier.java
```diff
@@ -93,6 +93,40 @@ public Qualifier(String name, List<SortKey> keys) {
...
+        if (repoIds.isEmpty()) {
+            repoIds.add(model.getDefaultRepo().toLowerCase());
+        }
+
+        List<TurboMilestone> allMilestones = model.getMilestones().stream()
+                .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
+                .sorted(Qualifier::milestoneDueDateComparator)
+                .collect(Collectors.toList());
+
+        Optional<Integer> currentMilestoneIndex = getCurrentMilestoneIndex(allMilestones);
```
dariusf: Maybe we could check this here and return `expr` if it's absent? We don't need to do the rest if there is some index, and the rest can just deal with an int.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -117,6 +148,36 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+            // current milestone must be open
+            if (checker.isOpen()) {
+                // if milestone is overdue but have no open issue, ignore
+                if (checker.getDueDate().isPresent() &&
+                        checker.getDueDate().get().isBefore(LocalDate.now()) &&
+                        checker.getOpenIssues() == 0) {
+                    // do nothing; using negative statement here
+                    // since positive statement would be much longer
+                    // and more complicated; this works since
+                    // milestones are sorted
```
dariusf: How about this?
```java
boolean overdue = checker.getDueDate().isPresent()
    && checker.getDueDate().get().isBefore(LocalDate.now());

boolean relevant = !(overdue && checker.getOpenIssues() > 0);

if (checker.isOpen() && relevant) {
    return Optional.of(currentIndex);
}
```
This isn't super important, just trying to write it so the comments aren't needed.

dariusf:
Yup, I think the tests are fine.

src/main/java/filter/expression/Qualifier.java
HansNewbie: Sure, will do. Thanks for the help!

src/main/java/filter/expression/Qualifier.java
```diff
@@ -93,6 +93,44 @@ public Qualifier(String name, List<SortKey> keys) {
...
+        List<TurboMilestone> allMilestones = model.getMilestones().stream()
+                .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
+                .sorted(Qualifier::milestoneDueDateComparator)
+                .collect(Collectors.toList());
+
+        Optional<Integer> currentMilestoneIndex = getCurrentMilestoneIndex(allMilestones);
+
+        if (!currentMilestoneIndex.isPresent()) {
+            return expr;
+        }
```
HansNewbie: Here is the check. And I changed the parameter of `convertMilestoneAliasQualifier` from `Optional<Integer>` to `int`.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -103,12 +141,9 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
         List<Qualifier> metaQualifiers = expr.find(Qualifier::isMetaQualifier);
 
         // Preprocessing for repo qualifier
-        boolean containsRepoQualifier = metaQualifiers.stream()
-                .map(Qualifier::getName)
-                .collect(Collectors.toList())
-            .contains("repo");
+        List<String> repoIds = getReposInMetaQualifiers(metaQualifiers);
```
HansNewbie: Sorry, missed this one. Will use the metaqualifier finder.

HansNewbie: Maybe I will return this one to its original, seems cleaner.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -117,6 +152,29 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+        int currentIndex = 0;
+
+        for (TurboMilestone checker : allMilestones) {
+            boolean overdue = checker.getDueDate().isPresent() &&
+                    checker.getDueDate().get().isBefore(LocalDate.now());
+            boolean relevant = !(overdue && (checker.getOpenIssues() == 0));
+
+            if (checker.isOpen() && relevant) {
+                return Optional.of(currentIndex);
+            }
```
HansNewbie: Here is the new one without continue or comments.

HansNewbie:
I am going to cleanup my commits and rebase.

HansNewbie:
@dariusf we got a problem. @ndt93's comparator and mine behaves differently. He sorts milestone from latest to earliest then to no due date. Mine sorts milestone from earliest to latest then to no due date. May not be able to reuse my comparator, unless my milestone alias behaviour is changed.

damithc:
Unify the two behavior?

HansNewbie:
Passed all tests locally.

dariusf:
Okay, looks good.

PR #1051, opened 2015-10-13
Move installer link to home README page

### LINKED ISSUE #1016: Move installer link to home README page
UX issue - Want to reduce the barriers for the user to install HubTurbo and improve intuitiveness when downloading

codemanmeet:
Fixes #1016.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -6,7 +6,9 @@ Are you a GitHub power user? Enhance your GitHub Issue Tracker experience with t
 
 ![](https://github.com/HubTurbo/HubTurbo/blob/release/docs/images/gettingStarted/bothViewsScreenshot.png?raw=true)
 
-To **start using HubTurbo**, check out our [User Guide](https://github.com/HubTurbo/HubTurbo/blob/release/docs/userGuide.md).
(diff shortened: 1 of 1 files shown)


damithc:
Before we move the link to the landing page, we should implement a check in HT to verify if the prerequisites are set up is correctly. Otherwise users will simply download the jar file without any idea of the prerequisites.

codemanmeet:
The pre reqs are mentioned in the main page as well. I have included "pre-requisities" and "Installer" as subheadings under "Quick installation" to make sure the users have the necessary pre-requisites. We can add a separate issue to perform a check when the jar file itself is opened.

codemanmeet:
Appears something like this - 
screen shot 2015-10-13 at 6 18 02 pm

damithc:
This has two problems:
1. duplication of information
2. README.md is served from the master branch which may be ahead of the release branch. So the pre-requisites given in README.md may not be compatible with the released version of the app.

codemanmeet:
@damithc @dariusf Do you know how I can serve it from the release branch? 

@damithc Concerning the duplication, shall I remove it from the main User Guide then?

damithc:
May be the easy solution to give a direct link from README to the place where it explains how to install? At this time we don't expect random people to start using HT, so it is not a big problem if the download link is not in the landing page.

codemanmeet:
@damithc Let me know if it looks good now.

PR #1055, opened 2015-10-14
milestone missing from some issues #1054

### LINKED ISSUE #1054: milestone missing from some issues
This happens for teammates (has a lot of milestones), so it could be a pagination issue
image

ndt93:
Fixes #1054

### DIFF
--- a/src/main/java/github/update/UpdateService.java
+++ b/src/main/java/github/update/UpdateService.java
@@ -80,44 +80,50 @@ public class UpdateService<T> extends GitHubService {
      * @return a list of requested items
      */
     public ArrayList<T> getUpdatedItems(IRepositoryIdProvider repoId) {
-
         // Return cached results if available
         if (updatedItems != null)  {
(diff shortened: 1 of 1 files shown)


ndt93:
After running a version with this fix, a cache refresh is also required for it to take effect.

ndt93:
@dariusf Can help review this? This bug is quite dangerous.

dariusf:
I get that the bug was caused by the same object being used twice, but why does it happen? Does the request fail if the object has already been mutated with some result?

If `request` is only needed lower down, maybe move it inside the else branch?

ndt93:
Yup, the header request mutates the `request` object and the `PageIterator` can only carry on from the page at which the header request left off. I will move it to the else branch.

ndt93:
@dariusf I've refactored nested `if/else` in `getUpdatedItems` also. You can check this again.

dariusf:
Okay, looks good.

PR #1059, opened 2015-10-14
Extend UITest.type so it accepts non-alphanumeric keystrokes #966

### LINKED ISSUE #966: Extend UITest.type so it accepts non-alphanumeric keystrokes
See discussion [here](https://github.com/HubTurbo/HubTurbo/pull/960#discussion_r39348801).

rd1992:
Fixes #966


src/test/java/guitests/IssuePanelTests.java
```diff
@@ -34,6 +34,7 @@
 public class IssuePanelTests extends UITest {
 
     private static final int EVENT_DELAY = 1000;
+    private UtilMethods util = new UtilMethods();
```
damithc: Perhaps be more specific in the class name? e.g. `GuiTestingRobot robot` `robot.typeString("sort:date")`

dariusf: Every test class which does keyboard input is already a subclass of UITest, so you don't need this object -- overriding `type` in UITest would allow you to call it as if it had been defined on all the test classes.

src/test/java/guitests/KeyboardShortcutsTest.java
```diff
@@ -213,11 +214,9 @@ public void keyboardShortcutsTest() {
 
         // testing corner case for mark as read where there is only one issue displayed
         click("#dummy/dummy_col1_filterTextField");
-        type("id");
-        press(KeyCode.SHIFT).press(KeyCode.SEMICOLON).release(KeyCode.SEMICOLON).release(KeyCode.SHIFT);
-        type("5");
+        util.typeString("id:5");
         push(KeyCode.ENTER);
-        push(KeyCode.SPACE).push(KeyCode.SPACE);
+        push(KeyCode.CONTROL).push(KeyCode.ENTER);
```
dariusf: You can `press` the KeyCodeCombination here. There's a new overload for that.

dariusf:
Are there any chars which can't be typed via string, but which don't require Shift to be pressed?

rd1992:
@dariusf type will type any character which generates a unicode on press. (at least according to their documentation).

rd1992:
@dariusf updated

src/test/java/guitests/UIComponentsTest.java
```diff
@@ -25,20 +25,8 @@ public void keywordCompletionTest() {
...
-        type("is");
-        press(KeyCode.SHIFT).press(KeyCode.SEMICOLON).release(KeyCode.SEMICOLON).release(KeyCode.SHIFT);
-        push(KeyCode.LEFT);
-        push(KeyCode.TAB);
-        push(KeyCode.RIGHT);
-        type("open OR is");
-        press(KeyCode.SHIFT).press(KeyCode.SEMICOLON).release(KeyCode.SEMICOLON).release(KeyCode.SHIFT);
-        type("closed");
-        press(KeyCode.SHIFT).press(KeyCode.DIGIT0).release(KeyCode.DIGIT0).release(KeyCode.SHIFT);
+        type("(is:open OR is:closed)");
```
dariusf: Are the parentheses needed?

src/test/java/guitests/UITest.java
```diff
@@ -44,7 +45,32 @@
...
+        chars.put('_', KeyCode.MINUS);
+        chars.put('+', KeyCode.EQUALS);
+        chars.put('{', KeyCode.OPEN_BRACKET);
+        chars.put('}', KeyCode.CLOSE_BRACKET);
+        chars.put(':', KeyCode.SEMICOLON);
+        chars.put('"', KeyCode.QUOTE);
+        chars.put('<', KeyCode.COMMA);
+        chars.put('>', KeyCode.PERIOD);
+        chars.put('?', KeyCode.SLASH);
+    }
```
dariusf: It might be better to put this in a private method below, then do `private static final Map<Character, KeyCode> chars = getKeysWhichCannotBeTyped();` (better names required). The name of the field should also be more specific.

The map can also be made immutable before it's returned from the method (Collections.unmodifiableMap).

src/test/java/guitests/UIComponentsTest.java
rd1992: No. But i want to test whether parentheses can be typed using our method.

src/test/java/guitests/UITest.java
rd1992: @dariusf made the relevant changes

src/test/java/guitests/UIComponentsTest.java
dariusf: Hmm, but if it failed to type parentheses there, the test would still pass. Is there some other way we could test the method?

Maybe a new test: start up the GUI, have it type a bunch of special characters into the filter box (doesn't have to be valid syntax), then read the value of the field and assert that everything expected is there?

rd1992: Alright added those also @dariusf

dariusf:
Okay, looks good.

PR #1061, opened 2015-10-14
 Improve or change panel element selection in tests #933 

### LINKED ISSUE #933: Improve or change panel element selection in tests

HyungJon:
Fixes #933 

Those tests involving more than 1 panels, especially if the ordering of the panels may change, were changed to use ```TextController``` to directly access panels instead of using IDs which are more fallible.

### DIFF
--- a/src/main/java/ui/issuepanel/FilterPanel.java
+++ b/src/main/java/ui/issuepanel/FilterPanel.java
@@ -51,6 +51,7 @@ public abstract class FilterPanel extends AbstractPanel {
     protected HBox nameBox;
     protected HBox menuBarNameArea;
     protected Label renameButton;
+    protected Label closeButton;
     protected PanelNameTextField renameTextField;
     protected FilterTextField filterTextField;
     
@@ -173,7 +174,7 @@ public abstract class FilterPanel extends AbstractPanel {
     private HBox createCloseButton() {
         HBox closeArea = new HBox();
         
-        Label closeButton = new Label(CLOSE_PANEL);
+        closeButton = new Label(CLOSE_PANEL);
         closeButton.setId(model.getDefaultRepo() + "_col" + panelIndex + "_closeButton");
         closeButton.getStyleClass().add("label-button");
         closeButton.setOnMouseClicked((e) -> {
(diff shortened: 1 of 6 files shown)


HyungJon:
That... is strange. An unrelated ```BoardSwitchTest``` fails on Travis. Passes locally. I see no reason why this should be unstable though.

HyungJon:
I found something I want to improve on. ```PanelsTest``` has a part where the two panels are swapped, but I found that they don't actually get swapped, at least most of the time. I'm going to add a test here now that we can directly access panels based on their indices, but I need to solve this UI issue first. The panel swapping seems kind of blocky, needs some retries until I can actually swap. Can panel swapping be improved?

dariusf:
We could put in a hotkey to swap panels. The mouse event has always been unreliable.

HyungJon:
Will do that in a separate issue then, and focus on the panel selection here.

As for ```BoardSwitchTest``` failing, I think it's Travis problem again. It only fails on Travis, and It's using the name "Dummy Board" instead of the assigned "Board 1" and such. I had similar issues before with Travis checking panel names with strange names like "Panel should have name quit", possibly some variables that somehow drifted in. Should this be considered an unstable test?

dariusf:
What do you mean by Travis using a different name?

Tests can be nondeterministic in that sometimes nodes can't be found, key events fail or repeat many times, etc. but variables 'drifting in' sounds like a problem with the logic of the test or related setup, e.g. wrong node id causing a field to have a different name than expected.

If the field has a different value because of the aforementioned nondeterminism (and we need to pinpoint the cause), I suppose we could make this unstable, but not unless.

HyungJon:
I've been looking into this more, I think it might be there's something wrong with how ```BoardSwitchTest``` is initiated. Possibly because of "testPrefs = new Preferences(true)"? It doesn't explain why this test is failing as a result of panel elements selection, but anyway something might be altering the starting state of ```Preferences```. Could you have a look at the problematic test?

As for the "drifting", I really have no idea. I had that problem once, where the name of the panel was something that's not supposed to be possible since there's no board with such a name.

HyungJon:
Found the cause.

HyungJon:
Right, solved.

HyungJon:
@dariusf Can I have a review? Also can the fall in coverage be considered a false positive? I'm quite sure the panel element getter methods are responsible.

dariusf:
I won't be able to get to this for a while. @lejolly could you help to review?

lejolly:
looks ok, how does this fare on travis? namely, what's the pass/fail ratio?

HyungJon:
Normally passes. I don't think I've seen it fail after I figured out what was causing the problem.

lejolly:
could you run it for at least 10 tries and report how many times it fails?

HyungJon:
@lejolly  Ran at least 10 times I think, passed every time.

lejolly:
ok, could you try moving `PanelsTest` out of `unstable` and run it on travis a few times to see if it passes too?

HyungJon:
Alright. I think I ran 3~4 times. All passed.

lejolly:
ok, @dariusf good enough to move it back to the stable tests?

dariusf:
Sure, should be fine.

lejolly:
ok, looks good then.

PR #1066, opened 2015-10-17
Some repos not saved as json files #1065

### LINKED ISSUE #1065: Some repos not saved as json files
I have powerpointlabs/powerpointlabs as a repo in my filters. Turns out it is never saved to the hard disk. Everytime I start HT, the repo is downloaded again.

HansNewbie:
Fixes #1065

### DIFF
--- a/src/main/java/github/update/UpdateService.java
+++ b/src/main/java/github/update/UpdateService.java
@@ -98,7 +98,7 @@ public class UpdateService<T> extends GitHubService {
                    The assumption is that updates are cheap and we can do them as frequently as needed. */
                 logger.warn(String.format("%s: error getting updated items", getClass().getSimpleName()));
             } else {
-                Optional<String> updatedETags = combineETags(etags.get().getLeft());
+                updatedETags = combineETags(etags.get().getLeft());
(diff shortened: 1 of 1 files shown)


damithc:
I'm also interested to know how this bug got through and how to prevent something like that happening again.

HansNewbie:
The bug was caused by using a local variable instead of the supposed class variable. If you would like to prevent it to be happening again, the solutions I can think of are for `checkStyle` to check either:
- no local variable has same name as class variable except in constructor.
- all class variable should use `this.` reference so the next developer know that they should use class variable.

I am not sure if they are possible to be included in `checkStyle`, though.

ndt93:
This bug and similar major bugs happening recently are from the download and update process i.e. interactions with GitHub which we technically don't have any test for. These bugs only crop up in scenarios which we could miss, e.g. for this one I only check for the updating part but never try downloading a new repo. So ideally we could have some kinds of mock-ups to test these parts.

Beside, how does the json files get deleted in the first place?

HansNewbie:
Using remove repo option. That removes the JSON file.

ndt93:
Ok, I see. This fix can be merged then.

HansNewbie:
Ok.

PR #1067, opened 2015-10-17
User can generate a 'milestones' board for a repo #1030

### LINKED ISSUE #1030: User can generate a 'milestones' board for a repo
... so that the user need not create board manually.

It can be under the 'Borads' menu.
It will generate a board with one panel each for five milestones, name the panels, and save the board.

HansNewbie:
Fixes #1030

### DIFF
--- a/src/main/java/filter/expression/Qualifier.java
+++ b/src/main/java/filter/expression/Qualifier.java
@@ -25,10 +25,11 @@ public class Qualifier implements FilterExpression {
     private static final String UPDATED = "updated";
 
     public static final Qualifier EMPTY = new Qualifier("", "");
+    public static final Qualifier FALSE = new Qualifier("false", "");
 
     public static final List<String> KEYWORDS = Collections.unmodifiableList(Arrays.asList(
         "assignee", "author", "body", "closed", "comments", "created", "creator",
-        "date", "desc", "description", "has", "id", "in", "involves",
+        "date", "desc", "description", "false", "has", "id", "in", "involves",
         "is", "issue", "keyword", "label", "labels", "merged", "milestone", "milestones",
         "no", "nonSelfUpdate", "open", "pr", "pullrequest", "read", REPO, SORT, "state", "status",
         "title", "type", "unmerged", "unread", UPDATED, "user"
@@ -224,12 +225,19 @@ public class Qualifier implements FilterExpression {
         return name.isEmpty() && content.isPresent() && content.get().isEmpty();
     }
 
+    public boolean isFalseQualifier() {
+        return name.equalsIgnoreCase("false") && content.isPresent() && content.get().isEmpty();
+    }
+
     @Override
     public boolean isSatisfiedBy(IModel model, TurboIssue issue, MetaQualifierInfo info) {
         assert name != null;
 
         // The empty qualifier is satisfied by anything
         if (isEmptyQualifier()) return true;
+
+        // The false qualifier is satisfied by nothing
+        if (isFalseQualifier()) return false;
 
         switch (name) {
         case "id":
@@ -282,6 +290,9 @@ public class Qualifier implements FilterExpression {
 
         // The empty qualifier should not be applied to anything
         assert !isEmptyQualifier();
+
+        // The false qualifier should not be applied to anything
+        assert !isFalseQualifier();
 
         switch (name) {
         case "title":
@@ -1031,11 +1042,9 @@ public class Qualifier implements FilterExpression {
             }
         }
 
-        // clamp
-        if (currentIndex >= allMilestones.size()) {
-            currentIndex = allMilestones.size() - 1;
-        } else if (currentIndex < 0) {
-            currentIndex = 0;
+        // if out of milestone range, don't convert alias
+        if (currentIndex >= allMilestones.size() || currentIndex < 0) {
+            return Qualifier.FALSE;
         }
 
         contents = allMilestones.get(currentIndex).getTitle().toLowerCase();
--- a/src/main/java/ui/BoardAutoCreator.java
+++ b/src/main/java/ui/BoardAutoCreator.java
@@ -1,0 +1,84 @@
+package ui;
+
+
+import javafx.scene.control.Menu;
+import javafx.scene.control.MenuItem;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import prefs.Preferences;
+import ui.issuepanel.FilterPanel;
+import ui.issuepanel.PanelControl;
+import util.DialogMessage;
+import util.Utility;
+import util.events.BoardSavedEvent;
+
+import java.util.stream.Collectors;
+
+public class BoardAutoCreator {
+    private static final Logger logger = LogManager.getLogger(MenuControl.class.getName());
+    private static final String MILESTONES = "Milestones";
+
+    private final PanelControl panelControl;
+    private final Preferences prefs;
+
+    public BoardAutoCreator(PanelControl panelControl, Preferences prefs) {
+        this.panelControl = panelControl;
+        this.prefs = prefs;
+    }
+
+    public Menu generateBoardAutoCreateMenu() {
+        Menu autoCreate = new Menu("Auto-create");
+
+        MenuItem milestone = new MenuItem(MILESTONES);
+        milestone.setOnAction(e -> createMilestoneBoard());
+        autoCreate.getItems().add(milestone);
+
+        return autoCreate;
+    }
+
+    private void createMilestoneBoard() {
+        logger.info("Creating " + MILESTONES + " board");
+
+        panelControl.closeAllPanels();
+
+        // reverse ordering so always addPanelAt(0)
+        FilterPanel panelCurrPlus3 = panelControl.addPanelAt(0);
+        panelCurrPlus3.setPanelName("Next Next Next Milestone");
+        panelCurrPlus3.setFilterByString("milestone:curr+3 sort:status");
+
+        FilterPanel panelCurrPlus2 = panelControl.addPanelAt(0);
+        panelCurrPlus2.setPanelName("Next Next Milestone");
+        panelCurrPlus2.setFilterByString("milestone:curr+2 sort:status");
+
+        FilterPanel panelCurrPlus1 = panelControl.addPanelAt(0);
+        panelCurrPlus1.setPanelName("Next Milestone");
+        panelCurrPlus1.setFilterByString("milestone:curr+1 sort:status");
+
+        FilterPanel panelCurr = panelControl.addPanelAt(0);
+        panelCurr.setPanelName("Current Milestone");
+        panelCurr.setFilterByString("milestone:curr sort:status");
+
+        FilterPanel panelCurrMinus1 = panelControl.addPanelAt(0);
+        panelCurrMinus1.setPanelName("Previous Milestone");
+        panelCurrMinus1.setFilterByString("milestone:curr-1 sort:status");
+
+        panelControl.selectPanel(1); // current
+
+        String boardName = Utility.generateName(MILESTONES,
+                prefs.getAllBoards().keySet().stream().collect(Collectors.toList()));
+
+        triggerBoardSaveEventSequence(boardName);
+
+        DialogMessage.showInformationDialog("Auto-create Board - " + MILESTONES,
+                MILESTONES + " board has been created and loaded.\n\n" +
+                        "It is saved under the name \"" + boardName + "\".");
+    }
+
+    private void triggerBoardSaveEventSequence(String boardName) {
+        prefs.addBoard(boardName, panelControl.getCurrentPanelInfos());
+        prefs.setLastOpenBoard(boardName);
+        TestController.getUI().triggerEvent(new BoardSavedEvent());
+        logger.info("Auto-created board, saved as \"" + boardName + "\"");
+        TestController.getUI().updateTitle();
+    }
+}
--- a/src/main/java/ui/issuepanel/PanelControl.java
+++ b/src/main/java/ui/issuepanel/PanelControl.java
@@ -21,6 +21,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 
 public class PanelControl extends HBox {
@@ -117,10 +119,15 @@ public class PanelControl extends HBox {
         return panel;
     }
 
+    public void selectPanel(int index) {
+        assert(index >= 0 && index < getNumberOfPanels());
+        setCurrentlySelectedPanel(Optional.of(index));
+        scrollToPanel(index);
+        getPanel(index).requestFocus();
+    }
+
     public void selectFirstPanel() {
-        setCurrentlySelectedPanel(Optional.of(0));
-        scrollToPanel(0);
-        getPanel(0).requestFocus();
+        selectPanel(0);
     }
 
     private void setCurrentlySelectedPanel(Optional<Integer> selectedPanel) {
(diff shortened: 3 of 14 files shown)


HansNewbie:
@dariusf can take a look if my implementation for Board Preset is okay?

What I did:
- abstract out board preset setting into JSON file saved in `resources`.
- I put board preset loader under pref.

I did this so that in the future if we have any other preset board, we can just add on to the JSON.

What I can improve:
- I copied a lot on the `gson`. Should I abstract it into some `Utils` that `ConfigFileReader` can use as well?

I will write the tests when my implementation is finalized.

HansNewbie:
@damithc Panel names of the milestone board right now are Previous Milestone, Current Milestone, Next Milestone, Next Next Milestone and Next Next Next Milestone. Should I put Next+1 and Next+2 instead?

Also, in the current milestone alias implementation, the alias is clamped, i.e. if there is no future milestone, `curr+1` and `curr+2` etc. will be the same as `curr`, thus the last 3 panels might have the same issues. Should the clamping be kept that way?

Currently, the first panel - Previous Milestone - will be selected. Should I make it such that for preset board, when it is opened, the selected panel can be set by developer? Or should I just change the order of the boards for milestone - like current, prev, next, next+1 and next+2?

damithc:
Allowing curr+1 to be equal to curr will be misleading. That panel should appear empty if there are no such milestone.
Keep the chronological order in panels. It's nice if the current panel can be auto-selected, but it's also ok to select the first panel.

damithc:
I don't see why generated panels have to be treated differently and put in a separate json file. 
Can you post a screenshot of the new menu? I'm expecting to see something like `Auto-create board --> Milestone plan`

HansNewbie:
Preset

I am thinking that if we are going to have many preset boards, extracting them as JSON would make it more manageable in adding and removing presets. Ideally, we can have a separate application to generate the JSON for a new preset. Running methods on-the-fly to generate board would make user wait and see the whole process while loading a JSON would be fast and hidden from user.

Do you have something else in mind?

damithc:
1. Is the word Preset widely used and intuitive? The sentence 'Preset Milestone' doesn't make sense same way 'Delete teammates/repo'.
2. Some boards we plan to generate cannot be specified in a static way e.g. 'work distribution' board depends on the project members. If not for the milestone aliases, even this board would require dynamic generation. So I'm not sure separating them into a different file makes sense. 
By the way, we can also add some panel presets.

HansNewbie:
Since we need the interactivity in the future, I will change the implementation. I will use `Auto-create` for the menu item.

src/main/java/prefs/BoardPresetsLoader.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+public class BoardPresetsLoader {
+    private static final Logger logger = LogManager.getLogger(BoardPresetsLoader.class.getName());
+    private static final String CHARSET = "UTF-8";
+
+    private Gson gson;
+    private Map<String, List<PanelInfo>> boardPresets = new LinkedHashMap<>();
```
dariusf: These two be final, and assigned in the constructor.

src/main/java/prefs/BoardPresetsLoader.java
```diff
@@ -0,0 +1,66 @@
...
+                        (JsonSerializer<LocalDateTime>) (src, typeOfSrc, context) -> {
+                            Instant instant = src.atZone(ZoneId.systemDefault()).toInstant();
+                            long epochMilli = instant.toEpochMilli();
+                            return new JsonPrimitive(epochMilli);
+                        })
+                .registerTypeAdapter(LocalDateTime.class,
+                        (JsonDeserializer<LocalDateTime>) (json, typeOfT, context) -> {
+                            Instant instant = Instant.ofEpochMilli(json.getAsJsonPrimitive().getAsLong());
+                            return LocalDateTime.ofInstant(instant, ZoneId.systemDefault());
+                        }
```
dariusf: Are these necessary? LocalDateTime instances can be serialised by default, judging from similar code in Read/WriteTask.

If they are, they should be pulled out into an inner class.

src/main/java/prefs/BoardPresetsLoader.java
```diff
@@ -0,0 +1,66 @@
...
+        try (Reader reader =
+                     new InputStreamReader(new FileInputStream(boardPresetsFile), CHARSET)) {
+            Type type = new TypeToken<Map<String, List<PanelInfo>>>(){}.getType();
+            boardPresets = gson.fromJson(reader, type);
+        } catch (IOException e) {
+            HTLog.error(logger, e);
+        }
+    }
+
+    public Map<String, List<PanelInfo>> getAllBoards() {
```
dariusf: This method name is a bit too general for what it does.

dariusf:
About the JSON approach: I was under the impression that we would have more dynamic presets in future. For example, generating a weekly sprint board with a panel for each team member, i.e. we would have to iterate through the collaborators of a repository. In that case we might want to express presets with source code rather than data.

If this isn't a legit concern then we can ignore this.

HansNewbie:
@dariusf yeah, sorry about that. I am changing everything.

HansNewbie:
@damithc on saving the auto-created board, how should it be handled? Automatically or prompt user for board name? My concern is when there is already a board with the same name.

damithc:
I guess it is easier to save automatically. The name can be same as shown in the menu e.g. Milestones.
If it exists, we can use the common technique of adding a number to the end. May be you can write a method in a utility class to generate the name? e.g. `generateName(String desiredName, ArrayList<String> existingNames)`

HansNewbie:
@damithc how should user be informed the board name that auto-created board is saved as? User might not know what it is saved as. They can look at the window title but I am not sure if that is obvious enough.

damithc:
Can show a dialog box with some info after the board is created?

HansNewbie:
I was thinking of using `Notification` instead, like undo label except without the undo button. The message will be `Auto-created [milestones - will be method parameter] board - saved as [board name]`. Is that okay? I think dialog will be too intrusive since in milestones board case there is nothing to be set (as compared to say 'work distribution' board). And this notification can be used for other auto-created boards.

damithc:
I favor a dialog because,
1. Notifications have very limited display area. I don't think it is suitable for showing lot of information to the user.  We can use the dialog to show more instructions to the user e.g. how to edit the board.
2. Notifications are for info that users may or may not look at. I think we want user to read this message, especially if this is the first time a board is created.
3. Generating a board is very infrequent, compared to adding a label for instance. So it is ok to use a dialog although it is an extra click. 

BTW, you automatically switch to the created board right? If not, we can provide that as an option in the dialog box?

damithc:
Anyway, the whole dialog box thing can be done as a separate issue.

HansNewbie:
Yes, I automatically switch to the created board. Or rather, I clear all the panels then do all the steps. If the current board is not saved, it will be lost. Is that fine? It's very difficult to save the current board before updating now though, since all board functionality is in `MenuControl` and I think it's weird to keep a reference to `MenuControl` to invoke board save (see #1074).

Just to confirm, for milestones board case, what should be shown in the dialog? The auto-created board type and the board name that it is saved as? I will see if I can do it in this PR. If not, I will create a new issue for it.

damithc:
For now, it is fine to overwrite the existing panels. In future, we should ask 'save the existing board/panels?' before generating the new board.
Yes, just inform the board has been created, loaded, and saved under the name ...'.

HansNewbie:
Alright.

HansNewbie:
@dariusf passed tests locally.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -1031,11 +1031,9 @@ private Qualifier convertMilestoneAliasQualifier(List<TurboMilestone> allMilesto
...
         }
 
-        // clamp
-        if (currentIndex >= allMilestones.size()) {
-            currentIndex = allMilestones.size() - 1;
-        } else if (currentIndex < 0) {
-            currentIndex = 0;
+        // if out of milestone range, don't convert alias
+        if (currentIndex >= allMilestones.size() || currentIndex < 0) {
+            return this;
```
dariusf: Say we get the input `milestone:curr+99`. In that case we would search for the milestone "curr+99"?

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -0,0 +1,84 @@
...
+import util.events.BoardSavedEvent;
+
+import java.util.stream.Collectors;
+
+public class BoardAutoCreator {
+    private static final Logger logger = LogManager.getLogger(MenuControl.class.getName());
+    private static final String MILESTONES = "Milestones";
+
+    PanelControl panelControl;
+    Preferences prefs;
```
dariusf: Can these be private and final?

src/main/java/ui/issuepanel/PanelControl.java
```diff
@@ -56,7 +58,7 @@ public void init(GUIController guiController, ScrollPane panelsScrollPane) {
         this.guiController = guiController;
         this.panelsScrollPane = panelsScrollPane;
         restorePanels();
-        selectFirstPanel();
+        selectPanel(0);
```
dariusf: We could leave selectFirstPanel in, just define it in terms of selectPanel?

src/main/java/util/Utility.java
```diff
@@ -299,4 +299,26 @@ public static boolean startsWithIgnoreCase(String source, String query) {
...
+            int index = 1;
+
+            while (existingSuffixes.contains(Integer.toString(index))) {
+                index++;
+            }
+
+            availableName = desiredName + Integer.toString(index);
+        }
+
+        return availableName;
```
dariusf: You could put the other, easier case first, so we get it out of the way quickly.

src/test/java/guitests/DialogMessageTests.java
```diff
@@ -38,4 +38,12 @@ public void showYesNoWarningDialogTest() throws ExecutionException, InterruptedE
         click("nO");
         assertEquals(false, noTask.get());
     }
+
+    @Test
+    public void showInformationDialogTest() {
+        Platform.runLater(() -> DialogMessage.showInformationDialog("Information", "Test Information Dialog"));
+        sleep(1000);
```
dariusf: Could we wait until `hasText` returns true instead to synchronise?

src/main/java/filter/expression/Qualifier.java
HansNewbie: Yes.

src/main/java/ui/BoardAutoCreator.java
HansNewbie: Yep, sorry missed that.

src/main/java/ui/issuepanel/PanelControl.java
HansNewbie: Sure.

src/main/java/filter/expression/Qualifier.java
dariusf: Is that a good idea? It seems unlikely that that is what the user would have intended.

src/main/java/util/Utility.java
HansNewbie: Oh, makes sense. Didn't realize that because I put the second statement afterwards due to problems.

src/test/java/guitests/DialogMessageTests.java
HansNewbie: So we wait until `hasText` is true then assert the `hasText` again? But implementing the wait will involve `sleep` as well, yes? And currently there is no method to `waitUntilNodeAppears(Matcher<>)`. Should I make one?

If this gets on, should I change the two tests in the same file as this to use that wait as well?

src/main/java/filter/expression/Qualifier.java
HansNewbie: It's not that good of an idea, but do we have any qualifier that would fail for all issues? I was thinking it might be the case that the user has such milestone (although highly unlikely) so might as well pass it as it is.

src/main/java/util/Utility.java
HansNewbie: Is it okay now?

src/main/java/filter/expression/Qualifier.java
dariusf: I thought the clamping from before was fine, actually. I don't think there's such a qualifier, but it seems like a useful addition.

src/test/java/guitests/DialogMessageTests.java
dariusf: I was thinking with `awaitCondition`, but sure, that sounds like a better idea. Yeah we just want to synchronise as much as possible.

src/main/java/util/Utility.java
dariusf: Hmm? Not sure what changed. What I meant was
```java
if (!existingNames.contains()) {
    return desiredName;
}
// build the name
return desiredName + Integer.toString(index);
```

HansNewbie: Oh, oops. I switched the `&&`. Sure.

src/main/java/filter/expression/Qualifier.java
HansNewbie: I will try to make one.

src/test/java/guitests/DialogMessageTests.java
HansNewbie: Okay.

src/main/java/filter/expression/Qualifier.java
dariusf: I guess you could call it the null (or false, if we don't want to overload the term) qualifier. Implementing it is mostly just figuring out a way to differentiate it from others, then adding it to isSatisfiedBy. I don't think we need to expose it to users.

HansNewbie: I don't want to overload the term. Will use false qualifier.

src/test/java/guitests/UITest.java
```diff
@@ -256,6 +257,20 @@ public void waitUntilNodeDisappears(String selector) {
         }
     }
 
+    public void waitUntilNodeAppears(Matcher<Object> matcher) {
+        while (!findQuiet(matcher).isPresent()) { // no `exists` for matcher, so using find
```
HansNewbie: This is the new function. I implemented `findQuiet` with `find(Matcher<Object>)`. I didn't use `exist` because there is no `exists(Matcher<Object>)` by testfx.

src/main/java/util/Utility.java
```diff
@@ -299,4 +299,28 @@ public static boolean startsWithIgnoreCase(String source, String query) {
...
     }
 
+    public static String generateName(String desiredName, List<String> existingNames) {
+        String availableName = desiredName;
+
+        if (!existingNames.contains(desiredName)) {
+            return availableName;
+        }
+
+        List<String> existingSuffixes = existingNames.stream()
```
HansNewbie: Like this?

src/main/java/filter/expression/Qualifier.java
HansNewbie: Do I need to test it? I am not sure how to test it.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -231,6 +236,9 @@ public boolean isSatisfiedBy(IModel model, TurboIssue issue, MetaQualifierInfo i
         // The empty qualifier is satisfied by anything
         if (isEmptyQualifier()) return true;
 
+        // The false qualifier is satisfied by nothing
+        if (isFalseQualifier()) return false;
```
HansNewbie: I implemented it as opposite of `Empty` qualifier. Its name would be `false` and its content  would be empty.

src/main/java/filter/expression/Qualifier.java
dariusf: Yes, just to cover it. There are some examples in FilterEvalTests. You can just check that it returns false on some issues.

HansNewbie: Okay, sure.

dariusf:
Looks good.

PR #1069, opened 2015-10-18
Intuitive closing of renaming Panel box

### LINKED ISSUE #1053: Intuitive closing of renaming Panel box
Currently when a user is renaming a panel, the cross sign just next to it is unintuitive in giving an impression of the renaming box being closed while in fact the entire panel would get closed, which can be dangerous. It happened to me the first time I used the renaming panel feature.

pencil_issue2

I was thinking of keeping the pencil in its place and using it as a toggle to go back to the non-naming mode (This is in addition to pressing the escape key which also takes you to the non-naming mode). This is shown below - 
pencil_issue

codemanmeet:
Fixes #1053.

### DIFF
--- a/src/main/java/ui/issuepanel/FilterPanel.java
+++ b/src/main/java/ui/issuepanel/FilterPanel.java
@@ -50,11 +50,14 @@ public abstract class FilterPanel extends AbstractPanel {
     protected Text nameText;
     protected HBox nameBox;
     protected HBox menuBarNameArea;
+    protected HBox menuBarCloseButton;
     protected Label renameButton;
     protected PanelNameTextField renameTextField;
     protected FilterTextField filterTextField;
     
     private String panelName = "Panel";
+    private static final int NAME_FIELD_WIDTH = PANEL_WIDTH - 15;
+    private static final int NAME_DISPLAY_WIDTH = PANEL_WIDTH - 70;
     private UI ui;
 
     protected FilterExpression currentFilterExpression = Qualifier.EMPTY;
@@ -120,13 +123,13 @@ public abstract class FilterPanel extends AbstractPanel {
     
     private Node createPanelMenuBar() {
         menuBarNameArea = createNameArea();
-        HBox menuBarCloseArea = createCloseButton();
+        menuBarCloseButton = createCloseButton();
         
         panelMenuBar = new HBox();
         panelMenuBar.setSpacing(2);
         panelMenuBar.setMinWidth(PANEL_WIDTH);
         panelMenuBar.setMaxWidth(PANEL_WIDTH);
-        panelMenuBar.getChildren().addAll(menuBarNameArea, menuBarCloseArea);
+        panelMenuBar.getChildren().addAll(menuBarNameArea, menuBarCloseButton);
         panelMenuBar.setPadding(new Insets(0, 0, 8, 0));
         return panelMenuBar;
     }
@@ -139,8 +142,8 @@ public abstract class FilterPanel extends AbstractPanel {
         
         nameBox = new HBox();
         nameBox.getChildren().add(nameText);
-        nameBox.setMinWidth(330);
-        nameBox.setMaxWidth(330);
+        nameBox.setMinWidth(NAME_FIELD_WIDTH);
+        nameBox.setMaxWidth(NAME_FIELD_WIDTH);
         nameBox.setAlignment(Pos.CENTER_LEFT);
         
         nameBox.setOnMouseClicked(mouseEvent -> {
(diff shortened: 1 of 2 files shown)


src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -50,6 +50,7 @@
     protected Text nameText;
     protected HBox nameBox;
     protected HBox menuBarNameArea;
+    protected HBox menuBarCloseArea;
```
dariusf: menuBarCloseButton? This is expressly meant to be clicked so it could be given a more specific name.

dariusf:
The panel looks like this now:

screen shot 2015-10-18 at 4 26 59 pm

Since the space for the close button is no longer needed, the field can stretch the entire length of the bar.

codemanmeet:
Sure, should I increase the `PANEL_MAX_NAME_LENGTH` also? Since there is more space to have a longer name?

dariusf:
Yes. Also it would be better to express the new width in terms of PANEL_WIDTH instead of using a literal.

codemanmeet:
Just realised, maximum length of the name should stay the same, since after the rename box is closed the pencil and close buttons come back so we don't want any overlap with them.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -50,11 +50,14 @@
     protected Text nameText;
     protected HBox nameBox;
     protected HBox menuBarNameArea;
+    protected HBox menuBarCloseButton;
     protected Label renameButton;
     protected PanelNameTextField renameTextField;
     protected FilterTextField filterTextField;
     
     private String panelName = "Panel";
+    private static final int TEXT_PANEL_WIDTH = 385;
```
dariusf: `PANEL_WIDTH - 15`

codemanmeet: If I'm not wrong it should be 385 only. The name box gets extended if that number increases so I optimized it to reach till the end.

dariusf: PANEL_WIDTH is a constant in AbstractPanel with value 400. Define this (and the other constant) in terms of that so the width of the field will increase if we ever change the width of the panel.

Also TEXT_PANEL_WIDTH isn't the most specific name for the width of a text field.

codemanmeet: Got it. Changed the name to `TEXT_BOX_WIDTH`.

codemanmeet:
Let me know if it looks good to merge now!

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -50,11 +50,14 @@
...
     protected HBox nameBox;
     protected HBox menuBarNameArea;
+    protected HBox menuBarCloseButton;
     protected Label renameButton;
     protected PanelNameTextField renameTextField;
     protected FilterTextField filterTextField;
     
     private String panelName = "Panel";
+    private static final int TEXT_BOX_WIDTH = PANEL_WIDTH - 15;
+    private static final int NAME_BOX_WIDTH = PANEL_WIDTH - 70;
```
dariusf: These are still very generic names. `NAME_FIELD_WIDTH` and `NAME_DISPLAY_WIDTH`?

src/test/java/guitests/PanelRenameTest.java
```diff
@@ -151,7 +152,34 @@ public void panelRenameTest() {
...
+        PanelNameTextField renameTextField8 = find("#dummy/dummy_col8_renameTextField");
+        renameTextField8.setText(randomName8);
+        push(KeyCode.ENTER);
+        boolean isPresentAfterEdit = exists("#dummy/dummy_col8_closeButton");
+        Text panelNameText8 = find("#dummy/dummy_col8_nameText");
+        assertEquals(true, isPresentBeforeEdit);
+        assertEquals(false, isPresentDuringEdit);
+        assertEquals(true, isPresentAfterEdit);
+        assertEquals(randomName8, panelNameText8.getText());
+        sleep(EVENT_DELAY);
```
dariusf: Don't use sleep in new tests. Find some other way to synchronise. If there might be lag, you can use PlatformEx.waitOnFxThread(), otherwise wait for some condition to be true or some node to be visible.

damithc:
Was this cleared for merging? I don't see the status change or the comment to that effect.

dariusf:
Not explicitly, but I looked it over after getting the notification and there are no more issues.

Anyway @codemanmeet just in case, do wait for the green light in future.

codemanmeet:
Oh I thought I could merge after we spoke on Slack. I will wait for the label in the PR to change to 'ToMerge' in the future.

PR #1071, opened 2015-10-18
Intuitive closing of renaming panel box

### LINKED ISSUE #1053: Intuitive closing of renaming Panel box

codemanmeet:
Fixes #1053.

### DIFF
--- a/src/main/java/ui/issuepanel/FilterPanel.java
+++ b/src/main/java/ui/issuepanel/FilterPanel.java
@@ -142,8 +142,8 @@ public abstract class FilterPanel extends AbstractPanel {
         
         nameBox = new HBox();
         nameBox.getChildren().add(nameText);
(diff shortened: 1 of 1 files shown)


codemanmeet:
@dariusf Tick mark creation has been shifted to a separate issue #1073. This PR relates to #1053.

dariusf:
Tell me when you squash. No need to wait for coverage.

Issue #1058, opened 2015-10-14
Create undo buffer


lejolly:
Which is basically a queue of size 1. 

When an undo object comes in, it waits there till it times out (then it gets acted on) or is canceled. If before the timeout occurs, and a new undo object comes in, both objects get reconciled and the resulting first object gets acted on, while the resulting second object remains in the queue. 

e.g.
Undo Object 1: +a +b +c
Undo Object 2: -a +d

After reconciliation:
Undo Object 1: +b +c (this gets sent off to github)
Undo Object 2: +d (this remains in the queue)
Undoing an action now, will only remove Undo Object 2, which means only "+d" will be removed. "-a" will not be undone. 

Obviously this needs to be linked up with notification controller which should no longer have any undo logic. 

Should fix #1049.

lejolly:
New scope:
1. No delay
2. No eager update
3. Notification for 30 sec

damithc:
3. Why not keep the notification till it is replaced by the subsequent one?

lejolly:
I think it's dangerous to keep it around for too long as the data it holds might get stale if someone else is editing the same issue.

damithc:
That's true. However, stale data is a problem no matter what the duration is, so there should be proper error handling. Once we have proper error handling, might as well keep it open as long as we need, right?

lejolly:
so far the only error handling we're doing is to check if the action to replace labels is successful, what other kind of error handling are you thinking of?

damithc:
Yes, that's enough. Just check if it was successful and give an error message if it wasn't.

PR #1075, opened 2015-10-19
Create object and actions for undo #1057

### LINKED ISSUE #1057: Create object and actions for undo
Actions are invertible. They have a type (e.g. label), field value (e.g. label name), and an action(?) (e.g. add/remove). Actions should also be immutable. 

The object holds the relevant issue and a list of actions. 

LabelPicker has to return a list of actions.

lejolly:
Fixes #1057

Not in use yet, but needed for #1058.

### DIFF
--- a/src/main/java/ui/components/pickers/LabelPicker.java
+++ b/src/main/java/ui/components/pickers/LabelPicker.java
@@ -8,8 +8,8 @@ import javafx.scene.control.Label;
 import javafx.stage.Stage;
 import ui.UI;
 import ui.components.Notification;
+import undo.ChangeLabels;
 import util.DialogMessage;
-import util.GitHubURL;
 import util.events.ShowLabelPickerEventHandler;
 
 import java.util.List;
@@ -49,18 +49,20 @@ public class LabelPicker {
     private void replaceLabels(TurboIssue issue, List<String> labels) {
         List<String> originalLabels = issue.getLabels().stream().sorted().collect(Collectors.toList());
         if (!labels.equals(originalLabels)) {
-            ui.logic.replaceIssueLabelsUI(issue, labels);
+            List<String> resultingLabels = createResultingLabelList(issue, labels);
+            assert(labels.equals(resultingLabels));
+            ui.logic.replaceIssueLabelsUI(issue, resultingLabels);
             Notification undoNotification = new Notification(createInfoOcticon(),
                     "Undo label change(s) for #" + issue.getId() + ": " + issue.getTitle(),
                     "Undo",
-                    () -> ui.logic.replaceIssueLabelsRepo(issue, labels, originalLabels)
-                            .thenApply(success -> postReplaceLabelActions(success, issue)),
+                    () -> ui.logic.replaceIssueLabelsRepo(issue, resultingLabels, originalLabels)
+                            .thenApply(success -> showErrorDialogOnFailure(success, issue)),
                     () -> ui.logic.replaceIssueLabelsUI(issue, originalLabels));
             ui.showNotification(undoNotification);
         }
     }
 
-    private boolean postReplaceLabelActions(Boolean success, TurboIssue issue) {
+    private boolean showErrorDialogOnFailure(Boolean success, TurboIssue issue) {
         if (!success) {
             // if not successful, show error dialog
             Platform.runLater(() -> DialogMessage.showErrorDialog(
@@ -81,6 +83,17 @@ public class LabelPicker {
         label.setPadding(new Insets(0, 0, 5, 0));
         label.getStyleClass().addAll("octicon");
         return label;
+    }
+
+    private List<String> createResultingLabelList(TurboIssue issue, List<String> newLabels) {
+        return (new ChangeLabels(
+                newLabels.stream()
+                        .filter(newLabel -> !issue.getLabels().contains(newLabel))
+                        .collect(Collectors.toList()),
+                issue.getLabels().stream()
+                        .filter(originalLabel -> !newLabels.contains(originalLabel))
+                        .collect(Collectors.toList())))
+                .act(issue).getLabels();
     }
 
 }
--- a/src/main/java/undo/Action.java
+++ b/src/main/java/undo/Action.java
@@ -1,0 +1,25 @@
+package undo;
+
+public interface Action<T> {
+
+    T act(T t);
+
+    T undo(T t);
+
+    @SuppressWarnings("unchecked")
+    default Action invert() {
+        Action that = this;
+        return new Action<T>() {
+            @Override
+            public T act(T t) {
+                return (T) that.undo(t);
+            }
+
+            @Override
+            public T undo(T t) {
+                return (T) that.act(t);
+            }
+        };
+    }
+
+}
--- a/src/main/java/undo/ChangeLabels.java
+++ b/src/main/java/undo/ChangeLabels.java
@@ -1,0 +1,38 @@
+package undo;
+
+import backend.resource.TurboIssue;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ChangeLabels implements Action<TurboIssue> {
+
+    private List<String> addedLabels;
+    private List<String> removedLabels;
+
+    public ChangeLabels(List<String> addedLabels, List<String> removedLabels) {
+        this.addedLabels = new ArrayList<>(addedLabels);
+        this.removedLabels = new ArrayList<>(removedLabels);
+    }
+
+    @Override
+    public TurboIssue act(TurboIssue turboIssue) {
+        List<String> newLabels = new ArrayList<>(turboIssue.getLabels());
+        newLabels.addAll(addedLabels);
+        newLabels.removeAll(removedLabels);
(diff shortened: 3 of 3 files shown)


src/main/java/ui/components/pickers/LabelPicker.java
```diff
@@ -49,6 +50,9 @@ private void showLabelPicker(TurboIssue issue) {
     private void replaceLabels(TurboIssue issue, List<String> labels) {
         List<String> originalLabels = issue.getLabels().stream().sorted().collect(Collectors.toList());
         if (!labels.equals(originalLabels)) {
+            List<Action> actions = getActions(originalLabels, labels);
+            List<String> resultingLabels = getResultingLabels(originalLabels, actions);
```
dariusf: These methods can be named better. This is sort of like computing a diff and applying a diff between sets of labels, right?

Also, does this part do anything? It seems that we still end up with `labels` as the result, and we don't use `actions`.

src/main/java/undo/Action.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Action action = (Action) o;
+        return type == action.type &&
+                value.equals(action.value) &&
+                operation == action.operation;
+    }
+
```
dariusf: hashCode needs to be overridden as well.

dariusf:
Instead of dispatching on an enum, Actions can be polymorphic in what they do:

```java
interface Action<T> {
    // Imperative actions. We may want a return value instead.
    public void act(T t);
    public void undo(T t);
}
```
This way we abstract over the type of operation, and Actions define their own behaviour rather than have it defined where they are used.

One way to use this, for getting the diff of a label set:
```java
class Diff<T> {
    // Not sure if we want to abstract over the type of collection as well
    private final List<Action<Set<T>>> actions;
    private Diff(List<Action<Set<T>>> actions) {
        this.actions = actions;
    }

    public void apply(Set<T> items) {
        actions.forEach(a -> a.act(items));
    }

    public void revert(Set<T> items) {
        actions.forEach(a -> a.undo(items));
    }

    public static <T> Diff<T> of(Set<T> as, Set<T> bs) {
        Stream<Action<Set<T>>> newA = as.stream()
            .filter(a -> !bs.contains(a))
            .map(a -> new Action<Set<T>>() {
                @Override
                public void act(Set<T> before) {
                    before.remove(a);
                }
                @Override
                public void undo(Set<T> after) {
                    after.add(a);
                }
            });
        Stream<Action<Set<T>>> newB = bs.stream()
            .filter(b -> !as.contains(b))
            .map(b -> new Action<Set<T>>() {
                @Override
                public void act(Set<T> before) {
                    before.add(b);
                }
                @Override
                public void undo(Set<T> after) {
                    after.remove(b);
                }
            });
        return new Diff<>(Stream.concat(newA, newB)
    }
}

    Set<String> labels = Sets.newHashSet("a", "b", "c");
    Set<String> labels2 = Sets.newHashSet("d", "b", "e");
    Diff<String> diff = Diff.of(labels, labels2);

    System.out.println(labels); // original

    diff.apply(labels);
    System.out.println(labels); // changed list, same as labels2

    diff.revert(labels);
    System.out.println(labels); // back to original list
}
```
Another Action implementation could create Tasks for adding and removing the labels from GitHub, etc.

I'm not sure of the exact style to use (the above is just an example), but I do think Actions could be more self-contained. Any better ideas?

src/main/java/ui/components/pickers/LabelPicker.java
lejolly: Nope it doesn't do anything for this PR.

lejolly:
perhaps hard code the `.add` and `.remove` in actions (so it'll only deal with lists/sets), making it a class with the constructor specifying the direction (so either add/remove or remove/add)? although i realise most of the potential writes to github would be a replacement of an item rather than a set/list of items, e.g. rename title, assigning milestone/assignee.

dariusf:
Yeah, that's why I was hoping to abstract over the concrete action. It needs to be done in such a way that we can do/undo in general, not just for adding/removing.

Maybe we could construct the action like this? It takes in new and old state, so it knows how to undo. We can add a method for inverting, so we only use one type of object. And here we pass parameters in via constructor rather than rely on lexical references (which in retrospect was bad style).

```java
interface Action<T> {
    public void act(T t);
    public void undo(T t);
    public Action<T> invert();
}

class RenameAction implements Action<TurboIssue> {

    private final String oldName;
    private final String newName;

    public RenameAction(String oldName, String newName) {
        this.oldName = oldName;
        this.newName = newName;
    }

    @Override
    public void act(TurboIssue t) {
        // Create task to change name from oldName to newName
    }

    @Override
    public void undo(TurboIssue t) {
        // Create task to change name from newName to oldName
    }

    @Override
    public Action<TurboIssue> invert() {
        return new RenameAction(newName, oldName);
    }
}
```

Now that I think about it more, it's possible to get `invert` for free. :smile: The cost is some type safety though (we get back an `Action` instead of whatever was implementing `Action`).

```java
interface Action<T> {
    public void act(T t);
    public void undo(T t);

    public default Action<T> invert() {
        Action<T> that = this;
        return new Action<T>() {
            @Override
            public void act(T t) {
                that.undo(t);
            }

            @Override
            public void undo(T t) {
                that.act(t);
            }
        };
    }
}
```

lejolly:
That would work too, just wondering if it makes sense to encapsulate the data changes within a `TurboIssue` or just implement `Action` on either a `String` or a `List<String>`.

dariusf:
Do you have an example of that? Right now the type parameter that the Action interface takes is the type of the value it operates on.

lejolly:
I guess just an `Action<String>` or `Action<List<String>>` since this seems to be the way logic handles write requests now? Actually, why do the `act` and `undo` methods require parameters? Shouldn't they be tied to a specific `TurboIssue` that can be stored when creating the `Action`?

Is this how it's supposed to be used?
1. User does something, say rename an issue
2. `RenameAction` which implements `Action<String>` is created, which holds `TurboIssue`, new and old names
3. `RenameAction` is passed to ui to update it, so ui calls `act` here?
4. `UndoController` holds the `RenameAction` till user closes the notification or it times out, then it calls `act` on it again? If user chooses to undo, then `undo` needs to be called on the ui?

Here we're still separating the ui from github.

dariusf:
The type parameter makes it generic, for example if we want to apply the same action to a bunch of things. But yeah, passing things in via constructor would work as well, and I agree that that should be done for TurboIssues. Maybe it's better to keep things simple.

I hadn't thought about the eager UI updates. If we're going to take care of them here by storing them separately, the Action for it would probably operate on a collection of some kind?

- User adds a label
- An Action to add a label to some collection for buffering eager updates is created, put somewhere the notification bar knows about, and run
- The next UI update displays the changes
- If user clicks undo, the Action is undone. Next UI update displays changes
- If notification bar times out, another Action is created and used to queue something up to be sent to GitHub
- The inverse Action is kept somewhere in case we want to remove the label from GitHub

lejolly:
So how many queues/buffers are we having? Do we need a queue for ui updates or can we just run them immediately/use `Platform.runLater`? I don't see the need for more than one queue, which will basically just hold `Actions` which are pending github writes since I assume all ui updates will be sent off immediately? Unless there's a reason to hold them back?

Will each `Action` then be responsible for updating the ui and github or are we separating that (e.g. different `Actions` for ui and github)?

dariusf:
I was assuming we would hold updates back. It allows for some optimisation (we can batch updates, or not send at all if the user undoes right away) but isn't essential.

Either way, since there will always be some delay between the write being sent and the updates coming back, we will need a second buffer for the eager UI update. Mutating the model to add stuff without it actually coming from GitHub leads to the consistency problems we have/forces us to reason about order of updates, so I was hoping to avoid that.

If we have two queues in different places, we should have two sets of actions, otherwise only one is needed.

lejolly:
So how should we deal with updating the ui with updates that have not been sent to github? Should the ui support the display of information that is not within the model?

dariusf:
Yes, those extra updates would be stored separately and applied before the model is displayed. I guess they would have a condition for removal as well, so they would know when the actual update comes in and be removed then.

lejolly:
right, ok the scope of this seems to be way larger than this pr, so we kinda need to take a step back to decide what needs to be changed? unfortunately, i don't think i can commit much time to implementing these changes for at least this and next week. 

unless you just want to focus now on getting `Actions` out?

dariusf:
This kind of large rewrite generally doesn't break down nicely into weekly issues. We can get the actions out now but might have to re-evaluate them as we go.

So the version we are using is the last one I posted, minus the type parameter and the parameters to `act` and `undo`?

lejolly:
ok, how's this?

lejolly:
ok, does this work?

dariusf:
Okay, looks good. Can merge if everything passes. We'll only test these later, right?

lejolly:
yup, ok.

PR #1079, opened 2015-10-19
User can load the related issue in bView when a PR is selected in pView using shortcut #1046

### LINKED ISSUE #1045: Users can load related PR using a keyboard shortcut
... while the issue is selected in pView, so that user can navigate to the relevant PR easily.

### LINKED ISSUE #1046: User can load the related issue in bView when a PR is selected in pView using shortcut
... so that user can easily navigate to the issue being fixed by the PR

rd1992:
Fixes #1046, fixes #1045

### DIFF
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -18,6 +18,8 @@ import util.events.testevents.JumpToCommentEvent;
 import util.events.testevents.SendKeysToBrowserEvent;
 
 import java.io.*;
+import java.util.NoSuchElementException;
+import java.util.Optional;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
 
@@ -493,5 +495,18 @@ public class BrowserComponent {
         if (GitHubURL.isPullRequestLoaded(getCurrentUrl())) {
             driver.findElement(By.xpath("//a[@data-container-id='" + tabName + "_bucket']")).click();
         }
+    }
+
+    public Optional<Integer> getPRNumberFromIssue() {
+        // will only get the first PR number if multiple PR's are mentioned
+        try {
+            WebElement element = driver.findElement(By.xpath("//div[contains(@id, 'ref-pullrequest')]"));
+            return Optional.of(Integer.parseInt(element.findElement(By.xpath(".."))
+                    .findElement(By.xpath(".//h3/a/span"))
+                    .getAttribute("innerHTML").substring(1)));
+        } catch (NoSuchElementException e) {
+            logger.info("no PR mention found");
+        }
+        return Optional.empty();
     }
 }
--- a/src/main/java/ui/listpanel/ListPanel.java
+++ b/src/main/java/ui/listpanel/ListPanel.java
@@ -263,18 +263,15 @@ public class ListPanel extends FilterPanel {
     }
 
     private void showRelatedIssueOrPR() {
-        if (listView.getSelectedItem().isPresent()) {
-            TurboIssue issue = listView.getSelectedItem().get();
-            Optional<Integer> relatedIssueNumber = listView.getSelectedItem().get().isPullRequest()
-                ? GithubURLPageElements.extractIssueNumber(listView.getSelectedItem().get().getDescription())
-                : ui.getBrowserComponent().getPRNumberFromIssue();
-            if (relatedIssueNumber.isPresent()) {
-                ui.triggerEvent(
-                        new IssueSelectedEvent(issue.getRepoId(), 
-                           relatedIssueNumber.get(), panelIndex, issue.isPullRequest())
-                );
-            }
-        }
+        if (!listView.getSelectedItem().isPresent()) return;
+        TurboIssue issue = listView.getSelectedItem().get();
+        Optional<Integer> relatedIssueNumber = listView.getSelectedItem().get().isPullRequest()
+            ? GithubURLPageElements.extractIssueNumber(listView.getSelectedItem().get().getDescription())
+            : ui.getBrowserComponent().getPRNumberFromIssue();
+        if(!relatedIssueNumber.isPresent()) return;
+        ui.triggerEvent(
+            new IssueSelectedEvent(issue.getRepoId(), relatedIssueNumber.get(), panelIndex, issue.isPullRequest())
+        );
     }
 
     private ContextMenu setupContextMenu() {
--- a/src/main/java/ui/listpanel/ListPanel.java
+++ b/src/main/java/ui/listpanel/ListPanel.java
@@ -20,6 +20,7 @@ import ui.components.IssueListView;
 import ui.components.KeyboardShortcuts;
 import ui.issuepanel.FilterPanel;
 import ui.issuepanel.PanelControl;
+import util.GithubURLPageElements;
 import util.KeyPress;
 import util.events.IssueSelectedEvent;
 import util.events.ShowLabelPickerEvent;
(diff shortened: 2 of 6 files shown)


rd1992:
Haven't added documentation yet. Only thing needed now is to decide which key to use to load related issue. Also this can be extended to fix #1045. @damithc what key would you like to have?
@dariusf have a look

damithc:
Can't think of a good one atm. any ideas?

src/main/java/backend/resource/TurboIssue.java
```diff
@@ -148,6 +153,18 @@ public TurboIssue(String repoId, Issue issue) {
         this.metadata = IssueMetadata.empty();
         this.repoId = repoId;
         this.markedReadAt = Optional.empty();
+        this.relatedIssueOrPRUrl = setrelatedIssueOrPRUrl(issue);
+    }
+
+    private Optional<String> setrelatedIssueOrPRUrl(Issue issue) {
+        if (isPullRequest(issue) && issue.getBodyText() != null) {
+            Pattern p = Pattern.compile("Fixes <a href=\"(.*?)\"", Pattern.DOTALL);
```
dariusf: Do we need DOTALL for this? There shouldn't be a newline in the URL. Is it possible to use `[^"]*` over `.*?`, so it's more explicit?

There are [lots of other strings GitHub accepts](https://help.github.com/articles/closing-issues-via-commit-messages/) to close issues.

The regex portion can probably be put in some GitHub-specific class, and should be tested.

src/main/java/backend/resource/TurboIssue.java
```diff
@@ -148,6 +153,18 @@ public TurboIssue(String repoId, Issue issue) {
...
+            Matcher m = p.matcher(issue.getBodyHtml());
+            while (m.find()) {
```
dariusf: An if statement can be used here.

dariusf:
I'm not sure if we should persist the URL. It seems fragile to save URLs (in general; ignoring the fact that GitHub's URLs are unlikely to ever change), plus we have to parse it if we want to get meaningful info out of it, compared to just saving that info and using that to build the URL.

An alternative is to get check the body text instead, and turn that into a URL with the issue number and a GitHubURL method. This won't work with [cross-repo issue references](https://github.com/blog/967-github-secrets) unless we account for that as well though...

rd1992:
@dariusf so if we want to account for cross-repo issue reference then the best way would be to go with parsing the html. Also checking the body text is fragile if the user mentions another issue in the issue body. What do you think?

dariusf:
My main reservation with saving the URL is that it won't be easy to extend in future, as we don't have a way to change schema. Parsing the body doesn't require us to save anything.

We should probably discuss this further tomorrow.

Multiple issues: I guess we can take the first one for now. If we want to support multiple linked issues we'll need a UI to choose between them as well.

rd1992:
@dariusf hmm okay cool. So you are only against storing the URL right? I mean you still have to parse the html to get the related issue.

dariusf:
Yeah, mainly for the reason that we can't change schema. Avoiding saving things isn't a long-term solution though.

damithc:
We still don't have a key combination for this. `g,e`? as in 'go to r**e**lated'? It's not used for anything else right?

rd1992:
@damithc yup. using `g, e` now. Updated docs also

rd1992:
We could use `e` but its also the default to mark issue as read. If you have other suggestions i can implement that. Otherwise can stick to `e`

damithc:
I can't think of a better alternative.

src/test/java/tests/GithubURLPageElementsTest.java
```diff
@@ -0,0 +1,18 @@
...
+
+import util.GithubURLPageElements;
+
+public class GithubURLPageElementsTest {
+    @Test
+    public void gitHubURLPageElementTest() {
+        assertEquals("123", GithubURLPageElements.extractIssueNumber("fixes #123").get());
+        assertEquals("123", GithubURLPageElements.extractIssueNumber("Closed #123. This is dummy").get());
+        assertEquals("123", GithubURLPageElements.extractIssueNumber("refer to #144. Closed #123. This is du").get());
+        assertNotEquals("131", GithubURLPageElements.extractIssueNumber("fixes #123").get());
```
damithc: What's the purpose of this line?

src/main/java/browserview/BrowserComponent.java
```diff
@@ -494,4 +496,15 @@ public void switchToTab(String tabName) {
             driver.findElement(By.xpath("//a[@data-container-id='" + tabName + "_bucket']")).click();
         }
     }
+
+    public Optional<String> getPRNumberFromIssue() {
+        try {
+            WebElement element = driver.findElement(By.xpath("//div[contains(@id, 'ref-pullrequest')]"));
+            return Optional.of(element.findElement(By.xpath("..")).findElement(By.xpath(".//h3/a/span"))
+                    .getAttribute("innerHTML").substring(1));
```
dariusf: Will this get the first PR if an issue mentions a few? Maybe we can add a comment saying what it does.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -148,7 +149,8 @@ private void setupListView() {
 
     private void setupKeyboardShortcuts() {        
         addEventHandler(KeyEvent.KEY_PRESSED, event -> {
-            if (KeyboardShortcuts.markAsRead.match(event)) {
+            // Temporary fix for now since markAsRead and Show Related Issue/PR have same keys
+            if (KeyboardShortcuts.markAsRead.match(event) && !SHOW_RELATED_ISSUE_OR_PR.match(event)) {
```
dariusf: This confused me until I saw the next part. :stuck_out_tongue: Maybe you could add to the comment, saying it will only work if the key for marking as read is not the default.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -248,9 +250,29 @@ private void setupKeyboardShortcuts() {
                     break;
                 }
             }
+            if (SHOW_RELATED_ISSUE_OR_PR.match(event) && ui.getBrowserComponent().isCurrentUrlIssue()) {
+                if (KeyPress.isValidKeyCombination(GOTO_MODIFIER.getCode(), event.getCode())) {
+                    showRelatedIssueOrPR();
+                } else { // only for default. can remove if default key for MARK_AS_READ_CHANGES
```
dariusf: You might want to check the key for mark as read here also. If it's changed and we press E, this will still trigger.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -248,9 +250,29 @@ private void setupKeyboardShortcuts() {
...
+                    markAsRead();
+                }
+            }
         });
     }
 
+    private void showRelatedIssueOrPR() {
+        TurboIssue issue = listView.getSelectedItem().get();
```
dariusf: You have to check here if there is a selected item. This would be like an NPE otherwise.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -248,9 +250,29 @@ private void setupKeyboardShortcuts() {
...
+        Optional<String> relatedIssueNumber = listView.getSelectedItem().get().isPullRequest()
+            ? GithubURLPageElements.extractIssueNumber(listView.getSelectedItem().get().getDescription())
+            : ui.getBrowserComponent().getPRNumberFromIssue();
+        if (relatedIssueNumber.isPresent()){
+            ui.triggerEvent(
+                    new IssueSelectedEvent(issue.getRepoId(), 
+                       Integer.parseInt(relatedIssueNumber.get()), panelIndex, issue.isPullRequest())
```
dariusf: The `parseInt` could be done in `extractIssueNumber` instead, then we can deal with an int here.

dariusf:
Not sure why commits from other PRs are showing up here. You might need to rebase.

rd1992:
@dariusf done all the changes

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -248,9 +251,32 @@ private void setupKeyboardShortcuts() {
...
+                // only for default. can remove if default key for MARK_AS_READ_CHANGES
+                } else if (KeyboardShortcuts.markAsRead.match(event)) {
+                    markAsRead();
+                }
+            }
         });
     }
 
+    private void showRelatedIssueOrPR() {
+        if (listView.getSelectedItem().isPresent()) {
```
dariusf: Maybe we could reduce nesting a bit with a guard clause here.

dariusf:
Could we cover getPRNumberFromIssue using the stub?

dariusf:
Can merge when it passes. We're not going to cover the new BrowserComponent method.

PR #1087, opened 2015-10-24
Panel UI becomes blank when switching repo #1056

### LINKED ISSUE #1056: Panel UI becomes blank when switching repo
This happens on both Windows and Linux without any filter, and I think might be a separate problem from #934 . Step to reproduce:

1. Open `teammates/repo` using repo selector box.
2. Switch to `hubturbo/hubturbo` using repo selector box. Right after that, try to scroll the Panel UI (by using the middle mouse button for example).
3. The Panel UI will be blank and a loud crash occurs (check the log).

This could have happened before, but may be I only realize it recently since it takes more time to switch repos now so I get impatient and scroll my mouse around.

ndt93:
Fixes #1056 

This only prevents the crash but doesn't solve the root of the problem, which is the race between loading and removing repos. I've tried various async solutions but they can still fail in some cases (switching rapidly twice between hubturbo and teammates, race problem happens about 1 in 10 trials). This is due to the interleaving between the entire repo loading/updating and repo removal process: 

<pre>
Load:    --------[used]--[load models]--------[update UI]--->
Remove:  [unused]---------------------[delete]-------------->
</pre>

Problem happens e.g. when `hubturbo` is in `[unused]` but then becomes `[used]` shortly after due to the rapid switching.

One solution is to make the **entire** load/update process mutually exclusive with the removal process. The order doesn't matter, supposed for a certain `used` repo:

* If `Remove` -> `Load`, then the repo is properly present.
* If `Load` -> `Remove`, then the repo will not be in the `[unused]` set, hence is not removed.

I could open a separate issue for this, any opinion?

### DIFF
--- a/src/main/java/ui/listpanel/ListPanelCell.java
+++ b/src/main/java/ui/listpanel/ListPanelCell.java
@@ -5,12 +5,15 @@ import backend.resource.Model;
 import backend.resource.TurboIssue;
 import javafx.geometry.Pos;
 import javafx.scene.control.ListCell;
+import org.apache.logging.log4j.Logger;
+import util.HTLog;
 
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 
 public class ListPanelCell extends ListCell<TurboIssue> {
+    private static final Logger logger = HTLog.get(ListPanelCell.class);
 
     private final IModel model;
     private final int parentPanelIndex;
@@ -40,8 +43,13 @@ public class ListPanelCell extends ListCell<TurboIssue> {
         }
         this.issue = issue;
         Optional<Model> currentModel = model.getModelById(issue.getRepoId());
-        assert currentModel.isPresent() : "Invalid repo id " + issue.getRepoId()
-            + " for issue " + issue.getId();
+        if (!currentModel.isPresent()) {
+            // TODO: see issue #1089
+            logger.error("Model is not present for issue " + issue.getId() + " at " + issue.getRepoId());
+            setGraphic(getGraphic());
+            return;
+        }
+
         setGraphic(new ListPanelCard(currentModel.get(), issue, parent, issuesWithNewComments));
         this.setId(issue.getRepoId() + "_col" + parentPanelIndex + "_" + issue.getId());
     }


ndt93:
@HansNewbie Can help review this? Thanks.

HansNewbie:
Confirm me if I am wrong: when rapid switching happens, it could be that the new model is just loaded but then it will be instantly deleted since the delete operation is late.

I don't get your solution. Do you mean mutually inexclusive? LIke the queue for `Load` and `Remove` is shared, so whichever enter the queue last should be executed? I think right now the processes are mutually exclusive, hence the problem.

ndt93:
Sorry I forget to clarify that I was talking about the solution of deferring the repo deletion until the replacement repo has been loaded. The original problem is not related to concurrency since deletion is executed right away in the main thread.

So, the happens when I try a solution of deferring the delete until the replacement repo has been loaded. A pair of `Load/Remove` actions is sequential in that case but interleave with other pairs of `Load/Remove` actions.

To summarize, we need to defer the deletion of a repo until the replacement repo has been loaded but also need to make `Load` and `Remove` mutually exclusive. A synchronized queue is one of the implementation to make that happen. We need to figure out how to encapsulate then whole loading or updating process as well as refactor all those places that trigger a load or update.

HansNewbie:
Okay, I think can open a separate issue for that. This PR can be merged as well.

dariusf:
Rather than queue the operations, I think a better solution is to prevent them from occurring together. Removal should fail if an update is in progress, and an update should not start if removal is in progress.

If we want to support deletion, some kind of mutual exclusion is inevitable. Another approach to consider is to not support deletion. I'm still not fully convinced that it helps.

- Model objects don't dominate memory usage.
- We do want to keep cached repos (on disk) for when they are next used.
- There are lots of stray references to model objects everywhere. If we can somehow guarantee that all references besides the one in MultiModel are transient/weak, this would be useful, but until then...

The last point is interesting. We may be able to enforce it statically with a good MultiModel API.

ndt93:
Ok, I will open an issue for this and copy our discussion here over.

dariusf:
Also, we should revert this when the root cause is actually fixed? I feel like that assertion is valuable. We should at the very least log when we encounter that situation, because it's an error on our part, not a runtime error.

HansNewbie:
I think so, too, on reverting this once this model deletion is cleared up.

ndt93:
Sure, I will put in a log statement first. Another thing is that the original problem is actually still related to concurrency since the `[delete]` step of `Remove` interleaves and happen before the `[update UI]` step of `Load` which happens asynchronously.

HansNewbie:
@ndt93 changed status to ongoing for now since you are adding log statement.

dariusf:
The repo would have to be 'locked' as long as an update to it is in progress then. The synchronisation point I have in mind is at the triggering of both operations. When the repo is locked, removal should just be a no-op. If we are deleting like we are now, it should be quickly detected as unused again (when the set of used repos changes).

ndt93:
@HansNewbie I've added in the log and also opened a related issue for this.

HansNewbie:
@ndt93 can add a comment behind the logger sth like `// TODO see issue 1089`? So that when people check the source of the logger they are not alarmed.

HansNewbie:
Looks good.

Issue #1078, opened 2015-10-19
Change repoId representation from raw string to object for safe comparison


HansNewbie:
User input on repo id might have wrong letter casing, but `Model` will have correct letter casing since it was taken directly from GitHub. The due diligence falls on the developer to ensure all checking of repo id ignores letter case, which can be easily missed (for example in using `List<String>.contains(repoId)`). Repo id representation should be made an object of itself which implement its own comparison.

This is to prevent #1060 from happening again.

PR #1090, opened 2015-10-24
Model is deleted when repo letter casing is wrong #1060

### LINKED ISSUE #1060: Model is deleted when repo letter casing is wrong
e.g. default repo is `TEAMMATES/repo` and filter is `(repo:TEAMMATES/repo || repo:hubturbo/hubturbo)`. Everytime something is added to the filter, the model of HubTurbo will be deleted and reloaded.

Will cause same issue with #1056, except that the only blank issues would be HubTurbo's issue and TEAMMATES's issue would show up just fine.

HansNewbie:
Fixes #1060

### DIFF
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -187,10 +187,14 @@ public class Logic {
         return repoIO.removeRepository(repoId);
     }
 
+    /**
+     * Recommended Pre-condition: normalize reposInUse to lower case
+     *                           - using Utility.convertSetToLowerCase()
+     */
     public void removeUnusedModels(Set<String> reposInUse) {
-        models.toModels().stream().map(Model::getRepoId).
-                filter(repoId -> !reposInUse.contains(repoId)).
-                forEach(repoIdNotInUse -> models.removeRepoModelById(repoIdNotInUse));
+        models.toModels().stream().map(Model::getRepoId)
+                .filter(repoId -> !reposInUse.contains(repoId.toLowerCase()))
+                .forEach(repoIdNotInUse -> models.removeRepoModelById(repoIdNotInUse));
     }
 
     /**
--- a/src/main/java/ui/MenuControl.java
+++ b/src/main/java/ui/MenuControl.java
@@ -14,8 +14,8 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import prefs.PanelInfo;
 import prefs.Preferences;
-import ui.issuepanel.FilterPanel;
 import ui.issuepanel.PanelControl;
+import util.Utility;
 import util.events.*;
 
 import java.util.List;
@@ -23,7 +23,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import static ui.components.KeyboardShortcuts.*;
 
@@ -336,9 +335,10 @@ public class MenuControl extends MenuBar {
     private void updateRepoRemoveList(Menu remove) {
         remove.getItems().clear();
 
-        Set<String> currentlyUsedRepos = ui.getCurrentlyUsedRepos();
-        Set<String> removableRepos = ui.logic.getStoredRepos();
-        removableRepos.removeAll(currentlyUsedRepos);
+        Set<String> currentlyUsedRepos = Utility.convertSetToLowerCase(ui.getCurrentlyUsedRepos());
+        Set<String> removableRepos = ui.logic.getStoredRepos()
+                .stream().filter(repoId -> !currentlyUsedRepos.contains(repoId.toLowerCase()))
+                .collect(Collectors.toSet());
 
         for (String repoId : removableRepos) {
             MenuItem removeItem = new MenuItem(repoId);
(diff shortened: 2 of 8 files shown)


HansNewbie:
This PR corrects 2 things:
- Model deletion should be case insensitive (as per issue).
- `Repos` > `Remove` listing is now also case insensitive 
  - this is a bug whereby the list might have double entries (with one not being marked as being `in use`) if using `repo` qualifier with the content being in wrong letter casing.

Wrote tests to ensure case sensitive problem does not happen again.

HansNewbie:
@ndt93 can help me review this PR?

Basically the two things I am tackling is:
Given that there is a repo in the system (say `HubTurbo/HubTurbo`) but it is not default repo, the following cases should not happen:
- When using `(repo:dummy/dummy || repo:hubturbo/hubturbo)` - notice the different letter casing - the model of HubTurbo/HubTurbo should not be loaded and deleted after every filter changing. This happened because the model loaded will follow the original `repoId`, which is with big `H` and `T`, but the one passed to remove model method is with small `h` and `t`.
- Different repo letter casing will cause them to be listed in `Repos > Remove` separately. Say for the same given case, my filter is `repo:hubturbo/hubturbo` - in different letter casing. This will cause `Repos > Remove` to list `HubTurbo/HubTurbo` as removable (i.e. not in use) while `hubturbo/hubturbo` as non-removable. Hence, user might delete the JSON of `HubTurbo/HubTurbo` even though it's in use.

I am running the tests locally now, will notify again if it all passes.

HansNewbie:
@ndt93 it passed local tests.

src/main/java/backend/Logic.java
```diff
@@ -188,9 +188,9 @@ public String getDefaultRepo() {
     }
 
     public void removeUnusedModels(Set<String> reposInUse) {
-        models.toModels().stream().map(Model::getRepoId).
-                filter(repoId -> !reposInUse.contains(repoId)).
-                forEach(repoIdNotInUse -> models.removeRepoModelById(repoIdNotInUse));
+        models.toModels().stream().map(Model::getRepoId)
+                .filter(repoId -> !reposInUse.contains(repoId.toLowerCase()))
+                .forEach(repoIdNotInUse -> models.removeRepoModelById(repoIdNotInUse));
```
ndt93: Although we take care of `toLowerCase` inside, it could be helpful to add a method's comment recommending at `reposInUse` should be normalized as a precondition.

src/main/java/ui/UI.java
```diff
@@ -350,14 +351,18 @@ private Parent createRootNode() {
     public Set<String> getCurrentlyUsedRepos() {
         Set<String> currentlyUsedRepos = new HashSet<>();
         String defaultRepo = logic.getDefaultRepo();
-        currentlyUsedRepos.add(defaultRepo);
         currentlyUsedRepos.addAll(panels.getRepositoriesReferencedOnAllPanels());
+        if (!currentlyUsedRepos.stream().map(String::toLowerCase)
+                .collect(Collectors.toSet()).contains(defaultRepo.toLowerCase())) {
+            currentlyUsedRepos.add(defaultRepo);
+        }
```
ndt93: `getCurrentlyUsedRepos` could have a method's comment mentioning a post-condition that all repo ids are normalized (spaces removed/lowercase etc.). Maybe we should have some sorts of a separate `normalize` method in `Utility` that does all the repo clean up. But this all these won't be necessary anymore if we solve #1078

src/main/java/ui/MenuControl.java
```diff
@@ -336,9 +334,11 @@ private MenuItem createRefreshMenuItem() {
...
         remove.getItems().clear();
 
-        Set<String> currentlyUsedRepos = ui.getCurrentlyUsedRepos();
-        Set<String> removableRepos = ui.logic.getStoredRepos();
-        removableRepos.removeAll(currentlyUsedRepos);
+        Set<String> currentlyUsedRepos = ui.getCurrentlyUsedRepos()
+                .stream().map(String::toLowerCase).collect(Collectors.toSet());
+        Set<String> removableRepos = ui.logic.getStoredRepos()
+                .stream().filter(repoId -> !currentlyUsedRepos.contains(repoId.toLowerCase()))
+                .collect(Collectors.toSet());
```
ndt93: If `getCurrentlyUsedRepos` already provides some guarantees, are all the extra `toLowerCase` necessary?

src/main/java/ui/UI.java
HansNewbie: Yep, will make normalize in Utility and add comment on it that should be removed once #1078 is done.

src/main/java/ui/MenuControl.java
HansNewbie: So `getCurrentlyUsedRepos` does not normalize the names to lower cases as we would like to show it as what the user has input. Just a few lines below this we will call non-normalize `getCurrentlyUsedRepos`. `toLowerCase` for `repoId` is called (instead of mapped) because we would like to use the original `repoId` casing instead of the normalized one as per the reason mentioned.

src/main/java/backend/Logic.java
HansNewbie: Added comment.

ndt93:
All right, this should be good to merge then.

PR #1092, opened 2015-10-25
Open issues is coloured green after some clicking

### LINKED ISSUE #1088: Open issues is coloured green after some clicking
screen shot 2015-10-24 at 5 29 06 pm

Not sure if it only happens to me. If I click on some issues, the green colouring is messed up, at least when you start scrolling up and down.

codemanmeet:
Fixes #1088.

### DIFF
--- a/src/main/java/ui/listpanel/ListPanelCell.java
+++ b/src/main/java/ui/listpanel/ListPanelCell.java
@@ -38,9 +38,7 @@ public class ListPanelCell extends ListCell<TurboIssue> {
         if (issue == null) {
             return;
         }
-        if (!issue.isOpen()){
-            getStyleClass().add("issue-cell-closed");
-        }
+        updateStyleToMatchStatus(issue);
         this.issue = issue;
         Optional<Model> currentModel = model.getModelById(issue.getRepoId());
         if (!currentModel.isPresent()) {
@@ -52,6 +50,20 @@ public class ListPanelCell extends ListCell<TurboIssue> {
 
         setGraphic(new ListPanelCard(currentModel.get(), issue, parent, issuesWithNewComments));
         this.setId(issue.getRepoId() + "_col" + parentPanelIndex + "_" + issue.getId());
+    }
+
+    private void updateStyleToMatchStatus(TurboIssue issue){
+        final String closedStyle = "issue-cell-closed";
+        boolean isCurrentStyleClosed = getStyleClass().contains(closedStyle);
+        if (!issue.isOpen()){
+            if (!isCurrentStyleClosed) {
+                getStyleClass().add(closedStyle);
+            }
+        } else {
+            if (isCurrentStyleClosed) {
+                getStyleClass().remove(closedStyle);
+            }
+        }
     }
 
     public List<String> getIssueLabels() {


codemanmeet:
@damithc Scrolling doesn't make issues green anymore! Let me know if it's still happening!

damithc:
Were you able to reproduce it before? Cannot reproduce it anymore after this fix? If the answer is yes t both then you can consider it fixed. 
W.r.t to the code, although the code is short, it sill takes some time to understand what you are doing there. Try to make it more self explanatory. some options: extract into a method with a self-explanatory name, add comments, replace negation by adding another method (e.g. `isClosed()`) etc.

codemanmeet:
Yup I was facing the same issue when scrolling through the list view. I am not facing that issue anymore. I have added an important comment that helped me solve the issue to prevent the green color from appearing for open issues.

src/main/java/ui/listpanel/ListPanelCell.java
```diff
@@ -39,7 +39,12 @@ public void updateItem(TurboIssue issue, boolean empty) {
             return;
         }
         if (!issue.isOpen()){
-            getStyleClass().add("issue-cell-closed");
+            if (!getStyleClass().contains("issue-cell-closed")) {
+                getStyleClass().add("issue-cell-closed");
+            }
```
damithc: What does this statement do?

src/main/java/ui/listpanel/ListPanelCell.java
```diff
@@ -39,7 +39,12 @@ public void updateItem(TurboIssue issue, boolean empty) {
...
+        } else {
+        // prevents the closed issue style class from being applied for open issues when scrolling through list view
+            getStyleClass().remove("issue-cell-closed");
```
damithc: What does this statement do? Don't explain it in terms of a bug. The bug shouldn't have been there in the first place so it should not be used to explain code.

src/main/java/ui/listpanel/ListPanelCell.java
```diff
@@ -39,7 +39,12 @@ public void updateItem(TurboIssue issue, boolean empty) {
             return;
         }
         if (!issue.isOpen()){
```
damithc: Can you extract this entire block into a method?

damithc:
Remember to change the status label when the status of the PR changes.

src/main/java/ui/listpanel/ListPanelCell.java
```diff
@@ -54,6 +52,18 @@ public void updateItem(TurboIssue issue, boolean empty) {
         this.setId(issue.getRepoId() + "_col" + parentPanelIndex + "_" + issue.getId());
     }
 
+    private void updateStyleForClosedIssues(TurboIssue issue){
```
damithc: Seems this is not just about closed issues (in your `else` clause you change the style of open issues).
Shouldn't it be called something like `updateStyleToMatchStatus`?

The logic should be like this?
if open, apply open style
else, apply closed style.

damithc: Alternative logic (if open issues do not require styling):
if open but styled as closed, remove closed style
else if closed but styled as open, add closed style.

codemanmeet: Yup, open issues do not require styling since it's *derived* from JavaFX source files for ListView. It would be pretty cumbersome to overwrite a lot of CSS associated with the default list view. I have refactored the code to check whether the open issues contain the closed styling. If so, it is removed. I have also renamed the function.

src/main/java/ui/listpanel/ListPanelCell.java
```diff
@@ -54,6 +52,20 @@ public void updateItem(TurboIssue issue, boolean empty) {
...
+            if (!getStyleClass().contains("issue-cell-closed")) {
+                getStyleClass().add("issue-cell-closed");
+            }
+        } else {
+            // For open issues, closed issue styling is removed if it is contained in the cell.
+            if (getStyleClass().contains("issue-cell-closed")) {
+                getStyleClass().remove("issue-cell-closed");
+            }
+        }
+    }
```
damithc: The style class string `"issue-cell-closed"` and `getStyleClass()` is repeated 4 times. May be declare as local variables? Something like this?
```
currentStyleClass = ...
String closedStyle = "issue-cell-closed";
boolean isCurrentStyleClosed = getStyleClass().contains(closedStyle);

if(issue.isOpen() && isCurrentStyleClosed ) {
    currentStyleClass.remove(closedStyle);
```

Now that the code is clearer, may be no need for the comments?
Get @lejolly to have a look after revising.

damithc: Do the suggested changes before getting @lejolly to review.

lejolly:
looks good, you can merge it after squashing your commits.

src/main/java/ui/listpanel/ListPanelCell.java
```diff
@@ -54,6 +54,18 @@ public void updateItem(TurboIssue issue, boolean empty) {
...
+        if (!issue.isOpen()){
+            if (!isCurrentStyleClosed) {
+                getStyleClass().add(closedStyle);
+            }
+        } else {
+            if (isCurrentStyleClosed) {
+                getStyleClass().remove(closedStyle);
+            }
+        }
+    }
```
damithc: I'm not sure if those variables should be at instance level. Can't we keep them inside the method?

src/main/java/ui/listpanel/ListPanelCell.java
```diff
@@ -20,6 +20,8 @@
     private final ListPanel parent;
     private final HashSet<Integer> issuesWithNewComments;
     private TurboIssue issue;
+    private final String closedStyle = "issue-cell-closed";
+    private boolean isCurrentStyleClosed = getStyleClass().contains(closedStyle);
```
lejolly: yes, it makes more sense if this was a method instead.

damithc: rather, I meant the two variables should be declared inside the method.  It's safer to keep the scope of variables as narrow as possible.

lejolly: oh alright then.

codemanmeet: done!

lejolly:
could you squash your commits?

codemanmeet:
Sure!

damithc:
No more comments from me.

Issue #1077, opened 2015-10-19
Prompt user to save current board before auto-creating a board


HansNewbie:
As of now, auto-creating a board will clear the panels then execute all panel adding then save it as a board. The problem is that user's current board will be cleared and it's not saved.

We should prompt user if they would like to save their current board before auto-creating a board.

Issue #1074, opened 2015-10-19
Abstract Board Save, Open, Delete functionality from MenuControl


HansNewbie:
Right now all functionality is built inside `MenuControl` but this makes it difficult to add functionality to `Board`, e.g. Board Auto Creator needs a reference to `MenuControl` to save a board which does not make sense.

HansNewbie:
Once this is done, saving board functionality should be abstracted from #1067 on `BoardAutoCreator`-milestone.

dariusf:
Yes, this is a good idea.

PR #1093, opened 2015-10-27
User can generate a 'work allocation' board automatically #1031

### LINKED ISSUE #1031: User can generate a 'work allocation' board automatically
... so that user can avoid the work required to create such a board manually.

HansNewbie:
Fixes #1031

### DIFF
--- a/src/main/java/ui/BoardAutoCreator.java
+++ b/src/main/java/ui/BoardAutoCreator.java
@@ -1,10 +1,12 @@
 package ui;
 
 
+import backend.resource.TurboUser;
 import javafx.scene.control.Menu;
 import javafx.scene.control.MenuItem;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import prefs.PanelInfo;
 import prefs.Preferences;
 import ui.issuepanel.FilterPanel;
 import ui.issuepanel.PanelControl;
@@ -12,16 +14,22 @@ import util.DialogMessage;
 import util.Utility;
 import util.events.BoardSavedEvent;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.stream.Collectors;
 
 public class BoardAutoCreator {
     private static final Logger logger = LogManager.getLogger(MenuControl.class.getName());
     private static final String MILESTONES = "Milestones";
+    private static final String WORK_ALLOCATION = "Work Allocation";
+    private static final int MAX_WORK_ALLOCATION_PANELS = 5;
 
+    private final UI ui;
     private final PanelControl panelControl;
     private final Preferences prefs;
 
-    public BoardAutoCreator(PanelControl panelControl, Preferences prefs) {
+    public BoardAutoCreator(UI ui, PanelControl panelControl, Preferences prefs) {
+        this.ui = ui;
         this.panelControl = panelControl;
         this.prefs = prefs;
     }
@@ -33,6 +41,10 @@ public class BoardAutoCreator {
         milestone.setOnAction(e -> createMilestoneBoard());
         autoCreate.getItems().add(milestone);
 
+        MenuItem workAllocation = new MenuItem(WORK_ALLOCATION);
+        workAllocation.setOnAction(e -> createWorkAllocationBoard());
+        autoCreate.getItems().add(workAllocation);
+
         return autoCreate;
     }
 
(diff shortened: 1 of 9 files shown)


src/test/java/guitests/BoardAutoCreatorTest.java
```diff
@@ -72,4 +72,43 @@ public void milestoneBoardAutoCreationTest() {
...
+
+        click("Boards");
+        push(KeyCode.DOWN); // Save
+        push(KeyCode.DOWN); // Save As
+        push(KeyCode.DOWN); // Open
+        push(KeyCode.DOWN); // Delete
+        push(KeyCode.DOWN); // Auto-create
+        push(KeyCode.RIGHT); // Open Auto-create, Milestones
+        push(KeyCode.DOWN); // Work Allocation
+        push(KeyCode.ENTER);
```
damithc: This is going to break when the menu changes in future. May be write a function that can find an item and click it irrespective of menu content? e.g. `clickMenu(String... items)` and call it `clickMenu("Boards","Auto-create","Work distribution")` Note the use of variable length argument lists.

HansNewbie: Yes, this has been the practice for all GUI menu tests and I think it will
break the menu changes. I will try to make such function.

src/test/java/guitests/UITest.java
```diff
@@ -308,6 +308,12 @@ private boolean existsQuiet(String selector) {
         }
     }
 
+    public void clickMenu(String... menuNames) {
+        for (int i = 0; i < menuNames.length; i++) {
+            click(menuNames[i]);
+        }
+    }
```
damithc: public method. Add comment?

HansNewbie: Will do. Didn't know that.

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -74,6 +86,46 @@ private void createMilestoneBoard() {
...
+        panelControl.selectFirstPanel();
+
+        String boardName = Utility.generateName(WORK_ALLOCATION,
+                prefs.getAllBoards().keySet().stream().collect(Collectors.toList()));
+
+        triggerBoardSaveEventSequence(boardName);
+
+        DialogMessage.showInformationDialog("Auto-create Board - " + WORK_ALLOCATION,
+                WORK_ALLOCATION + " board has been created and loaded.\n\n" +
+                        "It is saved under the name \"" + boardName + "\".");
```
damithc: Two higher levels can be extracted out to increase testability and reuse?
1. a method that takes a list of TurboUser objects and return a list of filter strings
2. a method that takes a list of filter strings and creates a board for it (this method can be reused at other places)

HansNewbie: For those methods, should I put them in this file or in a different file?

HansNewbie: I think for the second method, it should wait for #1074?

damithc: Something like this? It can be modified later when #1074 is done
```java
//panelData = pairs of filter and panel name
private void createBoard(List<Pair<String,String>> panelData, String boardName){
        //everything to do with board creation
}

```

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -74,6 +87,79 @@ private void createMilestoneBoard() {
...
+            return;
+        }
+
+        panelControl.closeAllPanels();
+
+        FilterPanel panelAdded;
+        int noOfPanelsToBeGenerated = Math.min(MAX_WORK_ALLOCATION_PANELS, userList.size());
+
+        List<String> filterStringsForPanels = generateListOfFilterStringsFromTurboUsers(
+                userList.subList(0, noOfPanelsToBeGenerated), "assignee:%s sort:milestone,status");
```
damithc: Enhance this method to generate filters as well as panel names in one shot?

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -74,6 +87,79 @@ private void createMilestoneBoard() {
...
+            DialogMessage.showInformationDialog("Auto-create Board - " + WORK_ALLOCATION,
+                    WORK_ALLOCATION + " board cannot be created as your" +
+                            " account has no push access to the default repo.");
+
+            return;
+        }
+
+        panelControl.closeAllPanels();
+
+        FilterPanel panelAdded;
```
damithc: These two lines can be moved further down?

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -74,6 +87,79 @@ private void createMilestoneBoard() {
...
+        for (int i = 0; i < noOfPanelsToBeGenerated; i++){
+            panelAdded = panelControl.addPanelAt(i);
+            panelAdded.setPanelName("Work allocated to " + assigneeNamesForPanelNames.get(i));
+            panelAdded.setFilterByString(filterStringsForPanels.get(i));
+        }
+
+        panelControl.selectFirstPanel();
+
+        String boardName = Utility.generateName(WORK_ALLOCATION,
+                prefs.getAllBoards().keySet().stream().collect(Collectors.toList()));
```
damithc: The 2nd parameter can be abstracted into a getBoardNames() method somewhere it belongs?

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -74,6 +87,79 @@ private void createMilestoneBoard() {
...
+
+        for (int i = 0; i < noOfPanelsToBeGenerated; i++){
+            panelAdded = panelControl.addPanelAt(i);
+            panelAdded.setPanelName("Work allocated to " + assigneeNamesForPanelNames.get(i));
+            panelAdded.setFilterByString(filterStringsForPanels.get(i));
+        }
+
+        panelControl.selectFirstPanel();
+
+        String boardName = Utility.generateName(WORK_ALLOCATION,
```
damithc: `generateName` name is not very indicative of what it does. Not sure what's a better name though.

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -74,6 +87,79 @@ private void createMilestoneBoard() {
...
+        List<String> userNames = new ArrayList<>();
+        TurboUser currUser;
+        String realName, loginName;
+
+        for (int i = 0; i < userList.size(); i++) {
+            currUser = userList.get(i);
+            realName = currUser.getRealName();
+            loginName = currUser.getLoginName();
+
+            userNames.add((!realName.isEmpty() ? realName : loginName));
```
damithc: The conditional logic in this line should be pushed to `TurboUser` class? e.g. `getFriendlierName()`

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -74,6 +87,79 @@ private void createMilestoneBoard() {
...
+                WORK_ALLOCATION + " board has been created and loaded.\n\n" +
+                        "It is saved under the name \"" + boardName + "\".");
+    }
+
+    private List<String> getNameListFromTurboUsers(List<TurboUser> userList) {
+        List<String> userNames = new ArrayList<>();
+        TurboUser currUser;
+        String realName, loginName;
+
+        for (int i = 0; i < userList.size(); i++) {
```
damithc: Can use a `foreach` loop instead?

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -74,6 +87,79 @@ private void createMilestoneBoard() {
...
+    }
+
+    private List<String> generateListOfFilterStringsFromTurboUsers(List<TurboUser> userList, String filterTemplate) {
+        List<String> generatedFilterStrings = new ArrayList<>();
+        TurboUser currUser;
+        String loginName, formatApplication;
+
+        for (int i = 0; i < userList.size(); i++) {
+            currUser = userList.get(i);
+            loginName = currUser.getLoginName();
```
damithc: These two lines can be merged?

src/main/java/ui/BoardAutoCreator.java
HansNewbie: I will use `List<PanelInfo>`. `PanelInfo` has variable of panel name and panel filter.

src/main/java/ui/BoardAutoCreator.java
HansNewbie: Will do.

src/main/java/ui/BoardAutoCreator.java
HansNewbie: Sure. Didn't notice that.

src/main/java/ui/BoardAutoCreator.java
HansNewbie: Yes. Missed that since I was reusing code.

src/main/java/ui/BoardAutoCreator.java
HansNewbie: Would `getNameClosestToDesiredName` be better?

src/main/java/ui/BoardAutoCreator.java
HansNewbie: Same case with `createBoard`. Currently it has no good place to stay until #1074 is solved. I will abstract it out into a method for now and add a comment to clean it up once board functionality is abstracted.

src/main/java/ui/BoardAutoCreator.java
HansNewbie: I will move `closeAllPanels()` to inside `createBoard()`. #1077 should watch out for this.

src/main/java/ui/BoardAutoCreator.java
HansNewbie: Okay, will do.

damithc:
Let's try to include this in this release. It is useful for CS2103 students.

damithc:
Don't forget to update documentation.

HansNewbie:
Passed local tests multiple times.

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -74,6 +74,70 @@ private void createMilestoneBoard() {
...
+                WORK_ALLOCATION + " board has been created and loaded.\n\n" +
+                        "It is saved under the name \"" + boardName + "\".");
+    }
+
+    private List<PanelInfo> generatePanelInfoFromTurboUsers(List<TurboUser> userList,
+                                                            String nameTemplate, String filterTemplate) {
+        List<PanelInfo> panelData = new ArrayList<>();
+        String panelName, panelFilter;
+
+        for (TurboUser eachUser : userList) {
```
dariusf: We could write this as a `map` over the collection of users. Also the parameter can probably be named `users`, so it isn't tied to its implementation.

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -74,6 +74,70 @@ private void createMilestoneBoard() {
...
+        }
+
+        panelControl.selectFirstPanel();
+
+        triggerBoardSaveEventSequence(boardName);
+    }
+
+    // TODO: extract out once #1074 (abstract board functionality from menu control) is done
+    private List<String> getBoardNames() {
+        return prefs.getAllBoards().keySet().stream().collect(Collectors.toList());
```
dariusf: Could we define `prefs.get[All]BoardNames()`? This is used in quite a few places. It can just do `new ArrayList<>(getAllBoards().keySet())`.

src/test/java/guitests/UITest.java
```diff
@@ -308,6 +308,17 @@ private boolean existsQuiet(String selector) {
         }
     }
 
+    /**
+     * Automate menu traversal by clicking them in order of input parameter
+     *
+     * @param menuNames array of strings of menu item names in sequence of traversal
+     */
+    public void clickMenu(String... menuNames) {
+        for (int i = 0; i < menuNames.length; i++) {
```
dariusf: Could we use a for each loop for this instead?
```java
for (String menuName : menuNames) {
    click(menuName);
}
```

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -74,6 +74,70 @@ private void createMilestoneBoard() {
...
+                "Work allocated to %s", "assignee:%s sort:milestone,status");
+
+        String boardName = Utility.getNameClosestToDesiredName(WORK_ALLOCATION, getBoardNames());
+
+        createBoard(panelData, boardName);
+
+        DialogMessage.showInformationDialog("Auto-create Board - " + WORK_ALLOCATION,
+                WORK_ALLOCATION + " board has been created and loaded.\n\n" +
+                        "It is saved under the name \"" + boardName + "\".");
+    }
```
damithc: Cleaner and the higher level logic is more visible now. :-)

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -74,6 +74,70 @@ private void createMilestoneBoard() {
...
+        return panelData;
+    }
+
+    // TODO: extract out once #1074 (abstract board functionality from menu control) is done
+    private void createBoard(List<PanelInfo> panelData, String boardName) {
+        panelControl.closeAllPanels();
+
+        FilterPanel panelAdded;
+
+        for (int i = 0; i < panelData.size(); i++) {
```
damithc: `foreach`?

damithc:
No more comments from me.

src/main/java/ui/BoardAutoCreator.java
HansNewbie: Sure. I am using `i` for the index to add the panel in for fear that `panelControl.getNoOfPanels()` would be slower in appending panel to the end, but will user `foreach` instead.

src/main/java/ui/BoardAutoCreator.java
HansNewbie: :+1:

src/test/java/guitests/UITest.java
HansNewbie: Oops, yes. I am too used to iterating with index :p

src/main/java/ui/BoardAutoCreator.java
HansNewbie: Sure, will do.

src/main/java/ui/BoardAutoCreator.java
HansNewbie: Ok.

PR #1098, opened 2015-10-28
Users can access HT using a universal hotkey #553

### LINKED ISSUE #553: Users can access HT using a universal hotkey
e.g. Win+H

rd1992:
Fixes #553


rd1992:
@dariusf hey i added the dependencies to build.gradle. could you run this on a mac? I cant test it. @ndt93 could you also take a look and test it on linux? Thanks

build.gradle
```diff
@@ -71,6 +71,7 @@ allprojects {
 
         testCompile "junit:junit:$junitVersion"
         testCompile "org.loadui:testFx:$testFxVersion"
+        compile files('src/main/resources/jkeymaster-1.1.jar')
```
dariusf: Could we use this instead?
```groovy
compile fileTree(dir: 'lib', include: '*.jar')
```
The jar would go in lib.

I searched a bit more and jars aren't really considered resources, so it would be a better idea to have a separate directory for them.

src/main/java/ui/UI.java
```diff
@@ -248,6 +258,8 @@ public void quit() {
             Platform.exit();
             System.exit(0);
         }
+        provider.reset();
+        provider.stop();
```
dariusf: What happens if this part doesn't execute? Does the hotkey process or something persist? Reason is that if `System.exit` is called, this is unreachable. Might be better to put it at the top of the method.

Also we should put this in its own method, so it's clear what it does.

src/main/java/ui/components/KeyboardShortcuts.java
```diff
@@ -134,6 +134,9 @@
     public static final KeyCodeCombination CLOSE_PANEL =
             new KeyCodeCombination(KeyCode.W, KeyCombination.SHORTCUT_DOWN);
     
+    // Global Hotkey
```
dariusf: This comment isn't really needed, since it's the same as the field name.

dariusf:
Maybe we could make a new class for all the additions to the UI? This is so it's all kept together. The class would have `init` and `quit` methods, as well as the `Provider` field.

I checked this out and tried it, and the hotkey does trigger. It doesn't bring the window into focus though, since that's not implemented for Mac.

rd1992:
@dariusf made the relevant changes. You can take a look again

dariusf:
Looks good.

PR #1099, opened 2015-10-28
 Create keyboard shortcuts for swapping panels #1062 

### LINKED ISSUE #1062: Create keyboard shortcuts for swapping panels
The function to swap panels using the mouse is rather blocky. Takes some tries to get it right, and it is especially troublesome for testing. ```PanelsTest``` has a part where panels are supposed to be swapped, but they don't get swapped, at least much of the time.

Therefore adding keyboard shortcuts to swap the currently selected panel with the left/right panel would be helpful. Would not be activated if there is <2 panel, and will have to wrap to the other side if it swaps to the left when it's the leftmost panel and vice versa.

What ```KeyCodeCombination``` would be appropriate?

HyungJon:
Fixes #1062


docs/keyboardShortcuts.md
```diff
@@ -72,6 +72,8 @@ These shortcuts can be used when the focus is on an issue card in a panel.
 | Force Refresh|<kbd>Ctrl</kbd> + <kbd>F5</kbd>|Clears HubTurbo's cache and re-downloads all issues from the repo|
 | Switch Board|<kbd>Ctrl</kbd> + <kbd>B</kbd>|Switches currently open board to another saved board|
 | Switch Default Repo|<kbd>Ctrl</kbd> + <kbd>R</kbd>|Switches currently default repo to another saved repo|
+| Swap to left|<kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Left</kbd>|Moves currently selected panel to the left|
+| Swap to right|<kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Right</kbd>|Moves currently selected panel to the right|
```
dariusf: Using left/right to swap but D/F to move seems a bit asymmetric.

src/main/java/ui/issuepanel/PanelControl.java
```diff
@@ -199,6 +199,36 @@ public void swapPanels(int panelIndex, int panelIndex2) {
         getChildren().set(panelIndex2, one);
     }
 
+    public void movePanelRight(int panelIndex) {
+        if (this.getNumberOfPanels() >= 2) {
```
dariusf: This can be a guard clause. It could also include the check for panelIndex?
```java
if (getNumberOfPanels() < 2 || panelIndex == 0) return;
```

src/main/java/ui/issuepanel/PanelControl.java
```diff
@@ -199,6 +199,36 @@ public void swapPanels(int panelIndex, int panelIndex2) {
...
+            int other = -1;
+            if (panelIndex == (this.getNumberOfPanels() - 1)) {
+                other = 0;
+            } else {
+                other = panelIndex + 1;
+            }
```
dariusf: We can use modulo here. The intent shouldn't be unclear.
```java
int other = (panelIndex + 1) % getNumberOfPanels();
```

src/main/java/ui/issuepanel/PanelControl.java
```diff
@@ -199,6 +199,36 @@ public void swapPanels(int panelIndex, int panelIndex2) {
...
+            if (panelIndex >= 0) {
+                swapPanels(panelIndex, other);
+                this.setCurrentlySelectedPanel(Optional.of(other));
```
dariusf: Could we use `selectPanel`, so it scrolls and updates focus?

src/main/java/ui/issuepanel/PanelControl.java
```diff
@@ -247,6 +277,17 @@ private void setupKeyEvents() {
                 handleKeys(KeyboardShortcuts.rightPanel.match(event));
                 assert currentlySelectedPanel.isPresent() : "handleKeys doesn't set selectedIndex!";
             }
+            if (KeyboardShortcuts.SWAP_PANEL_LEFT.match(event)) {
+                Optional<Integer> current = this.currentlySelectedPanel;
+                if (current.isPresent()) {
+                    movePanelLeft(current.get());
+                }
```
dariusf: This branch can be replaced with
```java
currentlySelectedPanel.ifPresent(this::movePanelLeft);
```

docs/keyboardShortcuts.md
HyungJon: Hmm that makes sense. Ctrl+Shift+D/F are kind of tricky to press though. Do you think it might be more desirable to change moving to left/right instead?

Or what other key combinations for swapping could be used? Alt+D/F maybe?

damithc: Actually, why do we need a shortcut for swapping panels? It's not something you do regularly. I assume drag-and-drop works already?

HyungJon: It works, but not very smoothly. And ```PanelsTest``` has a part where panels are swapped by dragging, but the swap actually doesn't happen. We figured in #1061 that we might want to do a keyboard shortcut, for the sake of this test at least. From my testing it does work smoother than dragging.

damithc: Understood.

dariusf: Okay, so we'll just ignore the asymmetry, since it's not commonly used.

HyungJon:
Alright, could I have a review now.

dariusf:
Could we rename getNumberOfPanels to getPanelCount?

Apart from that looks good, just housekeeping matters left.

- Is the drop in coverage legitimate?
- Not sure why other commits are appearing in here. You might have to rebase.

HyungJon:
Yes. The methods for swapping aren't directly tested, that's probably causing the coverage fall.

HyungJon:
Okay, resolved everything.

dariusf:
Dragging isn't tested any more? It needs to be covered as well. We don't have to assert that the swap actually happens, but it should at least not fail when used.

Also, I don't get why the extra commits from your other PR are showing up. In particular the commit for V3.12.0 isn't one of the three I committed (master, rc, release). Did you rebase over it?

HyungJon:
I thought I rebased those unwanted commits but some are still there. I recall there was a notification that my local branch is behind this remote branch and have to pull, that's probably where it started. I should have checked more.

Rebasing caused quite serious merge conflicts that I'm not confident I can safely resolve.

dariusf:
Never mind, I will squash at the point of merging. You should add the tests for dragging back. That's the last thing.

HyungJon:
Done.

dariusf:
Merged 818c3c5ac608659cc539c9325cdf666aee3e0f4f.

PR #1104, opened 2015-10-29
Universal shortcut does not work when minimized #1101

### LINKED ISSUE #1101: Universal shortcut does not work when minimized
If minimized, the shortcut should restore it?
Also, remember to document the shortcut this time.

rd1992:
Fixes #1101

### DIFF
--- a/docs/keyboardShortcuts.md
+++ b/docs/keyboardShortcuts.md
@@ -65,7 +65,7 @@ These shortcuts can be used when the focus is on an issue card in a panel.
 
 | **Action on the panel view**| **Shortcut**|**Explanation**|
 | ------------- |:-------------:| -----------|
-| Bring HubTurbo into focus|<kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>H</kbd>| Global Hotkey to make HubTurbo topmost window|
+| Toggle HubTurbo focus|<kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>H</kbd>| Global Hotkey to toggle between HubTurbo's window states (Focused or minimized)|
 | Close|<kbd>Ctrl</kbd> + <kbd>W</kbd>| Closes the panel in focus|
 | Ma**x**imize|<kbd>Ctrl</kbd> + <kbd>X</kbd>||
 | Mi**n**imize|<kbd>Ctrl</kbd> + <kbd>M</kbd>||
--- a/docs/keyboardShortcuts.md
+++ b/docs/keyboardShortcuts.md
@@ -63,8 +63,9 @@ These shortcuts can be used when the focus is on an issue card in a panel.
 
 ## Control the panel view window ...
 
-| **Action on the browser view**| **Shortcut**|**Explanation**|
+| **Action on the panel view**| **Shortcut**|**Explanation**|
 | ------------- |:-------------:| -----------|
+| Bring HubTurbo into focus|<kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>H</kbd>| Global Hotkey to make HubTurbo topmost window|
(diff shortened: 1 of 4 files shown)


rd1992:
@dariusf could you take a look at this real quick? thanks

docs/keyboardShortcuts.md
```diff
@@ -62,8 +62,9 @@ These shortcuts can be used when the focus is on an issue card in a panel.
 
 ## Control the panel view window ...
 
-| **Action on the browser view**| **Shortcut**|**Explanation**|
+| **Action on the panel view**| **Shortcut**|**Explanation**|
 | ------------- |:-------------:| -----------|
+| Bring HubTurbo into focus|<kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>H</kbd>| Global Hotkey to make HubTurbo topmost window|
```
damithc: Can we use this shortcut for both dismissing and restoring? i.e. if not in focus (or minimized), it restores both windows and give focus to pView. If pView is in focus, it minimized both windows.

rd1992: that would only be possible on Windows.

damithc: It wont work at all on other platforms? Or work partially? At least the restore part (i.e. the current behavior of Ctrl+Alt+H is supposed to work on all platforms right?)

rd1992: Sorry should have been more clear. We can minimize/maximize the panel view (the main java application). Minimizing maximizing the bview is not possible across platforms yet

damithc: I see, go ahead if it fails gracefully in other platforms. Can do as part of this or as a separate issue.

rd1992: i have done another commit. please check it out and see if this is what you had in mind. The minimizing/restoring of bview works on Windows only.

damithc: Update the documentation too.

rd1992: sure. but the functionality is fine?

damithc:
I tried the branch. The shortcut seems to work. However, I can seem to maximize or resize the pView using mouse. It's just stuck at 1-panel size.

damithc:
Ctrl+D and Ctrl+X still works, but cannot resize using mouse or normal windows title bar icons.

rd1992:
i can't reproduce this. I can minimize maximize using the mouse. @HyungJon can you try this out on windows?

rd1992:
okay there is a different bug. If you use `Ctrl + X` or `Ctrl + D` and then try to use mouse to maximize minimize it behaves weirdly. but that has nothing to do with this change. It is even on the master branch

damithc:
I guess so. Open a separate issue for it.

damithc:
How's this one coming along?

HyungJon:
Sorry I didn't see the request for test.

The shortcut ctrl+alt+H works on Windows, but it does show the problematic behavior prof mentioned. pView moves diagonally to occupy the top left corner, but it doesn't expand. I think it's caused by the same thing that's causing the inability to maximize/restore after using ctrl+X&ctrl+D shortcuts.

rd1992:
@dariusf can you test this on a mac again? And see if the minimize also works? Thanks

dariusf:
Both work. When minimised the window also gets restored when the hotkey is pressed. This is platform-specific behaviour though.

rd1992:
yeah but the pview should come into focus etc. right? (the bview focus is platform specific)

dariusf:
Here's what happens:

- When the window is in focus, the hotkey minimises and restores it you would expect.
- When the window is not in focus, nothing happens. The event is registered though, judging from the log.

Is that correct?

The code looks fine. I guess we won't be testing this.

By the way there are conflicts with master.

rd1992:
@dariusf yeah that is correct. So i can merge this after travis passes. I have rebased it

PR #1118, opened 2015-11-05
Cannot go to 'related PR' of issue #1115

### LINKED ISSUE #1115: Cannot go to 'related PR' of issue
The `g,e` shortcut doesn't work for issue #1005 of hubturbo repo

rd1992:
Fixes #1115

### DIFF
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -503,7 +503,7 @@ public class BrowserComponent {
             WebElement element = driver.findElement(By.xpath("//div[contains(@id, 'ref-pullrequest')]"));
             return Optional.of(Integer.parseInt(element.findElement(By.xpath(".."))
                     .findElement(By.xpath(".//h3/a/span"))
-                    .getAttribute("innerHTML").substring(1)));
+                    .getAttribute("innerHTML").trim().substring(1)));
         } catch (NoSuchElementException e) {
             logger.info("no PR mention found");
         }


damithc:
Any changes to tests?

dariusf:
There are no tests for this. It's similar to the problem of us not being able to test GitHub API interaction. This particular feature depends on HTML structure, so we need the actual pages to check that it works.

damithc:
There's no method that takes the html string and extracts the issue number? If not, try to extract a method like that. Then, it should have a unit test should a unit test for that feed it different variations of HTML and checks the output. It improves testability when logic is put inside stateless functions.

dariusf:
Okay, I guess such a method would go in GitHubPageElements.

rd1992:
There are two methods. One is to extract pull request number and one to extract Issue number. pull request we need the html so hence its part of browser component. So you are saying it should be moved to githuburlpageelements?

dariusf:
Hmm, there's no HTML string to operate on... We get only the contents of the tag from the browser.

rd1992:
yup. we need the input HTML from the issue page. Otherwise to test this we will need to generate our own

dariusf:
Yes, but now I'm wondering if it's possible to generate our own without going through the browser.

Or do we want to bite to bullet and use the browser in tests?

damithc:
Even if it is in an object not currently involved in testing, it can be a static method that can be tested without instantiating right?

damithc:
@rd1992 do you plan to finish this one soon?

rd1992:
@damithc the tests depends on #1134 . So as soon as that is integrated we can test this class better.

damithc:
#1134 is still at concept stage. Go ahead and merge this even if the tests are missing. Create a separate issue for testing, which someone else can do later.

PR #1119, opened 2015-11-06
Create undo buffer #1058

### LINKED ISSUE #1058: Create undo buffer

lejolly:
Fixes #1058


damithc:
Can add proper comments to significant public methods?

lejolly:
sure thing

dariusf:
- We should fill in the type parameters everywhere. There's no need for dynamic casts and runtime type checks when all that information is statically available.
- Things should be designed to leverage polymorphism, i.e. we want to minimise explicit instanceof checks. If a ChangeLabelsAction is required, we can just take that as a parameter, rather than check if the Action is an instance of it. Otherwise we should be able to do `action.act(thing)` and have it magically resolve based on what `action` is.
- If we always need an issue to apply an Action to, we should store them together, rather than have to do `obj.method(obj.getter())`. This probably requires a new class, since using tuples for data types is generally a bad idea (more so in Java, because it's verbose and we need a new class for each tuple size...).
- reconcile should be a static method of either ChangeLabelsAction or Action, seeing as it doesn't use any fields. The implementation is also really dense... Could we add comments, both to the implementation and the interface? We can try to simplify it from there.
- Are we going to remove the mutation and make the undo buffer a list instead of an optional?

Basically we need to design things in a more OO way. Call sites should be as simple as possible (just method calls with only the required arguments), and all type-based dispatch should be done via polymorphism rather than instanceof.

lejolly:
@dariusf i want to keep `reconcile` in the `Action` interface but i can't find a nice way to define a specific implementation of the interface, in this case `ChangeLabelsAction` as the parameter types without resorting to something like this: https://stackoverflow.com/questions/6639642/designing-an-interface-for-methods-with-different-parameters

The whole point is so that `UndoController.addAction` can call `reconcile` for any action. However, if we plan on implementing different `Actions` then there's no escaping from explicit instanceOf checks? We can't reconcile different types of `Actions` right?

dariusf:
Yes, it's tricky to express in Java. The root problem is that we can't dispatch on multiple parameter types, so `reconcile(Action, Action)` can't be overridden by `reconcile(ChangeLabelAction, ChangeLabelAction)` as we'd expect (it's an overload instead, which is resolved at compile time).

[Implementing our own method dispatch](http://martinsjava.blogspot.sg/2009/12/how-to-code-multiple-dispatch-in-java.html) will actually work, but it's a gross reinvention of instanceof, so yeah, it is okay to use instanceof here.

I was trying this alternative, which is boilerplate-heavy, but constrains instanceof use to one place. Any other ideas?

```java
interface Action {
}

abstract class AResult<T extends Action> {
    abstract T first();
    abstract T second();
}

class ChangeLabelsAction implements Action {

    // Each Action implements their reconcile method this way
    static Result reconcile(ChangeLabelsAction a, ChangeLabelsAction b) {
        // Implement algorithm here
        return new Result(a, b);
    }

    // Boilerplate. Needed because generic types are invariant in type parameters,
    // i.e. Pair<ChangeLabelAction, ChangeLabelAction> is not a subtype of
    // Pair<Action, Action>, hence we can't return the former above. An actual
    // subtype is needed.
    static class Result extends AResult<ChangeLabelsAction> {
        ChangeLabelsAction first, second;
        public Result(ChangeLabelsAction first, ChangeLabelsAction second) {
            this.first = first;
            this.second = second;
        }
        @Override public ChangeLabelsAction first() { return first; }
        @Override public ChangeLabelsAction second() { return second; }
    }
}

class UndoController {

    // The only place we use (unchecked) dynamic casts and instanceof
    @SuppressWarnings("unchecked")
    static <T extends AResult<? extends Action>> T reconcile(Action first, Action second) {
        if (first instanceof ChangeLabelsAction && second instanceof ChangeLabelsAction) {
            ChangeLabelsAction a = (ChangeLabelsAction) first;
            ChangeLabelsAction b = (ChangeLabelsAction) second;
            return (T) ChangeLabelsAction.reconcile(a, b);
        } else {
            throw new IllegalArgumentException("arguments of different types");
        }
    }
}

    ChangeLabelsAction.Result result =
        UndoController.reconcile(new ChangeLabelsAction(), new ChangeLabelsAction());

    // After all that, this compiles without any casts
    ChangeLabelsAction action = result.first();
}
```

lejolly:
Ok, reduced the scope of the PR to match the issue.

src/main/java/backend/Logic.java
```diff
@@ -184,35 +188,22 @@ public Model getRepo(String repoId) {
...
+    // Replaces labels of issue on GitHub, returns false if it fails
+    public CompletableFuture<Boolean> replaceIssueLabels(TurboIssue issue, Action<TurboIssue> action) {
+        logger.info("Applying " + action.getDescription() + " on " + issue + " in GitHub");
+        return repoIO.replaceIssueLabels(issue, action.act(new TurboIssue(issue)).getLabels())
+                .thenApply(labels -> {
+                    logger.info("Applying " + action.getDescription() + " on " + issue + " in UI");
+                    issue.setLabels(labels);
+                    refreshUI();
+                    return true;
+                })
```
dariusf: The `exceptionally` part below this can use `Futures.withResult(false)`.

src/main/java/undo/UndoController.java
```diff
@@ -0,0 +1,84 @@
...
+    public void undoCallback() {
+        if (undoBuffer.isPresent()) {
+            determineLogicMethod(undoBuffer.get().getKey(), undoBuffer.get().getValue());
+            undoBuffer = Optional.empty();
+        }
+    }
+
+    // Adds an action to the undo buffer.
+    @SuppressWarnings("unchecked")
+    public void addAction(TurboIssue issue, Action action) {
```
dariusf: We should use Action<TurboIssue> here, since this method only deals with issues (we can generalise this in future). Then we don't need the cast below and the SuppressWarnings above.

src/main/java/undo/UndoController.java
```diff
@@ -0,0 +1,84 @@
...
+        determineLogicMethod(issue, action);
+        undoBuffer = Optional.of(new Pair<>((TurboIssue) action.act(issue), action.invert()));
+        Notification notification = new Notification(createInfoOcticon(),
+                "Undo " + action.getDescription() + " for #" + issue.getId() + ": " + issue.getTitle(),
+                "Undo", this::undoCallback);
+        notificationController.showNotification(notification);
+    }
+
+    // Check to determine which logic method to call
+    @SuppressWarnings("unchecked")
```
dariusf: Same here.

src/main/java/undo/UndoController.java
```diff
@@ -0,0 +1,84 @@
...
+    private void determineLogicMethod(TurboIssue issue, Action action) {
+        if (action.getType() == Action.ActionType.CHANGELABELS) {
+            logic.replaceIssueLabels(issue, action)
```
dariusf: Instead of doing this, could the action call `logic.replaceIssueLabels` in its `act`? Maybe `act` should return `CompletableFuture<T>` as well, and we can chain the error handling onto it. Then we can replace `getType` with polymorphism.

src/main/java/undo/actions/Action.java
```diff
@@ -0,0 +1,43 @@
...
+    String getDescription();
+
+    T act(T t);
+
+    T undo(T t);
+
+    // Returns an inverse of the Action.
+    @SuppressWarnings("unchecked")
+    default Action invert() {
+        Action that = this;
```
dariusf: We can use Action<T> here and above, then we don't need to SuppressWarnings or cast.

src/main/java/undo/actions/Action.java
```diff
@@ -0,0 +1,43 @@
...
+public interface Action<T> {
+
+    @SuppressWarnings("SpellCheckingInspection")
+    enum ActionType { CHANGELABELS }
+
+    ActionType getType();
+
+    String getDescription();
+
+    T act(T t);
```
damithc: Name `act` was chosen to follow some convention? If not, `do` goes nicely with `undo` :-)

src/main/java/undo/actions/Action.java
damithc: A more self-explanatory name for `that`?

src/main/java/undo/actions/Action.java
```diff
@@ -0,0 +1,43 @@
...
+
+            @Override
+            public String getDescription() {
+                return that.getDescription();
+            }
+
+            @Override
+            public T act(T t) {
+                return (T) that.undo(t);
+            }
```
damithc: This is probably not relevant right now, but not all actions are undo-able. The eventual design needs to take that into account.

src/main/java/undo/actions/ChangeLabelsAction.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.stream.Collectors;
+
+public class ChangeLabelsAction implements Action<TurboIssue> {
+
+    public static final String DESCRIPTION = "change label(s)";
+
+    private final Set<String> addedLabels;
+    private final Set<String> removedLabels;
+
+    // Represents changes to the labels of a TurboIssue
```
damithc: Use proper javadoc comments?

src/main/java/undo/UndoController.java
lejolly: Still appears that the SuppressWarnings is needed for the `action.invert()` below. 

You meant this right:
```
public void addAction(TurboIssue issue, Action<TurboIssue> action)
```

src/main/java/undo/UndoController.java
lejolly: I don't really like calling `Logic` within the `Action` as I see the `Action` more as a diff rather than to encapsulate the actual method calls?

The `getType` was implemented because I couldn't get `instanceOf` to work reliably with the `Optional`. 
So you would prefer something like:
```
private void determineLogicMethod(TurboIssue issue, ChangeLabelsAction action)
```
and then a catch-all at the end?
```
private void determineLogicMethod(TurboIssue issue, Action action)
    // for unimplemented Actions?
```
would that even work?

src/main/java/undo/actions/Action.java
lejolly: I guess it was chosen because the method takes in an object which it then `acts` on? @dariusf any preference?

src/main/java/undo/UndoController.java
dariusf: Yeah, that wouldn't work, because it's overloading. The method to use will be decided at compile-time. There's no runtime decision based on object type.

I was thinking of this:

```java
private void determineLogicMethod(TurboIssue issue, Action<TurboIssue> action) {
    action.act(issue)
        .thenApply(success ->
            showErrorDialogOnFailure(success, issue, action));
}
```

which is the de facto way to decide what to do based on type. Was hoping to leverage that rather than reinvent it (and have to repeat it everywhere Actions are used, without the compiler helping us check exhaustiveness... which is my real concern).

We have to agree on what Actions represent: right now I'm thinking of them as the command object that we put on the undo stack, and you're thinking of them as the issue diff.

Since once we decide the type of command (i.e. labels, milestones, etc.), we know what type of diff to use, I'm more inclined to have Actions represent commands. The decision is really which resource we're dealing with. After `action.act(issue)` selects the type of command/Action, a specific type of diff can be applied in it without requiring further polymorphism. What do you think?

src/main/java/undo/UndoController.java
dariusf: Oops, yeah.

src/main/java/undo/actions/Action.java
dariusf: Yes, it's an action, so it `acts`. `do` is a keyword, unfortunately.

src/main/java/undo/UndoController.java
dariusf: Hmm, GitHub doesn't like unquoted angle brackets. All my `<T>`s were missing.

lejolly:
@dariusf anything else left to do?

src/main/java/ui/components/Notification.java
```diff
@@ -4,25 +4,24 @@
 
 public class Notification {
 
-    public static final int DEFAULT_NOTIFICATION_PANE_VISIBLE_PERIOD = 10;
+    public static final int DEFAULT_NOTIFICATION_PANE_VISIBLE_PERIOD = 30;
 
-    public enum NotificationType { NOACTION, ACTIONONBUTTON, ACTIONONTIMEOUT, ACTIONONBUTTONANDTIMEOUT }
+    @SuppressWarnings("SpellCheckingInspection")
```
dariusf: This is an IntelliJ thing, right? I'm not sure if we want to have it in source code, since it's not IDE-agnostic, and we're going to need it everywhere. Maybe it would be better to have some kind of project-wide configuration to disable spell check? Or let developers do it for themselves.

src/main/java/undo/UndoController.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @SuppressWarnings("SpellCheckingInspection")
+    private static final String OCTICON_INFO = "\uf059";
+
+    private final Logic logic;
+    private final NotificationController notificationController;
+    private Optional<Pair<TurboIssue, Action<TurboIssue>>> undoBuffer;
+
+    // Holds an undo buffer of size 1.
+    // Buffer stores the undo action.
```
dariusf: Might want to use javadoc comments here. Gets us a bit more tool support if we decide to generate docs someday.

dariusf:
Did you see [this](https://github.com/HubTurbo/HubTurbo/pull/1119#discussion_r49860759)? Wanted to get your opinions. That should be the last thing.

src/main/java/ui/components/Notification.java
lejolly: ah ok.

src/main/java/undo/UndoController.java
lejolly: ok

src/main/java/undo/UndoController.java
lejolly: So if an Action is a command, it is necessarily tied to a particular issue? And if so, should we just store a reference to the issue within the Action itself?

Also, if it's a command, then do we call the relevant Logic method within the Action itself or use something like another event?

Lastly, so `Action.act` should return a `CompletableFuture<Boolean>`? Should `Action.undo` return the same thing?

dariusf: Hmm, good point. It wouldn't make sense to do `action.act(a).undo(b)`.

I think it's fine to call the Logic method directly.

Depends if you need to wait on the result of the undo and do something on completion, etc. If you need to wait but don't need the result it can be `CompletableFuture<Void>`, otherwise `void` is okay. I don't think the two methods need to be symmetric because we don't have redo.

If we do this, we may not need `invert` either. Just put the action on the buffer and call `undo` there.

lejolly: Ok, so since it isn't a diff, i'm just going to store two sets of labels, the original list of labels and the new list. Since the relevant Logic method doesn't care about the diffs, it just replaces them wholesale.

lejolly:
ok, please check that it works for you. managed to strip out even more code/complexity, yay.

dariusf:
Nice. Thanks for making all the changes!

PR #1123, opened 2015-11-08
Abstract out `selectAll()` for GUI Test #1122

### LINKED ISSUE #1122: Abstract out `selectAll()` for GUI Test
And add comment properly.

HansNewbie:
Fixes #1122

### DIFF
--- a/src/test/java/guitests/InvalidLoginTest.java
+++ b/src/test/java/guitests/InvalidLoginTest.java
@@ -17,9 +17,7 @@ public class InvalidLoginTest extends UITest {
 
     @Test
     public void invalidLoginTest() throws InterruptedException {
-        TextField repoOwnerField = find("#repoOwnerField");
-        doubleClick(repoOwnerField);
-        doubleClick(repoOwnerField);
+        selectAll();
         push(KeyCode.BACK_SPACE).type("abc").push(KeyCode.TAB);
         push(KeyCode.BACK_SPACE).type("abc").push(KeyCode.TAB);
         push(KeyCode.BACK_SPACE).type("abc").push(KeyCode.TAB);
@@ -32,9 +30,7 @@ public class InvalidLoginTest extends UITest {
 
     @Test
     public void emptyLoginTest() throws InterruptedException {
-        TextField repoOwnerField = find("#repoOwnerField");
-        doubleClick(repoOwnerField);
-        doubleClick(repoOwnerField);
+        selectAll();
         push(KeyCode.BACK_SPACE).type("abc").push(KeyCode.TAB);
         push(KeyCode.BACK_SPACE).type("abc").push(KeyCode.TAB);
         push(KeyCode.BACK_SPACE).push(KeyCode.TAB).push(KeyCode.BACK_SPACE);
@@ -46,9 +42,7 @@ public class InvalidLoginTest extends UITest {
 
     @Test
     public void emptyUsernameTest() throws InterruptedException {
-        TextField repoOwnerField = find("#repoOwnerField");
-        doubleClick(repoOwnerField);
-        doubleClick(repoOwnerField);
+        selectAll();
         push(KeyCode.BACK_SPACE).type("abc").push(KeyCode.TAB);
         push(KeyCode.BACK_SPACE).type("abc").push(KeyCode.TAB);
         push(KeyCode.BACK_SPACE).push(KeyCode.TAB);
@@ -61,9 +55,7 @@ public class InvalidLoginTest extends UITest {
 
     @Test
     public void emptyPasswordTest() throws InterruptedException {
-        TextField repoOwnerField = find("#repoOwnerField");
-        doubleClick(repoOwnerField);
-        doubleClick(repoOwnerField);
+        selectAll();
         push(KeyCode.BACK_SPACE).type("abc").push(KeyCode.TAB);
         push(KeyCode.BACK_SPACE).type("abc").push(KeyCode.TAB);
         push(KeyCode.BACK_SPACE).type("abc").push(KeyCode.TAB);
--- a/src/test/java/guitests/LogoutTest.java
+++ b/src/test/java/guitests/LogoutTest.java
@@ -24,9 +24,7 @@ public class LogoutTest extends UITest {
 
     @Test
     public void logoutFunctionTest() {
-        TextField repoOwnerField = find("#repoOwnerField");
-        doubleClick(repoOwnerField);
-        doubleClick(repoOwnerField);
+        selectAll();
         type("dummy").push(KeyCode.TAB);
         type("dummy").push(KeyCode.TAB);
         type("test").push(KeyCode.TAB);
--- a/src/test/java/guitests/UITest.java
+++ b/src/test/java/guitests/UITest.java
@@ -3,6 +3,7 @@ package guitests;
 import static com.google.common.io.Files.getFileExtension;
 
 import java.awt.*;
+import java.awt.event.KeyEvent;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
@@ -19,6 +20,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Predicate;
 
+import javafx.application.Platform;
 import org.apache.commons.lang3.reflect.FieldUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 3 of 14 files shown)


src/test/java/guitests/UITest.java
```diff
@@ -337,4 +337,12 @@ public GuiTest type(String text) {
         return this;
     }
 
+    /**
+     * NOTE: please turn off NumLock as it will interfere with this method
+     *
+     * Used to select the whole filter text so that it can be replaced
+     */
+    public void selectAll() {
```
damithc: Is it possible to verify NumLock is turned off and throw an exception if it is on?

dariusf:
There are lots of instances of this across the tests, where we press Shortcut+A, triple-click, etc. Maybe take the chance to fix those as well?

Also, if we use one of the above, we may not need to worry about num lock.

src/test/java/guitests/UITest.java
HansNewbie: Yes, it is possible. Would using try and catch for just `selectAll` be good, though?

HansNewbie:
@dariusf I tried Shortcut+A but it didn't work in Mac (I think the test framework treat SHORTCUT as `Ctrl` in Mac for testing instead of `Cmd`, hence the fail). I tried Ctrl+A as well, it didn't work on Mac.

Double-click works for some, but for things with `:` like filters it won't work. Triple-click feels unreliable to me if the delay is not exact and the system click delay detection is different.

If it's fine as long as the test works on Windows and Travis, I will use Ctrl+A. Otherwise the only solution is this End->Shift+Home (haven't added the end, but will add if this implementation is okay).

dariusf:
If you use `pushKeys` or the KeyCodeCombination overload for `press`, we have a [helper method](https://github.com/HubTurbo/HubTurbo/blob/master/src/test/java/guitests/UITest.java#L183) to fix that. See if that works reliably?

HansNewbie:
@dariusf okay. Will try.

HansNewbie:
@dariusf okay, using Shortcut+A now. I checked all GUI Test and replaced all variations with `selectAll()`. Is it ok? If ok I squash commits and run local tests.

src/test/java/guitests/InvalidLoginTest.java
```diff
@@ -17,9 +17,8 @@ public void launchApp() {
 
     @Test
     public void invalidLoginTest() throws InterruptedException {
-        TextField repoOwnerField = find("#repoOwnerField");
-        doubleClick(repoOwnerField);
-        doubleClick(repoOwnerField);
+        find("#repoOwnerField");
```
dariusf: The `find` calls shouldn't be needed any more. Unless we want to click in those fields to make sure they're in focus?

src/test/java/guitests/UITest.java
```diff
@@ -337,4 +337,10 @@ public GuiTest type(String text) {
         return this;
     }
 
+    /**
+     * Used to select the whole filter text so that it can be replaced
+     */
+    public void selectAll() {
+        pushKeys(getKeyCodes(new KeyCodeCombination(KeyCode.A, KeyCombination.SHORTCUT_DOWN)));
```
dariusf: `pushKeys(new KeyCodeCombination(...))` should work.

src/test/java/guitests/InvalidLoginTest.java
```diff
@@ -32,9 +31,8 @@ public void invalidLoginTest() throws InterruptedException {
 
     @Test
     public void emptyLoginTest() throws InterruptedException {
-        TextField repoOwnerField = find("#repoOwnerField");
-        doubleClick(repoOwnerField);
-        doubleClick(repoOwnerField);
+        find("#repoOwnerField");
+        selectAll();
```
damithc: If these two lines always appear together, may be they should be combined?

src/test/java/guitests/InvalidLoginTest.java
HansNewbie: I will check if the line is needed.

src/test/java/guitests/InvalidLoginTest.java
HansNewbie: I think they should be kept separate. One reason would be readability.
Another is that sometimes only find is used (to check the text) or only
selectAll is used (to change, say, default repo or filter again and again).

dariusf: An abstraction that could be nice is to wait until a node appears, find, click, select all, then type a string. Seems like this is frequently done but I don't know exactly how much.

HansNewbie: I can make such function but I don't find so many of them from replacing
all with selectAll. In fact, I don't find any thay does it with waiting.

dariusf: Okay, never mind then.

src/test/java/guitests/UITest.java
HansNewbie: Ok.

src/test/java/guitests/InvalidLoginTest.java
HansNewbie: Ok removed the `find`s.

HansNewbie:
@dariusf the coverage drop seems very inconsistent. I improved one, then it gave me a drop in another file. Do you want me to ensure it passes or just let it be, as long as it can passed the test locally? I haven't test locally, will update you if I have done so.

dariusf:
Yes, coverage is inconsistent. It's from the tests not always being deterministic.

It's best to fix it gradually, since after we lose context it's hard to remember which parts were random. Here though, the drop seems to be in the fix for the race condition? We may not want to cover that anyway.

HansNewbie:
Yes. So just ignore it?

HansNewbie:
@dariusf aside from panels test, all other tests pass locally.

src/test/java/guitests/LogoutTest.java
```diff
@@ -24,9 +24,8 @@ public void launchApp() {
 
     @Test
     public void logoutFunctionTest() {
-        TextField repoOwnerField = find("#repoOwnerField");
-        doubleClick(repoOwnerField);
-        doubleClick(repoOwnerField);
+        find("#repoOwnerField");
```
dariusf: This probably isn't needed, unless we click on it?

src/test/java/guitests/UseGlobalConfigsTest.java
```diff
@@ -42,9 +42,8 @@ public void globalConfigTest() {
         UI ui = TestController.getUI();
         PanelControl panels = ui.getPanelControl();
 
-        TextField repoOwnerField = find("#repoOwnerField");
-        doubleClick(repoOwnerField);
-        doubleClick(repoOwnerField);
+        find("#repoOwnerField");
```
dariusf: This too.

src/test/java/guitests/UITest.java
```diff
@@ -337,4 +337,11 @@ public GuiTest type(String text) {
         return this;
     }
 
+    /**
+     * Used to select the whole filter text so that it can be replaced
+     */
+    public void selectAll() {
+        pushKeys(KeyCode.END);
```
dariusf: Do we still need the End key press?

HansNewbie: I added press end to make it usable for any purpose, in case the cursor is
not at the end of text box. Should I remove it? Seems safer to me.

dariusf: But whether or not it's at the end, wouldn't Ctrl+A select everything? I was just wondering why it was needed.

HansNewbie: OH. Haha. I thought I am still pressing sift+home. Alright, will remove.

dariusf:
Okay this is fine, just have to update to latest master. We're going to ignore the coverage drop.

HansNewbie:
Will do one more local test run. Will update here once it finishes.

HansNewbie:
@dariusf Passed tests locally. Do you want to take one last look?

HansNewbie:
@dariusf removed all swap hotkeys. I am running a local test again to ensure things are proper.

HansNewbie:
@dariusf passed tests locally.

HansNewbie:
@dariusf seems the coverage drop is legit. Giving one check.

HansNewbie:
@dariusf passed tests locally.

dariusf:
Could you update the docs as well? Just remove any mention of the shortcut.

Don't worry about the random coverage drops now I guess. Just cover only those which happen consistently.

HansNewbie:
Alright

HansNewbie:
Docs updated.

dariusf:
Looks good.

PR #1125, opened 2015-11-09
Fix failures in PanelsTest #1124

### LINKED ISSUE #1124: Fix failures in PanelsTest
This fails offline but passes on Travis. Blocking #1123.

dariusf:
Fixes #1124.

This makes the panel-dragging test reliable. @HyungJon you might also want to take a look.

### DIFF
--- a/src/test/java/guitests/PanelsTest.java
+++ b/src/test/java/guitests/PanelsTest.java
@@ -1,89 +1,111 @@
 package guitests;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static junit.framework.TestCase.assertTrue;
+import static ui.components.KeyboardShortcuts.*;
 
-import ui.TestController;
-import ui.issuepanel.FilterPanel;
-import ui.issuepanel.PanelControl;
-import static ui.components.KeyboardShortcuts.CREATE_RIGHT_PANEL;
-import static ui.components.KeyboardShortcuts.MAXIMIZE_WINDOW;
-import static ui.components.KeyboardShortcuts.SWAP_PANEL_LEFT;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.junit.Test;
-import org.loadui.testfx.exceptions.NoNodesFoundException;
 
-import javafx.scene.control.Label;
-import javafx.scene.text.Text;
+import javafx.scene.Node;
 import javafx.scene.input.KeyCode;
+import javafx.scene.layout.HBox;
+import ui.TestController;
 import ui.UI;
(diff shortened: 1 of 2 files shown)


HyungJon:
Yes, I think this is fine now, since both functions are tested.
Do you think there's a need for more tests regarding this function e.g. not just swapping with direct left/right but dragging over another panel (3 panels, drag leftmost to far right)?

src/test/java/guitests/PanelsTest.java
```diff
@@ -1,80 +1,63 @@
...
         click(name1);
-        assertTrue(eventTriggered.value);
+        TestUtils.awaitCondition(eventTriggered::get);
 
-        // Drag
-        // TODO find a way to actually reorder the panels, and make it compatible with swap key command
+        // Reorder panels by dragging, and wait for them to swap position
+        final AbstractPanel firstPanel = panels.getPanel(0);
+        Node from = ((FilterPanel) panels.getPanel(1)).getCloseButton();
+        Node to = ((FilterPanel) panels.getPanel(0)).getFilterTextField();
```
HansNewbie: Why don't we use close button as well here? Might be confusing in the future. If there is a reason why we use filter text field instead of close button, maybe put a comment about it?

HansNewbie:
I think what @HyungJon suggested sounds good. Since we test dragging to the left, maybe for this test drag to the right?

src/test/java/guitests/PanelsTest.java
dariusf: With the close button as a target, the swap doesn't actually trigger. It's due to the way the node hierarchy is set up.

I've pulled out the means of getting the node into a method, so we can change that if the node hierarchy changes.

dariusf:
Okay, I've added a third panel, and tests for dragging left/right as well as swapping left/right with hotkeys.

src/test/java/guitests/PanelsTest.java
```diff
@@ -82,8 +50,60 @@ public void panelsTest() {
         type("dummy2/dummy2");
         push(KeyCode.ENTER);
         pushKeys(CREATE_RIGHT_PANEL);
-        PlatformEx.waitOnFxThread();
-        // Actually a check. If #dummy2/dummy2_col1 did not exist, this would throw an exception.
-        click("#dummy2/dummy2_col1");
+
+        waitUntilNodeAppears(panel1);
```
HansNewbie: Is this the same as the original? To me it seems that originally we want to assert that a panel can be created after dragging other panels. But in this wait, we just check if the original panel is still present after creating a new panel.

dariusf: Good catch, it's different.

HansNewbie:
Good to go.

PR #1126, opened 2015-11-14
Tick mark and undo button for renaming panel

### LINKED ISSUE #1073: Tick mark and undo button for renaming panel
Add a tick mark to rename panels to make the editing process intuitive. This is a separate issue from #1053. Breaking it down to address @damithc's two comments at the end.

codemanmeet:
Fixes #1073.


codemanmeet:
renamed_panel

damithc:
Tick mark is fine, but the other icon looks like it means 'reload/refresh'. We want something to mean 'cancel' or 'back'

src/main/java/ui/components/PanelNameTextField.java
```diff
@@ -5,36 +5,50 @@
...
 import javafx.scene.control.TextField;
 import javafx.scene.input.KeyCode;
+import javafx.scene.layout.HBox;
 
 public class PanelNameTextField extends TextField {
 
     private final FilterPanel panel;
 
-    public PanelNameTextField(String panelName, FilterPanel panel) {
+    public PanelNameTextField(String panelName, HBox undoButton, HBox confirmButton, FilterPanel panel) {
```
dariusf: These aren't conceptually part of the text field, so they shouldn't be passed in here.

setNewName should be a method of FilterPanel which queries the text field for its value. That way this class is just a thin wrapper for a vanilla text field which handles key events and calls FilterPanel methods.

Querying for the text field value can be done via a new method I guess, since trim needs to be called. The method can return Optional<String>.

The button events can be bound in FilterPanel.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -287,13 +307,17 @@ private void activateInplaceRename() {
...
         renameTextField.setId(model.getDefaultRepo() + "_col" + panelIndex + "_renameTextField");
         renameTextField.setMaxWidth(NAME_FIELD_WIDTH);
         renameTextField.setMinWidth(NAME_FIELD_WIDTH);
         menuBarNameArea.getChildren().removeAll(nameBox, renameButton);
         panelMenuBar.getChildren().remove(menuBarCloseButton);
         menuBarNameArea.getChildren().addAll(renameTextField);
+        menuBarNameArea.setMinWidth(NAME_AREA_EDIT_WIDTH);
+        menuBarNameArea.setMaxWidth(NAME_AREA_EDIT_WIDTH);
+        panelMenuBar.getChildren().addAll(menuBarConfirmButton, menuBarUndoButton);
+        //panelMenuBar.setPadding(new Insets(0, 0, 8, 0));
```
dariusf: Can this be removed?

src/test/java/guitests/PanelRenameTest.java
```diff
@@ -104,6 +104,30 @@ public void panelRenameTest() {
...
+        // Expected: Panel is renamed after the button is pressed
+        pushKeys(CREATE_RIGHT_PANEL);
+        PlatformEx.runAndWait(() -> UI.events.triggerEvent(new ShowRenamePanelEvent(4)));
+        PlatformEx.waitOnFxThread();
+        type("Renamed panel with confirm tick");
+        click("#dummy/dummy_col4_confirmButton");
+        FilterPanel panel4 = (FilterPanel) panels.getPanel(4);
+        Text panelNameText4 = panel4.getNameText();
+        assertEquals("Renamed panel with confirm tick", panelNameText4.getText());
+        PlatformEx.waitOnFxThread();
```
dariusf: Are all these waits required?

dariusf:
FilterPanel is getting quite messy. Could you pull out everything related to editing titles into a new class? It would handle all the logic for swapping out the field for the label, confirming an edit, etc. It could even have a callback for informing FilterPanel on confirmation of the edit.

FilterPanel would use it very abstractly, by calling its methods and registering callbacks.

codemanmeet:
@damithc Let me know if this works - 

renamed_panel2

codemanmeet:
@dariusf Sure, I will add those functionalities to a new class!

damithc:
Yes, this one is better.

codemanmeet:
I shifted all the logic for button and keyboard events as well as logic for deleting/adding HBox's for menuBar items to PanelRenameTextField. `initializeRenameTextFieldAndEvents()` in FilterPanel.java assigns the object to `renameTextField`, sets its ID (FilterPanel specific). The events on `renameTextField` are then taken care of in PanelControl.java.

codemanmeet:
Removed unnecessary comments and `PlatformEx.waitOnFxThread()` as well.

codemanmeet:
I tried doing what you said but realised we cannot have the text field along with the other fields as the text field is dynamically generated only when the `ShowRenamePanelEventHandler` is called. All of our manipulations made on `menuBarConfirmButton`, `menuBarCloseButton` etc. are done only when the renameTextField is created. The close and pencil button are removed from one layer and replaced with undo, confirm button and *rename text field* on the same layer. All of this happens once the handle is given to it. If we have one more HBox/VBox on top, the button pressed and the renameTextField dynamically created cannot be done without avoiding boilerplate code being passed to it from FilterPanel.java.

But I agree the boilerplate is also an issue.  I suggest since there are only 4 methods, we can include them together to avoid the Boilerplate code. PanelRenameTextField adds only 1-2 additional CSS properties so it can be avoided altogether. The button and keyboard events can be tracked together in Filter Panel itself.

codemanmeet:
@dariusf Please ignore my previous comment. I managed to solve the boilerplate issue. As suggested, I created a new class called `PanelMenuBarWithRename.java` which extends `HBox` and presents a high level interface to `FilterPanel` for all event listeners and methods. Only one declaration is made in FilterPanel.java, which is an initialization to `panelMenuBar`.

damithc:
`PanelMenuBarWithRename` --> `HtPanelMenuBar` or something like that?

src/main/java/ui/components/PanelMenuBarWithRename.java
```diff
@@ -0,0 +1,105 @@
...
+import javafx.scene.control.Label;
+import ui.issuepanel.FilterPanel;
+import javafx.application.Platform;
+import javafx.scene.control.TextField;
+import javafx.scene.input.KeyCode;
+import javafx.scene.layout.HBox;
+
+/**
+ * Created by manmeetsinghsaluja on 23/11/15.
+ */
```
damithc: We don't usually show author name in the header comment. Write a proper header comment for this class.

src/main/java/ui/components/PanelMenuBarWithRename.java
```diff
@@ -0,0 +1,105 @@
...
+        this.panelIndex = panelIndex;
+        this.menuBarUndoButton = panel.menuBarUndoButton;
+        this.menuBarCloseButton = panel.menuBarCloseButton;
+        this.menuBarConfirmButton = panel.menuBarConfirmButton;
+        this.menuBarNameArea = panel.menuBarNameArea;
+        this.nameBox = panel.nameBox;
+        this.renameButton = panel.renameButton;
+    }
+
+    public void initializeRenameTextFieldAndEvents() {
```
damithc: Can this be just `init`?

damithc: Write header comments for non trivial public methods

src/main/java/ui/components/PanelMenuBarWithRename.java
```diff
@@ -0,0 +1,105 @@
...
+            renameTextField.requestFocus();
+            renameTextField.selectAll();
+        });
+        renameTextField.setText(panel.getCurrentName());
+        augmentRenameTextField();
+        buttonAndKeyboardEvents();
+        renameTextField.setPrefColumnCount(30);
+    }
+
+    private void buttonAndKeyboardEvents() {
```
damithc: A method name should indicate an action (i.e. a verb), not a noun.

src/main/java/ui/components/PanelMenuBarWithRename.java
```diff
@@ -0,0 +1,105 @@
...
+                panelNameSetterLogic();
+                closeRenameTextField();
+            }
+
+            e.consume();
+        });
+    }
+
+    // put the confirm and undo icons to the children of panelMenuBar and menuBarNameArea
+    // to show new icons when renaming
```
damithc: Write header comments in proper javadoc format

src/main/java/ui/components/PanelMenuBarWithRename.java
```diff
@@ -0,0 +1,105 @@
...
+        this.menuBarUndoButton = panel.menuBarUndoButton;
+        this.menuBarCloseButton = panel.menuBarCloseButton;
+        this.menuBarConfirmButton = panel.menuBarConfirmButton;
+        this.menuBarNameArea = panel.menuBarNameArea;
+        this.nameBox = panel.nameBox;
+        this.renameButton = panel.renameButton;
+    }
+
+    public void initializeRenameTextFieldAndEvents() {
+        renameTextField = new TextField();
```
damithc: `renameTextField` sounds like a method rather than a variable. Should use a noun. e.g. `renamableTextField`

src/main/java/ui/components/PanelMenuBarWithRename.java
```diff
@@ -0,0 +1,105 @@
...
+        menuBarNameArea.getChildren().remove(renameTextField);
+        this.getChildren().removeAll(menuBarConfirmButton, menuBarUndoButton);
+        this.getChildren().add(menuBarCloseButton);
+        menuBarNameArea.getChildren().addAll(nameBox, renameButton);
+        menuBarNameArea.setMinWidth(panel.NAME_AREA_WIDTH);
+        menuBarNameArea.setMaxWidth(panel.NAME_AREA_WIDTH);
+        panel.requestFocus();
+    }
+
+    private void panelNameSetterLogic(){
```
damithc: Not a good method name

damithc:
Added some comments about cosmetic issues.

codemanmeet:
Sure will address them!

src/main/java/ui/components/PanelMenuBarWithRename.java
codemanmeet: Oops, the IDE added it by default. I have added a javadoc explaining what the class does.

src/main/java/ui/components/PanelMenuBarWithRename.java
codemanmeet: Changed to `renamableTextField`.

codemanmeet:
@damithc @dariusf can we merge this?

src/main/java/ui/components/HTPanelMenuBar.java
```diff
@@ -0,0 +1,115 @@
...
+import javafx.scene.input.KeyCode;
+import javafx.scene.layout.HBox;
+
+/**
+ * An HTPanelMenuBar allows for the creation of a high level panel menu bar interface that
+ * contains logic for renaming the panel menu, handling the button and keyboard events
+ * and augmenting the children of panel menu bar defined in FilterPanel.java during the
+ * renaming process. The class is instantiated in FilterPanel.java.
+ */
+public class HTPanelMenuBar extends HBox {
```
damithc: @dariusf do we use `Hub` as the prefix or `Ht`? Use whatever we use in other places.

damithc:
No more comments from me.

src/main/java/ui/components/HTPanelMenuBar.java
dariusf: We use HT, but don't have many prefixed names. Did a quick search and there's only HTLog (logging specialised to us) and HTStatusBar (subclass of JavaFX StatusBar).

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -47,19 +49,22 @@
...
-    protected PanelNameTextField renameTextField;
+    public HTPanelMenuBar panelMenuBar;
+    private Text nameText;
+    public HBox nameBox;
+    public HBox menuBarNameArea;
+    public HBox menuBarCloseButton;
+    public HBox menuBarConfirmButton;
+    public HBox menuBarUndoButton;
+    public Label renameButton;
+    public Label closeButton;
```
dariusf: Are all these fields still needed? They could be contained inside the panel menu bar.

src/main/java/ui/issuepanel/PanelControl.java
```diff
@@ -46,7 +46,7 @@ public PanelControl(UI ui, Preferences prefs) {
         ui.registerEvent((PanelClickedEventHandler) e ->
                 setCurrentlySelectedPanel(Optional.of(e.panelIndex)));
         ui.registerEvent((ShowRenamePanelEventHandler) e ->
-                ((FilterPanel) getPanel(e.panelId)).showRenameTextField());
+                ((FilterPanel) getPanel(e.panelId)).initRenameHandler());
```
dariusf: Maybe the method could be named in a way that doesn't mention implementation details, like `startRename` or something?

src/test/java/guitests/UseGlobalConfigsTest.java
```diff
@@ -80,6 +78,7 @@ public void globalConfigTest() {
         pushKeys(KeyCode.SHIFT, KeyCode.SEMICOLON);
         type("issue");
         pushKeys(KeyCode.ENTER);
+        PlatformEx.waitOnFxThread();
```
dariusf: Is this needed? I think pushKeys might already wait a bit after it completes.

src/test/java/guitests/UseGlobalConfigsTest.java
```diff
@@ -96,10 +95,9 @@ public void globalConfigTest() {
         Label renameButton1 = filterPanel1.getRenameButton();
         click(renameButton1);
         type("Dummy 2 panel");
-        PanelNameTextField renameTextField1 = filterPanel1.getRenameTextField();
-        waitUntilNodeAppears(renameTextField1);
-        assertEquals("Dummy 2 panel", renameTextField1.getText());
-        pushKeys(KeyCode.ENTER);
+        push(KeyCode.ENTER);
+        assertEquals("Dummy 2 panel", filterPanel1.getNameText().getText());
+        PlatformEx.waitOnFxThread();
```
dariusf: Same here. We've already asserted the condition, so there might not be a need to wait.

src/test/java/guitests/UseGlobalConfigsTest.java
codemanmeet: I tried it without platform.wait too. The test failed :(

src/test/java/guitests/UseGlobalConfigsTest.java
codemanmeet: Same here!

src/main/java/ui/issuepanel/PanelControl.java
codemanmeet: sure will change it to `startRename`.

src/main/java/ui/issuepanel/FilterPanel.java
codemanmeet: These fields are needed. Their main CSS properties are still defined in FilterPanel. We are using them as instance variables of `panel` to make manipulations needed for rename.

src/test/java/guitests/UseGlobalConfigsTest.java
codemanmeet: This test works fine without platform.wait!

src/test/java/guitests/UseGlobalConfigsTest.java
dariusf: In that case, additional synchronisation may be needed. Which part fails?

src/main/java/ui/issuepanel/FilterPanel.java
dariusf: Can we move the related parts and initialisation into PanelMenuBar then? The new class is meant to encapsulate these, and we want to avoid having to know about them individually in FilterPanel.

src/main/java/ui/components/HTPanelMenuBar.java
dariusf: I guess we don't need the prefix here then? The name is unambiguous and tells us it's a component used by panels.

damithc: Yes, we should use the prefix only when there is another class with the same name.

src/main/java/ui/issuepanel/FilterPanel.java
codemanmeet: We are moving back to the same problem then. If we move all of FilterPanel's components to PanelMenuBar, PanelMenuBar will become messy (just like the original FilterPanel). I thought PanelMenuBar is only used for rename events and for manipulating it and its children in case of a rename. We still want all of the CSS initialisations in FilterPanel.

src/test/java/guitests/UseGlobalConfigsTest.java
codemanmeet: It passes now :)

codemanmeet: Ran it 4-5 times.

src/main/java/ui/issuepanel/FilterPanel.java
dariusf: I would say that the point of the menu bar class is exactly to abstract away details like CSS classes, what child objects are involved, etc. from FilterPanel.

FilterPanels are mainly concerned with filtering (dragging is related as its effects depend on the filter). The ability to be named is somewhat orthogonal, so we should try to keep it separate and abstract. It might even make sense in future to move it up to AbstractPanel, if we end up having panels with names that don't filter issues. Keeping things decoupled will help with that.

Anyway, if the menu bar component still turns out to be messy, it should be broken down further.

src/test/java/guitests/UseGlobalConfigsTest.java
dariusf: Hmm, but there is still the possibility that it might fail... Do take note of the error if it happens again, otherwise it will just surface at a later time.

By the way, `type` supports the colon character. You can put the colon directly in the string.

src/main/java/ui/issuepanel/FilterPanel.java
codemanmeet: Yup, agreed. Just figured won't be that messy as only the filter panel CSS parts are shifted. I have moved the CSS parts to the panel menu bar file and renamed it `PanelMenuBar.java`.

codemanmeet:
Let me know if it looks good to merge now :)

dariusf:
FilterPanel looks much neater. I would personally try to reduce the number of parameters to PanelMenuBar (using callbacks), but I think that's more work than it's worth right now, so it's fine.

Please squash the review commits. After that we have to get CI + coverage to pass.

codemanmeet:
Sure, will do!

codemanmeet:
It's ready to merge now!

PR #1141, opened 2015-12-13
Panel menu options for auto-creating custom panels (assignee, milestone, updated)

### LINKED ISSUE #1136: Panel menu options for auto-creating custom panels (assignee, milestone, updated)
The panel filters would be `is:open assignee:me` and `is:open updated:<24` respectively. The panel would be created on the right. @damithc Are the filters ok? Let me know!

codemanmeet:
Fixes #1136.

### DIFF
--- a/src/main/java/ui/PanelMenuCreator.java
+++ b/src/main/java/ui/PanelMenuCreator.java
@@ -21,7 +21,16 @@ import static ui.components.KeyboardShortcuts.CREATE_LEFT_PANEL;
 import static ui.components.KeyboardShortcuts.CREATE_RIGHT_PANEL;
 
 public class PanelMenuCreator {
-    private static final Logger logger = LogManager.getLogger(MenuControl.class.getName());
+    public static final Logger logger = LogManager.getLogger(MenuControl.class.getName());
+    public static final String MILESTONE_FILTER_NAME =  "milestone:curr sort:status";
+    public static final String MILESTONE_PANEL_NAME = "Current Milestone";
+
+    public static final String ASSIGNEE_FILTER_NAME =  "is:open ((is:issue assignee:me) OR (is:pr author:me))";
+    public static final String ASSIGNEE_PANEL_NAME = "Open issues and PR's";
+
+    public static final String UPDATED_FILTER_NAME = "assignee:me updated:<48";
+    public static final String UPDATED_PANEL_NAME = "Recently Updated issues";
+
     private final PanelControl panelControl;
     private final ScrollPane panelsScrollPane;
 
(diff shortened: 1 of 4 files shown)


src/main/java/ui/PanelMenuCreator.java
```diff
@@ -0,0 +1,133 @@
...
+
+    public MenuItem createMilestonePanel(){
+        MenuItem recentlyUpdatedPanel = new MenuItem("Current Milestone");
+        recentlyUpdatedPanel.setOnAction(e -> {
+            logger.info("Menu: Panels > Create Current Milestone panel");
+            generatePanelWithNameAndFilter("Current Milestone", "milestone:curr sort:status");
+            panelControl.selectLastPanel();
+        });
+        return recentlyUpdatedPanel;
+    }
```
damithc: The above three methods look very similar. Can they be abstracted into one method?

src/main/java/ui/PanelMenuCreator.java
```diff
@@ -0,0 +1,133 @@
...
+
+    public FilterPanel generatePanelWithNameAndFilter(String panelName, String filterName){
+        FilterPanel panelAdded = panelControl.addPanelAt(panelControl.getPanelCount());
+        panelAdded.setPanelName(panelName);
+        panelAdded.setFilterByString(filterName);
+        return panelAdded;
+    }
+}
```
damithc: @HansNewbie can code be reused across this PR and code for auto-creating boards, because creating boards involve creating panels? If not reuse, at least some consistency between the two?

HansNewbie: Yes, I think it's possible. But I think the method
`generatePanelWithNameAndFilter` would be better to be inside
`PanelControl`? This is to prevent the same issue with board currently
where board methods is unnecessarily inside menu instead of in its own
class, which in panel case would be `PanelControl`.

In reusing the code, should I do it in a separate issue once this PR is
done?

codemanmeet: Yup, `generatePanelWithNameAndFilter` should ideally be in `PanelControl.java` to avoid duplication. I'll put it there. The board auto creator class should then be able to use it!

src/main/java/ui/PanelMenuCreator.java
codemanmeet: Yup, I'll abstract them out!

codemanmeet:
Looks something like this - 
open_issues_pr

damithc:
Why not follow the same style as boards? i.e. `Auto-create --> panel name`

codemanmeet:
Yup, changed it to "Auto-create".

codemanmeet:
Let me know if this can be merged!

damithc:
Someone needs to review it. My comments don't count as a review. :-)

codemanmeet:
Sure :)

src/main/java/ui/PanelMenuCreator.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public Menu generatePanelMenu(){
+        Menu panelMenu = new Menu("Panels");
+        Menu autoCreatePanelMenu = new Menu("Auto-create");
+        List<MenuItem> items = new ArrayList<>();
+        List<MenuItem> autoCreateItems = new ArrayList<>();
+        items.add(createLeftPanel());
+        items.add(createRightPanel());
+        items.add(closePanel());
+        autoCreateItems.add(createCustomizedPanel(ASSIGNEE_PANEL_NAME, ASSIGNEE_FILTER_NAME));
```
ndt93: `createLeftPanel()` etc. trips me a bit when I'm reading this. Adding *Menu* or *MenuItem* suffix to menu related variables and methods could make things clearer.

ndt93:
If all the unstable tests (excluding the `RepositorySelectorTest`) can pass locally then this should be good.

src/main/java/ui/PanelMenuCreator.java
codemanmeet: Sure, will do that!

codemanmeet:
All unstable tests (excluding the `RepositorySelectorTest`) pass locally. Merging this now! :+1:

PR #1143, opened 2015-12-14
Integrate Mockito #1133

### LINKED ISSUE #1133: Integrate Mockito
http://mockito.org/

*Mockito* can be used to mock behaviors of classes which are not targets for a test case but are needed by classes under test. This is a step to work towards eliminating special variables for testing and manually constructed dummy classes.

ndt93:
Fixes #1133
* Add `mockito` dependency to gradle build
* Mock `Preferences`, `MultiModel`, `RepoSource`, `TurboIssue` in various test cases
* Remove most of stubbed classes.
* Rename tests to correspond to classes under test and move tests around accordingly.
* Refactor `RepoIO` and `Preferences` interface to be more injectable and remove correspondingly *test mode* variables. As a result, testing logic which was previously mixed with normal code is shifted to `TestController`.
* Fix bug in running `RepositorySelectorTest`.
* Minor refactoring of the tests due to above changes.

A common pattern among the tests is that the stubbed classes contain a large amount of emulated logic and partial stubbing (mixing between stubbed code and actual code). This is because we often try to test multiple things at once. Since we can easily mock classes now, ideally, we can try to test one thing at a time, return a simple value relevant to the test at hand in a stubbed method instead of emulating more complex logic and avoid partial stubbing. The [mockito docs](http://mockito.github.io/mockito/docs/current/org/mockito/Mockito.html) contain more info regarding various testing practices.


ndt93:
Another thing is that some of the unit test classes are named and coded according to the features and not the class under test, which gives a tendency to writing tests that cover multiple classes working together according to the feature instead of testing each class's method in isolation.

So for unit tests, we can try to create test classes with direct correspondence to classes in application code, which gives a focus to which class is under test. Also, an example of how `mockito` can help to test a class in isolation is shown in `RepoIOTests`.

damithc:
Good work @ndt93 
Let's see if anyone is available to review this.

damithc:
I guess our previous testing code is not ideal in the context of Mockito. Perhaps we can rewrite a small part of the test code to showcase the ideal way to use Mockito? That part of the test code can be used as the model for future test code. We should also refactor existing tests to improve the structure, but that could take some time to complete.

ndt93:
Yup, I'm thinking about rewriting a number of tests also, which will take some times to ensure coverage and all, so that should be on a separate issue. I could come back to converting more tests after experimenting with the `MockServer` part.

ndt93:
Anyone can help review this PR?

src/test/java/tests/MultiModelTest.java
```diff
@@ -23,7 +23,7 @@ public static void setup() {
         UI.status = new StatusUIStub();
     }
 
-    MultiModel multiModel = new MultiModel(new Preferences(true));
+    MultiModel multiModel = new MultiModel(mock(Preferences.class));
```
HansNewbie: Why do you use mock for some but test preference for others? Any rationale?

I agree that in unit testing we should use mock for other components, but will Mockito not mess with the application preference directory?

src/main/java/ui/UI.java
```diff
@@ -195,7 +186,7 @@ private void initPreApplicationState() {
                 logger.error(throwable.getMessage(), throwable));
 
         TestController.setUI(this, getParameters());
-        prefs = new Preferences(TestController.isTestMode());
+        prefs = new Preferences(TestController.getTestConfigFile(false));
```
HansNewbie: This looks confusing. Can we have `TestController.createRealPreference()` or `TestController.createApplicationPreference()` (can have a better name) just like for test preference?

src/main/java/prefs/Preferences.java
```diff
@@ -18,13 +17,8 @@
...
     public GlobalConfig global;
 
-    public Preferences(boolean isTestMode) {
-        if (isTestMode) {
-            this.fileHandler = new ConfigFileHandler(DIRECTORY, TEST_CONFIG_FILE);
-        } else {
-            this.fileHandler = new ConfigFileHandler(DIRECTORY, GLOBAL_CONFIG_FILE);
-        }
-
+    public Preferences(Optional<String> configFileName) {
```
HansNewbie: I prefer the original since the parameter can only be one or the other, but if we have a function that initialises this for both (as mentioned on my other comment), this should be fine.

HansNewbie:
Nice work! Aside from comments on the changes, a few more general comments:
- Does this fix the issue, or by-parts work for the issue? Because I didn't see any comment mentioning so.
- We will need a documentation about mocking, like when to use mock, an example of mocking (the `RepoIOTests`), and where to find the documentation.

One question is why are all the serializable being removed? From a quick glance seems that they are not being used anyway but I am not sure if that is true.

src/test/java/tests/MultiModelTest.java
ndt93: The mocks are used on tests that are not trying to test `Preferences` but need it to test other classes. E.g. all tests in `MultiModelTests` don't use `Preferences` but need it for the constructor.

`mockito` fakes the behavior of a class and doesn't use the original class's code unless partial mocking features like `spy` are used.

src/main/java/ui/UI.java
ndt93: Ok sure. Good idea.

src/main/java/prefs/Preferences.java
ndt93: This is changed to make the class more configurable and so testable in a way without having a explicit test mode variable that branches out testing and application logic. The test mode variable may then encourage further branching logic in other methods of the class in the future.

I think it's still simple enough here but if we have to do too much of these to make things more testable then it's an indication of other problems as discussed [here](http://gmoeck.github.io/2012/07/09/dont-make-your-code-more-testable.html).

ndt93:
Oh right, totally forget about the fix comment. I will add `mockito` to the docs.

I wrote the `toString` for all the `Serializable*` last time to debug the JSON explosion, but the same thing can be done with gson's pretty print. So I was revisiting test classes that are named according to features and not classes under test and take the chance to delete redundant codes :-P.

ndt93:
@HansNewbie I've made the changes and also removed all the confusing `get*` methods in `TestController`. You can check this again.

src/test/java/tests/MultiModelTest.java
HansNewbie: Sure. Can add it into documentation? It's quite basic in terms of unit testing but code copying might result in people using test pref when they should use mocking and vice versa.

src/main/java/prefs/Preferences.java
HansNewbie: Alright. I think this is fine.

HansNewbie:
Noted for serializable.

src/main/java/prefs/Preferences.java
HansNewbie: Thinking about it again, can we flip this? When configFileName is empty, it changes `TEST_CONFIG_FILE`. We will supply `GLOBAL_CONFIG_FILE` for real application. My concern is that if we have a bug on creating preferences in tests, it may mess up actual preference of the developer. Defaulting to test config would be safer.

What do you think?

src/main/java/ui/TestController.java
```diff
@@ -79,4 +84,46 @@ public static boolean isCloseOnQuit() {
...
+                          isTestJSONEnabled() ? Optional.of(new JSONStoreStub()) : Optional.empty(),
+                          isTestMode() ? Optional.of(RepoStore.TEST_DIRECTORY) : Optional.empty());
+    }
+
+    /**
+     * Creates a partially stubbed RepoIO used for testing
+     * @param enableJSONStore set to true if actual json file is to be used for storing repos' data
+     * @return
+     */
+    public static RepoIO createTestingRepoIO(boolean enableJSONStore) {
```
HansNewbie: I think `createTestingRepoIO(false)` can be misleading. Does it mean we are creating an actual repo IO (the convention you used previously in this PR)? Would it be better to pass `Optional<JSONStore> jsonStoreToBeUsed` instead for the parameter? It can use `Optional.empty()` by default.

src/main/java/prefs/Preferences.java
ndt93: Good point. I think in that case we should just ask for a `String` without an `Optional`.

src/main/java/ui/TestController.java
ndt93: Yeah. This is the parameter that is used in the previous `RepoIO`'s constructor. I got confused by this and the `isTestJSONEnabled()` a few times.

src/test/java/tests/MultiModelTest.java
ndt93: I think it's not a strict guideline to use mocks as long as it's clear what's under test and the real classes can be used without too much additional logic. E.g. it's fine to use the original Preference here too. But sure, I will add some pointers regarding this to the docs.

src/main/java/backend/RepoIO.java
```diff
@@ -37,20 +34,19 @@
...
+     * parameters for repos source and storage which are useful for testing purposes.
+     * @param repoSource optional source of repos. Default to GitHubSource if not present
+     * @param jsonStore optional storage for repos. Default to a new JSONStore if not present
+     * @param storeDirectory optional directory for storing repos. Default value is in RepoStore.
+     */
+    public RepoIO(Optional<RepoSource> repoSource, Optional<JSONStore> jsonStore,
+                  Optional<String> storeDirectory) {
+        this.repoSource = repoSource.orElseGet(() -> new GitHubSource());
+        storeDirectory.ifPresent((dir) -> RepoStore.changeDirectory(dir));
+        this.jsonStore = jsonStore.orElseGet(() -> new JSONStore());
```
dariusf: We can use `orElse` here, then we don't need the extra lambda. Also, maybe use a method reference for the second line? `ifPresent(RepoStore::changeDirectory)`

dariusf:
Nice work. The new interfaces are much more explicit about what's being stubbed. Also being able to delete so much old code is a good sign. :smile: 

Is there anything else you are planning to do for this PR? It seems open-ended and lots more tests could use mockito effectively. We should try to scope it so work can be picked up in future issues.

Docs [here](https://github.com/HubTurbo/HubTurbo/blob/1133-integrate-mockito/docs/settingUpDevEnvironment.md#tool-stack) can be updated also.

src/main/java/backend/RepoIO.java
ndt93: The first line is to avoid creating a new instance of `GitHubSource` (which takes quite some time) when it's not needed. I will change to method's reference in the second line.

ndt93:
I may write a couple of tests more to improve coverage but that should be about it for this PR. I'm not sure how could should scope the tests conversion yet beside going through tests one by one and find places where `mockito`can improve it. It could take a few tests to replace 1 existing tests since they tend to cover multiple classes at the same time.

I will update the docs.

ndt93:
@HansNewbie Ready for review again.

HansNewbie:
@ndt93 looks okay to me. However, the commit on update to doc seems to have some issues. I am not sure if this is going to affect the repo. If you check `File changed` above, all the files have their mode being changed (seems that now all `.java` becomes executable). Is it possible to correct that commit?

If that can be changed (or we can be sure it won't affect the repo), I will change the status to `toMerge`.

ndt93:
Oh right, I'm copying the repo around and the permissions were changed. I will revert the permissions back.

ndt93:
@HansNewbie Reverted back.

src/test/java/tests/RepositorySelectorPopulationTests.java
```diff
@@ -47,16 +49,17 @@ public void noJsonFiles() {
     public void oneInvalidJsonFile() throws IOException {
         File invalidJson = new File("store/test/dummy-dummy.json");
         assert invalidJson.createNewFile();
-        RepoIO testIO = new RepoIO(true, true);
+        RepoIO testIO = TestController.createTestingRepoIO(Optional.of(new JSONStoreStub()));
```
HansNewbie: Why don't this one and the above use `Optional.empty()` instead? Is it because there is no JSON Store method being called?

ndt93: `Optional.empty()`defaults to a normal `JSONStore`, not a `JSONStoreStub`.

src/main/java/ui/TestController.java
```diff
@@ -79,4 +85,46 @@ public static boolean isCloseOnQuit() {
...
+    }
+
+    /**
+     * Creates a RepoIO for the application that uses different components
+     * depending on various test options: --test, --testjson etc.
+     * @return
+     */
+    public static RepoIO createApplicationRepoIO() {
+        return new RepoIO(isTestMode() ? Optional.of(new DummySource()) : Optional.empty(),
+                          isTestJSONEnabled() ? Optional.of(new JSONStoreStub()) : Optional.empty(),
```
HansNewbie: I am confused about this line. In the original, if `isTestJSONEnabled()` is true, it will use `Optional.empty`. Otherwise, it will use the stub. This holds true if it is in test mode.

I think this can be made close to the original and more readable with this
```
if (isTestMode()) {
  return new RepoIO(Optional.of(new DummySource()),
      isTestJSONEnabled() ? Optional.empty() : Optional.of(new JSONStoreStub()),
      Optional.of(RepoStore.TEST_DIRECTORY));
} else {
  return new RepoIO(Optional.empty(), Optional.empty(), Optional.empty());
}
```

The above should be safe since if test json command line option is set to true, i.e. `isTestJSONEnabled()` is true, `isTestMode()` would return true. Hence if developer forgot to put test mode to true when testing json is set to true, the real JSONs will not be messed up.

src/test/java/tests/RepositorySelectorPopulationTests.java
HansNewbie: Wouldn't we want to use the normal `JSONStore` here? Since we are testing the JSON operations.

I am just ensuring the consistency since all the tests in this file used `RepoIO(true, true)` but you changed some to use `JSONStoreStub()` and some to `Optional.empty()`. Is there a reason behind that?

ndt93: Oh right, those 2 tests below are supposed to test the JSONStore itself with actual files (the above ones are invalid cases).  Not sure why the stub was used before. I can change it back to the original ones though but they will affect the coverage (that's how I found them).

HansNewbie: I think `noJsonFiles()` can use the stub, but `oneInvalidJsonFile()` should stay using proper JSONStore. Will that still affect the coverage?

ndt93: Ok I will check them.

src/main/java/ui/TestController.java
ndt93: Sure, that's more readable.

src/test/java/tests/RepositorySelectorPopulationTests.java
ndt93: I think this tests suite is weird in the first place. It's called `RepositorySelectorPopulationTests` but has nothing to do with that? It's more like `StoreTests`, and in that case, all the tests should use the real `JSONStore`.

HansNewbie: From what I understand, this test suite is used to test RepoIO repository list, which is based on the JSON. It tests populating the RepoIO. I don't think it's a `StoreTest`, but now that I relook at it, I agree that all the tests should use the real `JSONStore`.

src/main/java/ui/TestController.java
ndt93: Another thing though, the when `isTestJSONEnabled` is true (i.e. --testjson=true), the stub is supposed to be used since the default is false.

HansNewbie: I think from the original `RepoIO`, when `isTestJSONEnabled` is true _and_ is in test mode, it will use true JSON. If `isTestJSONEnabled` is false _and_ is in test mode, then it will use the stub (`isTestJSONEnabled` signifies that we will use some real JSON operation in test folder).

src/main/java/ui/TestController.java
```diff
@@ -79,4 +85,50 @@ public static boolean isCloseOnQuit() {
...
+
+    /**
+     * Creates a RepoIO for the application that uses different components
+     * depending on various test options: --test, --testjson etc.
+     * @return
+     */
+    public static RepoIO createApplicationRepoIO() {
+        if (isTestMode()) {
+            return new RepoIO(Optional.of(new DummySource()),
+                              isTestJSONEnabled() ? Optional.of(new JSONStoreStub()) : Optional.empty(),
```
HansNewbie: I really think this should be reversed. By default (i.e. `false`) in test mode, we would like not to use real JSON (since we would like to test other units aside from JSON), Hence, we should use the stub on `false`. When we would like to incorporate JSON part in testing or would like to test the JSON component, then we will set the `--testjson=true` so that we will use real JSON component (although in test folder).

ndt93:
@HansNewbie I added in the changes. Coverage decreases but those uncovered lines should not be caused by the changes.

src/main/java/ui/TestController.java
ndt93: `isTestJSONEnabled` returns the value of `--testjson`program argument which has a opposite meaning of the similarly named parameter in the original `RepoIO`.

src/main/java/ui/TestController.java
ndt93: Yeah, I think we should use an entirely different name along the line of `disable/enable`JSON store. However, the current meaning of `--testjson` has been used in all guitests. We can open another issue for that since it's quite error prone to refactor from my experience and isn't really related to this PR.

HansNewbie: Yep sure. Can you file an issue about it?

HansNewbie: Oops, one minor improvement. Can we use `return createTestingRepoIO(isTestJSONEnabled() ? Optional.of(new JSONStoreStub()) : Optional.empty())` here? Since we already have such method.

HansNewbie:
Good job for this PR! If you have resolved my latest comment on using `createTestingRepoIO` and this passes the tests, will be good to merge. Update me again if it is done and I will change the PR status.

ndt93:
@HansNewbie I've made the changes.

HansNewbie:
Looks good. The coverage drops seems to be false. @dariusf can help merge?

PR #1144, opened 2015-12-16
BoardSwitchTest uses real preferences and not test preferences #1111

### LINKED ISSUE #1111: BoardSwitchTest uses real preferences and not test preferences
Should be a quick fix; change all `testPrefs = new Preferences(false);` to `testPrefs = new Preferences(true);`.

Discovery: test suddenly fails in Mac after auto-creating work allocation boards. Log shows that it is referencing some work allocation board name instead of the testing board name.

nicholaslum444:
Fixes #1111

### DIFF
--- a/src/test/java/tests/BoardSwitchTest.java
+++ b/src/test/java/tests/BoardSwitchTest.java
@@ -18,7 +18,7 @@ public class BoardSwitchTest {
 
     @Test
     public void boardsSwitchTest() {
-        testPrefs = new Preferences(false);
+        testPrefs = new Preferences(true);
 
         List<PanelInfo> board1 = new ArrayList<PanelInfo>();
         List<PanelInfo> board2 = new ArrayList<PanelInfo>();
@@ -46,7 +46,7 @@ public class BoardSwitchTest {
 
     @Test
     public void noBoardSwitchTest() {
-        testPrefs = new Preferences(false);
+        testPrefs = new Preferences(true);
 
         testPrefs.switchBoard();
         assertEquals(false, testPrefs.getLastOpenBoard().isPresent());
@@ -54,7 +54,7 @@ public class BoardSwitchTest {
 
     @Test
     public void noBoardOpenSwitchTest() {
-        testPrefs = new Preferences(false);
+        testPrefs = new Preferences(true);
 
         List<PanelInfo> board1 = new ArrayList<PanelInfo>();
         List<PanelInfo> board2 = new ArrayList<PanelInfo>();
@@ -70,7 +70,7 @@ public class BoardSwitchTest {
 
     @Test
     public void oneBoardSwitchTest() {
-        testPrefs = new Preferences(false);
+        testPrefs = new Preferences(true);
 
         List<PanelInfo> board1 = new ArrayList<PanelInfo>();
         testPrefs.addBoard(BOARDNAME1, board1);


HansNewbie:
The PR is alright. Just 2 things:
- clear your commits (look at `git rebase`; usually I use `git rebase -i origin/master`)
- have you run the tests locally?

nicholaslum444:
Yes, I have run the tests locally.

guitests: 100% passed (with issues, see below)
tests: 100% passed
unstable: 7/20 passed

Having an issue with guitests. I am unable to run the entire test suite at once. I need to run them individually.

PanelsTest fails on its own because the mouse click to close the middle panel is slightly off target (time out exception since it doesn't close). If I manually close the panel, the test can continue and pass.

All the other guitests pass when run on their own. However when run them as a suite, many tests fail. Seems like the failures are caused by reading wrong values from test.json at the wrong time, which eventually crashes the app. After that, every test fails due to the app crash. [Click here to see the output as a gist.](https://gist.github.com/nicholaslum444/b25560ac5ea8836acb8c)

Please let me know if there is anything I should be aware of while running the guitests.

nicholaslum444:
Update: I was previously running the tests in IntelliJ. Now I figured out how to run the tests with gradle instead. The guitests are still slightly wonky though. Will update more again... Sorry for the hassle

HansNewbie:
Which platform are you at? Linux? Running tests on Windows and Mac should be fine. The caveat on Windows is that sometimes it fails randomly due to the window switching not working properly; just run the tests again.

Unstable at 7/20 is worry-some, as usually I have most of them passes, except 1 or 2 tests.

PanelsTest fail should not happen. I will try that out on my side, see if there is something wrong.

I am not sure about the gist. We'll look at it again if the other issues are resolved.

In case you didn't know yet, when running guitests, you can't move your mouse at all. Also, Caps lock should be off (from the gist it seems that you have it off, anyway). I don't think there is any other thing you need to be aware of.

About the screenshots, yes they are normal. This is because testFX will generate screenshot every time there is an exception, even if we catch or expect (with JUnit) the exception. Most of them are generated when we would like to wait for a UI element to appear (you can look at the code `waitUntilNodeAppears` if interested). You can delete them after running the tests.

HansNewbie:
Update: PanelsTest runs fine on my Mac. The click position is correct. Can you try again?

nicholaslum444:
I'm running on Windows.

The large number of unstable fails was due to me running in IDE. I just ran on gradle again; only guitests.PanelsTest and unstable.RepositorySelectorTest fails. 2/234 failed.

The Panels thing can't be helped on my com, I think. Might have something to do with the resolution. It properly selects the close button on the middle panel, but the mouse is just slightly off the clickable portion of the button.

I will re-run the tests again tonight to confirm. I will also try the rebase thing.

ndt93:
There is a bug in the `RepositorySelectorTest`. I've fixed it in my PR so don't worry about it.

damithc:
@ndt93 may be extract the fix into a separate PR? It's preferrable not to do multiple things in the same PR.

ndt93:
@damithc Ok sure.

nicholaslum444:
I re-ran the tests on Windows. Same results. Panels still facing the same issue. It cannot click the close button by itself, but if I help it, then it will pass. I will try again on linux when I get the chance. 

Also, I have rebased to the (only) commit that I made for this issue. However now it is behind the base branch. Should I click the "Update branch" button? Please advise on the best way to pull in changes.

HansNewbie:
I think Linux has troubles with some of the tests. For the Panels, let's ignore that for now since it passes Travis and works on Mac. I will try out on my Windows also and file an issue if it fails consistently. It might be your computer's configuration; we will see if it can be fixed next time.

On rebasing, generally before you rebase, you should update your master branch to the latest commit first (pull first), and then rebase your branch.

If you are using GitHub desktop app, you can use the "update from master" button on your branch if there is any changes to the master (basically it merges master to your branch). After that, you rebase your branch. Generally I use this workflow since it is fast and easy. The trouble here is when there is a merge conflict; if you "update from master" on your branch, resolve the conflict, then you rebase to master, you need to repeat the conflict resolution again. If you know that there is going to be conflict, I suggest to use the CLI git; pull master branch, rebase your branch, and resolve conflict.

HansNewbie:
Update me when you have rebased to the current master.

nicholaslum444:
I think I rebased correctly... Sorry for the hassle

HansNewbie:
This PR can be merged. The coverage drop have been verified to be false. Will need @dariusf help to merge.

damithc:
@HansNewbie can you merge?

HansNewbie:
@damithc the drop in coverage (even though false) will not allow me to push the merged master. Only the admin can merge and push. This is some sort of protection that PRs that fail to pass test or coverage will not be able to be merged without admin privilege.

damithc:
I see, @dariusf can merge it when he is back.

dariusf:
Okay, will merge. Thanks everyone.

PR #1150, opened 2015-12-17
Users can use aliases for qualifiers #1145

### LINKED ISSUE #1145: Users can use aliases for qualifiers
`milestone` : `m`
`status` : `s`

@damithc any more that we could include?

codemanmeet:
Fixes #1145.

### DIFF
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -152,6 +152,8 @@ Matches all issues associated with any milestones whose names contain the given 
 
 `current` or `curr` can be used to refer to an open milestone with earliest due date, or an overdue open milestone with open issues if any. `curr-2`, `curr-1`, `curr+1`, `curr+2`, etc. (no space before or after `-`/`+`) can then be used to refer to milestones before or after `current`, sorted by due date. Milestones with no due date will not be considered in resolving `curr`/`current`. If there is no open milestone, `current` will refer to no milestone, while `current-1` will refer to the last closed milestone.
 
+Aliases: `m`
+
 ### label
 
 *Expects a string in the form `name`, `group.name`, or `group.`*
@@ -164,11 +166,15 @@ Matches all issues with the given label. A group name may be used to qualify the
 
 Matches all issues assigned to the given person, identified by the given alias, GitHub username, or real name, in that priority.
 
+Aliases: `as`
+
 ### author
 
 *Expects a string*
 
 Matches all issues created by the given person, identified by the given alias, GitHub username, or real name, in that priority.
+
+Aliases: `au` or `creator`
 
 ### involves
 
@@ -181,6 +187,8 @@ Matches all issues involving (assigned to or created by) the given person, ident
 *Expects one of `open` or `closed`*
 
 Matches all issues of the given state.
+
+Aliases: `s`
 
 ### has
 
--- a/src/main/java/filter/expression/Qualifier.java
+++ b/src/main/java/filter/expression/Qualifier.java
@@ -28,10 +28,10 @@ public class Qualifier implements FilterExpression {
     public static final Qualifier FALSE = new Qualifier("false", "");
 
     public static final List<String> KEYWORDS = Collections.unmodifiableList(Arrays.asList(
-        "assignee", "author", "body", "closed", "comments", "created", "creator",
+        "assignee", "as", "author", "au", "body", "closed", "comments", "created", "creator",
         "date", "desc", "description", "false", "has", "id", "in", "involves",
-        "is", "issue", "keyword", "label", "labels", "merged", "milestone", "milestones",
-        "no", "nonSelfUpdate", "open", "pr", "pullrequest", "read", REPO, SORT, "state", "status",
+        "is", "issue", "keyword", "label", "labels", "merged", "milestone", "m", "milestones",
+        "no", "nonSelfUpdate", "open", "pr", "pullrequest", "read", REPO, SORT, "state", "status", "s",
         "title", "type", "unmerged", "unread", UPDATED, "user"
     ));
 
@@ -243,13 +243,16 @@ public class Qualifier implements FilterExpression {
         case "description":
             return bodySatisfies(issue);
         case "milestone":
+        case "m":
             return milestoneSatisfies(model, issue);
         case "label":
             return labelsSatisfy(model, issue);
         case "author":
         case "creator":
+        case "au":
             return authorSatisfies(issue);
         case "assignee":
+        case "as":
             return assigneeSatisfies(model, issue);
         case "involves":
         case "user":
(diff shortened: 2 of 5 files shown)


codemanmeet:
These are the qualifier aliases - 
`milestone` : `m`
`status` : `s`
`author` : `au`
`assignee` : `as`

src/test/java/tests/FilterEvalTests.java
```diff
@@ -455,6 +473,12 @@ public void author() {
         assertEquals(false, matches("author:alice", issue));
         assertEquals(true, matches("author:o", issue));
         assertEquals(false, matches("author:lic", issue));
+
+        assertEquals(true, matches("au:BOB", issue));
+        assertEquals(true, matches("au:bob", issue));
+        assertEquals(false, matches("au:alice", issue));
+        assertEquals(true, matches("au:o", issue));
+        assertEquals(false, matches("au:lic", issue));
```
damithc: Instead of duplicating the lines, can we abstract it to a method and call twice?

src/test/java/guitests/FilterTests.java
```diff
@@ -23,6 +23,14 @@ public void parseExceptionTest() {
         push(KeyCode.ENTER);
 
         assertEquals(issuePanel.getIssueCount(), 10);
+
+        // test parse exception returns Qualifier.EMPTY, i.e. all issues
+        click("#dummy/dummy_col0_filterTextField");
+        selectAll();
+        type("m:");
+        push(KeyCode.ENTER);
```
damithc: Abstract these two lines into a `typeAndEnter(string)` method?

src/main/java/filter/expression/Qualifier.java
```diff
@@ -28,10 +28,10 @@
...
     public static final List<String> KEYWORDS = Collections.unmodifiableList(Arrays.asList(
-        "assignee", "author", "body", "closed", "comments", "created", "creator",
+        "assignee", "as", "author", "au", "body", "closed", "comments", "created", "creator",
         "date", "desc", "description", "false", "has", "id", "in", "involves",
-        "is", "issue", "keyword", "label", "labels", "merged", "milestone", "milestones",
-        "no", "nonSelfUpdate", "open", "pr", "pullrequest", "read", REPO, SORT, "state", "status",
+        "is", "issue", "keyword", "label", "labels", "merged", "milestone", "m", "milestones",
+        "no", "nonSelfUpdate", "open", "pr", "pullrequest", "read", REPO, SORT, "state", "status", "s",
         "title", "type", "unmerged", "unread", UPDATED, "user"
     ));
```
damithc: Each string literal appears twice in this class. We can remove that duplication. e.g.
1. define an enumeration for all key words e.g. `KEYWORDS={ASSIGNEE, AUTHOR, ...}`
2. create a hasmap where the key is the string literal and the value is the enumeration value
3. during parsing, retrieve the enumeration value using the string literal and do the `switch` based on the enum value.
This can be done as a separate issue. Check with @dariusf too.

damithc:
Forgot to update docs?

src/test/java/tests/FilterEvalTests.java
codemanmeet: Yup, will abstract it out!

src/test/java/guitests/FilterTests.java
codemanmeet: Yup, will abstract this too.

src/main/java/filter/expression/Qualifier.java
codemanmeet: Sure, let's do that! This requires refactoring. I have created a separate issue to handle this #1152.

codemanmeet:
updated the docs!

docs/filters.md
```diff
@@ -150,6 +150,8 @@ Matches all issues with a body (or description) containing the given string.
 
 Matches all issues associated with any milestones whose names contain the given string.
 
+Qualifier alias `m` can also be used in addition to `milestone`.
```
damithc: May be use a shorter format? e.g.
Aliases: m

src/main/java/filter/expression/Qualifier.java
damithc: You may want to copy paste my comment to the issue. Otherwise diff comments could get lost through squashing.

src/test/java/tests/FilterEvalTests.java
```diff
@@ -124,6 +124,20 @@ public void in() {
         assertEquals(false, matches("in:something a is", issue));
     }
 
+    private void milestoneTest(String milestoneQualifier, TurboIssue issue, IModel model) {
```
ndt93: Could be helpful if tests and their helper methods like these have comments or more informative names saying what they are trying to test so it's easier to check for coverage and also to fix or refactor them in the future.

src/main/java/filter/expression/Qualifier.java
codemanmeet: :+1: I'll do that

src/test/java/tests/FilterEvalTests.java
codemanmeet: I'll add relevant comments to the helper methods!

src/test/java/tests/FilterEvalTests.java
```diff
@@ -124,6 +124,24 @@ public void in() {
         assertEquals(false, matches("in:something a is", issue));
     }
 
+    /**
+     * Helper method to test the parsing of the milestone qualifier and its values
+     * Uses the relevant milestone qualifier string and the stub issue and model as parameters
+     */
+    private void milestoneTest(String milestoneQualifier, TurboIssue issue, IModel model) {
```
damithc: Instead of the comment, you can name the method `testMilestoneParsing` or something like that? method should be named as actions not nouns.

codemanmeet: Sure!

codemanmeet:
@ndt93 let me know if I can go ahead and merge this!

src/main/java/filter/expression/Qualifier.java
dariusf: Yes, we should use an enum for keywords.

About this specific change: this keyword list is used for autocompletion, so I don't think we need to have the abbreviations here.

codemanmeet: Yup! removed the abbreviations from this list.

damithc:
May be the list of auto-complete words can be generated automatically? e.g. all keywords that are longer than 2 chars

ndt93:
@codemanmeet No more comments from me but you may want to discuss some issues with @damithc above before closing this.

damithc:
OK for me. I don't like how aliases result in a lot of duplicate test cases. After we replace magic strings with enums, we should try to reduce such duplication, or automate it (i.e. run the test case for each alias automatically)

PR #1155, opened 2015-12-19
Remove dead code in FilterPanel #958

### LINKED ISSUE #958: Remove dead code in FilterPanel
The methods ```getCurrentName``` and ```filterByString``` in ```FilterPanel``` are not accessed anywhere, and are redundant. These two also seem to be one of the causes of frequent coverage decrease. Unless there is any future use for these, we could delete them.

ravic1995:
Fixes #958


dariusf:
getCurrentFilterString is also unused.

Also, I don't think you need to specify method names in the commit message -- that kind of information can be found in the diff.

dariusf:
It'd be good if you could squash. Generally we try to keep the level of abstraction high in commits, so each commit represents a feature or self-contained change, rather than unrelated or partial changes.

Also, feel free to comment and notify the reviewer when you're ready.

ravic1995:
Ok I will keep that in mind. 
I will commit again . 

thank you

dariusf:
Do you need help with this? You can join the Slack group and ask there.

PR #1158, opened 2015-12-20
Removed dead/unused code from class FilterPanel #958

### LINKED ISSUE #958: Remove dead code in FilterPanel

ravic1995:
- Fixes #958


dariusf:
Good work.

PR #1159, opened 2015-12-20
Refactor keywords in the Qualifier class. #1152

### LINKED ISSUE #1152: Refactor keywords in the Qualifier class.
As discussed in #1150, keywords need to be refactored in order to avoid duplication in `Qualifier.java`.

### LINKED ISSUE #1157: Improve test cases for aliases in filters
At the moment test cases for filters have many duplicates due to aliases. i.e. copy-paste-modifying the test cases for each alias. We should make clean it up. options:
* separate alias testing from filter testing
* automate the alias testing (i.e. the same tests are run `for each` alias automatically, without having duplicate code).

#1152 needs to be done first.

dariusf:
Fixes #1152, fixes #1157

### DIFF
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -144,6 +144,8 @@ Matches all issues with a title containing the given string.
 
 Matches all issues with a body (or description) containing the given string.
 
+Aliases: `desc`, `description`
+
 ### milestone
 
 *Expects a string*
@@ -174,7 +176,7 @@ Aliases: `as`
 
 Matches all issues created by the given person, identified by the given alias, GitHub username, or real name, in that priority.
 
-Aliases: `au` or `creator`
+Aliases: `au`, `creator`
 
 ### involves
 
@@ -182,31 +184,41 @@ Aliases: `au` or `creator`
 
 Matches all issues involving (assigned to or created by) the given person, identified by the given alias, GitHub username, or real name, in that priority.
 
+Aliases: `user`
+
 ### state
 
 *Expects one of `open` or `closed`*
 
 Matches all issues of the given state.
 
-Aliases: `s`
+Aliases: `s`, `status`
 
 ### has
 
-*Expects one of `label`, `milestone`, `assignee`
+*Expects one of `label`, `milestone`, `assignee`*
 
-Matches all issues possessing the given attribute.
+Matches issues associated with the given type of resource.
+
+Aliases:
+
+- `label` can be written as `labels`
+- `milestone` can be written as `milestone` or `m`
+- `assignee` can be written as `assignee` or `as`
 
 ### no
 
-*Expects one of `label`, `milestone`, `assignee`
+*Expects one of `label`, `milestone`, `assignee`*
 
-The negation of `has`. Matches all issues without the given attribute.
+The negation of `has`. Matches issues not associated with the given type of resource.
+
+Aliases: see [has](#has).
 
 ### in
 
 *Expects one of `title` or `body`*
 
-Meta-qualifier. Changes the semantics of `keyword` so it only searches in either the title or body.
+Meta-qualifier. Changes the semantics of search terms to check only either the title or body.
 
 ### type
 
@@ -240,18 +252,20 @@ Matches issues of the given repository. If omitted, will match issues of the def
 
 ### sort
 
-*Expects a comma-separated list of sorting criteria. For example, `repo, ~updated, -comments`. Here are the available sort criteria:*
-- `comments` : sorts issues by number of comments (in descending order)
-- `repo` : sorts issues by repo. Only applicable if the panel is showing issues from more than one repo
-- `updated` (or `date`) : sorts issues by their updated time (latest updated issues are shown first)
-- `id` : sorts issues by id (in ascending order)
-- `assignee` : sorts issues by assignees (in alphabetical order of assignees' names)
-- `priority` : sorts issues within the group by name (high, low, etc.) (in alphabetical order)
-- `status` : sorts issues by status (`open` issues followed by `closed` issues)
-- `milestone` : sorts issues by their milestones (milestones with latest due date are shown first)
-- label group : sorts issues by the label group specified (in alphabetical order).Label groups can be disambiguated by appending a `.`. For example, `sort:priority.` will sort issues by their priorities in alphabetical order (Will show issues with `high` priority first, followed by `low` priority issues and then followed by `medium` priority)
+*Expects a comma-separated list of sorting criteria. For example, `repo, ~updated, -comments`.*
 
 Sorts a repository by the list of criteria, going from left to right. Negated criteria will reverse the ordering that they describe.
+
+Available sorting criteria:
+
+- `comments`: sorts by number of comments (in descending order)
+- `repo`: sorts by repo. Only applicable if the panel is showing issues from more than one repo
+- `updated` (or `date`): sorts issues by their updated time (latest updated issues are shown first)
+- `id`: sorts by issue id (in ascending order)
+- `assignee`: sorts by assignee (in alphabetical order of assignees' names). Aliases: `a`
+- `state`: sorts by status (open issues followed by closed issues). Aliases: `status`, `s`
+- `milestone`: sorts by milestone (latest due date first). Aliases: `m`
+- Anything else is interpreted as a *label group*: sorts by the label group specified in alphabetical order. Label groups can be disambiguated by appending a `.`, if there is a label group that clashes with one of the above names. For example, `sort:priority.` will sort issues by their priorities in alphabetical order (issues with `high` priority will come first, followed by `low` priority, then `medium` priority)
 
 ## Additional features
 
@@ -259,7 +273,7 @@ HubTurbo automatically downloads detailed information about issues when the [`up
 
 When the [`updated`](#updated) filter is specified, the issues to be displayed are also automatically sorted by the latest *non-self update* i.e. the last time someone other than the currently logged-in user makes a change to the issue. This order can be explicitly overridden by specifying another sort order through the [`sort`](#sort) filter.
 
-To use a reverse-non-self-update or combine sorting by non-self-update times with other sorting orders, use the `nonSelfUpdate` sorting key e.g. `sort:-nonSelfUpdate` or `sort:nonSelfUpdate,comments` 
+To use a reverse-non-self-update or combine sorting by non-self-update times with other sorting orders, use the `nonSelfUpdate` sorting key e.g. `sort:-nonSelfUpdate` or `sort:nonSelfUpdate,comments`
 
 ## Incompatibilities
 
--- a/src/main/java/filter/Parser.java
+++ b/src/main/java/filter/Parser.java
@@ -112,12 +112,12 @@ public class Parser {
         return result;
     }
 
-    private FilterExpression parseQuotedContent(QualifierType qualifierName) {
+    private FilterExpression parseQuotedContent(QualifierType type) {
         StringBuilder sb = new StringBuilder();
         while (!isQuoteToken(lookAhead())) {
             sb.append(consume().getValue()).append(" ");
         }
-        return new Qualifier(qualifierName, sb.toString().trim());
+        return new Qualifier(type, sb.toString().trim());
     }
 
     private FilterExpression parseKeyword(Token token) {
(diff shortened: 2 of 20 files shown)


dariusf:
@HansNewbie could you help to review? :v:

HansNewbie:
Sure.

damithc:
Does this cover keywords such as `open` `issue` etc.?

dariusf:
Yup, the qualifier types are now defined [here](https://github.com/HubTurbo/HubTurbo/blob/1152-refactor-qualifier/src/main/java/filter/expression/QualifierType.java#L14-L16). Completion keywords are [defined separately](https://github.com/HubTurbo/HubTurbo/blob/1152-refactor-qualifier/src/main/java/filter/expression/QualifierType.java#L42). Previously they were conflated in this  one collection for all the completions.

Keywords [still use strings](https://github.com/HubTurbo/HubTurbo/blob/1152-refactor-qualifier/src/main/java/filter/expression/Qualifier.java#L682). I'm not sure if we want to change this though, because qualifier content can be any arbitrary string, and the way we parse it depends on the qualifier (for example, `label` is more resilient to typos than `state`). In general this logic is localised to whatever function(s) are used to check particular qualifiers, so it shouldn't create too much duplication.

damithc:
I asked because I want to define aliases for keywords too. e.g. `i:o` to mean `is:open`

dariusf:
I guess we could implement this in an ad hoc (but clean) way for each qualifier. These keyword aliases wouldn't apply in general though: `label:o` would still stand for labels with 'o' in the name.

damithc:
Got it. I'm fine as long as the keyword abbreviation is substitutable for the keyword where the keyword is supposed to be recognized as a keyword.

src/main/java/filter/Parser.java
```diff
@@ -350,7 +344,7 @@ private SortKey parseSortKey() {
         }
     }
 
-    private FilterExpression parseRangeOperator(String name, Token token) {
+    private FilterExpression parseRangeOperator(QualifierType name, Token token) {
```
HansNewbie: Should we rename `name` to `type` to be consistent with other methods?

src/main/java/filter/Parser.java
```diff
@@ -106,26 +106,22 @@ private FilterExpression parseExpression(int precedence) {
...
         consume(TokenType.QUOTE);
         return result;
     }
 
-    private FilterExpression parseKeywords() {
-        return parseKeywords("keyword");
-    }
-
-    private FilterExpression parseKeywords(String qualifierName) {
+    private FilterExpression parseQuotedContent(QualifierType type) {
```
HansNewbie: I am not sure about the name. I feel that `content` means that it (the input) has been parsed, so we do not need to so-called parse content in this scenario - the `parseContent` can be used, for example, in resolving user alias or the `i:o` case in discussion. I can't come up with a better name, though. What do you think?

src/main/java/filter/expression/QualifierType.java
```diff
@@ -0,0 +1,78 @@
...
+     * @return
+     */
+    public static Set<String> getCompletionKeywords() {
+        Set<String> defaultCompletions = new ArrayList<>(Arrays.asList(QualifierType.values())).stream()
+            .map(QualifierType::name)
+            .map(String::toLowerCase)
+            .collect(Collectors.toSet());
+
+        defaultCompletions.addAll(Arrays.asList(
+            "creator", "user", "status", "body", "closed", "open", "issue", "pr", "pullrequest",
```
HansNewbie: I am wondering if we can use a stream of `aliases.keys` here; this would be more manageable when our aliases grow in size. Maybe filter the keys with lengths more than 2 or 3.

I noticed that `desc` is not inside completions. Is that intended?

src/main/java/filter/Parser.java
dariusf: Do you mean changing 'content' to something else?

This method is used in two contexts: `a:"|b c"` and `"|b c"`, where `|` denotes the position we've parsed until. `type` will be 'a' in the first case and 'keywords' in the second, and the loop will consume 'b' and 'c' from the token stream.

In the first case, we are parsing the content of a qualifier within quotes, and in the second,  keywords (search terms) within quotes, hence the names. :stuck_out_tongue: Maybe `parseQuotedQualifierContent`?

src/main/java/filter/Parser.java
dariusf: Yes, good catch.

src/main/java/filter/expression/QualifierType.java
dariusf: Okay, I'll use aliases for the common ones, and `remove` those which shouldn't appear. Filtering by length is a heuristic and works here, but it may not for other cases. Probably better to make exceptions explicitly.

'desc' is a prefix of 'description' (like 'is' and 'issue'), so I left it out. I suppose there's no real harm in including them.

src/main/java/filter/Parser.java
HansNewbie: Okay, I think the name cannot be better. Let's leave it as it is.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -214,16 +192,16 @@ public static void setCurrentTime(LocalDateTime dateTime) {
     }
 
     public boolean isEmptyQualifier() {
-        return name.isEmpty() && content.isPresent() && content.get().isEmpty();
+        return type == QualifierType.EMPTY;
     }
 
     public boolean isFalseQualifier() {
-        return name.equalsIgnoreCase("false") && content.isPresent() && content.get().isEmpty();
+        return type == QualifierType.FALSE;
```
HansNewbie: Just for consideration: would it be better to compare this and empty qualifier with `Qualifier.FALSE` and `Qualifier.EMPTY` instead? So if there is a bug where someone made uses an `QualifierType.EMPTY` or `QualifierType.FALSE` with content, they won't pass these two.

src/test/java/tests/QualifierTypeTests.java
```diff
@@ -0,0 +1,40 @@
...
+
+        // Some qualifiers are deliberately excluded
+        assertFalse(keywords.contains("is"));
+
+        // Other non-qualifier keywords are included
+        assertTrue(keywords.contains("repo"));
+    }
+
+    @Test
+    public void testParse() {
```
HansNewbie: Can we have some alias parse test as well? To ensure that `resolveAlias()` is called.

HansNewbie:
Aside from my latest comments, do you think a doc update is required? I look at the docs and seems that there is nothing to be changed. We might want to have a new part on qualifier alias resolution, though.

src/main/java/filter/expression/Qualifier.java
dariusf: The reason for this change is that since we use an enum now, we can't use a catch-all case to figure out the type of a qualifier. Previously we would return false [here](https://github.com/HubTurbo/HubTurbo/blob/1152-refactor-qualifier/src/main/java/filter/expression/Qualifier.java#L248) if we got an unrecognised string, but now we have to assign the qualifier a type upon parsing. I thought it would make most sense to assign unrecognised qualifiers `false`.

src/test/java/tests/QualifierTypeTests.java
dariusf: Okay, added.

dariusf:
Aliases were never documented before, so there are some missing ones. I've added them.

Additional stuff:

- The completion list is sorted, since now it isn't ordered manually.
- 'false' and 'empty' are prevented from parsing, so users can't access these via input.

src/main/java/filter/expression/Qualifier.java
HansNewbie: Oh. What I meant was instead of using `type == QualifierType.FALSE` and `type == QualifierType.EMPTY`, we use `this == Qualifier.FALSE` and `this == Qualifier.EMPTY`. In case someone accidentally add a bug where the type is wrongly set as `FALSE` or `EMPTY` but it has content.

src/main/java/filter/Parser.java
```diff
@@ -176,44 +172,42 @@ private FilterExpression parseQualifier(Token token) {
...
         qualifierName = qualifierName.substring(0, qualifierName.length() - 1).trim();
 
-        return parseQualifierContent(qualifierName, false);
+        QualifierType type = QualifierType.parse(qualifierName).orElse(QualifierType.FALSE);
+        return parseQualifierContent(type, false);
     }
 
-    private FilterExpression parseQualifierContent(String qualifierName, boolean allowMultipleKeywords) {
-        if (isSortQualifier(qualifierName)) {
+    private FilterExpression parseQualifierContent(QualifierType qualifier, boolean allowMultipleKeywords) {
```
HansNewbie: Just for clarity: can we use `type` or `qualifierType` here instead of `qualifier` for variable name? Especially that we have a `Qualifier` class, this might confuse people as to its type.

src/main/java/filter/expression/QualifierType.java
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Returns a list of keywords which may be completed when a prefix is typed in
+     * the filter field. Defaults to the string forms of all qualifier types. Exceptions
+     * may be added by modifying the collection before it is returned, for example, to
+     * include additional keywords or remove conflicting ones.
+     * @return a set of keywords to be used for completion
+     */
+    public static Set<String> getCompletionKeywords() {
+        Set<String> defaultCompletions =
```
HansNewbie: Shouldn't we remove `empty`, `false` and `keyword` as well here?

src/main/java/filter/expression/Qualifier.java
dariusf: That case does happen now if the qualifier is unrecognised. For example, when parsing `a:1`, we look at the 'a' and don't recognise it, so we return `new Qualifier(QualifierType.FALSE, 1)` (previously we would have "a" as the name and it would enter the default switch case, but now we need to use one of the enum values). That results in a false qualifier with content.

I'm actually not sure if we still need `Qualifier.EMPTY`. Previously we weren't able to distinguish the empty qualifier and a qualifier with name "empty", so we had to use an empty content field as well. Now we can, because user input cannot result in `QualifierType.EMPTY`. Maybe it would be better to replace `q == Qualifier.EMPTY` with `q.isEmptyQualifier()` instead?

src/main/java/filter/Parser.java
dariusf: Sure, will change.

src/main/java/filter/expression/QualifierType.java
dariusf: Good catch.

docs/filters.md
```diff
@@ -182,31 +184,41 @@ Aliases: `au` or `creator`
...
 
-Matches all issues possessing the given attribute.
+- `label` can be written as `labels`
+- `milestone` can be written as `milestone` or `m`
+- `assignee` can be written as `assignee` or `as`
 
 ### no
 
-*Expects one of `label`, `milestone`, `assignee`
+*Expects one of `label`, `milestone`, `assignee`*
```
damithc: The trailing `*` is intended?

src/main/java/filter/expression/Qualifier.java
HansNewbie: Ok I understand the first part.

On `Qualifier.EMPTY`, I think we can keep it. In many cases, we are returning `Qualifier` and not `QualifierType` - e.g. `map` function on `Negation`, `Conjunction` and `Disjunction`. If we remove `Qualifier.EMPTY`, then we need to construct it every time manually, which may introduce more bugs.

HansNewbie: If you decide to remove `Qualifier.EMPTY`, any solution to those methods returning `FilterExpression` or `Qualifier`?

dariusf: Okay, yeah, having to having to construct them manually would result in more boilerplate.

I'm still considering changing `== EMPTY` to `isEmptyQualifier`. Though we should use `Qualifier.EMPTY` whenever possible, `new Qualifier(QualifierType.EMPTY, "")` is an equally valid empty qualifier, and `==` would not allow it. Just seems a bit less brittle?

HansNewbie: Okay, sure then. We can do that.

dariusf:
@HansNewbie made the changes, you can check again.

HansNewbie:
Good to go.

dariusf:
Thanks for reviewing so thoroughly!

PR #1162, opened 2015-12-23
Integrate PMD #751

### LINKED ISSUE #751: Try findbugs on HT

dariusf:
Fixes #751.

This integrates PMD, a static analysis tool for source code (FindBugs works on bytecode, so they are better at picking up different kinds of problems).

The more widely-applicable rulesets are enabled (basic, empty, unnecessary, braces, strings, design, unusedcode; full set [here](https://pmd.github.io/pmd-5.4.1/pmd-java/rules/index.html)). About 700 violations were fixed. Future work would involve enabling more and fixing the violations that result.

Implementation: there are two custom rulesets (for tests and source code) which pull in the used rulesets and define exceptions. In general exceptions for tests and source code may differ: for example, empty catch blocks are acceptable in tests, but are generally a bad sign in source code.

Usage: there are different ways to suppress warnings for false positives.

- `// NOPMD` line comments. These disable PMD for specific lines.
- `@SuppressWarnings("PMD")` annotations. These disable rules at the class or method level.
- Exceptions in rulesets. This is used to turn off entire rules.

Suppressed warnings should generally come with an explanatory comment if the reasoning isn't obvious (to reviewers).


dariusf:
@ndt93 could you help to review? If the reason for any of the changes is unclear I can add comments or explain.

Also, I'm not sure if that last commit is the right way to fix that test.

Coverage drop is due to the new assert [here](https://github.com/HubTurbo/HubTurbo/blob/751-integrate-pmd/src/main/java/backend/stub/DummyRepo.java#L86), which I don't think we're going to cover.

ndt93:
I've commented on the last commit. All other refactoring improve the code's details overall so they shouldn't be a problem.

ndt93:
All right, no more comments from me. Should be good to merged if all unstable tests have passed too.

src/main/java/filter/lexer/Lexer.java
```diff
@@ -27,6 +27,7 @@
             new Rule("\\)", TokenType.RBRACKET),
             new Rule("\\\"", TokenType.QUOTE),
             new Rule(",", TokenType.COMMA),
+        new Rule("%", TokenType.PERCENT),
```
HansNewbie: Indentation :stuck_out_tongue:

HansNewbie:
Any need to update documentation on PMD suppression?

dariusf:
There was never much documentation before, but I've added the details in the first post to a new file.

damithc:
I think some tools such as PMD are missing from the dev doc section called 'Tool stack'

dariusf:
They are documented [here](https://github.com/HubTurbo/HubTurbo/blob/master/docs/staticAnalysis.md) now. Maybe we could put a link to that page in the Tool Stack section?

damithc:
In fact may be we should rejig the dev docs a little. Let me open another issue for it.

PR #1163, opened 2015-12-28
Grayed out title is hard to read when the issue is selected #1080

### LINKED ISSUE #1080: Grayed out title is hard to read when the issue is selected
When the title is grayed out (i.e. marked as read), it becomes hard to read when the issue is selected (blue background).
Steps to reproduce:
1. Select an issue in pView
2. Mark it as a 'read' (i.e. press `E`)
Here's an example:
image

gaieepo:
Fixes #1080.
Use different font weight to distinguish read/unread.
In this case, bold font is used.

Unread (not selected):
untitled

Unread (selected):
untitled

Read (not selected):
untitled

Read (selected):
untitled

### DIFF
--- a/src/main/resources/ui/hubturbo.css
+++ b/src/main/resources/ui/hubturbo.css
@@ -50,11 +50,12 @@
 }
 
 .issue-panel-name {
+    -fx-font-weight: bold;
     -fx-font-size: 14px;
 }
 
 .issue-panel-name-read {
-    -fx-text-fill: #808080;
+    -fx-font-weight: normal;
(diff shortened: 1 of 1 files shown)


dariusf:
The failures here are strange. I've restarted the build. Is this branching off the latest master?

By the way, there's no need to have the issue number in the commit message.

gaieepo:
Yes, this is from the latest master.
I will be careful on my phrasing later on.

dariusf:
Thanks!

PR #1169, opened 2016-01-02
Integrate MockServer #1134

### LINKED ISSUE #1134: Integrate MockServer
http://www.mock-server.com/

MockServer can be used to mock external services that HubTurbo need to interact with via HTTP and HTTPS. This is used to improve back-end testing for which there is currently no tests for interactions with *GitHub APIs*.

ndt93:
Fixes #1134 

For *unit tests*, `MockServer` is useful for components that interact closely with GitHub API. Other higher level components are preferably tested with`mockito` although `MockServer` might be used if it's more convenient (since we can just dump the request/response texts to be emulated) and clear which component is being tested. `MockServer` would be particularly useful for *system/GUI tests* since it can potentially replace all our remaining `Dummy*`stubs.

Sample usage of `MockServer` in this PR are in: `GitHubClientExTests`, `MilestoneUpdateServiceTests`, `PageHeaderIteratorTests`, `PullRequestServiceExTests`.


ndt93:
@dariusf Can help review this? Coverage decreases due to codes unrelated to this PR.

src/test/java/tests/PageHeaderIteratorTests.java
```diff
@@ -0,0 +1,131 @@
...
+     * @throws NoSuchElementException
+     */
+    @Test
+    public void testHeaderIterator() throws NoSuchElementException, IOException {
+        MockServerClient mockServer = ClientAndServer.startClientAndServer(8888);
+
+        HttpRequest page1Request = createMockServerPagedHeaderRequest("1");
+        String page1Etag = "aaf65fc6b10d5afbdc9cd0aa6e6ada4c";
+        String page1Link =
+             "<https://localhost:8888/repositories/20919534/pulls?state=all&per_page=100&page=2>; rel=\"next\", "
```
dariusf: Could these raw strings be factored out somewhere? We can make some kind of URL builder that abstracts away the differences between `repositories/20919534` and `repos/hubturbo/hubturbo`, the need to prefix `api/v3`, etc. At least we can have some constants which tests can share.

docs/testing.md
```diff
@@ -83,4 +83,10 @@ Unit tests are meant to extensively test the functionality of a HubTurbo compone
...
+- [mockito](http://mockito.org/) is used in HubTurbo's tests suite to create clean and verifiable stubbed classes.
+A common use case is to isolate a small component under test while mocking other required components without having to write complex stubbing logic and mixing between application and testing code. Refer to usage of *mockito* in our tests suite and [mockito's docs](http://mockito.github.io/mockito/docs/current/org/mockito/Mockito.html) for more information, guides and caveats when using *mockito*.
+- [MockServer](http://www.mock-server.com) is used in HubTurbo' tests suite to emulate and verify interactions with GitHub API services.
+Note that when passing in `hostname` other than `api.github.com` (usually `"localhost"` when MockServer is used) to `egit`'s `GitHubClient` constructor, subsequent request paths will be appended to the `/api/v3` prefix.
+Also note that if you receive a 404 for a valid request path and parameters, a possible cause is that the MockServer process may take while to build the expectations when a large amount of request or response data is involved.
```
dariusf: Does this make the tests unreliable? If so, maybe we need some kind of synchronisation, or should poll for success?

dariusf:
Nice work, very readable.

The possibility of doing away with the `Dummy*` stubs is appealing, but it seems like it would require a lot of different sample responses, since the stubs are configured to react in very specific ways to certain things, and some tests rely on a combination of these behaviours... Quite a few tests would probably have to be rewritten.

src/test/java/tests/PageHeaderIteratorTests.java
ndt93: Yup, we can abstract this here. It's only for the tests in this particular case though since the `repositories/*` part varies for each set of pagination results. Most of the time, we will use separate files to store responses/headers from GitHub API so we shouldn't have to do this too often.

docs/testing.md
ndt93: This only happens if we only get the headers from multiple requests (HEAD requests), which take very short time, while the body is still building up. If we request for the body (GET, POST etc.) then it will wait until the body is complete before we get a response (so no problems). So this can be completely avoided if we don't put body data into requests that we only care about headers. Otherwise, just make a full request then it will block until everything is available.

ndt93:
Yup. That will take considerable time to rewrite so we can keep them there since they are not really affecting the application codes much at the moment. We can move away from them in future tests.

ndt93:
@dariusf I've made some changes. You can check again.

docs/testing.md
dariusf: I see. Could we document this somewhere? Like under a section for possible causes of the 404 and what to do.

src/test/java/tests/TestUtils.java
```diff
@@ -77,4 +83,51 @@ public static void delay(double seconds) {
...
+    /**
+     * Creates a GitHub API v3 request that can be used with MockServer and GitHubClient
+     * @param method the HTTP request method
+     * @param page the page of the resource being requested
+     * @param repoStringId string version of the repo id
+     * @param repoNumericId numeric version of the repo id returned in the Link header
+     * @param apiSegments segments of the request path after the repo id
+     * @return
+     */
+    public static HttpRequest createMockServerRequest(String method, String page,
```
dariusf: This is great, more high-level than I thought it would be.

Would it be better for the page and numeric id here to be ints, and only later converted to strings? This way we leverage the type system a bit more, people can't pass in random strings or null, etc.

ndt93: GitHub API doesn't guarantee that pages are labeled with numeric values although we can see that they are. It's the same case with numeric id, we can observe from the urls that they look like a number but it's not documented that they must be or how large the number can be.

I think we can change the page to `int`, but the numeric id may better be left as `String`.

dariusf: Ah. I think we've had this conversation before. Okay, that sounds good.

ndt93:
@dariusf I've made the changes and updated docs.

dariusf:
Okay, looks good.

PR #1170, opened 2016-01-02
Count qualifier to view n issues in the list view #1140

### LINKED ISSUE #1140: Count qualifier to view n issues in the list view
For example, there should be a way to get the 10 issues that was most recently updated.

codemanmeet:
Fixes #1140.

### DIFF
--- a/src/main/java/backend/UpdateController.java
+++ b/src/main/java/backend/UpdateController.java
@@ -4,6 +4,7 @@ import backend.resource.MultiModel;
 import backend.resource.TurboIssue;
 import filter.expression.FilterExpression;
 import filter.expression.Qualifier;
+import org.apache.http.ParseException;
 import org.apache.logging.log4j.Logger;
 
 import filter.expression.QualifierType;
@@ -110,7 +111,7 @@ public class UpdateController {
                 List<TurboIssue> filteredSortedAndCountedIssues = allModelIssues.stream()
                         .filter(issue -> Qualifier.process(models, filterExprNoAlias, issue))
                         .sorted(determineComparator(filterExprNoAlias, hasUpdatedQualifier))
-                        .limit(determineCount(allModelIssues, filterExprNoAlias))
+                        .limit(Qualifier.determineCount(allModelIssues, filterExprNoAlias))
                         .collect(Collectors.toList());
 
                 filteredSortedAndCounted.put(filterExpr, filteredSortedAndCountedIssues);
@@ -123,7 +124,7 @@ public class UpdateController {
     /**
      * Produces a suitable comparator based on the given filter expression.
      *
-     * @param filterExpr The given filter expression.
+     * @param filterExpr          The given filter expression.
      * @param hasUpdatedQualifier Determines the behaviour of the sort key "nonSelfUpdate".
      * @return The comparator to use.
      */
@@ -144,21 +145,5 @@ public class UpdateController {
 
         // No sort or updated, return sort by descending ID, which is the default.
         return Qualifier.getSortComparator(models, "id", true, false);
-    }
-
-    /**
-     * Determines the count value to be taken from the first valid count qualifier.
-     *
-     * @param issueList Issue list obtained after filtering and sorting.
-     * @param filterExpr The filter expression of the particular panel.
-     * @return The count value in the qualifier
-     */
-    private int determineCount(List<TurboIssue> issueList, FilterExpression filterExpr){
-        for (Qualifier metaQualifier : filterExpr.find(Qualifier::isMetaQualifier)) {
-            if (metaQualifier.getType() == QualifierType.COUNT && metaQualifier.getNumber().isPresent()) {
-                return metaQualifier.getNumber().get();
-            }
-        }
-        return issueList.size();
     }
 }
--- a/src/main/java/backend/UpdateController.java
+++ b/src/main/java/backend/UpdateController.java
@@ -91,7 +91,7 @@ public class UpdateController {
      * Filters and sorts issues within the model according to the given filter expressions.
      *
      * @param filterExprs Filter expressions to process.
-     * @return Filter expressions and their corresponding issues after filtering and sorting.
+     * @return Filter expressions and their corresponding issues after filtering, sorting and counting.
      */
     private Map<FilterExpression, List<TurboIssue>> processFilter(List<FilterExpression> filterExprs) {
         MultiModel models = logic.getModels();
@@ -147,7 +147,7 @@ public class UpdateController {
     }
 
     /**
-     * Applies the count qualifier to the set of issues obtained after the sorting order is applied.
+     * Determines the count value to be taken from the first valid count qualifier.
      *
      * @param issueList Issue list obtained after filtering and sorting.
      * @param filterExpr The filter expression of the particular panel.
--- a/src/main/java/backend/UpdateController.java
+++ b/src/main/java/backend/UpdateController.java
@@ -50,10 +50,10 @@ public class UpdateController {
                             + results.size() + " repos"))
                     .thenCompose(n -> logic.getRateLimitResetTime())
                     .thenApply(logic::updateRemainingRate)
-                    .thenRun(() -> logic.updateUI(filterAndSort(filterExprs))); // Then filter the second time.
+                    .thenRun(() -> logic.updateUI(processFilter(filterExprs))); // Then filter the second time.
         } else {
             // If no issues requiring metadata update, just run the filter and sort.
-            logic.updateUI(filterAndSort(filterExprs));
+            logic.updateUI(processFilter(filterExprs));
         }
     }
 
@@ -93,30 +93,31 @@ public class UpdateController {
      * @param filterExprs Filter expressions to process.
      * @return Filter expressions and their corresponding issues after filtering and sorting.
      */
-    private Map<FilterExpression, List<TurboIssue>> filterAndSort(List<FilterExpression> filterExprs) {
+    private Map<FilterExpression, List<TurboIssue>> processFilter(List<FilterExpression> filterExprs) {
         MultiModel models = logic.getModels();
         List<TurboIssue> allModelIssues = models.getIssues();
 
-        Map<FilterExpression, List<TurboIssue>> filteredAndSorted = new HashMap<>();
+        Map<FilterExpression, List<TurboIssue>> filteredSortedAndCounted = new HashMap<>();
 
         filterExprs.forEach(filterExpr -> {
-            List<TurboIssue> filterAndSortedExpression = filteredAndSorted.get(filterExpr);
+            List<TurboIssue> filterAndSortedExpression = filteredSortedAndCounted.get(filterExpr);
 
             if (filterAndSortedExpression == null) { // If it already exists, no need to filter anymore
                 boolean hasUpdatedQualifier = Qualifier.hasUpdatedQualifier(filterExpr);
 
                 FilterExpression filterExprNoAlias = Qualifier.replaceMilestoneAliases(models, filterExpr);
 
-                List<TurboIssue> filteredAndSortedIssues = allModelIssues.stream()
+                List<TurboIssue> filteredSortedAndCountedIssues = allModelIssues.stream()
                         .filter(issue -> Qualifier.process(models, filterExprNoAlias, issue))
                         .sorted(determineComparator(filterExprNoAlias, hasUpdatedQualifier))
+                        .limit(determineCount(allModelIssues, filterExprNoAlias))
                         .collect(Collectors.toList());
 
-                filteredAndSorted.put(filterExpr, filteredAndSortedIssues);
+                filteredSortedAndCounted.put(filterExpr, filteredSortedAndCountedIssues);
             }
         });
 
-        return filteredAndSorted;
+        return filteredSortedAndCounted;
     }
 
     /**
@@ -143,5 +144,21 @@ public class UpdateController {
 
         // No sort or updated, return sort by descending ID, which is the default.
         return Qualifier.getSortComparator(models, "id", true, false);
+    }
+
+    /**
+     * Applies the count qualifier to the set of issues obtained after the sorting order is applied.
+     *
+     * @param issueList Issue list obtained after filtering and sorting.
+     * @param filterExpr The filter expression of the particular panel.
+     * @return The count value in the qualifier
+     */
+    private int determineCount(List<TurboIssue> issueList, FilterExpression filterExpr){
+        for (Qualifier metaQualifier : filterExpr.find(Qualifier::isMetaQualifier)) {
+            if (metaQualifier.getType() == QualifierType.COUNT && metaQualifier.getNumber().isPresent()) {
+                return metaQualifier.getNumber().get();
+            }
+        }
+        return issueList.size();
     }
 }
--- a/src/main/java/filter/expression/Qualifier.java
+++ b/src/main/java/filter/expression/Qualifier.java
@@ -243,8 +243,6 @@ public class Qualifier implements FilterExpression {
             return satisfiesUpdatedHours(issue);
         case REPO:
             return satisfiesRepo(issue);
-        case COUNT:
-                return satisfiesCount();
         default:
             assert false : "Missing case for " + type;
             return false;
@@ -408,6 +406,7 @@ public class Qualifier implements FilterExpression {
         switch (q.getType()) {
             case IN:
             case SORT:
+            case COUNT:
                 return true;
             default:
                 return false;
@@ -419,6 +418,7 @@ public class Qualifier implements FilterExpression {
         case SORT:
         case IN:
         case REPO:
+        case COUNT:
             return true;
         default:
             return isUpdatedQualifier(q);
@@ -437,15 +437,6 @@ public class Qualifier implements FilterExpression {
     public static boolean isUpdatedQualifier(Qualifier q) {
         switch (q.getType()) {
             case UPDATED:
-                return true;
-            default:
-                return false;
-        }
-    }
-
-    public static boolean isCountQualifier(Qualifier q) {
-        switch (q.getType()) {
-            case COUNT:
                 return true;
             default:
                 return false;
@@ -647,10 +638,6 @@ public class Qualifier implements FilterExpression {
         LocalDateTime dateOfUpdate = issue.getUpdatedAt();
         int hoursSinceUpdate = Utility.safeLongToInt(dateOfUpdate.until(getCurrentTime(), ChronoUnit.HOURS));
         return updatedRange.encloses(hoursSinceUpdate);
-    }
-
-    private boolean satisfiesCount(){
-        return number.isPresent() && number.get() >= 0;
     }
 
     private boolean satisfiesRepo(TurboIssue issue) {
--- a/src/main/java/filter/expression/Qualifier.java
+++ b/src/main/java/filter/expression/Qualifier.java
@@ -11,6 +11,7 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import backend.resource.*;
+import filter.ParseException;
 import util.Utility;
 import backend.interfaces.IModel;
 import filter.MetaQualifierInfo;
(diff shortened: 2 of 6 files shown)


damithc:
`sort` should be applied before `count` if both are present, am I right?.

codemanmeet:
Yup, sort takes precedence that way! `count` is applied to the final issue list that is to be displayed in the list view.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -243,6 +243,8 @@ public boolean isSatisfiedBy(IModel model, TurboIssue issue, MetaQualifierInfo i
             return satisfiesUpdatedHours(issue);
         case REPO:
             return satisfiesRepo(issue);
+        case COUNT:
+                return satisfiesCount();
```
dariusf: `count` is a meta-qualifier (like `in` and `sort`), so it should be [stripped](https://github.com/HubTurbo/HubTurbo/blob/master/src/main/java/filter/expression/Qualifier.java#L405) by the time we get here. This method shouldn't have to be changed.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -301,6 +303,22 @@ private ContextMenu updateContextMenu(ContextMenu contextMenu) {
         return contextMenu;
     }
 
+    private ObservableList<TurboIssue> applyCountQualifierToIssueList(ObservableList<TurboIssue> issueList,
```
dariusf: I'm not sure this is the right place to put this -- we want to try and keep logic out of the UI. By the time the issues get to the UI to be shown, all the effects of filters should already have been applied. You might want to look into how `sort` is implemented, and do something similar for this.

src/main/java/filter/expression/Qualifier.java
codemanmeet: Sure, will move it to the meta qualifier list and add it to the `stripped` method!

src/main/java/ui/listpanel/ListPanel.java
codemanmeet: Yup, looked at `sort`. I have added two methods in `UpdateController` that do the job of applying the count filter. `sort` is given preference over `count`.

codemanmeet:
@dariusf help me review it again when you are free!

src/main/java/backend/Logic.java
```diff
@@ -218,7 +218,7 @@ public void replaceIssueLabelsUI(TurboIssue issue, List<String> labels) {
      * and then sends the data to the GUI.
      */
     private void refreshUI() {
-        updateController.filterSortRefresh(getAllUIFilters());
+        updateController.filterSortCountRefresh(getAllUIFilters());
```
dariusf: Hmm, I don't think we need to change this method name -- `count` is kind of like a filter. It can't keep growing as we add more features.

src/main/java/backend/UpdateController.java
```diff
@@ -50,10 +51,10 @@ public void filterSortRefresh(List<FilterExpression> filterExprs) {
                             + results.size() + " repos"))
                     .thenCompose(n -> logic.getRateLimitResetTime())
                     .thenApply(logic::updateRemainingRate)
-                    .thenRun(() -> logic.updateUI(filterAndSort(filterExprs))); // Then filter the second time.
+                    .thenRun(() -> logic.updateUI(filterSortAndCount(filterExprs))); // Then filter the second time.
```
dariusf: Maybe use a more general name, like `processFilter`? `withFilter`?

src/main/java/backend/UpdateController.java
```diff
@@ -112,11 +113,12 @@ private void openRepositoriesInFilters(List<FilterExpression> filterExprs) {
                         .sorted(determineComparator(filterExprNoAlias, hasUpdatedQualifier))
                         .collect(Collectors.toList());
 
-                filteredAndSorted.put(filterExpr, filteredAndSortedIssues);
+                filteredSortedAndCounted.put(filterExpr,
```
dariusf: Instead of doing this, you can add `.limit(n)` after `.sorted(...)`. That will take care of truncating the stream so only n or fewer elements remain after the entire thing is collected.

src/main/java/backend/UpdateController.java
```diff
@@ -144,4 +146,40 @@ private void openRepositoriesInFilters(List<FilterExpression> filterExprs) {
...
     }
+
+    /**
+     * Applies the count qualifier to the set of issues obtained after the sorting order is applied.
+     *
+     * @param issueList Issue list obtained after filtering and sorting.
+     * @param expression The filter expression of the particular panel.
+     * @return The modified issue list with the count applied.
+     */
+    private List<TurboIssue> applyCountQualifierToIssueList(List<TurboIssue> issueList, FilterExpression expression){
```
dariusf: These two functions can be combined into one that just returns the number of elements to `limit` the stream to.

Also, instead of throwing an exception, maybe just use the first one, like `sort` does?

src/main/java/filter/expression/Qualifier.java
```diff
@@ -441,6 +443,15 @@ public static boolean isUpdatedQualifier(Qualifier q) {
         }
     }
 
+    public static boolean isCountQualifier(Qualifier q) {
+        switch (q.getType()) {
```
dariusf: This can just be `q.getType() == QualifierType.COUNT`. The use of switch is left over from when the type was a string and we had to account for aliases.

src/main/java/backend/UpdateController.java
codemanmeet: Sure will change to one function and will use the first one

src/main/java/filter/expression/Qualifier.java
codemanmeet: Yup, will change it

src/main/java/backend/UpdateController.java
codemanmeet: Using `limit` now!

src/main/java/backend/UpdateController.java
codemanmeet: changed it to `processFilter`.

src/main/java/backend/UpdateController.java
```diff
@@ -144,4 +145,22 @@ private void openRepositoriesInFilters(List<FilterExpression> filterExprs) {
...
+     * Applies the count qualifier to the set of issues obtained after the sorting order is applied.
+     *
+     * @param issueList Issue list obtained after filtering and sorting.
+     * @param filterExpr The filter expression of the particular panel.
+     * @return The maximum count of issues
+     */
+    private int determineMaxCount(List<TurboIssue> issueList, FilterExpression filterExpr){
+        for (Qualifier metaQualifier : filterExpr.find(Qualifier::isMetaQualifier)) {
+            if (metaQualifier.getType() == QualifierType.COUNT) {
+                int countQualifierNum = metaQualifier.getNumber().get();
```
dariusf: This would fail for an input like `count:a`, because then `getNumber()` would be empty. You should probably check for the presence of `getNumber()` in the condition. Unfortunately we don't have a way to statically rule out invalid inputs yet.

src/main/java/backend/UpdateController.java
```diff
@@ -144,4 +145,22 @@ private void openRepositoriesInFilters(List<FilterExpression> filterExprs) {
...
+                return countQualifierNum < issueList.size() && countQualifierNum >= 0 ?
```
dariusf: `limit` also handles this comparison for us. You can actually just return `countQualifierNum` here. The 'max' can be removed from the method name to reflect that.

codemanmeet:
Let me know if it can be merged!

src/main/java/backend/UpdateController.java
```diff
@@ -144,4 +145,20 @@ private void openRepositoriesInFilters(List<FilterExpression> filterExprs) {
...
+     * Applies the count qualifier to the set of issues obtained after the sorting order is applied.
+     *
+     * @param issueList Issue list obtained after filtering and sorting.
+     * @param filterExpr The filter expression of the particular panel.
+     * @return The count value in the qualifier
+     */
+    private int determineCount(List<TurboIssue> issueList, FilterExpression filterExpr){
+        for (Qualifier metaQualifier : filterExpr.find(Qualifier::isMetaQualifier)) {
+            if (metaQualifier.getType() == QualifierType.COUNT) {
+                return metaQualifier.getNumber().isPresent() ? metaQualifier.getNumber().get() : issueList.size();
```
dariusf: What I meant was to have the `isPresent` check in the if condition, so if the user input is

```
count:a count:1 count:2
```
it shows one issue, i.e. it takes the first *valid* qualifier, not the first one regardless of validity.

Right now it would see `count:a` then default to the size of the list.

You might want to have a test that has multiple qualifiers also, to demonstrate that this is what is intended.

dariusf:
Sorry, this was quite drawn out. Last round.

codemanmeet:
Oh, I didn't know I had to choose the first *valid* qualifier. I was choosing the first qualifier itself. I'll change that.

damithc:
Long drawn out reviews are a sign of high expectations in code quality. I'm happy to see those. :-p

codemanmeet:
Haha true true :-p

dariusf:
Okay, this looks good now. Thanks for making all the changes. Could you update to the latest master before merging?

src/test/java/guitests/FilterTests.java
```diff
@@ -64,6 +64,91 @@ public void milestoneAliasFilterTest() {
...
+
+        assertEquals(2, issuePanel.getIssueCount());
+
+        // Checking 7 issues shown for count:7
+        click("#dummy/dummy_col0_filterTextField");
+        selectAll();
+        type("count:7");
+        push(KeyCode.ENTER);
+
+        assertEquals(7, issuePanel.getIssueCount());
```
damithc: Why 2 and 7? Are they boundary values?

src/test/java/guitests/FilterTests.java
```diff
@@ -64,6 +64,91 @@ public void milestoneAliasFilterTest() {
...
+
+        //first count qualifier chosen
+        click("#dummy/dummy_col0_filterTextField");
+        selectAll();
+        type("count:6 count:9");
+        push(KeyCode.ENTER);
+
+        assertEquals(6, issuePanel.getIssueCount());
+
+        // Not-a-number returns the whole list
```
damithc: 'not a number' should return nothing instead?

src/test/java/guitests/FilterTests.java
```diff
@@ -64,6 +64,91 @@ public void milestoneAliasFilterTest() {
...
+        type("count:abcd");
+        push(KeyCode.ENTER);
+
+        assertEquals(10, issuePanel.getIssueCount());
+
+        // Return the value from the first valid count qualifier
+
+        click("#dummy/dummy_col0_filterTextField");
+        selectAll();
+        type("count:abc count:8 count:2");
```
damithc: There is an implicit AND between clauses, so should return the smallest number because count is interpreted as 'at most' ? i.e. 'at most 2 AND at most 5' should equal to 'at most 2' right?

src/test/java/guitests/FilterTests.java
```diff
@@ -64,6 +64,91 @@ public void milestoneAliasFilterTest() {
...
+        assertEquals(8, issuePanel.getIssueCount());
+        assertEquals("Issue 10", issuePanel.getIssueList().get(0).getTitle());
+        assertEquals("Issue 9", issuePanel.getIssueList().get(1).getTitle());
+
+        // Test with sort qualifier
+
+        click("#dummy/dummy_col0_filterTextField");
+        selectAll();
+        type("count:8 count:2 sort:unmerged");
+        push(KeyCode.ENTER);
```
damithc: Why 2 count clauses and sort are tested together?

src/test/java/guitests/FilterTests.java
```diff
@@ -64,6 +64,91 @@ public void milestoneAliasFilterTest() {
...
+
+        assertEquals(8, issuePanel.getIssueCount());
+        assertEquals("Issue 1", issuePanel.getIssueList().get(0).getTitle());
+        assertEquals("Issue 2", issuePanel.getIssueList().get(1).getTitle());
+    }
+
```
damithc: @dariusf can filters be tested without involving GUI?

src/main/java/backend/UpdateController.java
```diff
@@ -144,4 +145,20 @@ private void openRepositoriesInFilters(List<FilterExpression> filterExprs) {
...
     }
+
+    /**
+     * Applies the count qualifier to the set of issues obtained after the sorting order is applied.
+     *
+     * @param issueList Issue list obtained after filtering and sorting.
+     * @param filterExpr The filter expression of the particular panel.
+     * @return The count value in the qualifier
+     */
+    private int determineCount(List<TurboIssue> issueList, FilterExpression filterExpr){
```
damithc: The comment says 'Applies' and the method name says 'determines'. Which is it?

src/main/java/backend/UpdateController.java
```diff
@@ -93,30 +93,31 @@ private void openRepositoriesInFilters(List<FilterExpression> filterExprs) {
      * @param filterExprs Filter expressions to process.
      * @return Filter expressions and their corresponding issues after filtering and sorting.
      */
```
damithc: This header comment needs updating?

src/main/java/backend/UpdateController.java
```diff
@@ -93,30 +93,31 @@ private void openRepositoriesInFilters(List<FilterExpression> filterExprs) {
...
 
             if (filterAndSortedExpression == null) { // If it already exists, no need to filter anymore
                 boolean hasUpdatedQualifier = Qualifier.hasUpdatedQualifier(filterExpr);
 
                 FilterExpression filterExprNoAlias = Qualifier.replaceMilestoneAliases(models, filterExpr);
 
-                List<TurboIssue> filteredAndSortedIssues = allModelIssues.stream()
+                List<TurboIssue> filteredSortedAndCountedIssues = allModelIssues.stream()
                         .filter(issue -> Qualifier.process(models, filterExprNoAlias, issue))
                         .sorted(determineComparator(filterExprNoAlias, hasUpdatedQualifier))
```
damithc: not related to this issue, but shouldn't this be `sort`, not `sorted`?

src/main/java/backend/UpdateController.java
```diff
@@ -144,4 +145,20 @@ private void openRepositoriesInFilters(List<FilterExpression> filterExprs) {
         // No sort or updated, return sort by descending ID, which is the default.
         return Qualifier.getSortComparator(models, "id", true, false);
     }
+
+    /**
+     * Applies the count qualifier to the set of issues obtained after the sorting order is applied.
```
damithc: This method has nothing to do with sorting. It should not be mentioned in the header comment. If this was a proper description of the method, I should see something like ' ... else returns the size of the issueList'

src/test/java/guitests/FilterTests.java
dariusf: Yes for the true/false ones, but no for `sort` and `count`. This will require some refactoring. Should we do it now?

src/test/java/guitests/FilterTests.java
dariusf: Okay, this may be a better way to interpret it. Right now we're just doing the simplest thing and taking the first (leftmost, if we think of the expression as a tree) valid qualifier. Same for `sort` if there are multiple qualifiers.

src/test/java/guitests/FilterTests.java
damithc: I see. May be as a separate issue then. At the same time need to move some of the testing done in gui to lower level unit tests. Otherwise, future devs follow existing code and dump all testing into GUI tests.

src/main/java/backend/UpdateController.java
dariusf: That's a method from the new streams API. I think the convention is that verbs work on one element at a time, but adjectives work on the stream as a whole.

src/test/java/guitests/FilterTests.java
damithc: I see. First, we need to document it as users might expect a different behavior based on their intuition. Second, let's create issues to make the behavior more intuitive or make the limitation more explicit e.g. flag multiple sorts/counts as syntax errors.

src/main/java/backend/UpdateController.java
damithc: Got it.

src/test/java/guitests/FilterTests.java
codemanmeet: Nope, 0 and 10 are boundary values. Both have been tested. In addition, 2 and 7 are being tested as well.

src/test/java/guitests/FilterTests.java
codemanmeet: They are being tested to check for sort's precedence over count.

src/test/java/guitests/FilterTests.java
codemanmeet: For all qualifiers. `updated`, `sort` etc. if you enter the wrong data type (`string` instead of `number`), the entire list is returned.

src/main/java/backend/UpdateController.java
codemanmeet: Yup, will update the header comment.

src/main/java/backend/UpdateController.java
codemanmeet: It is `determines`. Will change it to that.

src/test/java/guitests/FilterTests.java
damithc: Should we reverse that? Otherwise it is misleading. @dariusf what do you think?

src/test/java/guitests/FilterTests.java
damithc: Having 2 count clauses is not necessary for that right? Instead, may be have test cases where sort comes before count in the filter, and vice versa.

src/test/java/guitests/FilterTests.java
damithc: Just one non boundary case is enough? If 2 is broken, 7 has to be broken as well right?

damithc: Did you test for a negative number?

codemanmeet: I added a test for negative numbers as well.

src/test/java/guitests/FilterTests.java
codemanmeet: Sorry, my bad. For `updated`, if you enter the wrong data type (string), empty list is returned. If all count qualifiers entered are invalid (for eg.), shall I return an empty list?

src/main/java/backend/UpdateController.java
```diff
@@ -144,4 +145,20 @@ private void openRepositoriesInFilters(List<FilterExpression> filterExprs) {
...
+     * @param filterExpr The filter expression of the particular panel.
+     * @return The count value in the qualifier
+     */
+    private int determineCount(List<TurboIssue> issueList, FilterExpression filterExpr){
+        for (Qualifier metaQualifier : filterExpr.find(Qualifier::isMetaQualifier)) {
+            if (metaQualifier.getType() == QualifierType.COUNT && metaQualifier.getNumber().isPresent()) {
+                return metaQualifier.getNumber().get();
+            }
+        }
+        return issueList.size();
```
damithc: This line doesn't match with the header comment.

codemanmeet: Changed it.

codemanmeet:
@damithc As discussed on Slack, I'll go ahead and implement `Parse Exception` for any invalid `count` qualifier found.

dariusf:
Note that on a parse exception, we show the entire list.

So in general, should we always throw exceptions on invalid inputs? A lot of qualifiers just default to empty (return `true` on everything) if there is something unexpected.

Also, any changes to docs needed for this?

codemanmeet:
Yup, that's how we want it. `Parse Exception` returns `Qualifier.EMPTY` and shows the entire list. The exception informs the user of the error at the bottom status bar of the pView. He/she should still see the issues and not be taken aback by an empty list. The user can then type in a fresh filter and then rectify the error.

damithc:
I'm inclined to show an error upon syntax errors. If we do show the full list on such cases, there should be a visual indication that the list is show due to an error, not as the filter result. Otherwise, we can show an empty list, but again we should indicate the empty list is due to the error, not because nothing matches the filter.
The alternative is to ignore incorrect part of the syntax. Is there any benefit of doing that? The downside is the user might be mislead.

dariusf:
Yes, I guess there's not much point in interpreting it wrongly.

codemanmeet:
@dariusf `Parse exception` has been added for `count` with relevant error messages. Commits have been squashed. Thanks for the review and creating the pertinent issues! :) As discussed, will go ahead and merge once the CI passes.

PR #1171, opened 2016-01-05
Add a 'new' option for the 'Boards' menu #1142

### LINKED ISSUE #1142: Add a 'new' option for the 'Boards' menu
This option is for creating an empty board. We can prompt for a board name before creating it.

zhongwei-z:
Add an option for new empty board. Current panels will be closed and the empty board will be saved. Fixes #1142 .

The codes could be refactored as the method has many overlaps with the existing `onBoardSaveAs()`. However, that method forbids a save of an empty board.

### DIFF
--- a/src/main/java/ui/MenuControl.java
+++ b/src/main/java/ui/MenuControl.java
@@ -87,7 +87,45 @@ public class MenuControl extends MenuBar {
         return file;
 
     }
-    
+
+    /**
+     * Creates an empty board. User will be prompted to
+     * confirm the action if there are unclosed panels.
+     */
+    private void onBoardNew() {
+        logger.info("Menu: Boards > New");
+
+        if (!isNewBoardCreationDialogConfirmed()) {
+            logger.info("New board creation cancelled");
+            return;
+        }
+
+        panels.closeAllPanels();
+        onBoardSaveAs();
+    }
+
+    /**
+     * Returns a boolean indicating whether we can proceed with
+     * the new board creation. If there are no panels open, it
+     * returns true; else a dialog is shown for the user to confirm.
+     */
+    private boolean isNewBoardCreationDialogConfirmed() {
+        List<PanelInfo> panelList = panels.getCurrentPanelInfos();
+        if (panelList.isEmpty()) {
+            return true;
+        }
+
+        Alert dlg = new Alert(AlertType.CONFIRMATION, "");
+        dlg.initModality(Modality.APPLICATION_MODAL);
+        dlg.setTitle("Confirmation");
+        dlg.getDialogPane().setHeaderText("Create a new board");
+        dlg.getDialogPane().setContentText("Are you sure you want to create a new board?" +
+                " All unsaved changes to the current board will be lost.");
+        Optional<ButtonType> response = dlg.showAndWait();
+        return response.isPresent() &&
+                response.get().getButtonData() == ButtonData.OK_DONE;
+    }
+
     private void onBoardSave() {
         logger.info("Menu: Boards > Save");
         
@@ -114,17 +152,11 @@ public class MenuControl extends MenuBar {
         logger.info("Menu: Boards > Save as");
 
         List<PanelInfo> panelList = panels.getCurrentPanelInfos();
-
-        if (panelList.isEmpty()) {
-            logger.info("Did not save new board");
-            return;
-        }
-
         BoardNameDialog dlg = new BoardNameDialog(prefs, mainStage);
         Optional<String> response = dlg.showAndWait();
         ui.showMainStage();
         panels.selectFirstPanel();
-        
+
         if (response.isPresent()) {
             String boardName = response.get().trim();
             prefs.addBoard(boardName, panelList);
@@ -179,6 +211,9 @@ public class MenuControl extends MenuBar {
     }
 
     private MenuItem[] createBoardsMenu() {
+        MenuItem newBoard = new MenuItem("New");
+        newBoard.setOnAction(e -> onBoardNew());
+
         MenuItem saveAs = new MenuItem("Save as");
         saveAs.setOnAction(e -> onBoardSaveAs());
         
@@ -211,7 +246,7 @@ public class MenuControl extends MenuBar {
 
         Menu autoCreate = boardAutoCreator.generateBoardAutoCreateMenu();
 
-        return new MenuItem[] {save, saveAs, open, delete, autoCreate};
+        return new MenuItem[] {newBoard, save, saveAs, open, delete, autoCreate};
     }
     
     public void switchBoard() {
--- a/src/test/java/unstable/MenuControlTest.java
+++ b/src/test/java/unstable/MenuControlTest.java
@@ -61,7 +61,7 @@ public class MenuControlTest extends UITest {
         // Testing board save when no board is open because nothing has been saved
         // Expected: prompts user to save as new board
         click("Boards");
-        push(KeyCode.DOWN).push(KeyCode.ENTER);
+        push(KeyCode.DOWN).push(KeyCode.DOWN).push(KeyCode.ENTER);
         type("Board 1");
         push(KeyCode.ESCAPE);
         PlatformEx.waitOnFxThread();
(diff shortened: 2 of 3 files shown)


dariusf:
Maybe make the New action clear all opened panels, but not save? The user can save manually after adding some panels and supply a name at that point. We can reuse the existing Save path and sidestep the empty board issue.

Being able to save an empty board makes sense to me, but can't remember if there was a technical reason for preventing it. @HyungJon you might be interested in this.

damithc:
Creating new board is not going to be frequent. But we should avoid users accidentally wiping out an existing set of panels that hasn't been saved. So, we should ask the user to confirm that it is ok to clear the current panels. That question need not be asked if there are unsaved changed to the panels, but that is an optimization we can do as a separate issue.

damithc:
And I too think we should allow saving empty boards.

dariusf:
@zhongwei-z How's this going? Do you need help?

zhongwei-z:
Sorry for the lack of updates! I was swamped by other duties and totally forgot about this! >.<

I added an alert box warning to user if there are unclosed panels when they try to create a new board.

src/main/java/ui/MenuControl.java
```diff
@@ -115,15 +136,14 @@ private void onBoardSaveAs() {
...
-            logger.info("Did not save new board");
-            return;
-        }
-
         BoardNameDialog dlg = new BoardNameDialog(prefs, mainStage);
         Optional<String> response = dlg.showAndWait();
         ui.showMainStage();
-        panels.selectFirstPanel();
+        if (panelList.isEmpty()) {
+            logger.info("Save an empty board");
```
damithc: `Save an ...` --> `Saved an ...`

src/main/java/ui/MenuControl.java
```diff
@@ -87,7 +87,28 @@ private Menu createFileMenu() {
         return file;
 
     }
-    
+
+    private void onBoardNewBoard() {
```
dariusf: To be consistent with the naming convention, maybe this could be `onBoardNew`?

src/main/java/ui/MenuControl.java
```diff
@@ -87,7 +87,28 @@ private Menu createFileMenu() {
...
+        logger.info("Menu: Boards > New");
+
+        List<PanelInfo> panelList = panels.getCurrentPanelInfos();
+        if (!panelList.isEmpty()) {
+            Alert dlg = new Alert(AlertType.CONFIRMATION, "");
+            dlg.initModality(Modality.APPLICATION_MODAL);
+            dlg.setTitle("Confirmation");
+            dlg.getDialogPane().setHeaderText("Create a new board");
```
dariusf: It'd be nice for SLAP if all this could be factored out into a method, maybe named `confirmNewBoardCreation`. It could return a boolean value which tells us whether we can safely close everything and save, etc.

zhongwei-z:
Ok I have refactored those :)

dariusf:
Code looks good, but it needs to pass CI as well (there's a PMD violation). The coverage drop is okay because that test is unstable.

It would be nice to have some kind of dirty flag to allow creating new boards in more scenarios, and simplify that huge stateful test, but those can be separate issues.

dariusf:
@damithc ready for approval.

src/main/java/ui/MenuControl.java
```diff
@@ -87,7 +87,35 @@ private Menu createFileMenu() {
...
+
+        if (confirmNewBoardCreation()) {
+            panels.closeAllPanels();
+            onBoardSaveAs();
+        } else {
+            logger.info("New board creation cancelled");
+        }
+    }
+
+    private boolean confirmNewBoardCreation() {
```
damithc: A better name that captures both boolean return and the use of dialog: `isNewBoardCreationDialogConfirmed`

src/main/java/ui/MenuControl.java
```diff
@@ -87,7 +87,35 @@ private Menu createFileMenu() {
...
+
+    private void onBoardNew() {
+        logger.info("Menu: Boards > New");
+
+        if (confirmNewBoardCreation()) {
+            panels.closeAllPanels();
+            onBoardSaveAs();
+        } else {
+            logger.info("New board creation cancelled");
+        }
```
damithc: Use guard conditions to deal with the low hanging fruit first. It reduces the complexity of the execution paths.
```java
if (!confirmNewBoardCreation()) {
    logger.info("New board creation cancelled");
    return;
}

panels.closeAllPanels();
onBoardSaveAs();
```

src/main/java/ui/MenuControl.java
```diff
@@ -87,7 +87,35 @@ private Menu createFileMenu() {
         return file;
 
     }
-    
+
+    private void onBoardNew() {
```
damithc: Write a header comment to explain what the method does.

src/main/java/ui/MenuControl.java
```diff
@@ -87,7 +87,35 @@ private Menu createFileMenu() {
...
+            onBoardSaveAs();
+        } else {
+            logger.info("New board creation cancelled");
+        }
+    }
+
+    private boolean confirmNewBoardCreation() {
+        List<PanelInfo> panelList = panels.getCurrentPanelInfos();
+        if (panelList.isEmpty()) {
+            return true;
```
damithc: Can use guard conditions here too (see my comment in the previous method).

src/main/java/ui/MenuControl.java
damithc: Write a header comment for the method. Remember to mention the behavior if there are no panels.

src/main/java/ui/MenuControl.java
```diff
@@ -115,15 +143,14 @@ private void onBoardSaveAs() {
...
-        }
-
         BoardNameDialog dlg = new BoardNameDialog(prefs, mainStage);
         Optional<String> response = dlg.showAndWait();
         ui.showMainStage();
-        panels.selectFirstPanel();
+        if (panelList.isEmpty()) {
+            logger.info("Saved an empty board");
+        } else {
+            panels.selectFirstPanel();
```
damithc: What's the purpose of this line? Is this the right place to have it. The SLAP of this method can be improved.

src/main/java/ui/MenuControl.java
```diff
@@ -114,27 +152,31 @@ private void onBoardSaveAs() {
...
+        prefs.addBoard(boardName, panelList);
+        prefs.setLastOpenBoard(boardName);
+        ui.triggerEvent(new BoardSavedEvent());
+        logger.info("New board " + boardName + " saved");
+    }
+
+    private void showFirstPanel(List<PanelInfo> panelList) {
+        ui.showMainStage();
+        if (!panelList.isEmpty()) {
+            panels.selectFirstPanel();
```
zhongwei-z: This line is to show up the first panel. It was added in [this commit](https://github.com/HubTurbo/HubTurbo/commit/ca91dbb84d91d0e735423cc560476643f08b544c#diff-c522a964929f2e9b705fba9068c9cf9dR212).

src/main/java/ui/MenuControl.java
```diff
@@ -87,7 +87,45 @@ private Menu createFileMenu() {
...
+
+    /**
+     * Returns a boolean indicating whether we can proceed with
+     * the new board creation. If there are no panels open, it
+     * returns true; else a dialog is shown for the user to confirm.
+     */
+    private boolean isNewBoardCreationDialogConfirmed() {
+        List<PanelInfo> panelList = panels.getCurrentPanelInfos();
+        if (!panelList.isEmpty()) {
+            Alert dlg = new Alert(AlertType.CONFIRMATION, "");
```
dariusf: I think @damithc meant this instead:

```java
if (panelList.isEmpty()) {
    return true;
}
// everything else
```
Makes the logic a bit simpler to read.

src/main/java/ui/MenuControl.java
dariusf: `panelList` originates from `panels`, right? It would be good if this check were done by a `panels` method instead.

dariusf:
You can comment when this is ready in future.

Anyway I think this is okay except for the tests, but those are being dealt with already.

src/test/java/unstable/MenuControlTest.java
```diff
@@ -150,14 +159,31 @@ public void menuControlTest() {
...
         PlatformEx.waitOnFxThread();
         assertEquals(2, panelControl.getNumberOfSavedBoards());
-        
+
+        // Testing board create new and confirm
+        click("Boards");
+        push(KeyCode.DOWN).push(KeyCode.ENTER);
+        click("OK");
+        assertEquals(0, panelControl.getPanelCount());
+        ((TextField) find("#boardnameinput")).setText("empty");
```
damithc: @dariusf can this be abstracted into a utility method somewhere (as a separate issue)? e.g. `setTextField(fieldName, text)`

dariusf: Yes, sounds good.

damithc: Great. Pls create an issue for it.

PR #1177, opened 2016-01-10
Space key doesn't work when editing filters #1176

### LINKED ISSUE #1167: F1 key doesn't work when the focus is in a filter box
F1 should work even when the focus is in a filter box. Even better if pressing F1 while in the filter box goes directly to the 'filters' section of the user docs.

### LINKED ISSUE #1176: Space key doesn't work when editing filters
When editing a filter, the space key doesn't work when the cursor is at certain locations of the filter.

dariusf:
Fixes #1176.

### DIFF
--- a/src/main/java/ui/components/FilterTextField.java
+++ b/src/main/java/ui/components/FilterTextField.java
@@ -68,9 +68,6 @@ public class FilterTextField extends TextField {
                 if (shouldTriggerCompletion) {
                     performCompletion(e);
                 }
-            } else if (typed == ' ' && (getText().isEmpty() || getText().endsWith(" "))) {
-                // Prevent consecutive spaces
-                e.consume();
             }
         });
         setOnKeyPressed(e -> {
--- a/src/main/java/ui/components/FilterTextField.java
+++ b/src/main/java/ui/components/FilterTextField.java
@@ -16,13 +16,21 @@ import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+/**
+ * A field augmented with the ability to revert edits, autocompletion, on-the-fly error
+ * checking, and callbacks for various actions.
+ */
 public class FilterTextField extends TextField {
+
+    // A character class that indicates the characters considered to be word boundaries.
+    // Specialised to HubTurbo's filter syntax.
+    private static final String WORD_BOUNDARY_REGEX = "[ (:)]";
 
     // Callback functions
     private Runnable cancel = () -> {};
     private Function<String, String> confirm = (s) -> s;
 
-    // For reversion of edits
+    // For reverting edits
     private String previousText;
 
     // The list of keywords which will be used in completion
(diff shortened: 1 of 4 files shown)


dariusf:
@HansNewbie could you help to review? Thanks!

HansNewbie:
For the space testing, can we have some boundary tests as well? Maybe test adding space at start, end, and middle of a word that has no space or start and end with space? For example,
`count` -> `_count`
`count` -> `count_`
`count` -> `cou_nt`
`_count` -> `__count` -> `__count_` -> `__count__`

dariusf:
Added the comment.

Added the last one -- insertion of spaces when the field contains trailing spaces. That's significant because of the legacy code removed. The rest are all in the same class of boundary case and should be covered already, since spaces are not treated specially now.

src/test/java/guitests/FilterTextFieldTest.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @Test
+    public void inputsTest() {
+        FilterTextField field = getFirstPanelField();
+        testCompletions(field);
+        testSpaces(field);
+    }
+
+//    @Test
```
HansNewbie: Is this meant to be commented? The `cancelTest` as well.

dariusf: No, good thing you caught it. :sweat:

src/main/java/ui/components/FilterTextField.java
```diff
@@ -87,22 +88,39 @@ private void setup() {
...
-        String word = getCurrentWord() + e.getCharacter();
+    /**
+     * Completion is only started when there's effectively nothing (only whitespace)
+     * after the caret, or if there is selected text (indicating either that we are in
+     * the midst of completion, or that the user does not want the selected text and is
+     * typing to replace it)
+     * @return true if completion should be started
+     */
+    private boolean shouldStartCompletion() {
+        return getCharAfterCaret().trim().isEmpty() || !getSelectedText().isEmpty();
```
HansNewbie: I need confirmation on this behaviour. What if the user select text in the middle of the filter? Should completion be performed? Say `assignee:me milestone:current` and the selected text is `ignee` of `assignee`.

dariusf: With `ignee` selected, if you type `i`, the result is `assi[gnee]:me` (selection represented with square brackets). Typing anything else (say, `a`) results in `assa:me`. Completion is performed, but only if the current word + currently-typed letter is a prefix of some keyword.

The main reason for this change was to [not have completion cancelled as the user continues typing](https://github.com/HubTurbo/HubTurbo/blob/1167-space-key/src/test/java/guitests/FilterTextFieldTest.java#L22-L32). The other scenario of typing with some text already selected is a side effect.

HansNewbie: Yep, I noticed about the completion toggling. Since this is an intended behaviour, sure. This is fine.

src/main/java/filter/expression/QualifierType.java
```diff
@@ -50,7 +50,7 @@
         defaultCompletions.removeAll(Arrays.asList("m", "as", "au", "s"));
 
         defaultCompletions.addAll(Arrays.asList("closed", "open", "issue", "pr",
-            "pullrequest", "read", "unread",  "merged", "unmerged", "comments",
+            "pullrequest", "read", "unread", "me", "merged", "unmerged", "comments",
```
HansNewbie: Does the alias `me` for `assignee` exist? I was using it in my previous comment for simplicity.

dariusf: Ugh, just checked and it seems to have been removed at some point. It used to use your login name. Never mind then.

HansNewbie:
Ok seems good.

PR #1178, opened 2016-01-10
Pass GUI elements from Logic to GUI #1131

### LINKED ISSUE #1131: Pass GUI elements from Logic to GUI
Instead of just passing TurboIssues, GUIElements, which encapsulate TurboIssues together with some relevant data from the Model to display the issue cards.

The goal is to remove GUI components' dependency on the Model to look up information necessary for displaying issues, thus allowing them to operate "on their own" without bugging out when the MultiModel in the backend experiences consistency problems.

One of two issues proposed to solve #1089

acruis:
Fixes #1131

With this change, instead of GUI nodes all containing references to the one `MultiModel`, they now contain references to smaller objects such as `TurboLabel`s and `TurboMilestone`s via their respective `GUIElement`s. This should allow garbage collection to clean up unnecessary objects while still allowing the GUI to display objects from deleted repos during race conditions without bugging out.


src/main/java/ui/GUIElement.java
```diff
@@ -0,0 +1,61 @@
...
+import backend.resource.TurboIssue;
+import backend.resource.TurboLabel;
+import backend.resource.TurboMilestone;
+import backend.resource.TurboUser;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+public class GUIElement {
```
damithc: Our convention prefers `GuiElement`

acruis: Alright, I'll fix that.

src/main/java/ui/GuiElement.java
```diff
@@ -0,0 +1,61 @@
...
+    private final List<TurboLabel> labels;
+    private final Optional<TurboMilestone> milestone;
+    private final Optional<TurboUser> assignee;
+
+    public GuiElement(TurboIssue issue,
+                      List<TurboLabel> labels,
+                      Optional<TurboMilestone> milestone,
+                      Optional<TurboUser> assignee) {
+        this.issue = issue;
+        this.labels = labels;
```
dariusf: Maybe this could be given a more specific name, to indicate that it includes labels referenced in events.

src/main/java/ui/GuiElement.java
```diff
@@ -0,0 +1,61 @@
...
+        return issue;
+    }
+
+    /**
+     * Returns all labels related to an issue (including labels that have been added and then removed).
+     * Use getLabelsOfIssue to retrieve an issue's current labels instead.
+     *
+     * @return Labels related to an issue, or null if the GuiElement has been constructed as such.
+     */
+    public List<TurboLabel> getLabels() {
```
dariusf: Doesn't seem like this getter is needed anywhere.

src/main/java/ui/listpanel/ListPanelCell.java
```diff
@@ -67,11 +56,14 @@ private void updateStyleToMatchStatus(TurboIssue issue){
...
+        return guiElement.getIssue().getLabels();
     }
 
     public TurboIssue getIssue() {
-        return issue;
+        return guiElement.getIssue();
     }
 
+    public GuiElement getGuiElement() {
+        return guiElement; }
```
dariusf: Formatting is a bit weird here.

src/main/java/util/events/ModelUpdatedEvent.java
```diff
@@ -1,20 +1,19 @@
...
+import ui.GuiElement;
 
 import java.util.List;
 import java.util.Map;
 
 public class ModelUpdatedEvent extends Event {
-    public final IModel model;
-    public final Map<FilterExpression, List<TurboIssue>> issuesToShow;
+    public final Map<FilterExpression, List<GuiElement>> elementsToShow;
+    public final List<TurboUser> users;
```
dariusf: Maybe we should have the comments on what these are for here instead, where the fields are, rather than at the places where this object is constructed. That would put them more in context and also not have them repeated.

src/main/java/backend/Logic.java
```diff
@@ -274,8 +272,8 @@ private boolean insertMetadata(Map<Integer, IssueMetadata> metadata, String repo
     /**
      * Carries the current model in Logic, as well as issues to be displayed in panels, to the GUI.
      */
-    public void updateUI(Map<FilterExpression, List<TurboIssue>> issuesToShow) {
-        uiManager.update(models, issuesToShow);
+    public void updateUI(Map<FilterExpression, List<GuiElement>> elementsToShow) {
```
dariusf: Outdated comment.

src/main/java/backend/UpdateController.java
```diff
@@ -146,4 +152,46 @@ private void openRepositoriesInFilters(List<FilterExpression> filterExprs) {
...
+     * @param model The model containing the relevant issue.
+     * @param issue The issue, the labels of which are to be retrieved.
+     * @return A list of labels needed for displaying the issue properly.
+     */
+    private List<TurboLabel> produceRelevantLabels(Model model, TurboIssue issue) {
+        Stream<TurboLabel> removedLabels = issue.getMetadata().getEvents().stream()
+                .filter(issueEvent -> issueEvent.getType() == IssueEventType.Unlabeled)
+                .map(unlabeledEvent -> {
+                    Optional<TurboLabel> label = model.getLabelByActualName(unlabeledEvent.getLabelName());
+                    assert label.isPresent() : issue + " Label \"" + unlabeledEvent.getLabelName() + "\" not found.";
```
dariusf: Hmm, just to confirm -- it's not possible that we trigger this assertion because of an old label being deleted or renamed on GitHub, right? If metadata is showing, it must be up to date, so we would never have an old label in the events?

src/main/java/ui/GuiElement.java
acruis: I was actually thinking of splitting the current labels and the removed labels lists into two lists, since we do need the list of current labels to display as `issueDetails`. What do you think?

src/main/java/ui/listpanel/ListPanelCell.java
acruis: Lol, thanks for the catch!

src/main/java/backend/UpdateController.java
acruis: Renaming shouldn't be a problem, but deletion is.

I tried to create a new label, assign it to an issue, and then deleted it. This results in the label being automatically removed from the issue. This changes the ETag for labels, and so HT will replace the current list of labels in the model with the updated list, which no longer contains the deleted label. However, the TurboIssueEvents for adding and deleting the label still stays.

Thus, this assertion will trigger. Let's say we remove the assertion and ignore the disappearance of the labels, then in `TurboIssueEvent.display` under the `Unlabeled` case, there is a check if the label is present in the GuiElement, and if no, it will display a new label with the given label name. However, the color will be the default label color (`ffffff`), and this is what happens when `master` is run:

image
versus:

image

The TurboIssueEvent should already contain the label's actual color which is retrievable with the `getLabelColor` method, so we can remove the assert and instead contruct the deleted label from the data in the TurboIssueEvent instead.

src/main/java/util/events/ModelUpdatedEvent.java
acruis: Alright, will do that.

src/main/java/backend/UpdateController.java
acruis: I think I can make the changes to `TurboIssueEvent.display` on this PR, would that be OK?

dariusf: Yup, sure. So with these changes, a GuiElement stores the TurboIssue, its TurboLabels (not including those referenced in events), and events? And the problem of labels not being found is avoided by using the info from events to display them.

acruis: Yup. The structure of the GuiElement itself won't change as the events are already being stored in the TurboIssue's metadata.

I'm not sure if we need to include removed events under the `labels` field of GuiElement. Instead, the field can hold only the current labels of the issue, and we just use `TurboIssueEvent.getLabelColor` to display unlabeling events.

dariusf: Yup, that's what I meant. Sounds good.

src/main/java/ui/GuiElement.java
acruis: Now that we've removed labels referenced in events, I think we can keep the name as is?

dariusf: Yup, okay.

acruis:
Ready for review again!

src/test/java/tests/TurboIssueEventTests.java
```diff
@@ -17,11 +17,11 @@
 
     private static TurboIssueEvent createLabelUpdateEvent(
             String userName, IssueEventType eventType,
-            GregorianCalendar time, String labelName) {
+            GregorianCalendar time, String labelName, String labelColour) {
```
dariusf: Is GregorianCalendar necessary? Could we use LocalDateTime (timezone-independent)?

dariusf:
Everything else looks good. :+1: Probably should squash the less significant commits before merging.

src/test/java/tests/TurboIssueEventTests.java
acruis: I don't think GregorianCalendar makes much of a difference; we just need the Date object, anyway.

dariusf:
Didn't know you had updated this already. I restarted the CI a few times and it's passed. Anything left to squash? If not, you could rebase this and merge when it passes after that.

acruis:
Yup, I've been pretty busy these days so I couldn't touch this PR, even though it was just squashing and rebasing. It'll be done by this week's meeting.

damithc:
See if you can finish by this week's release (tonight).

acruis:
Rebasing seems to be acting up due to there being too many merges in between commits, but merging to master on my local branch looks fine. Is it ok if I go ahead and merge?

dariusf:
Okay, no point in redoing all the merges.

PR #1179, opened 2016-01-12
Ensure MultiModel correctness #1132

### LINKED ISSUE #1132: Ensure MultiModel correctness
Via the use of a logical queue, which simply keeps track of all async operations. When an async operation is to be carried out (in its own worker thread), it checks against the logical queue to see if it can be carried out immediately, or whether it should wait until some prior tasks should finish first.

One of two issues proposed to solve #1089

dariusf:
Fixes #1132.

### DIFF
--- a/src/test/java/tests/RepoOpControlTest.java
+++ b/src/test/java/tests/RepoOpControlTest.java
@@ -1,0 +1,159 @@
+package tests;
+
+import static junit.framework.TestCase.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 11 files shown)


src/test/java/tests/RepoOpControlTest.java
```diff
@@ -0,0 +1,159 @@
...
+    private static final Logger logger = LogManager.getLogger(RepoOpControlTest.class.getName());
+
+    private static final String REPO = "test/test";
+
+    private final Executor executor = Executors.newCachedThreadPool();
+
+    @Test
+    public void opsWithinMultipleRepos() throws ExecutionException, InterruptedException {
+
+        // Operations on the same repository cannot execute concurrently
```
acruis: This should be about operations on multiple repos.

acruis:
Great PR, really not much to say here.

acruis:
Good to merge.

Issue #1146, opened 2015-12-17
RepositorySelectorTest always fails


ndt93:
This is due to missing UI's resource for unstable tests.

ndt93:
Closing as this shouldn't happen for everyone and can be solved manually.

@nicholaslum444 Try `gradle clean` or copy the resource directory from `guitests` to `unstable` if your `RepositorySelectorTest` is still failing.

PR #1186, opened 2016-01-14
Set up GUI test resources for unstable tests in `check` or `test` tasks #1185

### LINKED ISSUE #1185: GUI test resources not found for unstable tests when running `gradlew clean check`
Encountered `NullPointerException` for `repositorySelectorTest` when running `gradlew clean check`. 

It seems that running the `check` task is different from running the `unstableTests` task, as it does not setup the test resources for the unstable tests. (As in, copy the `resources/guitests` to `resources/unstable`). 

Creating a soft link from `guitest` to `unstable` does the trick. But I'm not sure whether it works on Windows.

xinan:
Fixes #1185, #1146 

This is done by adding a dependency to the `test` task.
This will fix the failure of `repositorySelectorTest` during `check` or `test` task.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -128,6 +128,8 @@ allprojects {
     assemble.dependsOn shadowJar
 }
 
+test.dependsOn generateUnstableTestResources
+
 task checkstyleHtml << {
     ant.xslt(in: checkstyleMain.reports.xml.destination,
              style: file('config/checkstyle/checkstyle-noframes-sorted.xsl'),


build.gradle
```diff
@@ -128,6 +128,9 @@ allprojects {
     assemble.dependsOn shadowJar
 }
 
+test.dependsOn generateUnstableTestResources
+check.dependsOn generateUnstableTestResources
```
dariusf: Is the second line needed? `test` is a dependency of `check`, so `generateUnstableTestResources` should transitively be a dependency also.

dariusf:
Nice. Is the `dependsOn` in `generateUnstableTestResources` still needed?

@ndt93 does this agree with the fix you had in mind?

ndt93:
That should work if the tests are run with `gradle`. There is still a problem if we want to run the tests through IDEs though. But symlink won't work with Windows and I don't think pushing a separate copy of the resources for unstable to the repo is a good idea. So anyone who wants to run through the IDE will need to make a copy/symlink in their local copy.

@xinan Looks good. Can you check the tests again after rebasing with the current master? Thanks.

xinan:
@dariusf Yes you are right! :+1: The second line is unnecessary. However the line is still needed in the `task unstableTests` though, because it is a different task from `test`, nor does it depends on `test`.

@ndt93 Yes I works fine with the latest master :smile:

ndt93:
Okay :+1: . Should be good to be merged then.

PR #1187, opened 2016-01-14
Added guide to run a single test using gradlew #1189

### LINKED ISSUE #1189: Add guide to run a single test using gradlew

jinified:
Fixes #1189

### DIFF
--- a/docs/workflow.md
+++ b/docs/workflow.md
@@ -25,6 +25,12 @@ This will run only the stable tests. To run the unstable tests separately,
 ./gradlew clean unstableTests
 ```
 
+To run only a single test i.e. PanelFocusTest of the unstable tests.
+
+```sh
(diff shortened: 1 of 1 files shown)


dariusf:
Thanks! Couple of minor things:

- PRs are supposed to fix issues, so there needs to be an accompanying issue for this. I know it's a bit trivial in this case, but could you create that issue and link it to this PR? Fixes #x, title, etc. Just so you're familiar with the whole process.
- If you're comfortable with rebasing, it'd be good to do that for updating to the latest master, especially if it's not a difficult merge and doesn't need to be preserved. Helps keep history a bit cleaner.

jinified:
Fixes #1189

dariusf:
The 'fixes' string has to be in the description of the PR for the issue to close automatically. You have to edit and insert it in there.

The title should also have the issue number at the back.

dariusf:
Thanks!

PR #1193, opened 2016-01-17
Adding labels: deleting last word removes two labels #1129

### LINKED ISSUE #787: Backspacing the last space should undo the application of the last label
If the user type space and immediately realizes the wrong label got applied, pressing backspace should send the label picker to the state it was in just before typing space. Currently the user has to type the wrong label again and then type the right label.

### LINKED ISSUE #1129: Adding labels: deleting last word removes two labels
e.g.
1. Go to label picker (assume there are no labels in that issue)
2. Type `med enh perf` which results in picking three labels, as expected.
3. Backspace `perf`
Expected: the label `performance` is removed from the selected labels. 
Actual: the label 'performance' is removed from the selected labels and `enhancement` is stricken off.

m133225:
Fixes #1129 and fixes #787.

- Changes large part of the logic of label picker. Most of the initialisation and helper functions remained.
- Since its method toggleSelectedLabel(String) was modified to toggleHighlightedLabel(), the appropriate change was also made in LabelPickerLogicTests.

Definitely, the code is rather messy and I think there still needs to be more refactoring and testing.
This still assumes that the user only types a character or backspaces to undo his/her changes in the label picker's text field. Any other changes (e.g. pasting) are still not handled.

Perhaps it is better to revamp the whole label picker UI?

### DIFF
--- a/src/main/java/ui/components/pickers/LabelPickerUILogic.java
+++ b/src/main/java/ui/components/pickers/LabelPickerUILogic.java
@@ -8,8 +8,27 @@ import java.util.*;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
-/*
- * Handles the logic when typing in the text field of the label picker UI.
+/**
+ * Handles the logic of the label picker UI.
+ *
+ * The crux of the logic is as follows:
+ * 1. When the text in the LabelPicker's text field changes, processTextFieldChange is called.
+ * 2. If the text is not empty, this method will try to determine if the the user typed in a character or backspaced.
+ *   2a. Typing is handled by addCharHandler
+ *     i. If the user typed ' ', it will attempt to add the current highlighted label (if there is).
+ *     ii. Else it is an ongoing query
+ *     Bottom labels that do not fit the query will be faded.
+ *     The first of the unfaded labels will be 'highlighted'.
+ *     The highlighted bottom label will then be displayed as a 'possible label' at the top.
+ *   2b. Backspacing is handled by removeCharHandler
+ *     i. When the current last character is ' ', there is no current query, therefore any 'possible label'
+ *     is removed
+ *     ii. When the removed character is ' ', and the last character is not ' ', the most recently specified label is
+ *     toggled. In addition, there is an existing query that will result in a 'possible label'
+ *
+ * listOfSpecifiedLabels is maintained to keep track of the mapping between what the user has entered in the
+ * text field, and the labels that have been added/removed due to them.
+ * listOfRemovedExclusiveLabels is also maintained so that the removed labels can be added back when needed.
  */
 public class LabelPickerUILogic {
 
@@ -22,10 +41,12 @@ public class LabelPickerUILogic {
     private final Map<String, Boolean> resultList = new HashMap<>();
     private Optional<String> targetLabel = Optional.empty();
 
+    // variables to handle text field changes
     private int previousTextFieldLength = 0;
     private Character previousLastChar = null;
     private final ArrayList<String> listOfSpecifiedLabels = new ArrayList<>();
     private final ArrayList<ArrayList<String>> listOfRemovedExclusiveLabels = new ArrayList<>();
+
 
     LabelPickerUILogic(TurboIssue issue, List<TurboLabel> repoLabels, LabelPickerDialog dialog) {
         this.issue = issue;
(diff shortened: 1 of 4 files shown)


dariusf:
While we want to keep the level of abstraction high in commits, it's helpful to break up large changes, especially if they can be considered separately. For example, you might want to keep input-handling changes and the other small refactoring changes separate.

Also, a lot of work went into the existing UI. :stuck_out_tongue: We should have a thorough discussion about what's wrong with it and what we want to change before considering a revamp.

Maybe you could summarise the changes to behaviour here? Any besides the bugs you fixed?

damithc:
Yup, the PR should do only one thing, which is to fix the issue in concern. If there are unrelated code improvements in this PR, move them to separate PRs.

m133225:
### The Problem
Most of the issues come with trying to backspace where it does not return to the *previous state* before the backspace.
This makes the UI rather unintuitive.

### Pull Request
The PR attempts to rectify the following interactions (as far as I can recall...):
A. Exclusive label interactions
Example: 
1. Using any HubTurbo issue with no labels.
2. Typing 'high' and then 'critical' will remove the `critical.high` label.
Text field: 'high critical '
**3a. Backspacing once does not clear the critical label (#787)**
**3b. Backspacing twice incorrectly strikes out the critical label**
**3c. Backspacing until it reaches the 'c' in 'priority.critical' incorrectly retains the `priority.critical` label**
**3d. Further backspacing will retain the `priority.critical` label instead of `priority.high`**

B. Adding 2 labels then removing both of them
Example:
1. Using any HubTurbo issue with no labels.
2. Typing 'enc med ' will add type.enhancement and priority.medium
**3a. Backspacing 'med ' will incorrectly strike off `type.enhancement`**
*I think this should be under #1129, but the description differs a little. Removing the last word does not remove the last label, but only the second last.*
**3b. Backspacing all the way will incorrectly retain both labels**

C. Removing existing labels
Example:
1. Using any HubTurbo issue with at least 1 label.
2. Try to add the same label again (to remove it)
**3. Backspace all the way until the first character is left, incorrectly strikes off the existing label**
*However, I think this problem is fundamentally the same as problem B, where there is an unexpected behaviour trying to remove the last label.*

### Possible Changes
Even with the PR, I think the Label Picker UI is still prone to unexpected behaviour (when the user pastes stuff into the text field)

I think there are a couple of ways to go about changing this.
1. Of course, we could always clear the text field when a label is added to solve this problem easily, but I do not think that is intended/wanted as it could end up being a hassle to the user.
2. Change the UI to the following:
comment
*Taken from Odoo 8.0*

**Should we combine all of these into 1 issue and deal with them all together?**

src/main/java/ui/components/pickers/LabelPickerUILogic.java
```diff
@@ -60,58 +64,107 @@ private void populatePanes() {
...
+                    if (lastLabel != null) {
+                        updateTopLabels(lastLabel, !resultList.get(lastLabel));
+                    }
 
-                if (!isBottomLabelsUpdated) {
-                    updateBottomLabels(query);
+                    ArrayList<String> listOfRemovedLabels = listOfRemovedExclusiveLabels.get(lastIndex);
+                    listOfRemovedExclusiveLabels.remove(lastIndex);
+                    if (listOfRemovedLabels != null) {
+                        listOfRemovedLabels.stream()
```
lejolly: one line lambda doesn't need the curly braces
```java
listOfRemovedLabels.stream().forEach(labelName -> updateTopLabels(labelName, true));
```

src/main/java/ui/components/pickers/LabelPickerUILogic.java
```diff
@@ -373,6 +466,12 @@ private void ______BOILERPLATE______() {}
                 .collect(Collectors.toList());
     }
 
+    public boolean isExistingLabel(PickerLabel targetLabel) {
```
lejolly: is this being used somewhere?

lejolly:
good work on the fixes, backspacing works much more intuitively now. 

however i think you should split up `processTextFieldChange` into smaller methods, it really is quite large now. more comments that clearly show the different scenarios within `processTextFieldChange` would be great, or you could just name the (new) sub-methods descriptively enough to reflect those scenarios. 

changing the ui would be the scope of another pr, but yes i do like the tagging ui design pattern (and it was considered in the initial version of the label picker, but i guess we just wanted an mvp out first), so by all means do go ahead to see how feasible that would be too. 

lastly, could you include just a paragraph or two of comments that describe the flow of how things work within `LabelPickerUILogic`, i guess for a newcomer it would seem rather disorganised/daunting (even to me 😝), feel free to move methods around if that better reflects the flow of things.

src/main/java/ui/components/pickers/LabelPickerUILogic.java
```diff
@@ -60,58 +64,107 @@ private void populatePanes() {
...
-                    String delimiter = TurboLabel.getDelimiter(query).get();
-                    String[] queryArray = query.split(Pattern.quote(delimiter));
-                    if (queryArray.length == 1) {
-                        isBottomLabelsUpdated = true;
-                        updateBottomLabels(queryArray[0], "");
-                    } else if (queryArray.length == 2) {
-                        isBottomLabelsUpdated = true;
-                        updateBottomLabels(queryArray[0], queryArray[1]);
+        int curTextFieldLength = text.length();
+        if (curTextFieldLength > 0) {
```
damithc: Consider putting the shorter branch first to improve understandability (i.e. pick low hanging fruits first). In this case, `if` is very long but `else` is very short. You can also use early returns to reduce nesting. e.g.
```
if (something){
   do this;
   return;
}

File: src/main/java/ui/components/pickers/LabelPickerUILogic.java [hunk 4]
```diff
@@ -8,6 +8,28 @@
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+/**
+ * Handles the logic of the label picker UI.
+ *
+ * The crux of the logic is as follows:
```
damithc: From this line onward, you are explaining a specific method, and hence should be given as the method comment, not the class comment.

src/main/java/ui/components/pickers/LabelPickerUILogic.java
```diff
@@ -8,6 +8,28 @@
...
+ * 1. When the text in the LabelPicker's text field changes, processTextFieldChange is called.
+ * 2. If the text is not empty, this method will try to determine if the the user typed in a character or backspaced.
+ *   2a. Typing is handled by addCharHandler
+ *     i. If the user typed ' ', it will attempt to add the current highlighted label (if there is).
+ *     ii. Else it is an ongoing query
```
damithc: You may want to explain somewhere what a `query` is. That's the sort of thing a comment should explain because the code itself cannot tell the read what that term means.

src/main/java/ui/components/pickers/LabelPickerUILogic.java
```diff
@@ -8,6 +8,28 @@
...
+ *   2b. Backspacing is handled by removeCharHandler
+ *     i. When the current last character is ' ', there is no current query, therefore any 'possible label'
+ *     is removed
+ *     ii. When the removed character is ' ', and the last character is not ' ', the most recently specified label is
+ *     toggled. In addition, there is an existing query that will result in a 'possible label'
+ *
+ * listOfSpecifiedLabels is maintained to keep track of the mapping between what the user has entered in the
+ * text field, and the labels that have been added/removed due to them.
+ * listOfRemovedExclusiveLabels is also maintained so that the removed labels can be added back when needed.
+ */
```
damithc: Need for extensive comments to explain the algorithm is an indication that the code is not self-explanatory. Consider this code:
```java
if(text.isEmpty()) return;

if(isBackSpace(text){
    handleBackspace(text);
} else if (isSpace(text)){
    handleSpace(text);
} else {
    handleChar(text);
}
```
Now the high level logic is clear and there is no need for comments. Use the same approach for each of the methods called. In other words, apply SLAP. The comment should explain WHAT the method does and WHY it does it, but not HOW it does it (the HOW should be clear from the code itself).

m133225: Ok, understood. Sorry, I was afraid of creating too many small functions which may bloat the class unnecessarily. If that's the case, for the comments that @lejolly has asked to include, what should they contain?

lejolly: I guess you can just highlight the starting method (`processTextFieldChange`), and that method should be clear and high level enough to eliminate the need for the big chunk of comments. 

you can also move parts 2a and 2b to their respective methods and do it in the proper javadoc style?

damithc: Make the code as self explanatory as possible. After that, add comments to mention things that are hard to get from the code. The comment explains WHAT (i.e. what is the purpose of the method) and WHY (when you need to justify something). The code itself should show HOW it is done (i.e. the algorithm).

damithc:
@m133225 When the code is ready to review again, post a comment to let the reviewer know.

src/main/java/ui/components/pickers/LabelPickerUILogic.java
```diff
@@ -8,6 +8,15 @@
...
 import java.util.stream.Collectors;
 
+/**
+ * Handles the logic of the label picker UI.
+ *
+ * This class provides the following key methods:
+ * 1. processTextFieldChange: this should be called after every change in the UI's text field
+ * 2. toggleLabel
+ * 3. toggleHighlightedLabel
+ * 4. moveHighlightOnLabel
```
damithc: Not sure if this listing of methods adds any value. Instead, add good header comments to each key method.

src/main/java/ui/components/pickers/LabelPickerUILogic.java
```diff
@@ -59,58 +71,146 @@ private void populatePanes() {
         if (dialog != null) dialog.populatePanes(getExistingLabels(), getNewTopLabels(), bottomLabels, groups);
     }
 
-    public void toggleLabel(String name) {
-        addRemovePossibleLabel("");
-        preProcessAndUpdateTopLabels(name);
-        updateBottomLabels(""); // clears search query, removes faded-out overlay on bottom labels
+    public ArrayList<String> toggleLabel(String name) {
```
damithc: Add a header comment to this method, and each non-trivial public method.

src/main/java/ui/components/pickers/LabelPickerUILogic.java
```diff
@@ -59,58 +71,146 @@ private void populatePanes() {
...
+        removePossibleLabel();
+        ArrayList<String> removedLabels = processAndUpdateTopLabels(name);
+        restoreBottomLabels(); // clears search query, removes faded-out overlay on bottom labels
```
damithc: See if you can rename `top` and `bottom` to more meaningful names based on what they represent rather than where they are located.

src/main/java/ui/components/pickers/LabelPickerUILogic.java
```diff
@@ -59,58 +71,146 @@ private void populatePanes() {
...
-                        isBottomLabelsUpdated = true;
-                        updateBottomLabels(queryArray[0], "");
-                    } else if (queryArray.length == 2) {
-                        isBottomLabelsUpdated = true;
-                        updateBottomLabels(queryArray[0], queryArray[1]);
-                    }
-                }
+        if (isEmpty(text)) {
+            removePossibleLabel();
+            restoreBottomLabels();
```
damithc: Add comments to explain WHY you are doing this. The code clearly says HOW you perform that action (i.e. 'if there is no text, we remove possible label and restore bottom labels'), but it doesn't tell us why we do it like that.

src/main/java/ui/components/pickers/LabelPickerUILogic.java
```diff
@@ -59,58 +71,146 @@ private void populatePanes() {
...
+
+    private boolean isEmpty(String text) {
+        return text.length() == 0;
+    }
+
+    private boolean isLongerTextFieldThanPrevious(String text) {
+        return text.length() > previousTextFieldLength;
+    }
+
+    private void updateVariables(String text) {
```
damithc: This name is too generic. Update what variables? `updatePreviousCharInfo`?

src/main/java/ui/components/pickers/LabelPickerUILogic.java
```diff
@@ -59,58 +71,146 @@ private void populatePanes() {
...
+        } else {
+            processChangeAndUpdateVariables(text);
+        }
+        populatePanes();
+    }
 
-                if (!isBottomLabelsUpdated) {
-                    updateBottomLabels(query);
-                }
+    private void processChangeAndUpdateVariables(String text) {
```
damithc: Does name have to mention `AndUpdateVariables`? Isn't updating variables an integral part of part of `processChange`?

src/main/java/ui/components/pickers/LabelPickerUILogic.java
```diff
@@ -59,58 +71,146 @@ private void populatePanes() {
...
+            listOfRemovedLabels.stream()
+                    .forEach(labelName -> updateTopLabels(labelName, true));
+        }
+    }
+
+    private boolean isEmpty(String text) {
+        return text.length() == 0;
+    }
+
+    private boolean isLongerTextFieldThanPrevious(String text) {
```
damithc: I think the purpose of this method is to check if a char was added. Checking the length is the mechanism for doing that. The method name should reflect the intent (WHAT it does), not mechanism. You can name it `isCharAdded(String previousText, String newText)`. By feeding both parameters, you remove a hidden dependency between the method and an instance variable.

src/main/java/ui/components/pickers/LabelPickerUILogic.java
```diff
@@ -59,58 +71,146 @@ private void populatePanes() {
...
+        }
+        populatePanes();
+    }
 
-                if (!isBottomLabelsUpdated) {
-                    updateBottomLabels(query);
-                }
+    private void processChangeAndUpdateVariables(String text) {
+        if (isLongerTextFieldThanPrevious(text)) {
+            addCharHandler(text);
```
damithc: Did you mean `handleCharAddition`?. The current name sounds like you are adding a char handler.

src/main/java/ui/components/pickers/LabelPickerUILogic.java
```diff
@@ -59,58 +71,146 @@ private void populatePanes() {
...
+        updateVariables(text);
+    }
+
+    /**
+     * This method will either toggle the highlighted label, or update both the bottom labels and the possible label
+     * in the top pane, depending on the last character of text (added character)
+     */
+    private void addCharHandler(String text) {
+        if (isLastCharSpace(text)) {
+            // space-toggle
```
damithc: What's the use of this comment?

src/main/java/ui/components/pickers/LabelPickerUILogic.java
```diff
@@ -59,58 +71,146 @@ private void populatePanes() {
...
+            spaceToggle();
```
damithc: Name methods as action verbs. e.g. `toggleSpace`

damithc:
@m133225 I've added a few comments. Refactor based on those comments and let us know when ready for review again. I'm using this PR is train you on our coding style, so brace yourself. It will take quite a few revisions before we are done. 
Normally, we don't fix bugs and refactor in the same PR, but let's make an exception here because this is a good opportunity to train you.

src/main/java/ui/components/pickers/LabelPickerUILogic.java
m133225: Hmm... Space-toggle is a verb here though, and I don't think that toggleSpace explain what it does. Would toggleUsingSpace or something along those lines be better?

damithc: What does the method do?

m133225: It toggles the highlighted label if it the last character is a space. Handled differently from a normal toggle (clicking) as label history has to be stored to support backspace.

damithc: When you encounter a space, what is the thing you have to do, at a high level? You need to specify the algorithm in a top-down way, not bottom up. For example, can it be like this?  
``` java
if space //user is done specifying the next label
    update label selection;
else //user is not yet done specifying the next label
    update label suggestion;
```

m133225:
Ready for review!

damithc:
I'll discuss this at tomorrow session.

lejolly:
@damithc is this on hold because of #1215?

damithc:
Yes. I'll add the label now.

damithc:
Time to close this PR? @m133225 I assume the bug is fixed in the new label picker?

m133225:
@damithc I suppose issues #1129 and #787 can be closed too?

PR #1195, opened 2016-01-19
Make sure most qualifiers throw Exception for invalid input #1174

### LINKED ISSUE #1174: Unexpected inputs should trigger parse exceptions instead of failing silently
All qualifiers should be changed to throw exceptions on unexpected conditions instead of returning `EMPTY`. `EMPTY` is what is defaulted to *upon* an exception, and should not be used directly as much as possible.

jinified:
Fixes #1174 

- [X] create FilterException and subclassed by ParseException and SemanticException 
- [X] change applyStringFilter to catch `FilterException`
- [x] change id, state, has, no, in, type, is, created, updated, repo,

### DIFF
--- a/src/main/java/filter/FilterException.java
+++ b/src/main/java/filter/FilterException.java
@@ -1,0 +1,8 @@
+package filter;
+
+@SuppressWarnings("serial")
+public class FilterException extends RuntimeException {
+    public FilterException(String message) {
+        super(message);
+    }
+}
--- a/src/main/java/filter/FilterException.java
+++ b/src/main/java/filter/FilterException.java
@@ -1,7 +1,12 @@
 package filter;
 
+/**
+ * Thrown to indicate failure while processing a filter expression
+ * Exceptions that extend this class should provide an appropriate error message
+ * and reason that it occurs
+ */
 @SuppressWarnings("serial")
-public class FilterException extends RuntimeException {
+public abstract class FilterException extends RuntimeException {
     public FilterException(String message) {
         super(message);
     }
--- a/src/main/java/filter/ParseException.java
+++ b/src/main/java/filter/ParseException.java
@@ -1,8 +1,8 @@
 package filter;
 
 @SuppressWarnings("serial")
-public class ParseException extends RuntimeException {
+public class ParseException extends FilterException {
       public ParseException(String message) {
-         super(message);
+         super("Error in filter: " + message);
       }
  }
--- a/src/main/java/filter/ParseException.java
+++ b/src/main/java/filter/ParseException.java
@@ -1,5 +1,9 @@
 package filter;
 
+/**
+ *  Thrown to indicate failure to parse a filter expression due to 
+ *  unsupported syntax
+ */
 @SuppressWarnings("serial")
 public class ParseException extends FilterException {
       public ParseException(String message) {
--- a/src/main/java/filter/SemanticException.java
+++ b/src/main/java/filter/SemanticException.java
@@ -2,13 +2,18 @@ package filter;
 
 import filter.expression.QualifierType;
 
+/**
+ * Thrown to indicate failure to interpret a filter expression which may 
+ * be resulted from invalid qualifier inputs, multiple qualifiers which 
+ * should only appear once or simply an empty input.
+ */
 @SuppressWarnings("serial")
 public class SemanticException extends FilterException {
     
-    public static final String ERR_MSG = "\"%s\" expects %s";
+    public static final String ERROR_MESSAGE = "\"%s\" expects %s";
+
     public SemanticException(QualifierType type, String details) {
-        super(String.format(ERR_MSG, type, details));
-        System.out.println(String.format(ERR_MSG, type, details));
+        super(String.format(ERROR_MESSAGE, type, details));
     }
 
 }
--- a/src/main/java/filter/SemanticException.java
+++ b/src/main/java/filter/SemanticException.java
@@ -12,8 +12,8 @@ public class SemanticException extends FilterException {
     
     public static final String ERROR_MESSAGE = "\"%s\" expects %s";
 
-    public SemanticException(QualifierType type, String details) {
-        super(String.format(ERROR_MESSAGE, type, details));
+    public SemanticException(QualifierType type) {
+        super(String.format(ERROR_MESSAGE, type, type.getDescriptionOfValidInputs()));
     }
 
 }
--- a/src/main/java/filter/SemanticException.java
+++ b/src/main/java/filter/SemanticException.java
@@ -1,0 +1,14 @@
+package filter;
+
+import filter.expression.QualifierType;
+
+@SuppressWarnings("serial")
+public class SemanticException extends FilterException {
+    
+    public static final String ERR_MSG = "\"%s\" expects %s";
+    public SemanticException(QualifierType type, String details) {
+        super(String.format(ERR_MSG, type, details));
+        System.out.println(String.format(ERR_MSG, type, details));
+    }
+
+}
--- a/src/main/java/filter/expression/Qualifier.java
+++ b/src/main/java/filter/expression/Qualifier.java
@@ -339,11 +339,13 @@ public class Qualifier implements FilterExpression {
         case TITLE:
         case DESCRIPTION:
         case KEYWORD:
-            throw new QualifierApplicationException("Unnecessary filter: issue text cannot be changed by dragging");
+            throw new QualifierApplicationException(
+                "Unnecessary filter: issue text cannot be changed by dragging");
         case ID:
             throw new QualifierApplicationException("Unnecessary filter: id is immutable");
         case CREATED:
-            throw new QualifierApplicationException("Unnecessary filter: cannot change issue creation date");
+            throw new QualifierApplicationException(
+                "Unnecessary filter: cannot change issue creation date");
         case HAS:
         case NO:
         case IS:
@@ -358,9 +360,11 @@ public class Qualifier implements FilterExpression {
             applyAssignee(issue, model);
             break;
         case AUTHOR:
-            throw new QualifierApplicationException("Unnecessary filter: cannot change author of issue");
+            throw new QualifierApplicationException(
+                "Unnecessary filter: cannot change author of issue");
         case INVOLVES:
-            throw new QualifierApplicationException("Ambiguous filter: cannot change users involved with issue");
+            throw new QualifierApplicationException(
+                "Ambiguous filter: cannot change users involved with issue");
         case STATE:
             applyState(issue);
             break;
@@ -480,12 +484,12 @@ public class Qualifier implements FilterExpression {
 
     private static boolean shouldBeStripped(Qualifier q) {
         switch (q.getType()) {
-            case IN:
-            case SORT:
-            case COUNT:
-                return true;
-            default:
-                return false;
+        case IN:
+        case SORT:
+        case COUNT:
+            return true;
+        default:
+            return false;
         }
     }
 
@@ -503,19 +507,19 @@ public class Qualifier implements FilterExpression {
 
     public static boolean isMilestoneQualifier(Qualifier q) {
         switch (q.getType()) {
-            case MILESTONE:
-                return true;
-            default:
-                return false;
+        case MILESTONE:
+            return true;
+        default:
+            return false;
         }
     }
 
     public static boolean isUpdatedQualifier(Qualifier q) {
         switch (q.getType()) {
-            case UPDATED:
-                return true;
-            default:
-                return false;
+        case UPDATED:
+            return true;
+        default:
+            return false;
         }
     }
 
@@ -549,82 +553,82 @@ public class Qualifier implements FilterExpression {
         boolean isLabelGroup = false;
 
         switch (key) {
-            case "comments":
-                comparator = (a, b) -> a.getCommentCount() - b.getCommentCount();
-                break;
-            case "repo":
-                comparator = (a, b) -> a.getRepoId().compareTo(b.getRepoId());
-                break;
-            case "updated":
-            case "date":
+        case "comments":
+            comparator = (a, b) -> a.getCommentCount() - b.getCommentCount();
+            break;
+        case "repo":
+            comparator = (a, b) -> a.getRepoId().compareTo(b.getRepoId());
+            break;
+        case "updated":
+        case "date":
+            comparator = (a, b) -> a.getUpdatedAt().compareTo(b.getUpdatedAt());
+            break;
+        case "nonSelfUpdate":
+            if (isSortableByNonSelfUpdates) {
+                comparator = (a, b) ->
+                    a.getMetadata().getNonSelfUpdatedAt().compareTo(b.getMetadata().getNonSelfUpdatedAt());
+            } else {
                 comparator = (a, b) -> a.getUpdatedAt().compareTo(b.getUpdatedAt());
-                break;
-            case "nonSelfUpdate":
-                if (isSortableByNonSelfUpdates) {
-                    comparator = (a, b) ->
-                        a.getMetadata().getNonSelfUpdatedAt().compareTo(b.getMetadata().getNonSelfUpdatedAt());
+            }
+            break;
+        case "assignee":
+        case "as":
+            comparator = (a, b) -> {
+                Optional<String> aAssignee = a.getAssignee();
+                Optional<String> bAssignee = b.getAssignee();
+
+                if (!aAssignee.isPresent() && !bAssignee.isPresent()) {
+                    return 0;
+                } else if (!aAssignee.isPresent()) {
+                    return 1;
+                } else if (!bAssignee.isPresent()) {
+                    return -1;
                 } else {
-                    comparator = (a, b) -> a.getUpdatedAt().compareTo(b.getUpdatedAt());
+                    return aAssignee.get().compareTo(bAssignee.get());
                 }
-                break;
-            case "assignee":
-            case "as":
-                comparator = (a, b) -> {
-                    Optional<String> aAssignee = a.getAssignee();
-                    Optional<String> bAssignee = b.getAssignee();
+            };
+            break;
+        case "milestone":
+        case "m":
+            comparator = (a, b) -> {
+                Optional<TurboMilestone> aMilestone = model.getMilestoneOfIssue(a);
+                Optional<TurboMilestone> bMilestone = model.getMilestoneOfIssue(b);
 
-                    if (!aAssignee.isPresent() && !bAssignee.isPresent()) {
+                if (!aMilestone.isPresent() && !bMilestone.isPresent()) {
+                    return 0;
+                } else if (!aMilestone.isPresent()) {
+                    return 1;
+                } else if (!bMilestone.isPresent()) {
+                    return -1;
+                } else {
+                    Optional<LocalDate> aDueDate = aMilestone.get().getDueDate();
+                    Optional<LocalDate> bDueDate = bMilestone.get().getDueDate();
+
+                    if (!aDueDate.isPresent() && !bDueDate.isPresent()) {
                         return 0;
-                    } else if (!aAssignee.isPresent()) {
+                    } else if (!aDueDate.isPresent()) {
                         return 1;
-                    } else if (!bAssignee.isPresent()) {
+                    } else if (!bDueDate.isPresent()) {
                         return -1;
                     } else {
-                        return aAssignee.get().compareTo(bAssignee.get());
+                        return -(TurboMilestone.getDueDateComparator()
+                            .compare(aMilestone.get(), bMilestone.get()));
                     }
-                };
-                break;
-            case "milestone":
-            case "m":
-                comparator = (a, b) -> {
-                    Optional<TurboMilestone> aMilestone = model.getMilestoneOfIssue(a);
-                    Optional<TurboMilestone> bMilestone = model.getMilestoneOfIssue(b);
-
-                    if (!aMilestone.isPresent() && !bMilestone.isPresent()) {
-                        return 0;
-                    } else if (!aMilestone.isPresent()) {
-                        return 1;
-                    } else if (!bMilestone.isPresent()) {
-                        return -1;
-                    } else {
-                        Optional<LocalDate> aDueDate = aMilestone.get().getDueDate();
-                        Optional<LocalDate> bDueDate = bMilestone.get().getDueDate();
-
-                        if (!aDueDate.isPresent() && !bDueDate.isPresent()) {
-                            return 0;
-                        } else if (!aDueDate.isPresent()) {
-                            return 1;
-                        } else if (!bDueDate.isPresent()) {
-                            return -1;
-                        } else {
-                            return -(TurboMilestone.getDueDateComparator()
-                                    .compare(aMilestone.get(), bMilestone.get()));
-                        }
-                    }
-                };
-                break;
-            case "id":
-                comparator = (a, b) -> a.getId() - b.getId();
-                break;
-            case "state":
-            case "status":
-            case "s":
-                comparator = (a, b) -> Boolean.compare(b.isOpen(), a.isOpen());
-                break;
-            default:
-                // Doesn't match anything; assume it's a label group
-                isLabelGroup = true;
-                break;
+                }
+            };
+            break;
+        case "id":
+            comparator = (a, b) -> a.getId() - b.getId();
+            break;
+        case "state":
+        case "status":
+        case "s":
+            comparator = (a, b) -> Boolean.compare(b.isOpen(), a.isOpen());
+            break;
+        default:
+            // Doesn't match anything; assume it's a label group
+            isLabelGroup = true;
+            break;
         }
 
         if (isLabelGroup) {
@@ -697,7 +701,7 @@ public class Qualifier implements FilterExpression {
         } else if (numberRange.isPresent()) {
             return numberRange.get().encloses(issue.getId());
         }
-        throw new SemanticException(type, type.getDescriptionOfValidInputs());
+        throw new SemanticException(type);
     }
 
     private boolean satisfiesUpdatedHours(TurboIssue issue) {
@@ -708,7 +712,7 @@ public class Qualifier implements FilterExpression {
         } else if (number.isPresent()) {
             updatedRange = new NumberRange(null, number.get(), true);
         } else {
-            throw new SemanticException(type, type.getDescriptionOfValidInputs());
+            throw new SemanticException(type);
         }
 
         LocalDateTime dateOfUpdate = issue.getUpdatedAt();
@@ -717,7 +721,7 @@ public class Qualifier implements FilterExpression {
     }
 
     private boolean satisfiesRepo(TurboIssue issue) {
-        if (!content.isPresent()) return false;
+        if (!content.isPresent()) throw new SemanticException(type);
         return issue.getRepoId().equalsIgnoreCase(content.get());
     }
 
@@ -728,12 +732,12 @@ public class Qualifier implements FilterExpression {
         } else if (dateRange.isPresent()) {
             return dateRange.get().encloses(creationDate);
         } else {
-            throw new SemanticException(type, type.getDescriptionOfValidInputs());
+            throw new SemanticException(type);
         }
     }
 
     private boolean satisfiesHasConditions(TurboIssue issue) {
-        if (!content.isPresent()) return false;
+        if (!content.isPresent()) throw new SemanticException(type);
         switch (content.get()) {
         case "label":
         case "labels":
@@ -749,7 +753,7 @@ public class Qualifier implements FilterExpression {
             assert issue.getAssignee() != null;
             return issue.getAssignee().isPresent();
         default:
-            throw new SemanticException(type, type.getDescriptionOfValidInputs());
+            throw new SemanticException(type);
         }
     }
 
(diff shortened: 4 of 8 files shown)


src/main/java/filter/ParseException.java
```diff
@@ -1,8 +1,8 @@
 package filter;
 
 @SuppressWarnings("serial")
-public class ParseException extends RuntimeException {
+public class ParseException extends FilterException {
       public ParseException(String message) {
-         super(message);
+         super("Parse error in filter: " + message);
```
damithc: Is this going to be shown to the user? In that case, you may want to say 'Error in filter'.

src/main/java/filter/SemanticException.java
```diff
@@ -0,0 +1,9 @@
+package filter;
+
+@SuppressWarnings("serial")
+public class SemanticException extends FilterException {
+    public SemanticException(String message) {
+        super("Semantic error in filter: " + message);
```
damithc: If this is going to be shown to the user, use a more user friendly phrase.

src/main/java/undo/UndoController.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private Label createInfoOcticon() {
+        Label label = new Label(OCTICON_INFO);
+        label.setPadding(new Insets(0, 0, 5, 0));
+        label.getStyleClass().addAll("octicon");
+        return label;
+    }
+
+}
```
damithc: Is this part of this PR, or came from another PR? If the latter, need to clean up the PR first.

src/main/java/filter/ParseException.java
jinified: I was just using the format of the previous ParseException.

src/main/java/undo/UndoController.java
jinified: It came from another PR, I will cleanup the PR before I ask for a review

src/main/java/filter/SemanticException.java
jinified: I am already planning to change the message and add suggestion of supported inputs

src/main/java/filter/SemanticException.java
```diff
@@ -0,0 +1,11 @@
+package filter;
+
+import filter.expression.QualifierType;
+
+@SuppressWarnings("serial")
+public class SemanticException extends FilterException {
+    public SemanticException(QualifierType type) {
+        super(String.format("Invalid input in filter for type \"%s\"", type));
```
dariusf: We don't have a lot of room in the status bar, so we might want to make the preamble shorter.

Also, this constructor could take another parameter for the details of the error. We should explain why the input causes a problem.

src/test/java/tests/FilterEvalTests.java
```diff
@@ -9,19 +9,28 @@
...
+import filter.expression.QualifierType;
 import prefs.Preferences;
 
 public class FilterEvalTests {
 
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
     private final IModel empty;
+    private static final String SEMANTIC_ERROR_MSG = "Invalid input in filter for"
```
dariusf: Could this field be in one of the exception classes? Otherwise we have to duplicate it here.

src/main/java/filter/SemanticException.java
jinified: Ok will work making the preamble shorter. The explanation can be universal like the input is not supported because a more specific explanation like what type of inputs are only supported would be too long to fit in the status bar also

src/test/java/tests/FilterEvalTests.java
jinified: I can put this in the SemanticException class

jinified:
What do @dariusf  think about these error message ? 

error_msg

dariusf:
Maybe we can drop the preamble entirely, and make the message as friendly as possible. For example,

```
"id" expects a number or a number range
"is" expects "open", "closed", "pr", "issue", or "read"
```
(Also the distinction between a data type and a string literal is made clearer.)

We can use the extra room to show the panel name instead of the number. Might need additional truncation on our end.

@damithc how do you find this?

By the way, we should find a way to integrate this new type of error into the [existing display](https://github.com/jinified/HubTurbo/blob/1174-unexpected-input-trigger-exception/src/main/java/ui/components/FilterTextField.java#L54) (which right now is quite coupled and only picks up parse errors), but that can be a separate issue.

damithc:
I agree.

jinified:
ready for review @dariusf

src/main/java/filter/expression/QualifierType.java
```diff
@@ -31,6 +32,27 @@
         return Collections.unmodifiableMap(aliases);
     }
 
+    // Ignore Qualifiers that expects a string where any user input would be valid
+    private static Map<QualifierType, String> initialiseValidInputs() {
+        Map<QualifierType, String> validInputs = new HashMap<>();
+        validInputs.put(ID, "a number or a number range");
+        validInputs.put(UPDATED, "a number or  a number range");
```
dariusf: Extra space here. :stuck_out_tongue:

src/main/java/filter/expression/QualifierType.java
```diff
@@ -31,6 +32,27 @@
...
+        validInputs.put(STATE, "\"open\" or \"closed\"");
+        validInputs.put(HAS, "\"label\", \"milestone\", or \"assignee\"");
+        validInputs.put(NO, "\"label\", \"milestone\", or \"assignee\"");
+        validInputs.put(IN, "\"title\" or \"body\"");
+        validInputs.put(TYPE, "\"issue\" or \"pr\"");
+        validInputs.put(CREATED, "a date or a date range");
+        validInputs.put(REPO, "repo id");
+        
+        //Show only most common inputs to safe space
```
dariusf: Typo here ("save space"). Also, aren't these all the keywords? Do you mean the frequently-used ones are ordered first?

src/main/java/filter/expression/QualifierType.java
```diff
@@ -31,6 +32,27 @@
...
+        validInputs.put(IS, "\"open\", \"closed\", \"pr\", \"issue\", \"merged\","
+                + " \"read\", \"unread\", or \"unmerged\"");
+        validInputs.put(SORT, "\"comments\", \"repo\", \"updated\", \"id\", "
+                + " \"assignee\", \"label\", \"state\", or \"milestone\"");
```
dariusf: `sort` doesn't expect the string "label", but a label name.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -174,11 +173,11 @@ private void applyStringFilter(String filterString) {
             } else {
                 this.applyFilterExpression(Qualifier.EMPTY);
             }
-        } catch (ParseException ex) {
+        } catch (FilterException ex) {
             this.applyFilterExpression(Qualifier.EMPTY);
             // Overrides message in status bar
             UI.status.displayMessage("Panel " + (panelIndex + 1)
-                + ": Parse error in filter: " + ex.getMessage());
+                + ": " + ex.getMessage());
```
dariusf: This is a bit tangential, but maybe we could add the panel name here, defaulting to the index only if it's unnamed?

jinified: Sure. I was afraid to edit that since it is not part of the issue. I will edit that

src/main/java/filter/expression/QualifierType.java
jinified: Miss out on that. Thanks.

src/main/java/filter/expression/QualifierType.java
jinified: Yup

dariusf: Okay, we should update the comment.

src/test/java/tests/FilterEvalTests.java
```diff
@@ -37,6 +44,17 @@ public boolean matches(String filterExpr, TurboIssue issue) {
         return Qualifier.process(empty, Parser.parse(filterExpr), issue);
     }
 
+    /**
+     * Helper method for testing SemanticException thrown by different qualifiers
+     */
+    public void expectSemanticException(QualifierType type, String invalidInput) {
```
damithc: `verifySemanticExceptionThrown`?

jinified: Yes. That actually seems like a better method name

jinified:
ready for review @dariusf

src/main/java/filter/SemanticException.java
```diff
@@ -0,0 +1,14 @@
+package filter;
+
+import filter.expression.QualifierType;
+
+@SuppressWarnings("serial")
+public class SemanticException extends FilterException {
+    
+    public static final String ERR_MSG = "\"%s\" expects %s";
```
dariusf: I know this is more or less unambiguous in context, but we should try not to abbreviate names, just as a matter of style.

src/main/java/filter/expression/QualifierType.java
```diff
@@ -31,6 +32,26 @@
         return Collections.unmodifiableMap(aliases);
     }
 
+    // Set valid inputs for each keyword except those that takes in any string
```
dariusf: This can be a javadoc comment.

src/main/java/filter/expression/QualifierType.java
```diff
@@ -93,4 +114,11 @@ private static boolean isSpecialQualifier(String input) {
     public String toString() {
         return name().toLowerCase();
     }
+    
+    public String getValidInputs() {
```
dariusf: We should add a javadoc comment to this.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -174,14 +173,21 @@ private void applyStringFilter(String filterString) {
...
             // Overrides message in status bar
-            UI.status.displayMessage("Panel " + (panelIndex + 1)
-                + ": Parse error in filter: " + ex.getMessage());
+            UI.status.displayMessage(getPanelNameForErrorMsg(panelMenuBar.getPanelName())
+                + ": " + ex.getMessage());
         }
     }
 
+    private String getPanelNameForErrorMsg(String panelName) {
+        if (panelName.equals("Panel")) {
```
dariusf: Is there some default panel name constant that we can compare against here?

dariusf:
Mostly just style issues. Also, needs to be updated to the latest master. You might want to squash away some of the extraneous commits, like things added in the course of review, things which should have been there to begin with, etc.

src/test/java/tests/FilterEvalTests.java
```diff
@@ -37,6 +44,17 @@ public boolean matches(String filterExpr, TurboIssue issue) {
...
     }
 
+    /**
+     * Helper method for testing SemanticException thrown by different qualifiers
+     */
+    public void verifySemanticException(QualifierType type, String invalidInput) {
+        TurboIssue issue = new TurboIssue(REPO, 1, "");
+        thrown.expect(SemanticException.class);
+        thrown.expectMessage(String.format(SemanticException.ERR_MSG, type, 
+                type.getValidInputs()));
```
damithc: Line wrapping can be better, to show the nested relationship. e.g.
```java
thrown.expectMessage(
          String.format(SemanticException.ERR_MSG, type, type.getValidInputs()));
```

damithc: See line wrapping guidelines in this doc https://github.com/nus-cs2103/website/blob/master/contents/coding-standards-java.md

jinified:
can review @dariusf

dariusf:
Okay, no more comments from me.

src/main/java/filter/SemanticException.java
```diff
@@ -0,0 +1,14 @@
+package filter;
+
+import filter.expression.QualifierType;
+
+@SuppressWarnings("serial")
+public class SemanticException extends FilterException {
```
damithc: Write a header comment to explain the purpose of this class.

src/main/java/filter/SemanticException.java
```diff
@@ -0,0 +1,14 @@
...
+    
+    public static final String ERROR_MESSAGE = "\"%s\" expects %s";
+    public SemanticException(QualifierType type, String details) {
+        super(String.format(ERROR_MESSAGE, type, details));
+        System.out.println(String.format(ERROR_MESSAGE, type, details));
```
damithc: Is this intended?

src/main/java/filter/expression/QualifierType.java
```diff
@@ -32,6 +33,30 @@
     }
 
     /**
+     * 
+     * Set valid inputs for each keyword except those that takes in any string
+     * @return
```
damithc: Incomplete comment

src/main/java/filter/expression/QualifierType.java
```diff
@@ -32,6 +33,30 @@
...
+     */
+    private static Map<QualifierType, String> initialiseValidInputs() {
+        Map<QualifierType, String> validInputs = new HashMap<>();
```
damithc: To be precise, these are 'descriptions of valid input', not 'valid inputs'

src/main/java/filter/expression/QualifierType.java
```diff
@@ -93,4 +118,15 @@ private static boolean isSpecialQualifier(String input) {
     public String toString() {
         return name().toLowerCase();
     }
+    
+    /**
+     * Retrieves acceptable inputs of a qualifier 
+     * @return
+     */
+    public String getValidInputs() {
```
damithc: Method name not precise enough.

src/main/java/filter/expression/QualifierType.java
```diff
@@ -93,4 +118,15 @@ private static boolean isSpecialQualifier(String input) {
...
+        String defaultExpectedInput = "string";
+        if (VALID_INPUTS.containsKey(this)) return VALID_INPUTS.get(this);
+        return defaultExpectedInput;
```
damithc: Instead of having a Map defined and populated elsewhere, can't we use a switch statement inside this method. Seems this is the only place where that Map is used.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -179,12 +178,19 @@ private void applyStringFilter(String filterString) {
...
             this.applyFilterExpression(Qualifier.EMPTY);
             // Overrides message in status bar
-            UI.status.displayMessage("Panel " + (panelIndex + 1)
-                + ": Parse error in filter: " + ex.getMessage());
+            UI.status.displayMessage(getPanelNameForErrorMsg(panelMenuBar.getPanelName())
+                + ": " + ex.getMessage());
+        }
+    }
+
+    private String getPanelNameForErrorMsg(String panelName) {
```
damithc: Don't define method names based on how the method is used.

src/main/java/filter/expression/QualifierType.java
```diff
@@ -32,6 +32,20 @@
...
+    private static Map<QualifierType, String> initialiseDescriptionOfValidInputs() {
+        Map<QualifierType, String> validInputs = new HashMap<>();
+        
+        validInputs.put(IS, "\"open\", \"closed\", \"pr\", \"issue\", \"merged\","
+                + " \"read\", \"unread\", or \"unmerged\"");
+        validInputs.put(SORT, "\"comments\", \"repo\", \"updated\", \"id\", "
+                + " \"assignee\", label name, \"state\", or \"milestone\"");
+        return Collections.unmodifiableMap(validInputs);
+    }
+
```
damithc: Is this method uses somewhere?

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -179,12 +178,19 @@ private void applyStringFilter(String filterString) {
...
             // Overrides message in status bar
-            UI.status.displayMessage("Panel " + (panelIndex + 1)
-                + ": Parse error in filter: " + ex.getMessage());
+            UI.status.displayMessage(resolvePanelInfo(panelMenuBar.getPanelName())
+                + ": " + ex.getMessage());
+        }
+    }
+
+    private String resolvePanelInfo(String panelName) {
+        if (panelName.equals("Panel")) {
```
damithc: Magic string

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -179,12 +178,19 @@ private void applyStringFilter(String filterString) {
...
             this.applyFilterExpression(Qualifier.EMPTY);
             // Overrides message in status bar
-            UI.status.displayMessage("Panel " + (panelIndex + 1)
-                + ": Parse error in filter: " + ex.getMessage());
+            UI.status.displayMessage(resolvePanelInfo(panelMenuBar.getPanelName())
+                + ": " + ex.getMessage());
+        }
+    }
+
+    private String resolvePanelInfo(String panelName) {
```
damithc: Header comment for this method?

src/main/java/ui/issuepanel/FilterPanel.java
jinified: Should I add the a string constant to [PanelMenuBar.java](https://github.com/HubTurbo/HubTurbo/blob/master/src/main/java/ui/components/PanelMenuBar.java#L43) where the panel name is initialized ?

damithc: Sure.

src/test/java/tests/FilterEvalTests.java
```diff
@@ -37,6 +44,17 @@ public boolean matches(String filterExpr, TurboIssue issue) {
         return Qualifier.process(empty, Parser.parse(filterExpr), issue);
     }
 
+    /**
+     * Helper method for testing SemanticException thrown by different qualifiers
+     */
```
damithc: Umm... not a very good description of the method. Write it is the form of `Does abc`. e.g. `Confirms that the the given input causes a SemanticException with the correct error message. Causes a test failure if it doesn't.`

damithc: When exactly is the expected exception is thrown. It's not clear from the method.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -179,12 +179,24 @@ private void applyStringFilter(String filterString) {
...
+            UI.status.displayMessage(resolvePanelName(panelMenuBar.getPanelName())
+                + ": " + ex.getMessage());
+        }
+    }
+
+    /**
+     * Shows panel name if renamed by user. Panel index is shown otherwise.
+     * @param panelName
+     * @return final panel name shown to users
+     */
```
damithc: Still not clear what this method does (or why this method exists).

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -179,12 +179,24 @@ private void applyStringFilter(String filterString) {
...
+    private String resolvePanelName(String panelName) {
+        if (panelName.equals(PanelMenuBar.DEFAULT_PANEL_NAME)) {
```
damithc: Feels like `PanelMenuBar.DEFAULT_PANEL_NAME` should be a parameter to this method.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -179,12 +179,24 @@ private void applyStringFilter(String filterString) {
...
+            return PanelMenuBar.DEFAULT_PANEL_NAME + " " + (panelIndex + 1);
```
damithc: And this one `panelIndex`. Try to give all the "inputs" the method need as parameters. That makes the method signature more self explanatory.

src/main/java/filter/SemanticException.java
```diff
@@ -0,0 +1,16 @@
...
+
+import filter.expression.QualifierType;
+
+/**
+ * Thrown to indicate an invalid input has been provided to a Qualifier
+ */
+@SuppressWarnings("serial")
+public class SemanticException extends FilterException {
+    
+    public static final String ERROR_MESSAGE = "\"%s\" expects %s";
```
damithc: Add blank line here?

jinified: After the class header comment ?

damithc: Between the variable declaration and the method declaration.

jinified: Seems like I already changed it on my local copy. Sorry for the confusion

jinified:
review

src/main/java/filter/FilterException.java
```diff
@@ -0,0 +1,8 @@
+package filter;
+
+@SuppressWarnings("serial")
```
dariusf: I guess we could document this as well, as being a general filter-related error. Maybe it should be abstract, so we ensure that something that's actually thrown has a concrete type/reason for happening?

src/main/java/filter/ParseException.java
```diff
@@ -1,8 +1,11 @@
 package filter;
 
+/**
+ *  Thrown to indicate failure to interpret a filter expression
```
dariusf: "Failure to parse" might be a bit more precise. This exception indicates a problem with syntax.

src/main/java/filter/SemanticException.java
```diff
@@ -0,0 +1,17 @@
+package filter;
+
+import filter.expression.QualifierType;
+
+/**
+ * Thrown to indicate an invalid input has been provided to a Qualifier
```
dariusf: This is a little specific. It can indicate other sorts of failures when trying to interpret these expressions. Maybe write something along those lines (problems with interpretation), and give a few examples? "For example, invalid qualifier inputs, multiple qualifiers which should only appear once, etc."

src/main/java/filter/expression/QualifierType.java
```diff
@@ -93,4 +93,32 @@ private static boolean isSpecialQualifier(String input) {
     public String toString() {
         return name().toLowerCase();
     }
+    
+    /**
+     * Retrieves acceptable inputs of a qualifier
```
dariusf: Maybe this could be something like: "returns a short description of the class of valid inputs a type of qualifier may be used with".

src/main/java/filter/expression/QualifierType.java
```diff
@@ -93,4 +93,32 @@ private static boolean isSpecialQualifier(String input) {
...
+            case IN:
+                return "\"title\" or \"body\"";
+            case TYPE:
+                return "\"issue\" or \"pr\"";
+            case CREATED:
+                return "a date or a date range";
+            case REPO:
+                return "repo id";
+            default:
+                return "string";
```
dariusf: "a repo id", and "a string"?

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -179,12 +179,26 @@ private void applyStringFilter(String filterString) {
...
+        }
+    }
+
+    /**
+     * Appends panel index to panel name if a panel is unnamed.
+     * This allows user to identify an otherwise unnamed panel.
+     * @param panelName
+     * @return final panel name shown to users
+     */
+    private String resolvePanelName(String panelName, String defaultName, int panelIndex) {
```
dariusf: Hmm, I'm not sure about these being parameters. They are all properties of the class that we're in. How about something like `getFriendlyName()`? Basically it returns a nice representation to be shown to users based on the panel's state.

jinified: I think removing the default name which is a constant and panelIndex which belongs to the class is fine. However, keeping the panelName as parameter because it is not a direct property of the class

dariusf: But it's transitively a property. :stuck_out_tongue: It can be accessed from there too.

damithc:
Test methods don't follow the new naming convention, but we can correct it later.

dariusf:
Almost there. Is the test coverage sufficient?

jinified:
It did cover all possibilities.

dariusf:
I mean, you should check coveralls and see where the drop is coming from.

jinified:
The drop mainly comes from MenuControl class which is not related to this pr at all

dariusf:
No idea why MenuControl was appearing earlier, but it seems like the drops are from the `in` and `repo` alias cases. Is it possible to cover them them now without too much effort? I know they aren't like the other qualifiers.

I don't think the FilterPanel parts are significant to test.

jinified:
I already added test for `in` qualifier. I am not too sure about `repo` because technically it accepts any string and shouldn't throw any SemanticException. I don't we should write a test for `repo`.

dariusf:
Hmm, there are no tests because they both don't throw semantic exceptions. Shouldn't they? If they didn't, we wouldn't need entries for them in `getDescriptionOfValidInputs`. Those cases are not covered because they aren't used.

On side note, just noticed that semantic exceptions are all created like so: `new SemanticException(type, type.getDescriptionOfValidInputs())`. If every qualifier type encapsulates the description of its inputs, we shouldn't need the second parameter.

Sorry for the last minute changes! We should sort out the coverage issues now, otherwise they'll never be fixed.

jinified:
Haha it's ok. It is needed for `IN` but not for `REPO`, and the description is added for the sake of completion. Come to to think we can remove it until there is a use case for handling valid repo id

dariusf:
Hmm, shouldn't we validate the repo id as well? Any string is permitted (given that repo aliases are coming), but other things like dates and ranges shouldn't be.

jinified:
Because the way we parse `QualifierContent`, the `content` will only be assigned if it is a normal keyword. If i were to throw a SemanticException when the content is empty, it may also throw when there is no input which is supposed to be a ParseException.

jinified:
Never mind I found a way to identify valid repo id by simply checking if the other fields are empty i.e `date`, `dateRange`, `number` and `numberRange`.

dariusf:
What's an example where there's no input?

jinified:
"repo:"

dariusf:
Hmm, but doesn't that currently get caught before it ever gets to the content part? It will be a parse exception, yes, but it will happen even before the filter expression is evaluated.

jinified:
Sorry my bad. You are right. Guess I will go ahead and throw the exception when the content is empty instead of returning false. Will need to change some of the tests for metaqualifier .

dariusf:
Okay, no worries.

jinified:
@dariusf  is it ok ?

src/main/java/filter/expression/Qualifier.java
```diff
@@ -210,43 +212,60 @@ public boolean isSatisfiedBy(IModel model, TurboIssue issue, MetaQualifierInfo i
...
-        case UPDATED:
-            return satisfiesUpdatedHours(issue);
-        case REPO:
-            return satisfiesRepo(issue);
-        default:
-            assert false : "Missing case for " + type;
-            return false;
+            case ID :
+                return idSatisfies(issue);
+
```
dariusf: Hmm... do we use these whitespace conventions everywhere? We should follow the rest of the codebase, which is like how it was.

jinified: I changed it because it was mentioned in the coding guideline which follows the TEAMMARE guideline. So I thought it would be a good idea to follow it.  The whitespace is not common in the code base

damithc:
Hmm... The space use in `switch` is inconsistent within the PR itself. Our coding standard says 

`case ID :` 

but I think that's an error in the standard itself. Should be `case ID:`?

jinified:
actually it is more natural without the space after each case i.e `case ID:` which is also recommended by Google's Java guideline. But what do you think about the newline after each case ? I think the newline after each case is good to follow since it clearer this way.

damithc:
Regarding the extra blank line above each `case`, while it makes the code look tidier, we are always trying to reduce vertical height (that's the the main reason why we don't align start and end braces). Extra blank lines can help to separate logical code blocks when all code is indented at the same level. However, each `case` is indented at a different level than the code that follows it, providing a visual break in the code even without the blank line. Therefore, we can do without the extra blank line.

damithc:
Coding standard updated.

jinified:
Reverted the changes. review please

src/main/java/filter/expression/Qualifier.java
```diff
@@ -726,28 +732,28 @@ private boolean satisfiesCreationDate(TurboIssue issue) {
...
-            assert issue.getAssignee() != null;
-            return issue.getAssignee().isPresent();
-        default:
-            return false;
+            case "label":
+            case "labels":
+                return issue.getLabels().size() > 0;
+            case "milestone":
+            case "milestones":
+            case "m" :
```
damithc: extra space before `:`. Noticed it in a few other places too.

src/test/java/tests/FilterEvalTests.java
```diff
@@ -726,4 +755,24 @@ public void filterLabelMatching() {
         assertFalse(Qualifier.labelMatches("pi.hi", "p.high"));
         assertFalse(Qualifier.labelMatches(".", "p.high"));
     }
+
+    /**
+     * Helper method for testing an issue against a filter string in the context
+     * of an empty model.
+     */
```
damithc: Follow the same style when commenting. e.g. ~~`Helper method for testing `~~ `Confirms that ...`

damithc:
@jinified 
While you are at it, can you remove the indentation of `case` clauses again?. I know it was it our previous coding standard, but the common convention seems to align `case` to `switch` which makes sense because it is just like an if-else-if statement.
I updated the coding standard.
@dariusf can merge after above is fixed. no further approval needed.

dariusf:
:tada:

jinified:
lol

PR #1201, opened 2016-01-20
Use consistent Chrome user profile #1200

### LINKED ISSUE #1200: Consistently use a chrome profile instead of creating new ones every time
By default, ChromeDriver will create a new temporary profile for each session. This means that if we close the companion chrome window and open a new one, it will create a new user profile such that all user data (caches, cookies, databases, etc) is gone. So it will need to login again, which takes some time, and it is a nightmare for those who has 2FA enabled (See #977). 

One way to improve this would be to ask `chromedriver` to use a consistent profile, probably store it in `store/chrome_profile`. I do not recommend to use the user's personal Chrome profile directly as some settings or plugins must break our functionalities. 

Since it is using the same Chrome profile each time, there is no need to login each time (good news for 2FA users!) This can be a temporary workaround for #977, until we completely eliminate the browser view. It also speeds up the loading time for a new Chrome window as the login page is immediately skipped.

I wonder if there is any drawback for doing this. If no, I would proceed working on it.

xinan:
Fixes #1200 

This is done by setting `user-data-dir` chrome option to `store/chrome_profile`.
See #1200 for more details.

Comments are welcome.


HansNewbie:
This PR has to wait until the chrome driver issue is resolved. I will try to update the bug in the chrome driver forum.

PR #1204, opened 2016-01-22
Fix unexpected behaviour when parsing number in quotes #1194

### LINKED ISSUE #1194: Fix parser not parsing number in quote as string
Causes filter expression to return false when matching with issue. For example: 

- I have an issue with body that contains the string `normalize output into (0 - 255)`.
- Filter expression: `body:"0"`

jinified:
Fixes #1194

### DIFF
--- a/src/main/java/filter/Parser.java
+++ b/src/main/java/filter/Parser.java
@@ -13,6 +13,8 @@ import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import util.Utility;
+
 public final class Parser {
 
     private Parser(ArrayList<Token> input) {
@@ -56,7 +58,6 @@ public final class Parser {
         // Prefix
 
         FilterExpression left;
-
         switch (token.getType()) {
         case LBRACKET:
             left = parseGroup();
@@ -67,8 +68,8 @@ public final class Parser {
         case QUALIFIER:
             left = parseQualifier(token);
             break;
-        case QUOTE:
-            left = parseQuotedKeywords();
+        case QUOTED_CONTENT:
+            left = parseQuotedContent(QualifierType.KEYWORD, token);
             break;
         case SYMBOL:
             left = parseKeyword(token);
@@ -107,18 +108,9 @@ public final class Parser {
         return left;
     }
 
-    private FilterExpression parseQuotedKeywords() {
-        FilterExpression result = parseQuotedContent(QualifierType.KEYWORD);
-        consume(TokenType.QUOTE);
-        return result;
-    }
-
-    private FilterExpression parseQuotedContent(QualifierType type) {
-        StringBuilder sb = new StringBuilder();
-        while (!isQuoteToken(lookAhead())) {
-            sb.append(consume().getValue()).append(' ');
-        }
-        return new Qualifier(type, sb.toString().trim());
+    private FilterExpression parseQuotedContent(QualifierType type, Token token) {
+        String keywords = token.getValue();
+        return new Qualifier(type, Utility.stripQuotes(keywords));
     }
 
     private FilterExpression parseKeyword(Token token) {
@@ -174,7 +166,7 @@ public final class Parser {
         qualifierName = qualifierName.substring(0, qualifierName.length() - 1).trim();
 
         QualifierType type = QualifierType.parse(qualifierName).orElse(QualifierType.FALSE);
-        return parseSeparatedContent(() -> parseQualifierContent(type, false));
+        return parseSeparatedContent(() -> parseQualifierContent(type));
     }
 
     private FilterExpression parseSeparatedContent(Supplier<FilterExpression> contentParser) {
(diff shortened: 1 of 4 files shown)


jinified:
ready for review @dariusf

dariusf:
Hmm, this may not be the best fix. For one, `allowMultipleKeywords` in that method will never be true, so [this case](https://github.com/jinified/HubTurbo/blob/1194-fix-parsing-of-content-in-quote/src/main/java/filter/Parser.java#L210) will never be entered.

The fix I was thinking of was to move all the quote-handling from the parser to the lexer. That way quoted strings are parsed as a single token, and we can just check for that here, without having to consume tokens in a loop.

jinified:
@dariusf  should we change the TokenType of QUOTE -> QUOTED_CONTENT since one token is essentially includes not only the quote.

dariusf:
Yes, that would be good.

jinified:
ready for review @dariusf

src/main/java/filter/lexer/Lexer.java
```diff
@@ -26,7 +26,7 @@
 
             new Rule("\\(", TokenType.LBRACKET),
             new Rule("\\)", TokenType.RBRACKET),
-            new Rule("\\\"", TokenType.QUOTE),
+            new Rule(".*\\\"(.*)\\\".*", TokenType.QUOTED_CONTENT),
```
dariusf: Hmm, this is a bit too permissive. Right now things like `body:@#$%^"is"` pass tests. Try to match the smallest possible set of strings which constitute 'something that's quoted'. For example, we probably don't want to allow `"""` (which also works now).

More tests for this would also be nice.

src/main/java/filter/Parser.java
```diff
@@ -107,18 +106,10 @@ private FilterExpression parseExpression(int precedence) {
...
-
-    private FilterExpression parseQuotedContent(QualifierType type) {
-        StringBuilder sb = new StringBuilder();
-        while (!isQuoteToken(lookAhead())) {
-            sb.append(consume().getValue()).append(' ');
-        }
-        return new Qualifier(type, sb.toString().trim());
+    private FilterExpression parseQuotedContent(QualifierType type, Token token) {
+        String keywords = token.getValue();
+        keywords = keywords.substring(1, keywords.length() - 1);
```
dariusf: Instead of this (which doesn't always remove the correct characters), we can use `Utility.stripQuotes` and `trim`.

src/main/java/filter/Parser.java
```diff
@@ -223,7 +206,7 @@ private boolean isKeywordToken(Token token) {
     }
 
     private boolean isQuoteToken(Token token) {
-        return token.getType() == TokenType.QUOTE;
+        return token.getType() == TokenType.QUOTED_CONTENT;
```
dariusf: This method name can be updated, to be more specific.

jinified:
@dariusf  I have been thinking about the pattern for a quoted content. My approach is to whitelist a character instead of simply negating a symbol such as ". Though this is more restrictive, considering the use case, not may users would match symbol such as "?" , "[" , "]". However, it will still permit whitespace char so "     " is will be matched.

dariusf:
Both have upsides and downsides. Users may want to quote characters which are normally taken as syntax, e.g. `a:":"`, so these should be allowed.

If the regex ends up duplicating all the others, it may be better to go with the negation approach, but otherwise being more restrictive is fine.

jinified:
ready for review

src/main/java/filter/lexer/Lexer.java
```diff
@@ -26,7 +26,7 @@
 
             new Rule("\\(", TokenType.LBRACKET),
             new Rule("\\)", TokenType.RBRACKET),
-            new Rule("\\\"", TokenType.QUOTE),
+            new Rule("\\\"([^(\\\")]+)\\\"", TokenType.QUOTED_CONTENT),
```
dariusf: The parentheses around the `[...]+` are unnecessary I think, because this simple lexer doesn't use capturing groups.

Are the parentheses inside necessary? Did you intend to disallow `label:"(abc)"`?

jinified: Nope

dariusf: Just noticed: escaping inside the `[]` may be unnecessary as well.

jinified: Is it because inside the [ ] it will just matches the character itself right ?

jinified: This is because inside the [ ] it will just match the character right ?

dariusf: Yes. Actually... `new Rule("\"[^\"]+\"", ...)` will work. The escaping from before was redundant. :stuck_out_tongue:

dariusf:
Same, are the tests adequate here? Seeing as coverage isn't green.

jinified:
I have tested different cases for parsing content in quotes.

src/main/java/filter/lexer/Lexer.java
```diff
@@ -26,7 +26,7 @@
 
             new Rule("\\(", TokenType.LBRACKET),
             new Rule("\\)", TokenType.RBRACKET),
-            new Rule("\\\"", TokenType.QUOTE),
+            new Rule("\\\"[^\"]+\\\"", TokenType.QUOTED_CONTENT),
```
dariusf: The escaping for the surrounding quotes is not needed. This can be simplified to `new Rule("\"[^\"]+\"", ...)`.

dariusf:
Looks like the coverage drop is a false positive. No more comments from me.

PR #1208, opened 2016-01-23
Saving boards: warning for existing board should be stronger #1203

### LINKED ISSUE #1203: Saving boards: warning for existing board should be stronger
image

Once I hit enter before I noticed the warning.
Instead of showing a warning below the text box, there should be another dialog box for the user to confirm. The message can be 'Overwrite the existing board with the same name?' Yes/No

m133225:
Fixes #1203.

Added an event filter to display a warning dialog upon clicking the confirm button if the name already exists.
image


damithc:
We can make the warning more detailed.
```
A board by the name XYZ already exists. 
Overwrite? Yes | No
```

damithc:
Add a comment when the PR is ready for review again.

m133225:
Ready for review!

damithc:
Can post a screenshot of the new dialog?

m133225:
screen shot 2016-01-31 at 11 34 58 am

ndt93:
@HansNewbie Can you help review this? Thanks.

HansNewbie:
@damithc do you want to keep the original warning message in the naming dialog? i.e., there will be warning of duplicate name in the naming dialog *and* there will be prompt about the duplicate name.

HansNewbie:
@m133225 can you write a test for this behaviour? You can use `unstable`>`MenuControlTest.java` as reference, though I think it would be good if you can create a new test class inside `guitests` under the name of `BoardTests` or something similar. We can move all tests related to board to this class later (right now the tests for board is still all over the place).

src/main/java/ui/BoardNameDialog.java
```diff
@@ -44,60 +49,71 @@ public BoardNameDialog(Preferences prefs, Stage mainStage) {
...
             if (submit == submitButtonType) {
                 return nameField.getText();
             }
             return null;
         });
-        
+
+    }
+
+    private boolean showWarningDuplicateName() {
```
HansNewbie: Since we are getting a return value here, `show` might not be concise enough. Maybe we can rename this method into `confirmDuplicateBoardName` or something along the way.

damithc:
No harm keeping both. The original warning can be made more noticeable too (e.g. red color)? Can be done as a separate issue though. Either fix in this issue or create a new issue for that.

src/test/java/guitests/BoardTests.java
```diff
@@ -0,0 +1,64 @@
...
+        click("OK");
+        assertEquals(1, panelControl.getPanelCount());
+
+        // Create a new panel, then save with the same name
+        // Expected: Dialog shown to confirm duplicate name
+        clickMenu("Panels", "Create");
+        clickMenu("Boards", "Save as");
+        type("board 1");
+        click("OK");
+        assertNodeExists(hasText("A board by the name 'board 1' already exists."));
```
HansNewbie: It's better to use `waitUntilNodeAppears` here so that if there is any lag in the window appearing, it doesn't fail immediately. This line causes the Travis to fail right now.

src/main/java/ui/BoardNameDialog.java
```diff
@@ -44,60 +49,71 @@ public BoardNameDialog(Preferences prefs, Stage mainStage) {
...
             if (submit == submitButtonType) {
                 return nameField.getText();
             }
             return null;
         });
-        
+
+    }
+
+    private boolean confirmDuplicateBoardName() {
```
damithc: Boolean methods should be named as such. e.g. `isDuplicateBoardName`

src/main/java/ui/BoardNameDialog.java
```diff
@@ -44,60 +49,71 @@ public BoardNameDialog(Preferences prefs, Stage mainStage) {
...
-    
+
     protected void setupGrid() {
         GridPane grid = new GridPane();
         setupGridPane(grid);
-        
+
         setTitle("Save board as");
-        
+
```
damithc: `Save board as ...`?

src/test/java/guitests/BoardTests.java
```diff
@@ -0,0 +1,64 @@
...
+        waitUntilNodeAppears(hasText("A board by the name 'board 1' already exists."));
+
+        // Overwrite previous board, then open the board again
+        // Expected: the board should contain 2 panels
+        click("Yes");
+        clickMenu("Boards", "Open", "board 1");
+        assertEquals(2, panelControl.getPanelCount());
+
+
+        // Create a new panel, then save with the same name to show warning dialog
```
damithc: Comma missing at the end of this line.

src/main/java/ui/BoardNameDialog.java
damithc: Use a better name. The current one sounds like 'confirm it is a duplicate name'. What it should sound like is 'confirm OK to overwrite existing board'

damithc:
No more comments from me.

src/test/java/guitests/BoardTests.java
```diff
@@ -0,0 +1,64 @@
...
+        click("Yes");
+        clickMenu("Boards", "Open", "board 1");
+        assertEquals(2, panelControl.getPanelCount());
+
+
+        // Create a new panel, then save with the same name to show warning dialog,
+        // don't save and try to reopen board
+        // Expected: Board is not overwritten, should contain 2 panels
+        clickMenu("Panels", "Create");
+        clickMenu("Boards", "Save as");
```
HansNewbie: Can we have an `assert` here before `Board save as` that the panel count is at 3? Just to ensure that the panel count at 2 is because of cancelled overwriting instead of failure to create panel (hence panel count is still at 2 but overwriting might have happened).

HansNewbie:
Since we are touching this file and working on duplicate name, can you rename the variable in [BoardNameDialog.java line 36](https://github.com/m133225/HubTurbo/blob/1203-boards-stronger-warning/src/main/java/ui/BoardNameDialog.java#L36) from `ERROR_DUPLICATE_NAME` to `WARNING_DUPLICATE_NAME`?

Help me to comment again when it's ready for review. Thanks!

m133225:
@HansNewbie should BoardTests be under unstable? It seems to work on local but not on Travis, where it waits indefinitely at waitUntilNodeAppears. I've tried to restart the Travis build about 3-4 times and they produced the same result.

HansNewbie:
@dariusf any suggestion on this? I am not sure why the wait fails for board
test. Maybe because of dialog pop up?

dariusf:
Yes, that could be the cause. @m133225 maybe you could try to find out where it gets stuck, for example, if the dialog never closes? If we really can't find a way to make this work, I'm fine with it being unstable. We should note the reason in a comment somewhere.

m133225:
Ready for review!

In the end, the problem wasn't that the warning dialog didn't appear. Instead, it was because it couldn't type 'board 1' as the name in the board name dialog.
At first I suspected that it was due to the focus of the text field, but even when I changed the test to manually click on the text field before typing, the name wasn't changed for some reason. To work around this temporarily, I've left it to the default 'New Board' name.

src/test/java/guitests/BoardTests.java
```diff
@@ -0,0 +1,70 @@
...
+
+        List<String> boardNames = testPref.getAllBoardNames();
+        boardNames.stream().forEach(testPref::removeBoard);
+    }
+
+    @Test
+    public void duplicateNameTest() {
+        // Save the current board
+        clickMenu("Boards", "Save as");
+        click("OK");
```
HansNewbie: Can add a comment above this line that `board name will default to "New Board"` so that there is no confusion on what name the board is saved as?

HansNewbie:
Aside from my last comment, please update the branch. Update me again once that's done. The rest seems fine to me.

src/test/java/guitests/BoardTests.java
dariusf: We should probably also say why the board name was left as the default (click on CI doesn't focus on the field, etc.).

m133225:
@HansNewbie Added comments and updated the branch!

HansNewbie:
@damithc I have reviewed this PR. I have no more comments.

Issue #1173, opened 2016-01-06
Multiple sort qualifiers should trigger a parse exception


dariusf:
Instead of taking the first one, we should explicitly indicate an error. Depends on #1172.

dariusf:
Maybe we can generalise this to all of the qualifiers which should only appear once (repo, count).

PR #1210, opened 2016-01-25
filter syntax: Support a shorter form of OR #1137

### LINKED ISSUE #1137: filter syntax: Support a shorter form of OR
e.g. `repo:repo1,repo2,repo3` or `repo:repo1|repo2|repo3`
This can be particularly useful when we create a panel with selected issues. `id:124,3245,6452,7689`

raisfathin:
Fixes #1137 

Still has problems in parsing quoted content because the parser currently doesn't parse quoted content exactly as typed by the user as described at #1137.


src/main/java/filter/expression/QualifierContent.java
```diff
@@ -0,0 +1,66 @@
+package filter.expression;
+
+import java.time.LocalDate;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+public class QualifierContent {
+
+    //Only one of these will exist in a QualifierContent object
```
damithc: Is there a more elegant way to handle this 'only one can be present' condition? e.g. inheritance, enums

dariusf: [Inheritance](http://programmers.stackexchange.com/a/163987) should be nicer here, since the values are parameterised, and each type of content should know how to create its corresponding qualifier (good use case for method dispatch).

Anyway, this (anti-)pattern is repeated inside Qualifier as well, but there it's a bit harder to get rid of without resorting to instanceof.

raisfathin: Isn't it non-trivial to determine the qualifier type based on only the content? For example, `involves` and `authors` use the same value type.

Is it fine to store the `QualifierType` inside `QualifierContent` (or its subtype) class? I found it to be quite odd to implement it that way. If that's the case, won't it contain the same data as the full `Qualifier` class?

dariusf: Hmm, true. The intention was for us to be able to treat the output of these parsing functions in a polymorphic way. If we can do this using just qualifiers, that would be ideal.

raisfathin: I think we can just use the existing Qualifier class for this purpose.

Actually, can we just move the `parseSeparatedContent` part to `parseQualifier` as below? This way, it would be able to parse qualifier content which might need different parsing function to be parsed (for example, to parse `id:<5;6 .. 7;10` we need to use `parseRangeOperator` and `parseNumberOrNumberRange`).

```java
private FilterExpression parseQualifier(Token token) {
    String qualifierName = token.getValue();

    // Strip the : at the end, then trim
    qualifierName = qualifierName.substring(0, qualifierName.length() - 1).trim();

    QualifierType type = QualifierType.parse(qualifierName).orElse(QualifierType.FALSE);
    if (type == QualifierType.SORT) { //special case for sort
        return new Qualifier(type, parseSortKeys());
    } else {
        return parseSeparatedContent(type, () -> parseQualifierContent);
    }
}
```

dariusf: Yup, that would be great.

raisfathin: Done, ready for review.

src/main/java/filter/Parser.java
```diff
@@ -173,13 +174,25 @@ private FilterExpression parseQualifier(Token token) {
         qualifierName = qualifierName.substring(0, qualifierName.length() - 1).trim();
 
         QualifierType type = QualifierType.parse(qualifierName).orElse(QualifierType.FALSE);
-        return parseQualifierContent(type, false);
+        if (type == QualifierType.SORT) {
+            //special case for sort
```
damithc: Explain further why it is a special case?

raisfathin: Because the semicolon operator doesn't make sense only when applied to sort qualifier.

We can also just parse it as it is and leave it to `UpdateController` to determine what to do when there are more than 1 sort qualifier (#1173?). Either that or we can prevent it from the parser by raising `ParseException`.

I just realised that I forgot to set the priority of semicolon type to handle a few corner cases. Will fix them along with this issue and write some tests as well.

dariusf: The original special case was that sorting keys would only parse if preceded by `sort`.

More than one sort qualifier is a semantic error, no need to handle that in the parser. Ideally it could be handled somewhere down the line, but before it gets to UpdateController (which handles the logic of updates, and shouldn't be concerned with checking its inputs).

We should consider what happens when another qualifier is used with sorting keys though, e.g. `id:~id,repo`. Is this a parse error or a semantic error? Maybe it should be the latter, since we have the distinction now?

damithc: I meant 'add a comment in the code to explain why it is a special case' :-)

raisfathin: @dariusf I agree. For the distinction, are you referring to #1195? Btw, for the second issue, should we try to solve it together in this PR or just create another issue?

@damithc Oops, my bad :(

dariusf: Yes, I'm referring to that issue.

Another issue, since I don't think we'll remove that edge case now.

raisfathin: Ok. Ready for another review.

src/main/java/filter/Parser.java
```diff
@@ -126,6 +127,8 @@ private FilterExpression parseKeyword(Token token) {
 
     private int getInfixPrecedence(Token token) {
         switch (token.getType()) {
+        case SEMICOLON:
+            return Precedence.SHORTER_DISJUNCTION;
```
dariusf: Hmm, is this needed? This is for operators between qualifiers, which are taken care of by the main loop in `parseExpression`. `;` isn't handled there as it's only ever used in the context of a single qualifier.

The parse exception in the `sort:label;assignee` case is due to `;` not being a recognised infix operator, not because `sort` doesn't parse properly.

src/test/java/tests/FilterParserTests.java
```diff
@@ -123,6 +147,23 @@ public void operators() {
...
+        try {
+            Parser.parse("&&");
+            fail("Operator by itself should cause a parse exception.");
+        } catch (ParseException ignored) {}
+
+        //might move this test in the future when proper
+        //semantic exception is implemented
+        try {
+            Parser.parse("sort:label;assignee");
+            fail("Operator by itself should cause a parse exception.");
```
dariusf: This message can be changed to reflect that.

src/test/java/tests/FilterParserTests.java
```diff
@@ -123,6 +147,23 @@ public void operators() {
...
+        } catch (ParseException ignored) {}
+
+        try {
+            Parser.parse("&&");
+            fail("Operator by itself should cause a parse exception.");
+        } catch (ParseException ignored) {}
+
+        //might move this test in the future when proper
+        //semantic exception is implemented
```
dariusf: In that case, you can prefix the comment with "TODO" and include the number of the issue it's referring to.

src/test/java/tests/FilterParserTests.java
```diff
@@ -72,6 +72,10 @@ public void unexpectedEOFs() {
             Parser.parse("a(b) ||");
             fail("Inputs which end unexpectedly should throw a parse exception");
         } catch (ParseException ignored) {}
+        try {
+            Parser.parse("a:b; c:d;e");
+            fail("Inputs which end unexpectedly should throw a parse exception");
```
dariusf: This isn't really a case of the input ending unexpectedly. Rather, it's because some kind of content is expected as the input to qualifier `a`, and `c:` (the start of a new qualifier) can't be taken as content.

It's a useful test, but we should make the assertion message more accurate and place it near the other related tests.

dariusf:
On second thought, I don't think this depends on #1194, since after transforming the semicolon, invalid inputs to qualifiers would sort themselves out.

It depends on #1174 only if we are removing the `sort` edge case and making it a semantic error... are we doing that?

src/main/java/filter/Parser.java
raisfathin: I think a better solution would be to just pass everything to `parseSeparatedContent` (and still retain the special case to sort qualifier) and let the semantic error handler in the future to handle it.

src/test/java/tests/FilterParserTests.java
raisfathin: Noted, will be done in the next push.

src/test/java/tests/FilterParserTests.java
raisfathin: Hmm, didn't know why I put it there in the first place. Thanks for the correction.

src/main/java/filter/Parser.java
dariusf: (This is a response to another comment?)

But okay, I think it may not be worth the trouble. We can leave it as a parse error and just take `sort` to be part of the syntax.

raisfathin:
Yeah, I think you're right. Shall I edit the pull request description?

Isn't it going to be a separate issue? It's also related to #1173.

dariusf:
Okay, sure. Yeah, and also I think it may not be worth the trouble. We can leave it as a parse error and just take `sort` to be part of the syntax.

raisfathin:
Ready for review.

dariusf:
Okay, this is good. Could you squash the outdated commits? Like those which added things which have since been removed in the course of review.

dariusf:
Thanks!

PR #1212, opened 2016-01-25
Add aliases for Qualifiers and Keywords #1151

### LINKED ISSUE #1151: Aliases for qualifiers - sort, type, body, title
Alias for `sort` could be `so`. Aliases for the other qualifiers can be discussed!

jinified:
Fixes #1151 

- [x] add aliases for Qualifier 
- [x] add aliases for keywords using the same method for milestones
- [x] unit tests 
- [x] update documentation

### DIFF
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -94,7 +94,7 @@ This will not work for ambiguous expressions (containing OR or NOT operators) an
 - [`id`](#id)
 - [`keyword`](#keyword)
 - [`title`](#title)
-- [`body`](#body)
+- [`description`](#description)
 - [`milestone`](#milestone)
 - [`label`](#label)
 - [`assignee`](#assignee)
@@ -139,13 +139,15 @@ Matches all issues with text containing the given string. Same as not specifying
 
 Matches all issues with a title containing the given string.
 
-### body
+Aliases: `t`
+
+### description
 
 *Expects a string*
 
 Matches all issues with a body (or description) containing the given string.
 
-Aliases: `desc`, `description`
+Aliases: `body`, `de`, `desc`
 
 ### milestone
 
@@ -155,7 +157,7 @@ Matches all issues associated with any milestones whose names contain the given 
 
 `current` or `curr` can be used to refer to an open milestone with earliest due date, or an overdue open milestone with open issues if any. However, if there is only one open milestone, it will be considered as the `current` milestone even if it does not have due date. `curr-2`, `curr-1`, `curr+1`, `curr+2`, etc. (no space before or after `-`/`+`) can then be used to refer to milestones before or after `current`, sorted by due date. If there is no open milestone, `current` will refer to no milestone, while `current-1` will refer to the last closed milestone.
 
-Aliases: `m`
+Aliases: `m`, `milestones`
 
 ### label
 
@@ -193,7 +195,10 @@ Aliases: `user`
 
 Matches all issues of the given state.
 
-Aliases: `s`, `status`
+Aliases: `st`, `status`
+
+- `open` can be written as `o`
+- `closed` can be written as `c`
 
 ### has
 
@@ -201,10 +206,10 @@ Aliases: `s`, `status`
 
 Matches issues associated with the given type of resource.
 
-Aliases:
+Aliases: `h`
 
-- `label` can be written as `labels`
-- `milestone` can be written as `milestone` or `m`
+- `label` can be written as `labels` or `l`
+- `milestone` can be written as `milestones` or `m`
 - `assignee` can be written as `assignee` or `as`
 
 ### no
@@ -213,7 +218,7 @@ Aliases:
 
 The negation of `has`. Matches issues not associated with the given type of resource.
 
-Aliases: see [has](#has).
+Aliases: `n`. For input aliases, see [has](#has).
 
 ### in
 
@@ -227,11 +232,23 @@ Meta-qualifier. Changes the semantics of search terms to check only either the t
 
 Matches issues of a given issue type. Pull requests are loosely considered issues in that the same operations work on both; this predicate allows users to distinguish them.
 
+Aliases: `ty`
+
+- `issue` can be written as `i`
+- `pr` can be written as `pullrequest` or `p`
+
 ### is
 
 *Expects one of `open`, `closed`, `pr`, `issue`, `merged`, `unmerged`, `read`, `unread`*
 
 Matches issues which are either open or closed, a pull request or an issue, depending on their merged status if they are pull requests, or read or unread. Is partially an alias for `state` and `type`.
+
+Aliases: 
+
+- `merged` can be written as `mg`
+- `unmerged` can be written as `um`
+- `read` can be written as `rd`
+- `unread` can be written as `ur`
 
 ### created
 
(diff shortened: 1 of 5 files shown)


damithc:
Don't forget to update tests and documentation too.

jinified:
No problem will do that. Just thinking of better way to add alias for keywords for now. `curr` is the only alias for keywords right now

damithc:
`curr` is different from other aliases as in it is not 'a shorter version of a longer term`. It's more or a dynamic alias (it can resolve into a different value at different times) while the rest are static aliases.

ndt93:
@jinified Let us know when when this PR (or part of it) is ready for review so a reviewer can be assigned. Thanks.

jinified:
Here are some concerns I have about how aliases are handled:

1. Though `resolveAlias` is supposed to replace alias during parsing. Most of the checks for qualifier and keywords still rely on [case fallthrough in switch statement](70d567c1dc597b17e9fc8d19ab6bddcb0fd6d9ea)
2.  `resolveAlias` only works on QualifierType which does not include keyword such as `closed`, `open` which does not qualify as a QualifierType for now.

Here are several alternatives to handle aliases for all keywords:
- replace qualifier content before `switch` so we can eliminate multiple `case`
- convert qualifier content to qualifiertype where we perform alias conversion in constructor.Match on `QualifierType` instead of string which is rather messy

jinified:
@damithc 
Just realized there are multiple way a user refer to a qualifier or keyword. Here are some of them:

- labels -> label (plural form)
- milestones -> milestone (plural form)
- pullrequest -> pr (only checks by `type` qualifier)

I figured they should be included as aliases to centralize all aliases together instead of having to consistently account for them in `switch` statement which is very error prone. In my opinion, an alias does not necessarily be shorter but just another way to represent something which depends on users.

dariusf:
Yes, `resolveAliases` only handles *qualifier* aliases. We can't really use an enum for *input* aliases because there isn't a fixed set of inputs. I wouldn't put them together in `resolveAliases` for those reasons. The method could probably be given a more specific name, and the distinction between these two kinds of aliases made clearer.

It would be nice to centralise the input aliases somewhere, but I don't think we can do it in general, because certain qualifiers are expressly not supposed to expand aliases (`label` inputs are taken verbatim). Maybe we can write a method which expands them, and call it for the qualifiers for which aliases work.

jinified:
I have written a method `replaceKeywordAliases` which will be called by specific `qualifier` to expand the input so we only need to match for one case per keyword.

dariusf:
Okay, sounds good.

src/main/java/filter/expression/QualifierType.java
```diff
@@ -19,15 +19,27 @@
 
     private static Map<String, QualifierType> initialiseAliases() {
```
damithc: Does this Map contains all aliases? If not, at least add a comment to explain what it excludes and why.

src/test/java/tests/QualifierTests.java
```diff
@@ -324,6 +324,13 @@ public void repoOrdering() {
         assertEquals(Arrays.asList(0, 1, 2, 3, 4, 5, 6, 7, 8, 9), renderedIssues.stream()
             .map(TurboIssue::getId)
             .collect(Collectors.toList()));
+
+        Collections.sort(renderedIssues,
+            Qualifier.getSortComparator(model, "r", false, false));
+
+        assertEquals(Arrays.asList(5, 6, 7, 8, 9, 0, 1, 2, 3, 4), renderedIssues.stream()
+            .map(TurboIssue::getId)
+            .collect(Collectors.toList()));
```
damithc: It's better if you can add a comment to explain what is being tested.

docs/filters.md
```diff
@@ -227,50 +232,72 @@ Meta-qualifier. Changes the semantics of search terms to check only either the t
...
 *Expects a number*
 
 The maximum number of issues that can be displayed in a panel. For example, `count:4` would display a maximum of 4 issues in the panel.
 
+Aliases: `cn`
+
 Available sorting criteria:
 
-- `comments`: sorts by number of comments (in descending order)
+- `comments`: sorts by number of comments (in descending order). Aliases: `cm`.
```
dariusf: Could you help to shift this back under `sort` heading? Seems to have been wrongly moved down here.

dariusf:
Looks good. Are the tests complete? There are a few TODOs, and I didn't check to see if every alias is covered.

jinified:
Both `date` and `count` is  used only in UpdateController. They are currently not tested in FilterEvalTest. Perhaps another pull request when we move `sort` and `count` to Qualifier. i have removed the TODO.

dariusf:
No more comments from me. We should get CI to pass though.

jinified:
Should I update the branch ?

src/test/java/tests/FilterEvalTests.java
```diff
@@ -587,6 +601,13 @@ public void type() {
         assertEquals(true, matches("type:pr", issue));
         assertEquals(false, matches("type:sldkj", issue));
 
+        // test: qualifier alias
+        assertEquals(false, matches("ty:issue", issue));
```
damithc: Can use `assertFalse`?

src/test/java/tests/QualifierTests.java
```diff
@@ -503,5 +530,12 @@ public void statusOrdering() {
         assertEquals(Arrays.asList(4, 5, 6, 7, 0, 1, 2, 3), renderedIssues.stream()
                 .map(TurboIssue::getId)
                 .collect(Collectors.toList()));
+
+        Collections.sort(renderedIssues,
+                Qualifier.getSortComparator(model, "st", false, false));
+        
+        assertEquals(Arrays.asList(0, 1, 2, 3, 4, 5, 6, 7), renderedIssues.stream()
+                .map(TurboIssue::getId)
+                .collect(Collectors.toList()));
```
damithc: What's being tested here?

damithc: BTW, the three code segments you added seem very similar. Can extract as a method?

dariusf:
Yeah, update whenever you can.

jinified:
review

src/test/java/tests/QualifierTests.java
```diff
@@ -490,18 +473,38 @@ public void statusOrdering() {
...
+                         model, "st", false, false);
+
+        // inverted sort order
+        testSortedIssues(renderedIssues, Arrays.asList(4, 5, 6, 7, 0, 1, 2, 3), 
+                         model, "status", true, false);
+
+    }
+    
+    /**
+     * Test order of TurboIssues after sorted according to a sortCriteria
```
damithc: `Tests the order ...`

src/main/java/filter/expression/Qualifier.java
```diff
@@ -113,6 +113,64 @@ public static FilterExpression replaceMilestoneAliases(IModel model, FilterExpre
     }
 
     /**
+     * Expands user input to a matching keyword, returns original input otherwise
+     * Includes aliases for Qualifier that also functions as keyword 
+     * @param input
+     * @return 
+     */
+    public static String replaceKeywordAliases(String input) {
```
dariusf: Maybe this could be: "expands aliases for qualifier keywords in user input"? The verb in the method name could be changed to "expand" also. Feels like a more high-level name.

src/test/java/tests/QualifierTests.java
```diff
@@ -490,18 +473,38 @@ public void statusOrdering() {
...
+        // inverted sort order
+        testSortedIssues(renderedIssues, Arrays.asList(4, 5, 6, 7, 0, 1, 2, 3), 
+                         model, "status", true, false);
+
+    }
+    
+    /**
+     * Tests the order of TurboIssues after sorted according to a sortCriteria
+     */
+    private void testSortedIssues(List<TurboIssue> issues, List<Integer> expectedIds,
```
dariusf: The method name and comment can be a bit more specific. `assertSorted`? "Checks that TurboIssues are ordered in a particular way given some sorting criteria".

jinified:
Ready for review

dariusf:
I think this is fine. Needs an update to latest master though.

jinified:
@dariusf  ready to be merged

PR #1213, opened 2016-01-26
Added an indicator to show that a panel is loading repo #1113

### LINKED ISSUE #1113: Show an indicator while a panel is loading
Sometimes a panel is blank because it is still loading the issues. It is better if we show some sort of 'loading ...' indicator during that period.

nicholaslum444:
Fixes #1113 

1. RepoOpeningEvent is triggered when a repo is going to be opened
2. FilterPanel receives event and will apply panel-loading indicator if it is concerned with that repo
3. RepoOpenedEvent is triggered when a repo is loaded
4. FilterPanel receives event and will remove panel-loading indicator if it is concerned with that repo


nicholaslum444:
Some problems:
- concurrency issues when dealing with multiple of the same event being triggered at once, sometimes causes serious exceptions
- primitive panel loading indication (grey panel color), to be improved when there is a more stable backend fix

damithc:
Are you planning to solve those problems in this PR?
Let others know when the PR is ready for review.

nicholaslum444:
I'm not sure how to handle the concurrency problem. Also, I need advice on how to solve the PMD issue as I require that method in the constructor.

I can continue to work on the UI for panel loading indication. The plan is to put a loading spinner gif somewhere on the panel

acruis:
I think you're on the right direction. I'm not sure if this works, but you can synchronize event firing by queuing them on the UI thread using Platform.runLater.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -101,6 +107,34 @@ private void setUpEventHandler() {
             ui.triggerEvent(new PanelClickedEvent(this.panelIndex));
             requestFocus();
         });
+
+        ui.registerEvent((RepoOpeningEventHandler) e -> {
+            logger.info("Received RepoOpeningEvent");
```
dariusf: I'm not sure this is useful, since we already have logging for when events are triggered. Maybe something more high-level, like 'panel loading', 'panel loaded'?

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -183,6 +217,28 @@ private void applyStringFilter(String filterString) {
     }
 
     /**
+     * Triggered when a RepoOpeningEvent is received by the filter panel
+     *
+     */
+    private void addPanelLoadingIndication() {
+        logger.info("Preparing to add panel loading indication");
+        synchronized (this) {
```
dariusf: Generally we try to avoid this sort of low-level synchronisation, as it is ad-hoc and doesn't compose well.

In this case Platform.runLater should be the solution. You have to either make sure these events are always triggered from the UI thread (like @acruis suggested), or received on the UI thread. I don't think there's a big difference here, but you should use the one that leads to less duplication (so we can't forget to synchronise somewhere).

src/main/java/util/events/RepoOpenedEvent.java
```diff
@@ -2,7 +2,10 @@
 
 public class RepoOpenedEvent extends UnusedStoredReposChangedEvent {
     public final String repoId;
-    public RepoOpenedEvent(String repoId) {
+    public boolean isPrimaryRepo;
```
dariusf: Should be final.

src/main/java/util/events/RepoOpeningEvent.java
```diff
@@ -0,0 +1,15 @@
+package util.events;
+
+import ui.issuepanel.AbstractPanel;
+
+public class RepoOpeningEvent extends Event {
+
+    public String repoId;
+    public boolean isPrimaryRepo;
```
dariusf: Same here, for both fields.

dariusf:
About the PMD violation, it can't tell that the use of the overridable method is within a lambda. You can silence it by factoring the lambda out into a method.

Maybe it would be a good idea to also factor out other common behaviour, since the handlers for the opening/opened events look similar.

nicholaslum444:
I've solved this issue. Each loading panel shows a spinner in the middle of the listview. Please take a look, thanks!

damithc:
@nicholaslum444 change the label when ready for review.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -370,4 +375,32 @@ private void changeLabels() {
             ui.triggerEvent(new ShowLabelPickerEvent(getSelectedIssue().get()));
         }
     }
+
+    /**
+     * Triggered when a RepoOpeningEvent is received by the filter panel.
```
damithc: A method has not control over when it is called. So the method comment should explain only what the method does.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -182,6 +188,32 @@ private void applyStringFilter(String filterString) {
...
 
+    private void indicatePanelLoading(RepoOpeningEvent e) {
+        togglePanelLoading(e.isPrimaryRepo, e.repoId, true);
+    }
+
+    private void unindicatePanelLoading(RepoOpenedEvent e) {
+        togglePanelLoading(e.isPrimaryRepo, e.repoId, false);
+    }
+
+    private void togglePanelLoading(boolean isPrimaryRepo, String repoId, boolean toIndicate) {
```
damithc: Try to name boolean parameters as such . e.g. `isIndicatorApplicable`

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -182,6 +188,32 @@ private void applyStringFilter(String filterString) {
...
+        HashSet<String> allReposInFilterExpr =
+                Qualifier.getMetaQualifierContent(getCurrentFilterExpression(), QualifierType.REPO);
+
+        // either primary repo changed and no repo filters, or repo filters changed
+        if (isPrimaryRepo && allReposInFilterExpr.isEmpty() || allReposInFilterExpr.contains(repoId)) {
```
damithc: Rather than add a comment, use explanatory intermediate variables. e.g.
`isPrimaryRepChanged = isPrimaryRepo && allReposInFilterExpr.isEmpty()`

nicholaslum444:
Noted, thanks!

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -182,6 +188,32 @@ private void applyStringFilter(String filterString) {
         }
     }
 
+    private void indicatePanelLoading(RepoOpeningEvent e) {
+        togglePanelLoading(e.isPrimaryRepo, e.repoId, true);
+    }
+
+    private void unindicatePanelLoading(RepoOpenedEvent e) {
+        togglePanelLoading(e.isPrimaryRepo, e.repoId, false);
+    }
```
damithc: For the above two methods, a better structure is:

```java
private void indicatePanelLoading(RepoOpeningEvent e) {
       if(isIndicatorApplicable(e.isPrimaryRepo, e.repoId){
          showLoadingIndicator();
      }
 }
```

In the current code, you have abstracted the wrong action. It's checking the applicability that is common, not the toggling.

src/main/java/util/events/RepoOpeningEvent.java
```diff
@@ -0,0 +1,15 @@
+package util.events;
+
+import ui.issuepanel.AbstractPanel;
+
+public class RepoOpeningEvent extends Event {
```
damithc: Always add a header comment at the top of the class

src/main/java/ui/issuepanel/FilterPanel.java
nicholaslum444: Actually, there are two different events. Repo OpenING Event and Repo OpenED Event. That's why I needed two different methods to handle them

src/main/java/util/events/RepoOpeningEvent.java
nicholaslum444: No other event does this though... Should I still do so?

damithc: Yes, it wasn't done until now, but let's make it a standard.

nicholaslum444: Okay. The corresponding handler too?

src/main/java/ui/issuepanel/FilterPanel.java
damithc: My comment is that you have extracted the wrong method. Instead of abstracting the toggle method, you should abstract to isIndicatorApplicable method.

nicholaslum444:
Please review the PR (I cant change the label, no rights)

src/main/java/backend/Logic.java
```diff
@@ -124,10 +126,17 @@ public void refresh(boolean isNotificationPaneShowing) {
             } else {
                 logger.info("Opening " + repoId);
                 UI.status.displayMessage("Opening " + repoId);
+                Platform.runLater(() -> UI.events.triggerEvent(new RepoOpeningEvent(repoId, isPrimaryRepository)));
```
ndt93: The `else` clause is not needed here. Not your code in this PR but would be nice if you help refactor this part :-P.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -370,4 +374,27 @@ private void changeLabels() {
...
+    @Override
+    protected void addPanelLoadingIndication() {
+        logger.info("Preparing to add panel loading indication");
+        listView.getStyleClass().add("listview-loading");
+
+        // Remove the items in the issue list because they are not relevant to the filter anymore.
+        // This also makes the listview background visible, since we intend to show a loading indicator on the
+        // background.
+        listView.setItems(null);
+    }
```
ndt93: So if there are for example 2 repos in the filter and we're opening a 3rd repo, from what I understand, this will cause the whole panel to go blank? I'm not sure if that's the desirable behaviour.

ndt93:
@nicholaslum444 Can you give more details on the concurrency problems? A scenario where there are unexpected behaviours and exceptions for example.

Also, would be nice if you can outline specific scenarios (series of actions) where this indicator comes into play so others can comment if that's the expected behaviour.

nicholaslum444:
@ndt93 The concurrency problems were because I wasn't running the correct things on the correct thread. But they have been solved by Platform.runLater(), so that's fine.

nicholaslum444:
@ndt93 

**Primary repo**
1. Primary repo is changed
2. All panels with no repos in filter are affected by the change
3. If the repo has to load/download, then the loading indicator will appear.
4. Once the repo is loaded, then the indicator disappears.

**Repo in filter**
1. Repos that are selected by the filter are changed
2. Only that panel is affected
3. If any of the repos need to be loaded, then the indicator appears.
4. Once all the repos have been loaded, then the indicator disappears.

A repo does not need to "load" if it is already opened in another panel, or is the primary repo. For all other repos, the "loading" will take place.

I realise I forgot to account for the case where one repo is loaded much earlier than the other. Currently when this happens, the loading indicator disappears the moment the first repo is completed. This will be fixed.

I chose to display the indicator on the listview because it is obvious to the user that some form of loading is taking place. Also, the panel has no other empty space left. However I agree that the removal of all the issues might disrupt the user's flow.

One alternative I can think of is replacing the X button on the panel with the loading spinner. On hover, the X button comes back and can still be used as per normal. But that sounds a little difficult...

Another alternative is simply changing the color of the entire panel to a greyish color. However it may not be intuitive to most users, so it would require additional user guide documentation, which I don't think is desirable either.

What are your opinions?

ndt93:
I don't have much problems with the whole panel blank out while some of the repos in the filter are being loaded since it might help to hide some inconsistencies during processing.

@damithc Is that what you would expect from the loading behaviour?

@nicholaslum444 Let me know when you finish the fix then I will check this again.

nicholaslum444:
@ndt93 I doubt I can fix this soon. I may have started out on the wrong foot, seeing that I seem to be fighting the backend a lot. I will be leaving this issue aside for now and starting on another issue.

damithc:
Let's go with 'blank while loading' approach now and see how it feels in usages. But we should not blank the panel during periodic refreshes.

damithc:
The minimal fix for this issue is to show the loading icon when a panel is being populated for the first time. That will make sure the icon shows up when the user opens HT, especially the very first time. In the current version HT opens with blank panel and the user is not sure if something is happening or not. Showing the icon when the filter is changed can be done as an enhancement later.

ndt93:
@nicholaslum444 Sure, do you want to do a simplified version as suggested by Damith or complete everything when you come back to this in the future?

nicholaslum444:
@ndt93 @damithc Yeah I think I can attempt the simplified version. The loading indicator will only appear on the change/loading of primary repo. For filter change, there will be no loading indicator. I will notify again when ready to review.

nicholaslum444:
@ndt93 Please review, thanks!

ndt93:
@nicholaslum444 No other comments from me. Please rebase/merge with latest master then I will change the PR status.

nicholaslum444:
@ndt93 Just rebased!

ndt93:
Ok. Can merge then.

PR #1220, opened 2016-01-30
HT not refreshing #1214

### LINKED ISSUE #1214: HT not refreshing
I've noticed this a few times in the recent days. HT goes into a stage where it doesn't seem to refresh anymore. This panel is like this for a long time now, although the filter has `is:unread`
image

ndt93:
Fixes #1214 

* Remove refresh block when undo pane is showing.
* Make `replaceIssueLabels` mutually exclusive with other repo operations.
* Eager update is now more immediate.

### DIFF
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -77,12 +77,7 @@ public class Logic {
         return repoIO.isRepositoryValid(repoId);
     }
 
-    public void refresh(boolean isNotificationPaneShowing) {
-        // TODO fix refresh to take into account the possible pending actions associated with the notification pane
-        if (isNotificationPaneShowing) {
-            logger.info("Notification Pane is currently showing, not going to refresh. ");
-            return;
-        }
+    public void refresh() {
         String message = "Refreshing " + models.toModels().stream()
                 .map(Model::getRepoId)
                 .collect(Collectors.joining(", "));
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -199,33 +199,42 @@ public class Logic {
      * @return True if label replacement on GitHub was a success, false otherwise.
      */
     public CompletableFuture<Boolean> replaceIssueLabels(TurboIssue issue, List<String> newLabels) {
-        eagerlyUpdateIssueLabels(issue, newLabels);
+        boolean isLabelsChanged = updateIssueLabels(issue, newLabels);
 
+        if (isLabelsChanged) {
+            logger.info("Changing labels for " + issue + " on UI");
+            refreshUI();
+        }
+
+        return updateIssueLabelsOnRepo(issue, newLabels);
+    }
+
+    /**
+     * Updates an issue with a set of labels
+     * @return true if changes are made to the labels on the issue
+     */
+    private Boolean updateIssueLabels(TurboIssue issue, List<String> labels) {
+        boolean isLabelsUnchanged = issue.getLabels().size() == labels.size() && issue.getLabels().containsAll(labels);
+        if (isLabelsUnchanged) {
+            logger.info("No changes to labels of " + issue);
+            return false;
+        }
+
+        issue.setLabels(labels);
+        return true;
+    }
+
+    /**
+     * Updates an issue on the repository with a set of labels
+     * @param issue
+     * @param newLabels
+     * @return
+     */
+    private CompletableFuture<Boolean> updateIssueLabelsOnRepo(TurboIssue issue, List<String> newLabels) {
         logger.info("Changing labels for " + issue + " on GitHub");
         return repoOpControl.replaceIssueLabels(issue, newLabels)
                 .thenApply(labels -> labels.containsAll(newLabels))
                 .exceptionally(Futures.withResult(false));
-    }
-
-    /**
-     * Assigns a list of labels to an issue offline and immediately reflect the change on the UI without
-     * concerning whether the labels have been updated on the remote repo's data source
-     *
-     * The UI will not be force refreshed if the set of labels are the same as the existing labels
-     * assigned to the issue.
-     * @param issue
-     * @param labels
-     */
-    private void eagerlyUpdateIssueLabels(TurboIssue issue, List<String> labels) {
-        logger.info("Changing labels for " + issue + " on UI");
-
-        if (issue.getLabels().size() == labels.size() && issue.getLabels().containsAll(labels)) {
-            return;
-        }
-
-        issue.setLabels(labels);
-        refreshUI();
-
     }
 
     /**
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -203,10 +203,7 @@ public class Logic {
 
         logger.info("Changing labels for " + issue + " on GitHub");
         return repoOpControl.replaceIssueLabels(issue, newLabels)
-                .thenApply(labels -> {
-                    eagerlyUpdateIssueLabels(issue, labels);
-                    return true;
-                })
+                .thenApply(labels -> labels.containsAll(newLabels))
                 .exceptionally(Futures.withResult(false));
     }
 
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -199,15 +199,36 @@ public class Logic {
      * @return True if label replacement on GitHub was a success, false otherwise.
      */
     public CompletableFuture<Boolean> replaceIssueLabels(TurboIssue issue, List<String> newLabels) {
+        eagerlyUpdateIssueLabels(issue, newLabels);
+
         logger.info("Changing labels for " + issue + " on GitHub");
-        return repoIO.replaceIssueLabels(issue, newLabels)
+        return repoOpControl.replaceIssueLabels(issue, newLabels)
                 .thenApply(labels -> {
-                    logger.info("Changing labels for " + issue + " on UI");
-                    issue.setLabels(labels);
-                    refreshUI();
+                    eagerlyUpdateIssueLabels(issue, labels);
                     return true;
                 })
                 .exceptionally(Futures.withResult(false));
+    }
+
+    /**
+     * Assigns a list of labels to an issue offline and immediately reflect the change on the UI without
+     * concerning whether the labels have been updated on the remote repo's data source
+     *
+     * The UI will not be force refreshed if the set of labels are the same as the existing labels
+     * assigned to the issue.
+     * @param issue
+     * @param labels
+     */
+    private void eagerlyUpdateIssueLabels(TurboIssue issue, List<String> labels) {
+        logger.info("Changing labels for " + issue + " on UI");
+
+        if (issue.getLabels().size() == labels.size() && issue.getLabels().containsAll(labels)) {
+            return;
+        }
+
+        issue.setLabels(labels);
+        refreshUI();
+
     }
 
     /**
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -192,44 +192,21 @@ public class Logic {
     }
 
     /**
-     * Replaces labels of issue on GitHub
+     * Replaces existing labels with new labels in the issue object, the UI, and the server, in that order.
+     * Server update is done after the local update to reduce the lag between the user action and the UI response
      *
-     * @param issue The issue whose labels are to be replaced on GitHub.
-     * @param newLabels The list of new labels to be applied on the issue.
-     * @return True if label replacement on GitHub was a success, false otherwise.
+     * @param issue The issue object whose labels are to be replaced.
+     * @param newLabels The list of new labels to be assigned to the issue.
+     * @return true if label replacement on GitHub was a success, false otherwise.
      */
     public CompletableFuture<Boolean> replaceIssueLabels(TurboIssue issue, List<String> newLabels) {
-        boolean isLabelsChanged = updateIssueLabels(issue, newLabels);
-
-        if (isLabelsChanged) {
-            logger.info("Changing labels for " + issue + " on UI");
-            refreshUI();
-        }
+        logger.info("Changing labels for " + issue + " on UI");
+        issue.setLabels(newLabels);
+        refreshUI();
 
         return updateIssueLabelsOnRepo(issue, newLabels);
     }
 
-    /**
-     * Updates an issue with a set of labels
-     * @return true if changes are made to the labels on the issue
-     */
-    private Boolean updateIssueLabels(TurboIssue issue, List<String> labels) {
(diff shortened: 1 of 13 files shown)


ndt93:
@lejolly Can you help review this?

src/main/java/backend/control/ops/ReplaceIssueLabelsOp.java
```diff
@@ -0,0 +1,37 @@
+package backend.control.ops;
```
lejolly: maybe `operations` is a better package name?

lejolly:
if `isNotificationPaneShowing` is not used anywhere else maybe you can remove that too. 

so what will happen if we update the UI but it fails to update on GitHub, say due to either connection being offline or failed authentication, etc.

ndt93:
For the failure case, it will need to wait until the next periodic refresh to be synced with GitHub. However, there are a few reasons why it was done this way.

There is is a bug I noticed when the UI update is deferred until after the API response and a `replace` request is quickly followed by an `undo` request. It occurs relatively often that the response from the first request is received last and hence the local model becomes inconsistent with both local user's expectation and the remote data for quite a while until the next periodic refresh.

Another reason is that although this does not always guarantee synchronisation with remote data, it has a strong local consistency i.e.how the model's states should be from the local user point of view.

Also this seems to fit the idea of eager update more, and a network failure should instead result in a reattempt to carry out the action (or a more reliable fallback) and a notifications to the user e.g. similar to how gmail behaves. These are not done here but should be done in the near future not only for this but all network operations (download and updates).

@damithc What's your opinion on this? To summarise, previously the eager UI update is actually deferred until a response from GitHub confirms successfully request to change label. Now it immediately updates the offline model and UI first.

ndt93:
@lejolly I renamed the package. `isNotificationPaneShowing` is still used in `MenuControl` to hide the pane when refresh menu is pressed so I still leave it there.

src/main/java/backend/Logic.java
```diff
@@ -196,18 +191,36 @@ public Model getRepo(String repoId) {
...
+                .thenApply(labels -> labels.containsAll(newLabels))
                 .exceptionally(Futures.withResult(false));
     }
 
     /**
+     * Assigns a list of labels to an issue offline and immediately reflect the change on the UI without
+     * concerning whether the labels have been updated on the remote repo's data source
+     *
+     * The UI will not be force refreshed if the set of labels are the same as the existing labels
+     * assigned to the issue.
```
damithc: This sentence is not actually part of this method. i.e. the code doesn't match this description. Based on the code, it simply refreshes the UI. On a related note, I wonder if `refreshUI();` should be part of this method. The abstraction level doesn't seem right.

damithc: After another reading, I realized the code matches the description more than I realized at first. After some refactoring, the code is like this.

```java
/**
  * If there are changes to the labels, the issue and the UI are updated.
  */
  private void updateIssueLabelsAndUpdateUi(TurboIssue issue, List<String> labels) {
        
        boolean isLabelsUnchanged = issue.getLabels().size() == labels.size() && issue.getLabels().containsAll(labels);
        if (isLabelsUnchanged) { 
             logger.info("No changes to labels of " + issue + " on UI");
             return; 
        }

        logger.info("Changing labels for " + issue + " on UI");
        issue.setLabels(labels);
        refreshUI();

    }
```
It doesn't seem to be related to `eager update` concept. It just updates the issue and the UI. I prefer it did just one of those things. Combining the two things should be done at a higher level.

ndt93: The `isLabelsUnchanged` check is mostly used to avoid doing `refreshUI` so they need to tag along together. One way is to make the Ui "observe" the issue somehow, then when we call `issue.setLabel` etc. it knows to update, but that's kind of far off. Another way is just to bring the whole block back up i.e. to `replaceIssueLabels`. What do you think?

damithc: One option is to make the `setLabels` method return `true` if there were any changes to the labels list. Would that work?

ndt93: Ah ok, that would make things nicer.

dariusf:
Being a bit paranoid, but can the consistency problems from before still happen? Is there a case where we could write something wrong to the cache or ignore a legitimate change to an eagerly-updated issue?

ndt93:
I think the only place where the in-memory store is written to disk is in `RepoIO.updateModel`, which is only called at the end of each periodic update as far as I'm aware of. That's the only valid time to do it at the moment since it guarantees to be consistent with remote data at some point of time.

So this shouldn't cause cache inconsistencies provided the above are correct.

ndt93:
As for the second point, yup, we might ignore the legitimate change. The periodic update will always override the eagerly updated changes, which is not fatal, but due to the queue, the chance is really small and should last only for a short moment.

ndt93:
@lejolly @damithc I've refactored the method. I decided not to change `TurboIssue.setLabels` since someone can use it on a large set of issues in the future and doesn't expect the overhead. The refactored method in `Logic` returns a `Boolean` instead.

src/test/java/tests/LogicTests.java
```diff
@@ -0,0 +1,85 @@
...
+        CompletableFuture<List<String>> resultLabels = new CompletableFuture<>();
+        resultLabels.complete(newLabels);
+        when(mockedRepoIO.replaceIssueLabels(any(TurboIssue.class), anyListOf(String.class)))
+                .thenReturn(resultLabels);
+
+        boolean status = logic.replaceIssueLabels(issue, newLabels).get();
+
+        assertTrue(status);
+        assertEquals(resultLabels.get(), issue.getLabels());
+    }
```
damithc: What about other interesting cases? e.g.  
1. original labels == new labels
2. new labels is a subset of original labels
If we are going to have more tests, consider creating a utility method and calling it repeatedly with different parameters. The above two test methods seems very similar already.

src/main/java/backend/control/operations/ReplaceIssueLabelsOp.java
```diff
@@ -0,0 +1,37 @@
...
+
+import backend.RepoIO;
+import backend.resource.TurboIssue;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+import static util.Futures.chain;
+
+public class ReplaceIssueLabelsOp implements RepoOp<List<String>> {
```
damithc: Header comment?

src/test/java/tests/LogicTests.java
ndt93: All right, sure. Hopefully I can add enough test cases by tomorrow :-P

ndt93:
@lejolly @damithc I've added the header and several more test cases.

The offline UI update gets overridden by a stale periodic update for 1 cycle when:
1. The offline update happens after the page containing the modified issue is downloaded and the UI refresh due to the periodic refresh.
2. Someone else modifies the issue less than 1 minute before the offline UI update.
Are the odds acceptable? Say window for (1) is 2s, then chance is 1/30 * [chance of (2)]. Otherwise we have to cancel any on going periodic refresh, which takes a bit more works to make it clean and safe.

src/main/java/backend/Logic.java
```diff
@@ -206,14 +201,41 @@ public Model getRepo(String repoId) {
...
     public CompletableFuture<Boolean> replaceIssueLabels(TurboIssue issue, List<String> newLabels) {
+        boolean isLabelsChanged = updateIssueLabels(issue, newLabels);
+
+        if (isLabelsChanged) {
+            logger.info("Changing labels for " + issue + " on UI");
+            refreshUI();
+        }
+
+        return updateIssueLabelsOnRepo(issue, newLabels);
+    }
```
damithc: * This reads better?
```java
    public CompletableFuture<Boolean> replaceIssueLabels(TurboIssue issue, List<String> newLabels) {
        
        if (isLabelsChanged(issue, newLabels)) {
            logger.info("Changing labels for " + issue + " on UI");
            issue.setLabels(newLabels)
            refreshUI();
        }

        return updateIssueLabelsOnRepo(issue, newLabels);
    }
```
* Why update on the repo even if there is no change in labels? If there is a reason, it should be given as a comment.

damithc: If you follow the suggested code,
1. `isLabelsChanged(issue, newLabels)` should be tested for the correctness of the boolean condition. e.g. new labels is a subset of originals
2. this method has only two paths, and needs two test cases only. i.e. labels changed case and labels unchanged case.

damithc: Here's the same code with a slight tweak:
```java
    public CompletableFuture<Boolean> replaceIssueLabels(TurboIssue issue, List<String> newLabels) {

        if (!issue.isLabelsSame(newLabels)) {
            logger.info("Changing labels for " + issue + " on UI");
            issue.setLabels(newLabels)
            refreshUI();
        }

        return updateIssueLabelsOnRepo(issue, newLabels);
    }
```

ndt93: Ok sure. I will refactor these. Also, this "almost correct" behaviour doesn't feel so good so I will try to deal with the clash with periodic update and put in some fallback measures, but likely won't make it for this milestone though.

ndt93:
@damithc I've refactored the relevant parts. You can take a look at this again.

src/main/java/backend/resource/TurboIssue.java
```diff
@@ -413,6 +413,16 @@ public void markAsUnread(Preferences prefs) {
     }
 
     /**
+     * Checks if the labels assigned to this issue is the same as {@code labels}
+     * @param labels
+     * @return
+     */
+    public Boolean isLabelsSame(List<String> labels) {
+        List<String> currentLabels = this.getLabels();
+        return currentLabels.size() == labels.size() && currentLabels.containsAll(labels);
```
damithc: Any built-in method to compare two lists?

src/test/java/tests/TurboIssueTests.java
```diff
@@ -111,4 +111,80 @@ public void testCombinePullRequest() {
...
+
+    /**
+     * Tests that isSameLabels returns false when the argument labels partially overlap with the issue's labels
+     */
+    @Test
+    public void testIsLabelsSamePartialOverlap() {
+        TurboIssue issue = createIssueWithLabels(Arrays.asList("label1", "label2", "label3"));
+        List<String> labels = Arrays.asList("label2", "label3", "label4");
+        assertFalse(issue.isLabelsSame(labels));
+    }
```
damithc: Not sure if it's worth having many test methods to test this simple method. Can you write a utility method like this?
`verifyIsLabelsSame(List<String> originalLabels, List<String> newLabels, boolean expected)`
Then you can call it multiple times in the same `@Test`.

src/main/java/backend/Logic.java
damithc: You missed this comment:
Why update on the repo even if there is no change in labels? If there is a reason, it should be given as a comment.

src/main/java/backend/resource/TurboIssue.java
ndt93: I've been looking around for this. There is an `equals` method on `List` but the elements must be in the same order. Sorting or adding them to a set first are more complicated and less efficient provided the few number of labels on each issue. So this seems to be the simplest way I can think of so far when we don't have to worry about duplicates.

src/main/java/backend/Logic.java
ndt93: Ok sure. This is done so that a user can force the action again when the issue and UI are updated but the labels are still not updated on GitHub after a while.

src/main/java/backend/resource/TurboIssue.java
damithc: Ok, got it.

src/main/java/backend/Logic.java
```diff
@@ -204,14 +199,21 @@ public Model getRepo(String repoId) {
      * @return True if label replacement on GitHub was a success, false otherwise.
```
damithc: The description need to be complete. e.g.
`If the given set of labels are different from the ones in the issue, updates the issue object and the UI with newLabels. Pushes the labels to the GitHub server irrespective of whether the labels are changed. Reason: so that this method can be called again to resolve persistent inconsistencies between the repo's state and the UI's state. Remote update is done after the local update, to avoid the lag between the user action and UI response.`

Now, if the last sentence means this is an 'eager' update, then you should put the 'eager' back into the method name.

More importantly, see how messy the method description is. That's a sign that this method is not well defined. What if we remove the equivalence test?  Then the description becomes cleaner.

`Replaces existing labels with new labels in the issue object, the UI, and the server, in that order. Server update is done after the local update to reduce the lag between the user action and the UI response.`

src/main/java/backend/Logic.java
```diff
@@ -204,14 +199,21 @@ public Model getRepo(String repoId) {
...
      */
     public CompletableFuture<Boolean> replaceIssueLabels(TurboIssue issue, List<String> newLabels) {
+        if (!issue.isLabelsSame(newLabels)) {
+            logger.info("Changing labels for " + issue + " on UI");
+            issue.setLabels(newLabels);
+            refreshUI();
+        }
+
+        // Always updates the repo so that a user can force the action again when there are
+        // persistent inconsistencies between the repo's state and the Ui's state
```
damithc: Having to mention a user action in a Logic method is not good.

src/main/java/backend/Logic.java
ndt93: I think removing the equivalence test is a good idea. There is a low chance that a user set the same set of labels (typically when the user want to force the action on GitHub again), so there is no significant benefit of avoiding the `refreshUI` in this case (which properly takes less than half a sec).

ndt93:
@damithc I've removed the labels equivalence check.

damithc:
No more comments from me. @lejolly ?

ndt93:
@lejolly @damithc Can this be merge for this coming milestone?

damithc:
I'm OK. @lejolly ?

ndt93:
@dariusf lejolly might be busy at the moment. Can you take a look at this?

dariusf:
Code looks good.

About the FindBugs warning, it's squelched in several places, but those aren't really the root issue. I don't think there's a case where we *don't* want to stub out `UI.events` and `UI.status`, right? If so we should stub it for all tests in some common place.

dariusf:
Hmm, this can be merged, since it's already approved.

ndt93:
I'm not sure if they should be stubbed out in all cases though esp. in `guitests`, so letting them get stubbed out where they need to be might be a more explicit option?

dariusf:
Good point, I was only thinking of unit tests. Maybe we could find a way to cover all the stubbing out with a more general rule.

PR #1222, opened 2016-01-30
Change repoId representation from raw string to object for safe comparison #1078

### LINKED ISSUE #1078: Change repoId representation from raw string to object for safe comparison

garbanzos:
Fixes #1078

- [x] write RepoID class
- [x] write RepoIDTest class 
- [ ] replace string to object


src/main/java/backend/RepoID.java
```diff
@@ -0,0 +1,53 @@
...
+    private String repoName;
+
+    public static Optional<RepoID> getRepoIDInstance(String repoIDString){
+        String lowerCaseRepoIDString = repoIDString.toLowerCase();
+        String[] repoIDComponents = lowerCaseRepoIDString.split("/");
+        if(repoIDComponents.length == 2){
+            RepoID repoID = new RepoID(lowerCaseRepoIDString, repoIDComponents[0], repoIDComponents[1]);
+            return Optional.of(repoID);
+        }
+        else{
```
damithc: Coding standard violation

src/main/java/backend/RepoID.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public int hashCode() {
+        return repoIDString.hashCode();
+    }
+
+    private RepoID(String lowerCaseRepoIDString, String repoOwner, String repoName){
+        this.repoIDString = lowerCaseRepoIDString;
+        this.repoOwner = repoOwner;
+        this.repoName = repoName;
+    }
```
damithc: Be sure to take the subtleties of equals method into account e.g. http://stackoverflow.com/questions/27581/what-issues-should-be-considered-when-overriding-equals-and-hashcode-in-java

damithc:
@garbanzos post a comment when the code is ready for review.

src/main/java/backend/RepoID.java
```diff
@@ -0,0 +1,65 @@
+package backend;
+
+import util.Utility;
+
+import java.util.Optional;
+
+public final class RepoID {
```
damithc: Add header comments to every class, and every non trivial public method.

garbanzos:
The classes RepoID and RepoIDTest is ready for review.

ndt93:
@acruis Can you help review this? It's not completed yet at the moment but some sanity check along the way would be good.

src/main/java/backend/RepoID.java
```diff
@@ -0,0 +1,68 @@
...
+import util.Utility;
+
+import java.util.Optional;
+
+/**
+ * Represents the ID of a repository. To be used to ensure safe comparison
+ */
+public final class RepoID {
+
+    private final String repoIDString;
```
dariusf: We can compute this from the other two fields. Maybe we could use a getter that appends them instead of storing this separately?

src/main/java/backend/RepoID.java
```diff
@@ -0,0 +1,68 @@
...
+    private final String repoOwner;
+    private final String repoName;
+
+    /**
+     * Return an Optional with a new instance of repoID if repoIDString is valid
+     * @param repoIDString
+     * @return Optional<RepoID>
+     */
+    public static Optional<RepoID> createRepoIDInstance(String repoIDString){
```
dariusf: It might be better to indicate the error with an exception instead of an Optional. The latter is generally for indicating that a value can be absent for reasons callers cannot control, and requiring that they handle that case. Here, it's the caller's responsibility to provide a valid string. We can reasonably expect that, so we don't have to make every caller go through the trouble of unwrapping the Optional.

Also, the name is a bit long. Maybe just `create`?

src/main/java/backend/RepoID.java
```diff
@@ -0,0 +1,68 @@
...
+        String lowerCaseRepoIDString = repoIDString.toLowerCase();
+        boolean isValidRepoIDString = Utility.isWellFormedRepoId(repoIDString);
```
dariusf: Maybe we could move this utility method in here? This class can handle everything related to repo ids.

Also, it might be useful to have a method for parsing repo ids, and define isWellFormedRepoId in terms of that. More DRY.

src/main/java/backend/RepoID.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public RepoID(String repoIDString) throws IllegalArgumentException {
+        boolean isValidRepoID = RepoID.isWellFormedRepoId(repoIDString);
+        if (isValidRepoID) {
+            String[] repoIDComponents = repoIDString.toLowerCase().split("/");
+            this.repoOwner = repoIDComponents[0];
+            this.repoName = repoIDComponents[1];
+        } else {
+            throw new IllegalArgumentException();
+        }
```
damithc: Reversing the condition will reduce nesting. This is also what John (from Code Bender) when he said 'don't use else' in his talk.
```java
if (!isValidRepoID) {
    throw new IllegalArgumentException();
}
```

src/main/java/backend/RepoID.java
```diff
@@ -0,0 +1,83 @@
...
+        boolean isValidRepoID = RepoID.isWellFormedRepoId(repoOwner, repoName);
+        if (isValidRepoID) {
+            this.repoOwner = repoOwner.toLowerCase();
+            this.repoName = repoName.toLowerCase();
+        } else {
+            throw new IllegalArgumentException();
+        }
+    }
+
+    public static boolean isWellFormedRepoId(String owner, String repo) {
```
damithc: Add a header comment and use it to define what 'well-formed` means.

damithc: This method doesn't seem to be doing what it advertises. If it was, it shouldn't need the `owner` parameter.

src/test/java/tests/RepoIDTest.java
```diff
@@ -0,0 +1,81 @@
+package tests;
+
+import backend.RepoID;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class RepoIDTest {
```
damithc: Our convention (although it is broken in many places) is to use strict camel case. i.e. `RepoIdTest`

acruis:
Sorry, I missed this PR. I'll review it now, do leave a comment when you're done.

src/main/java/backend/RepoID.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+public final class RepoID {
+
+    private final String repoOwner;
+    private final String repoName;
+
+    /**
+     * Constructor
+     * @param repoIDString repoOwner and repoName separated by slash
+     * @throws IllegalArgumentException If repoISString is invalid
```
acruis: Misspelling of `repoIDString`.

src/main/java/backend/RepoID.java
acruis: What @damithc means is, the parameter names are ambiguous. You can either clarify them in the method's javadoc comments, or rename them.

damithc: Ok, I realized later what the `owner` parameter is for. The missing header comment could have avoided my confusion.

src/main/java/backend/RepoID.java
```diff
@@ -0,0 +1,83 @@
...
+    public static boolean isWellFormedRepoId(String owner, String repo) {
+        return !(owner == null || owner.isEmpty() || repo == null || repo.isEmpty())
+                && isWellFormedRepoId(RepositoryId.create(owner, repo).generateId());
+    }
+
+    public static boolean isWellFormedRepoId(String repoId) {
+        RepositoryId repositoryId = RepositoryId.createFromId(repoId);
+        return repoId != null && !repoId.isEmpty() && repositoryId != null
+                && repoId.equals(repositoryId.generateId());
+    }
```
damithc: I can see you trying to reuse one method in the other, which is good. But the logic seem to be complicated than necessary. I can't comment further because you haven't defined what 'well-formed' means. I suspect you can use a utility method `isNonEmptyString(String)` inside these methods.

garbanzos:
The classes RepoId and RepoIdTest are ready for review again.

src/main/java/backend/RepoId.java
```diff
@@ -0,0 +1,100 @@
...
+        boolean isWellFormed = RepoId.isWellFormedRepoIdString(repoIdString);
+        throwExceptionIfNotWellFormed(isWellFormed);
+
+        String[] repoIDComponents = repoIdString.toLowerCase().split("/");
+        this.repoOwner = Utility.removeAllWhitespace(repoIDComponents[0]);
+        this.repoName = Utility.removeAllWhitespace(repoIDComponents[1]);
+    }
+
+    /**
+     * Constructor
```
damithc: Redundant comment

src/main/java/backend/RepoId.java
```diff
@@ -0,0 +1,100 @@
...
+     * @throws IllegalArgumentException If repoOwner or repoName is invalid
+     */
+    public RepoId(String repoOwner, String repoName) throws IllegalArgumentException {
+        boolean isWellFormed = RepoId.isWellFormedRepoIdString(repoOwner, repoName);
+        throwExceptionIfNotWellFormed(isWellFormed);
+
+        this.repoOwner = Utility.removeAllWhitespace(repoOwner.toLowerCase());
+        this.repoName = Utility.removeAllWhitespace(repoName.toLowerCase());
+
+    }
```
damithc: To reduce code (and duplication), you can remove this constructor altogether. i.e. have only one constructor that takes the full repo name as a single string.

src/main/java/backend/RepoId.java
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * Check if repoOwner and repoName is non-null and non-empty
```
damithc: Should be `Checks if repoOwner ...`

src/main/java/backend/RepoId.java
```diff
@@ -0,0 +1,100 @@
...
+     * and able to generate a well formed repoIdString of form <repoOwner>/<repoName>
+     * @param repoOwner Owner of the repo
+     * @param repoName Name of the repo
+     * @return True if the above conditions are met else False is returned
+     */
+    public static boolean isWellFormedRepoIdString(String repoOwner, String repoName) {
+        return !(repoOwner == null || repoOwner.isEmpty() || repoName == null || repoName.isEmpty())
+                && isWellFormedRepoIdString(RepositoryId.create(repoOwner, repoName).generateId());
+    }
```
damithc: This needs to be public?

damithc: Can you explain the logic here? Seems convoluted.

src/main/java/backend/RepoId.java
```diff
@@ -0,0 +1,100 @@
...
+     * Check if repoIdString non-null, non-empty
+     * and is of form <repoOwner>/<repoName>
+     * @param repoIdString
+     * @return True if the above conditions are met else False is returned
+     */
+    public static boolean isWellFormedRepoIdString(String repoIdString) {
+        RepositoryId repositoryId = RepositoryId.createFromId(repoIdString);
+        return repoIdString != null && !repoIdString.isEmpty() && repositoryId != null
+                && repoIdString.equals(repositoryId.generateId());
+    }
```
damithc: Again, get rid of one of the overloaded methods to reduce code and reduce duplication?

garbanzos:
@damithc 
The overloaded `isWellFormedRepoIdString` methods were previously in the `Utility` class and were used in the other classes like `RepositorySelector`. I moved them to the `RepoId` class because there was a suggestion that `RepoId` should do all the processing related to RepoIds.  

Since they are used in other classes, the `isWellFormedRepoIdString` methods are public for now. However, I think I will eventually change it to private after replacing raw strings of `RepoId` to the `RepoId` object. 

It is acceptable that let the two `isWellFormedRepoIdString` methods be public for now so that the transition can be smooth?

damithc:
Yes, ok to be public if they are used from elsewhere. Do we need two methods? Where the two-parameter is called now, we can call the one-parameter method using `(owner+"/"+repo)` right?

garbanzos:
@damithc I was thinking of deleting `isWellFormedRepoIdString(String repoOwner, String repoName)` eventually after the replacements. 

But I will take up your suggestion.

garbanzos:
RepoId is ready for review.

src/main/java/backend/LoginController.java
```diff
@@ -61,7 +61,8 @@ public void getPreviousLoginDetails() {
 
     public boolean attemptLogin() {
 
-        boolean validRepoId = Utility.isWellFormedRepoId(owner, repo) && !username.isEmpty() && !password.isEmpty();
+        boolean validRepoId = RepoId.isWellFormedRepoIdString(owner + "/" + repo)
```
damithc: Should be `isValidRepoId`

src/main/java/backend/LoginController.java
```diff
@@ -61,7 +61,8 @@ public void getPreviousLoginDetails() {
...
+                && !username.isEmpty() && !password.isEmpty();
         boolean loginSuccessful = login(username, password);
 
         if (validRepoId && loginSuccessful) {
```
damithc: Something is not quite right here. Does the validity of the repo ID matters to the login at all?

src/main/java/backend/RepoId.java
```diff
@@ -0,0 +1,68 @@
...
+
+import org.eclipse.egit.github.core.RepositoryId;
+import util.Utility;
+
+/**
+ * Represents the ID of a repository. To be used to ensure safe comparison
+ */
+public final class RepoId {
+
+    private final String repoOwner;
```
damithc: This can be org ID or username right? At least mention that in a comment?

src/main/java/backend/RepoId.java
```diff
@@ -0,0 +1,68 @@
...
+        String[] repoIDComponents = repoIdString.toLowerCase().split("/");
+        this.repoOwner = Utility.removeAllWhitespace(repoIDComponents[0]);
+        this.repoName = Utility.removeAllWhitespace(repoIDComponents[1]);
+    }
+
+    /**
+     * Checks if repoIdString non-null, non-empty
+     * and is of form <repoOwner>/<repoName>
+     * @param repoIdString
+     * @return True if the above conditions are met else False is returned
```
damithc: You can simplify the whole comment into this sentence 
`Returns true if the repoIdString is of the form <repoOwner>/<repoName> e.g. dave/foo-project`

src/main/java/backend/RepoId.java
```diff
@@ -0,0 +1,68 @@
...
+     */
+    public static boolean isWellFormedRepoIdString(String repoIdString) {
+        RepositoryId repositoryId = RepositoryId.createFromId(repoIdString);
```
damithc: Why do you need this?

src/test/java/tests/RepoIdTest.java
```diff
@@ -0,0 +1,79 @@
...
+        new RepoId("invalidTest");
+    }
+
+    @Test
+    public void isWellFormed() {
+        // Test not well formed
+        assertFalse(RepoId.isWellFormedRepoIdString(""));
+
+        // Test well formed
+        assertTrue(RepoId.isWellFormedRepoIdString("hubturbo/hubturbo"));
```
damithc: Check test coverage of the method to see if you have enough test cases here.

src/test/java/tests/RepoIdTest.java
```diff
@@ -0,0 +1,79 @@
...
+        // Test hashcode of objects initialised with different cases
+        assertEquals(repoId.hashCode(), repoId.hashCode());
+        assertEquals(repoId.hashCode(), repoIdAllCaps.hashCode());
+        assertEquals(repoId.hashCode(), repoIdMixed.hashCode());
+        assertEquals(repoIdAllCaps.hashCode(), repoIdMixed.hashCode());
+    }
+
+    @Test
+    public void getters(){
+        //repoIDString
```
damithc: The comment should be more detailed: What exact situation are you testing?

src/main/java/backend/RepoId.java
garbanzos: @damithc which method or line of code are you refering to?

src/main/java/backend/LoginController.java
garbanzos: @damithc I'm not sure about this part of the code as I didn't write the LogicController. (I only changed the `Utilty` to `RepoId` because `isWellFormedRepoIdString` was moved from the `Utility` class to the `RepoId` class).

But from what I see it is not checking if the validity repo ID (eg. whether it exists on github). It is really doing it checking if the 4 fields at login is non-empty. I guess I can refactor this to `isFieldsNotEmpty`.

src/main/java/backend/RepoId.java
dariusf: These `isWellFormed` methods delegate to EGit's repo id class to check validity, parse, etc. The intention of creating this class was partially to decouple us from the API code, so I guess we should implement our own checks.

damithc: That's what I guessed, but it's not clear from the code. I don't mind (in fact prefer) delegating to Egit if that works. Need to add comment to explain what's going on. Better variable names can help too. e.g. `idGeneratedByEgit`.

src/main/java/backend/RepoId.java
```diff
@@ -0,0 +1,68 @@
...
+        return repoIdString != null && !repoIdString.isEmpty() && repositoryId != null
+                && repoIdString.equals(repositoryId.generateId());
```
damithc: Is this check needed? `repoIdString != null && !repoIdString.isEmpty()`. Egit doesn't check for these two things? If not, shouldn't the check be done before delegating to Egit?

src/main/java/backend/LoginController.java
damithc: @dariusf what should be the purpose of this method? Do we have to check validity of repo id?

dariusf: @garbanzos is right about what the check essentially does.

That method is doing multiple things, and probably should only be concerned with supplying the login details to EGit. We can factor out the validation logic so it's checked before this point.

src/main/java/backend/RepoId.java
dariusf: I had a look inside the class and it actually does check. Guess we don't need them.

dariusf:
If we're okay with delegating all the work to EGit, this class exists for two reasons: equality ignoring case, and better interfaces. Could I suggest that we follow how EGit's [`RepositoryId`](https://github.com/eclipse/egit-github/blob/master/org.eclipse.egit.github.core/src/org/eclipse/egit/github/core/RepositoryId.java) is designed, and mostly just wrap its methods and constructors? We don't have to wrap the ones we don't use.

It can also implement `IRepositoryIdProvider`, which allows it to be used directly with EGit classes.

damithc:
Yes, I'm OK with delegating to EGit as much as we can.

dariusf:
Could we aim to complete this by next week's release? @garbanzos if you're not sure about the approach, feel free to ask!

garbanzos:
RepoId is ready for review. 
@dariusf Should I go ahead and change repoId representation from raw string to the RepoId object?

src/main/java/backend/RepoId.java
```diff
@@ -0,0 +1,69 @@
...
+ * Represents the ID of a repository. To be used to ensure safe comparison
+ */
+public final class RepoId {
+
+    //repoOwner can be an org ID or username
+    private final String repoOwner;
+    private final String repoName;
+
+    public RepoId(String repoIdString) throws IllegalArgumentException {
+        throwExceptionIfNotWellFormed(repoIdString);
```
dariusf: Don't think we need the `throws` declaration here, since it's a runtime exception.

Also, in general we can leave out the types of variables from their names, since that's meta-information, and it's usually apparent from the declaration anyway. :stuck_out_tongue:

src/main/java/backend/RepoId.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * Delegates to EGit's RepositoryId to check if repoString is
+     * of the correct format 
+     * @param repoIdString
+     * @return true if the repoIdString is of the form
+     * <repoOwner>/<repoName> e.g. dave/foo-project
+     */
+    public static boolean isWellFormedRepoIdString(String repoIdString) {
```
dariusf: Could this be isValidRepoId? Was suggested above.

src/main/java/backend/RepoId.java
```diff
@@ -0,0 +1,69 @@
...
+        RepositoryId idGeneratedByEgit = RepositoryId.createFromId(repoIdString);
+        return idGeneratedByEgit != null && repoIdString.equals(idGeneratedByEgit.generateId());
```
dariusf: Is it possible that RepositoryId.create* returns null?

src/main/java/backend/RepoId.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public String getRepoOwner() {
+        return repoOwner;
+    }
+
+    public String getRepoName() {
+        return repoName;
+    }
+
+    public String getRepoIdString() {
```
dariusf: Maybe override toString? Then we don't need this.

dariusf:
I actually think it might be useful to have a two-argument constructor, mostly so client code won't have to concatenate strings, and be coupled to a representation (not that it'll ever change, but still...). Maybe we could revisit it if the two-argument use case is prevalent in existing code?

Anyway I'm okay with the current interface. @damithc @acruis

src/main/java/backend/RepoId.java
```diff
@@ -0,0 +1,69 @@
...
+
+import org.eclipse.egit.github.core.RepositoryId;
+import util.Utility;
+
+/**
+ * Represents the ID of a repository. To be used to ensure safe comparison
+ */
+public final class RepoId {
+
+    //repoOwner can be an org ID or username
```
damithc: Use proper javadoc comment format?

src/main/java/backend/RepoId.java
```diff
@@ -0,0 +1,69 @@
...
+    private final String repoOwner;
+    private final String repoName;
+
+    public RepoId(String repoIdString) throws IllegalArgumentException {
+        throwExceptionIfNotWellFormed(repoIdString);
+
+        String[] repoIDComponents = repoIdString.toLowerCase().split("/");
```
damithc: If the values are kept in lower case, this behavior needs to be documented in the header comment of class/method.

damithc:
I'm not against having a two parameter constructor if there isn't too much duplication, but let's try to do without one first. If there's too much concatenation, we can add an extra constructor.

acruis:
I'm OK either way, but I'm sure by the end of this PR, we can tell whether repoID concatenation is done in many places.

dariusf:
How's this going? It's been open for a while.

dariusf:
Any progress on this? If you need help do ask.

src/main/java/backend/RepoId.java
garbanzos: Yes `RepositoryId.createFromId` returns null when the `repoId` passed is invalid.

garbanzos:
Closed due to too many merge conflicts.

PR #1224, opened 2016-01-31
Prompt user to save current board if there are unsaved changes #1199

xinan:
Fixes #1198
Fixes #1199 

1. Save the initial panel info when a board is opened. 
2. When switching to another board, check if the current panel info is from the initial one. If so, prompt for save.

PR #1225, opened 2016-01-31
Ctrl+B doesn't work when the open board is deleted #1198

### LINKED ISSUE #1198: Ctrl+B doesn't work when the open board is deleted
`Ctrl+B` stops switching boards when the currently open board is deleted.

garbanzos:
Fixes #1198

### DIFF
--- a/src/main/java/prefs/Preferences.java
+++ b/src/main/java/prefs/Preferences.java
@@ -118,6 +118,16 @@ public class Preferences { // NOPMD
     public void removeBoard(String name) {
         global.removeBoard(name);
     }
+
+    public void setLastDeletedIndex(String deletedBoardName){
+        List<String> boardNames = getAllBoardNames();
+        int index = boardNames.indexOf(deletedBoardName);
+        global.setLastDeletedBoardIndex(index);
+    }
+
+    public Optional<Integer> getLastDeletedBoardIndex() {
+        return global.getLastDeletedBoardIndex();
+    }
     
     public void setLastOpenBoard(String board) {
         global.setLastOpenBoard(board);
@@ -128,15 +138,18 @@ public class Preferences { // NOPMD
     }
     
     public Optional<String> switchBoard() {
+
         List<String> boardNames = getAllBoardNames();
         if (getLastOpenBoard().isPresent() && getAllBoardNames().size() > 1) {
             int lastBoard = boardNames.indexOf(getLastOpenBoard().get());
             int index = (lastBoard + 1) % boardNames.size();
             
             setLastOpenBoard(boardNames.get(index));
-        } else if (!getLastOpenBoard().isPresent() && getAllBoardNames().size() > 1) {
+        } else if (getLastDeletedBoardIndex().isPresent() && getAllBoardNames().size() > 1) {
+            int lastDeletedBoardIndex = getLastDeletedBoardIndex().get();
+            int index = lastDeletedBoardIndex % boardNames.size();
 
-            setLastOpenBoard(boardNames.get(0));
+            setLastOpenBoard(boardNames.get(index));
         }
         
         return getLastOpenBoard();
--- a/src/main/java/prefs/Preferences.java
+++ b/src/main/java/prefs/Preferences.java
@@ -128,12 +128,15 @@ public class Preferences { // NOPMD
     }
     
     public Optional<String> switchBoard() {
-        if (getLastOpenBoard().isPresent() && getAllBoards().size() > 1) {
-            List<String> boardNames = getAllBoardNames();
+        List<String> boardNames = getAllBoardNames();
+        if (getLastOpenBoard().isPresent() && getAllBoardNames().size() > 1) {
             int lastBoard = boardNames.indexOf(getLastOpenBoard().get());
             int index = (lastBoard + 1) % boardNames.size();
             
             setLastOpenBoard(boardNames.get(index));
+        } else if (!getLastOpenBoard().isPresent() && getAllBoardNames().size() > 1) {
+
+            setLastOpenBoard(boardNames.get(0));
         }
         
         return getLastOpenBoard();
--- a/src/main/java/prefs/Preferences.java
+++ b/src/main/java/prefs/Preferences.java
@@ -138,21 +138,32 @@ public class Preferences { // NOPMD
     }
     
     public Optional<String> switchBoard() {
-
-        List<String> boardNames = getAllBoardNames();
-        if (getLastOpenBoard().isPresent() && getAllBoardNames().size() > 1) {
-            int lastBoard = boardNames.indexOf(getLastOpenBoard().get());
-            int index = (lastBoard + 1) % boardNames.size();
-            
-            setLastOpenBoard(boardNames.get(index));
-        } else if (getLastDeletedBoardIndex().isPresent() && getAllBoardNames().size() > 1) {
-            int lastDeletedBoardIndex = getLastDeletedBoardIndex().get();
-            int index = lastDeletedBoardIndex % boardNames.size();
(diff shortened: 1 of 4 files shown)


garbanzos:
When a board is deleted, it becomes a "none" board which is not considered an open board. 
Hence, I fixed this issue by allowing one to switch to the first board in the board list even if there is no board open (and if there is a saved board to switch to). However, this fails the noBoardOpenSwitchTest() in BoardSwitchTest. 

It seems that the test implies that you should not be able to switch boards when there is no board open. May I know the purpose of this?

damithc:
My view is that it should be allowed. I can't see a reason not to.

garbanzos:
@damithc 
`Ctrl+B` may have 2 different behaviours when it is pressed in this case. 

**Behaviour 1**
When the open board is deleted and `Ctrl+B` is pressed, HubTurbo will switch to the first board on the board lists. 

For example, there may be 4 boards on the list: 
- board1 
- board2
- board3 
- board4 
If board2 was open and deleted, `Ctrl+B` will switch to board1. 

**Behaviour 2** 
When the open board is deleted and `Ctrl+B` is pressed, HubTurbo will switch to the board that was right after the deleted board in the board list. 

For example, there may be 4 boards on the list: 
- board1 
- board2
- board3 
- board4 
If board2 was open and deleted, `Ctrl+B` will switch to board3. 

I have already implemented behaviour 1. However, behaviour 2 seems to be more useful.
Which behaviour is the intended behaviour of HubTurbo?

damithc:
The 2nd option seems more in line with normal behavior of Ctrl+B. You can do in this PR or use option 1 in this PR and create a new PR for upgrading the behavior to option 2.

src/main/java/prefs/Preferences.java
```diff
@@ -128,12 +128,15 @@ public void setLastOpenBoard(String board) {
     }
     
     public Optional<String> switchBoard() {
-        if (getLastOpenBoard().isPresent() && getAllBoards().size() > 1) {
-            List<String> boardNames = getAllBoardNames();
+        List<String> boardNames = getAllBoardNames();
+        if (getLastOpenBoard().isPresent() && getAllBoardNames().size() > 1) {
```
damithc: Use an explanatory variable to clarify the intention of complicated conditions.

src/main/java/prefs/Preferences.java
```diff
@@ -128,12 +128,15 @@ public void setLastOpenBoard(String board) {
...
             int lastBoard = boardNames.indexOf(getLastOpenBoard().get());
             int index = (lastBoard + 1) % boardNames.size();
             
             setLastOpenBoard(boardNames.get(index));
+        } else if (!getLastOpenBoard().isPresent() && getAllBoardNames().size() > 1) {
```
damithc: same here

damithc:
Any change to functional code should come with changes to test code. Add tests if they are missing.

dariusf:
How's this going? Need help?

garbanzos:
Ready for review

src/main/java/prefs/Preferences.java
```diff
@@ -118,6 +118,16 @@ public void addBoard(String name, List<PanelInfo> panels) {
     public void removeBoard(String name) {
         global.removeBoard(name);
     }
+
+    public void setLastDeletedIndex(String deletedBoardName){
```
damithc: Add header comments. 
This method name is not specific enough as it doesn't mention Boards.

garbanzos: @damithc Do I still need to add header comments if i refactor it to `setLastDeletedBoardIndex`?

damithc: Can omit header comment if the method signature is explanatory enough (i.e. everything you are going to say in the comment is already contained in the method signature), which is usually the case for the most trivial methods only.

src/main/java/prefs/Preferences.java
```diff
@@ -128,12 +138,18 @@ public void setLastOpenBoard(String board) {
     }
     
     public Optional<String> switchBoard() {
-        if (getLastOpenBoard().isPresent() && getAllBoards().size() > 1) {
-            List<String> boardNames = getAllBoardNames();
+
+        List<String> boardNames = getAllBoardNames();
+        if (getLastOpenBoard().isPresent() && getAllBoardNames().size() > 1) {
```
damithc: Introduce a local variable to explain the meaning of the boolean condition.

dariusf:
Do comment when this is ready for review again.

garbanzos:
Ready for review

garbanzos:
Ready for review

src/test/java/tests/BoardSwitchTest.java
```diff
@@ -45,6 +45,31 @@ public void boardsSwitchTest() {
     }
 
     @Test
+    public void deletedBoardOpenSwitchTest() {
```
damithc: Follow the new naming convention for tests (refer slack conversation)

src/main/java/prefs/Preferences.java
```diff
@@ -128,15 +138,32 @@ public void setLastOpenBoard(String board) {
...
-        if (getLastOpenBoard().isPresent() && getAllBoards().size() > 1) {
-            List<String> boardNames = getAllBoardNames();
+        setBoardToSwitch();
+        return getLastOpenBoard();
+    }
+
+    /**
+     * Sets the lastOpenBoard to the board to switch to
+     */
+    private void setBoardToSwitch(){
```
damithc: Board switching should not even be in this class! :-o This class is for managing preferences!! Oh man. @garbanzos I know you are not the one who put it there, but can you refactor board switching out of here? @dariusf can you advise @garbanzos on the direction to take? It should be similar to label picker, but doesn't have a UI at the moment.

dariusf: Here's a rough design I would use.

1. `UI` would contain a `Boards` class, which can be serialised. It keeps track of board state (which is open, unsaved changes, etc.) and has high-level operations like creating a new board, etc. It isn't concerned with how boards are persisted.
2. We represent boards directly with a `Board` class, which has a name, `List<String>` of filters, etc. and is also meant to be serialised.
3. `Boards` knows about `PanelControl` (to create panels), and `MenuControl` knows about `Boards` (as a UI to board operations). We can look into decoupling them in future, but right now it doesn't seem worth the trouble of adding a lot of new events/callbacks/interfaces.
4. `Preferences` knows about `Boards` and decides when and how to serialise it. `Boards` controls what to serialise. Not clear on the details here as preferences are being changed too.

@garbanzos feel free to improve on this!

There are quite a few related PRs now, which might lead to conflicts. #1255 (board equality) and #1224 (unsaved changes) depend on this now. #1272 will change how preferences work, and #1262 changes the related tests. We can continue to discuss the redesign, but maybe we should defer implementation until a few more of the rest are done?

damithc: Yes, we can delay a implementation while conflicting PRs get through (hopefully, respective devs will push them through quickly).
By 'serialize' you mean convert to json right?

dariusf: Yes.

garbanzos: @dariusf So I wait for the other PRs to be merged before the refacor?

dariusf: Only the preferences one is unmerged and making progress, so I guess you can go ahead. You can look at the PR to see what the interface might be like. Probably better for the others to adapt to the new design here anyway.

In any case, do try to get feedback early and iterate on the design!

garbanzos:
Closed due to too many merge conflicts.

PR #1226, opened 2016-01-31
If there is only one open milestone, consider it as the current milestone #1147

### LINKED ISSUE #1147: If there is only one open milestone, consider it as the current milestone
If there is only one milestone open, it can be considered the `current` milestone even if it doesn't have a date assigned to it. i.e.alias `curr` should resolve to that milestone

raisfathin:
Fixes #1147

### DIFF
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -153,7 +153,7 @@ Aliases: `desc`, `description`
 
 Matches all issues associated with any milestones whose names contain the given string.
 
-`current` or `curr` can be used to refer to an open milestone with earliest due date, or an overdue open milestone with open issues if any. `curr-2`, `curr-1`, `curr+1`, `curr+2`, etc. (no space before or after `-`/`+`) can then be used to refer to milestones before or after `current`, sorted by due date. Milestones with no due date will not be considered in resolving `curr`/`current`. If there is no open milestone, `current` will refer to no milestone, while `current-1` will refer to the last closed milestone.
+`current` or `curr` can be used to refer to an open milestone with earliest due date, or an overdue open milestone with open issues if any. However, if there is only one open milestone, it will be considered as the `current` milestone even if it does not have due date. `curr-2`, `curr-1`, `curr+1`, `curr+2`, etc. (no space before or after `-`/`+`) can then be used to refer to milestones before or after `current`, sorted by due date. Milestones with no due date will not be considered in resolving `curr`/`current`. If there is no open milestone, `current` will refer to no milestone, while `current-1` will refer to the last closed milestone.
 
 Aliases: `m`
 
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -153,7 +153,7 @@ Aliases: `desc`, `description`
 
 Matches all issues associated with any milestones whose names contain the given string.
 
-`current` or `curr` can be used to refer to an open milestone with earliest due date, or an overdue open milestone with open issues if any. However, if there is only one open milestone, it will be considered as the `current` milestone even if it does not have due date. `curr-2`, `curr-1`, `curr+1`, `curr+2`, etc. (no space before or after `-`/`+`) can then be used to refer to milestones before or after `current`, sorted by due date. Milestones with no due date will not be considered in resolving `curr`/`current`. If there is no open milestone, `current` will refer to no milestone, while `current-1` will refer to the last closed milestone.
+`current` or `curr` can be used to refer to an open milestone with earliest due date, or an overdue open milestone with open issues if any. However, if there is only one open milestone, it will be considered as the `current` milestone even if it does not have due date. `curr-2`, `curr-1`, `curr+1`, `curr+2`, etc. (no space before or after `-`/`+`) can then be used to refer to milestones before or after `current`, sorted by due date. If there is no open milestone, `current` will refer to no milestone, while `current-1` will refer to the last closed milestone.
 
 Aliases: `m`
 
--- a/src/main/java/backend/resource/TurboMilestone.java
+++ b/src/main/java/backend/resource/TurboMilestone.java
@@ -5,7 +5,9 @@ import org.eclipse.egit.github.core.Milestone;
 import util.Utility;
 
 import java.time.LocalDate;
+import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 @SuppressWarnings("unused")
 public class TurboMilestone {
@@ -81,6 +83,33 @@ public class TurboMilestone {
     }
 
     private void ______METHODS______() {
+    }
+
+    public boolean isOverdue() {
+        return dueDate.isPresent() && dueDate.get().isBefore(LocalDate.now());
+    }
+
+    public boolean hasOpenIssues() {
+        return openIssues > 0;
+    }
+
+    /**
+     * A milestone is ongoing if it is not due yet or it is overdue but still has open issues.
+     */
+    public boolean isOngoing() {
+        return !isOverdue() || hasOpenIssues();
+    }
+
+    public static List<TurboMilestone> getMilestonesOfRepos(List<TurboMilestone> milestoneList, List<String> repoIds) {
+        return milestoneList.stream()
+                .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
+                .collect(Collectors.toList());
+    }
+
+    public static List<TurboMilestone> getOpenMilestones(List<TurboMilestone> milestoneList) {
+        return milestoneList.stream()
+                .filter(TurboMilestone::isOpen)
+                .collect(Collectors.toList());
     }
 
     @Override
--- a/src/main/java/backend/resource/TurboMilestone.java
+++ b/src/main/java/backend/resource/TurboMilestone.java
@@ -101,7 +101,7 @@ public class TurboMilestone {
         return isOpen() && (!isOverdue() || hasOpenIssues());
     }
 
-    public static List<TurboMilestone> filterMilestonesOfGivenRepos(List<TurboMilestone> milestones, List<String> repoIds) {
+    public static List<TurboMilestone> filterMilestonesOfGivenRepoIds(List<TurboMilestone> milestones, List<String> repoIds) {
         return milestones.stream()
                 .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
                 .collect(Collectors.toList());
--- a/src/main/java/backend/resource/TurboMilestone.java
+++ b/src/main/java/backend/resource/TurboMilestone.java
@@ -108,7 +108,7 @@ public class TurboMilestone {
                 .collect(Collectors.toList());
     }
 
-    public static List<TurboMilestone> getOpenMilestones(List<TurboMilestone> milestones) {
+    public static List<TurboMilestone> filterOpenMilestones(List<TurboMilestone> milestones) {
         return milestones.stream()
                 .filter(TurboMilestone::isOpen)
                 .collect(Collectors.toList());
@@ -127,6 +127,11 @@ public class TurboMilestone {
         };
     }
 
+    /**
+     * Sort a List<TurboMilestone> by due date. Milestones without due date are considered to
+     * have an imaginary due date in the far future. The sorting algorithm used is stable
+     * (i.e. relative ordering of 2 milestones with the same due date will be retained)
+     */
     public static List<TurboMilestone> sortByDueDate(List<TurboMilestone> milestones) {
         List<TurboMilestone> milestonesWithDueDate = milestones.stream()
                 .filter(ms -> ms.getDueDate().isPresent())
--- a/src/main/java/backend/resource/TurboMilestone.java
+++ b/src/main/java/backend/resource/TurboMilestone.java
@@ -95,13 +95,14 @@ public class TurboMilestone {
     }
 
     /**
-     * A milestone is ongoing if it is not due yet or it is overdue but still has open issues.
+     * A milestone is ongoing if it is open and not due yet or if it is overdue but still open and has open issues.
      */
     public boolean isOngoing() {
         return isOpen() && (!isOverdue() || hasOpenIssues());
     }
 
-    public static List<TurboMilestone> filterMilestonesOfGivenRepoIds(List<TurboMilestone> milestones, List<String> repoIds) {
+    public static List<TurboMilestone> filterMilestonesOfGivenRepoIds(List<TurboMilestone> milestones,
+                                                                      List<String> repoIds) {
         return milestones.stream()
                 .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
                 .collect(Collectors.toList());
--- a/src/main/java/backend/resource/TurboMilestone.java
+++ b/src/main/java/backend/resource/TurboMilestone.java
@@ -98,7 +98,7 @@ public class TurboMilestone {
      * A milestone is ongoing if it is not due yet or it is overdue but still has open issues.
      */
     public boolean isOngoing() {
-        return !isOverdue() || hasOpenIssues();
+        return isOpen() && (!isOverdue() || hasOpenIssues());
     }
 
     public static List<TurboMilestone> getMilestonesOfRepos(List<TurboMilestone> milestones, List<String> repoIds) {
--- a/src/main/java/backend/resource/TurboMilestone.java
+++ b/src/main/java/backend/resource/TurboMilestone.java
@@ -5,6 +5,7 @@ import org.eclipse.egit.github.core.Milestone;
 import util.Utility;
 
 import java.time.LocalDate;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -100,16 +101,42 @@ public class TurboMilestone {
         return !isOverdue() || hasOpenIssues();
     }
 
-    public static List<TurboMilestone> getMilestonesOfRepos(List<TurboMilestone> milestoneList, List<String> repoIds) {
-        return milestoneList.stream()
+    public static List<TurboMilestone> getMilestonesOfRepos(List<TurboMilestone> milestones, List<String> repoIds) {
+        return milestones.stream()
                 .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
                 .collect(Collectors.toList());
     }
 
-    public static List<TurboMilestone> getOpenMilestones(List<TurboMilestone> milestoneList) {
-        return milestoneList.stream()
+    public static List<TurboMilestone> getOpenMilestones(List<TurboMilestone> milestones) {
+        return milestones.stream()
                 .filter(TurboMilestone::isOpen)
                 .collect(Collectors.toList());
+    }
+
+    /**
+     * Condition: milestone must have due dates
+     */
+    public static Comparator<TurboMilestone> getDueDateComparator() {
+        return (a, b) -> {
+            assert a.getDueDate().isPresent();
+            assert b.getDueDate().isPresent();
+            LocalDate aDueDate = a.getDueDate().get();
+            LocalDate bDueDate = b.getDueDate().get();
+            return aDueDate.compareTo(bDueDate);
+        };
+    }
+
+    public static List<TurboMilestone> getSortedMilestonesByDueDate(List<TurboMilestone> milestones) {
+        List<TurboMilestone> milestonesWithDueDate = milestones.stream()
+                .filter(ms -> ms.getDueDate().isPresent())
+                .sorted(getDueDateComparator())
+                .collect(Collectors.toList());
+        List<TurboMilestone> milestonesWithoutDueDate = milestones.stream()
+                .filter(ms -> !ms.getDueDate().isPresent())
+                .collect(Collectors.toList());
+        List<TurboMilestone> result = milestonesWithDueDate;
+        result.addAll(milestonesWithoutDueDate);
+        return result;
     }
 
     @Override
--- a/src/main/java/backend/resource/TurboMilestone.java
+++ b/src/main/java/backend/resource/TurboMilestone.java
@@ -101,7 +101,7 @@ public class TurboMilestone {
         return isOpen() && (!isOverdue() || hasOpenIssues());
     }
 
-    public static List<TurboMilestone> getMilestonesOfRepos(List<TurboMilestone> milestones, List<String> repoIds) {
+    public static List<TurboMilestone> filterMilestonesOfGivenRepos(List<TurboMilestone> milestones, List<String> repoIds) {
         return milestones.stream()
                 .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
                 .collect(Collectors.toList());
--- a/src/main/java/backend/resource/TurboMilestone.java
+++ b/src/main/java/backend/resource/TurboMilestone.java
@@ -101,8 +101,8 @@ public class TurboMilestone {
         return isOpen() && (!isOverdue() || hasOpenIssues());
     }
 
-    public static List<TurboMilestone> filterMilestonesOfGivenRepoIds(List<TurboMilestone> milestones,
-                                                                      List<String> repoIds) {
+    public static List<TurboMilestone> filterMilestonesOfRepos(List<TurboMilestone> milestones,
+                                                               List<String> repoIds) {
         return milestones.stream()
                 .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
                 .collect(Collectors.toList());
@@ -127,7 +127,7 @@ public class TurboMilestone {
         };
     }
 
-    public static List<TurboMilestone> getSortedMilestonesByDueDate(List<TurboMilestone> milestones) {
+    public static List<TurboMilestone> sortByDueDate(List<TurboMilestone> milestones) {
         List<TurboMilestone> milestonesWithDueDate = milestones.stream()
                 .filter(ms -> ms.getDueDate().isPresent())
                 .sorted(getDueDateComparator())
--- a/src/main/java/filter/expression/Qualifier.java
+++ b/src/main/java/filter/expression/Qualifier.java
@@ -164,15 +164,15 @@ public class Qualifier implements FilterExpression {
      * i.e. 0 for "current" milestone, -1 for "current-1" milestone, etc.
      */
     private static Map<Integer, TurboMilestone> getMilestoneAliasIndex(List<TurboMilestone> milestones) {
-        Collections.sort(milestones, getMilestoneDueDateComparator());
-        Optional<Integer> currentMilestoneIndex = getCurrentMilestoneIndex(milestones);
+        List<TurboMilestone> sortedMilestones = TurboMilestone.getSortedMilestonesByDueDate(milestones);
+        Optional<Integer> currentMilestoneIndex = getCurrentMilestoneIndex(sortedMilestones);
 
         return IntStream
-                .range(0, milestones.size())
+                .range(0, sortedMilestones.size())
                 .boxed()
                 .collect(Collectors.toMap(
                         i -> i - currentMilestoneIndex.get(),
-                        i -> milestones.get(i)));
+                        i -> sortedMilestones.get(i)));
     }
 
     private static Optional<Integer> getCurrentMilestoneIndex(List<TurboMilestone> milestones) {
@@ -180,7 +180,7 @@ public class Qualifier implements FilterExpression {
             return Optional.empty();
         }
 
-        long openMilestonesCount = milestones.stream().filter(ms -> ms.isOpen()).count();
+        long openMilestonesCount = milestones.stream().filter(TurboMilestone::isOpen).count();
         if (openMilestonesCount == 1) {
             return IntStream
                     .range(0, milestones.size())
@@ -188,7 +188,7 @@ public class Qualifier implements FilterExpression {
                     .filter(i -> milestones.get(i).isOpen())
                     .findFirst();
         } else {
-            // Look for the first milestone in the (sorted) list which is both open and ongoing.
+            // Look for the first milestone in the (sorted) list that is both open and ongoing.
             int currentIndex = 0;
             for (TurboMilestone checker : milestones) {
                 if (checker.isOpen() && checker.isOngoing()) {
@@ -575,7 +575,7 @@ public class Qualifier implements FilterExpression {
                         } else if (!bDueDate.isPresent()) {
                             return -1;
                         } else {
-                            return -(getMilestoneDueDateComparator()
+                            return -(TurboMilestone.getDueDateComparator()
                                     .compare(aMilestone.get(), bMilestone.get()));
                         }
                     }
@@ -1041,7 +1041,11 @@ public class Qualifier implements FilterExpression {
 
         int offset = 0;
         if (aliasMatcher.group(2) != null && aliasMatcher.group(3) != null) {
-            offset = Integer.parseInt(aliasMatcher.group(3));
+            if (aliasMatcher.group(2).equals("+")) {
+                offset = Integer.parseInt(aliasMatcher.group(3));
+            } else {
+                offset = -Integer.parseInt(aliasMatcher.group(3));
+            }
         }
 
         // if out of milestone range, don't convert alias
@@ -1052,20 +1056,6 @@ public class Qualifier implements FilterExpression {
         contents = milestoneAliasIndex.get(offset).getTitle().toLowerCase();
 
         return new Qualifier(type, contents);
-    }
-
-    /**
-     * Condition: milestone must have due dates
-     */
-    public static Comparator<TurboMilestone> getMilestoneDueDateComparator() {
-        return (a, b) -> {
-            assert a.getDueDate().isPresent();
-            assert b.getDueDate().isPresent();
-            LocalDate aDueDate = a.getDueDate().get();
-            LocalDate bDueDate = b.getDueDate().get();
-
-            return aDueDate.compareTo(bDueDate);
-        };
     }
 
     /**
--- a/src/main/java/filter/expression/Qualifier.java
+++ b/src/main/java/filter/expression/Qualifier.java
@@ -185,7 +185,7 @@ public class Qualifier implements FilterExpression {
             return Optional.empty();
         }
 
-        long openMilestonesCount = milestones.stream().filter(TurboMilestone::isOpen).count();
+        int openMilestonesCount = TurboMilestone.getOpenMilestones(milestones).size();
         if (openMilestonesCount == 1) {
             return IntStream
                     .range(0, milestones.size())
--- a/src/main/java/filter/expression/Qualifier.java
+++ b/src/main/java/filter/expression/Qualifier.java
@@ -92,8 +92,8 @@ public class Qualifier implements FilterExpression {
             repoIds.add(model.getDefaultRepo().toLowerCase());
         }
 
-        List<TurboMilestone> milestonesOfReposInPanel = TurboMilestone.filterMilestonesOfGivenRepoIds(
-                model.getMilestones(), repoIds);
+        List<TurboMilestone> milestonesOfReposInPanel = TurboMilestone.filterMilestonesOfRepos(
+                                                                                    model.getMilestones(), repoIds);
         List<TurboMilestone> milestonesWithinAliasRange = getMilestonesWithinAliasRange(milestonesOfReposInPanel);
         Map<Integer, TurboMilestone> milestoneAliasIndex = getMilestoneAliasIndex(milestonesWithinAliasRange);
 
@@ -136,6 +136,9 @@ public class Qualifier implements FilterExpression {
 
     /**
      * Get all milestones that can be aliased with current-[n] to current+[n]
+     *
+     * A milestone can be aliased if and only if it is open and has due date or it does not have due date but it is
+     * the only open milestone.
      */
     private static List<TurboMilestone> getMilestonesWithinAliasRange(List<TurboMilestone> milestonesOfReposInPanel) {
         List<TurboMilestone> milestones = new ArrayList<>();
@@ -165,7 +168,7 @@ public class Qualifier implements FilterExpression {
      * i.e. 0 for "current" milestone, -1 for "current-1" milestone, etc.
      */
     private static Map<Integer, TurboMilestone> getMilestoneAliasIndex(List<TurboMilestone> milestones) {
-        List<TurboMilestone> sortedMilestones = TurboMilestone.getSortedMilestonesByDueDate(milestones);
+        List<TurboMilestone> sortedMilestones = TurboMilestone.sortByDueDate(milestones);
         Optional<Integer> currentMilestoneIndex = getCurrentMilestoneIndex(sortedMilestones);
 
         assert currentMilestoneIndex.isPresent() || !currentMilestoneIndex.isPresent() && sortedMilestones.isEmpty();
--- a/src/main/java/filter/expression/Qualifier.java
+++ b/src/main/java/filter/expression/Qualifier.java
@@ -92,7 +92,7 @@ public class Qualifier implements FilterExpression {
             repoIds.add(model.getDefaultRepo().toLowerCase());
         }
 
-        List<TurboMilestone> milestonesOfReposInPanel = TurboMilestone.filterMilestonesOfGivenRepos(model.getMilestones(), repoIds);
+        List<TurboMilestone> milestonesOfReposInPanel = TurboMilestone.filterMilestonesOfGivenRepoIds(model.getMilestones(), repoIds);
         List<TurboMilestone> milestonesWithinAliasRange = getMilestonesWithinAliasRange(milestonesOfReposInPanel);
         Map<Integer, TurboMilestone> milestoneAliasIndex = getMilestoneAliasIndex(milestonesWithinAliasRange);
 
@@ -166,6 +166,8 @@ public class Qualifier implements FilterExpression {
     private static Map<Integer, TurboMilestone> getMilestoneAliasIndex(List<TurboMilestone> milestones) {
         List<TurboMilestone> sortedMilestones = TurboMilestone.getSortedMilestonesByDueDate(milestones);
         Optional<Integer> currentMilestoneIndex = getCurrentMilestoneIndex(sortedMilestones);
+
+        assert currentMilestoneIndex.isPresent() || (!currentMilestoneIndex.isPresent() && sortedMilestones.isEmpty());
 
         return IntStream
                 .range(0, sortedMilestones.size())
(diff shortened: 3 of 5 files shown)


src/main/java/filter/expression/Qualifier.java
```diff
@@ -97,6 +97,16 @@ public static FilterExpression replaceMilestoneAliases(IModel model, FilterExpre
...
 
+        List<TurboMilestone> openMilestones = model.getMilestones().stream()
+                .filter(ms -> ms.isOpen())
+                .collect(Collectors.toList());
+
+        if (openMilestones.size() == 1 && !openMilestones.get(0).getDueDate().isPresent()) {
+            //if there is only one open issue in the repo and it doesn't have a due date, add it
+            //to the list of all milestones
+            allMilestones.addAll(openMilestones);
+        }
```
damithc: Do we really need this? Besides, this code doesn't look like it belongs in this method. i.e. the method looks like it's doing more than one thing.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -160,6 +172,12 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
         return Optional.of(allMilestones.size());
     }
 
+    private static boolean isRelevantMilestone(TurboMilestone milestone) {
```
damithc: Add a comment to explain the purpose of the method.

src/main/java/filter/expression/Qualifier.java
raisfathin: I think we need it, because the case where a milestone without due date can be considered by the alias `current` is when there is no open milestone with due date, right? Is it the intended behaviour?

For the second part, I think we can just create another method to generate the `allMilestones` list.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,23 +131,55 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
         return exprWithNormalQualifiers.isSatisfiedBy(model, issue, new MetaQualifierInfo(metaQualifiers));
     }
 
+    // Get all milestones that can be aliased with the keyword "current".
```
damithc: Use javadoc comment format for header comments.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,23 +131,55 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+                .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
+                .sorted((a, b) -> getMilestoneDueDateComparator().compare(a, b))
+                .collect(Collectors.toList()));
+
+        // Special case: if there is only one open milestone, it should be included regardless of whether
+        // it has due date or not.
+        // In this case, if it is not included already (the open milestone does not
+        // have due date), append it to the milestone list.
+        List<TurboMilestone> openMilestones = getOpenMilestones(model, repoIds);
+        if (openMilestones.size() == 1 && !openMilestones.get(0).getDueDate().isPresent()) {
```
damithc: Add an explanatory variable to clarify the intent of complicated conditions.

raisfathin:
Ready for review.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,23 +131,61 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
         return exprWithNormalQualifiers.isSatisfiedBy(model, issue, new MetaQualifierInfo(metaQualifiers));
     }
 
+    /**
+     * Get all milestones that can be aliased with the keyword "current".
+     */
+    private static List<TurboMilestone> getMilestonesWithAlias(IModel model, List<String> repoIds) {
```
damithc: Hmm... the description doesn't match exactly with the method name.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,23 +131,61 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+            boolean hasDueDate = openMilestone.getDueDate().isPresent();
+            if (!hasDueDate) {
+                milestones.add(openMilestone);
+            }
+        }
+
+        return milestones;
+    }
+
+    private static List<TurboMilestone> getOpenMilestones(IModel model, List<String> repoIds) {
```
damithc: This method is doing more than advertised. I think you need two methods:
```java
getMilestonesOfRepos(List<TurboMilestone>, List<String> repoIds): List<TurboMilestone>
getOpenMilestones(List<TurboMilestone> milestones): List<TurboMilestone>
```
There is no need for this method to receive an `IModel` object. Should not receive more than needed. 
I wonder if both methods should be static methods in `TurboMilestone` class.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,23 +131,61 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+        return model.getMilestones().stream()
+                .filter(ms -> ms.isOpen())
+                .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
+                .collect(Collectors.toList());
+    }
+
     private static Optional<Integer> getCurrentMilestoneIndex(List<TurboMilestone> allMilestones) {
```
damithc: Why is the parameter named `allMilestones`? As far as this method is concerned, it is just a list of milestones. The 'all' part is not relevant.

damithc: What's the high level logic of this method? Can you state it in pseudo code, for my reference? It's not immediately apparent from the code, and that's a failure of SLAP.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -160,6 +193,15 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
         return Optional.of(allMilestones.size());
     }
 
+    /**
+     * Checks whether a milestone still has open issues or it is not due yet.
+     */
+    private static boolean isRelevantMilestone(TurboMilestone milestone) {
```
damithc: 'Relevant' to what? This method name cannot be explained without referring to the caller. That should not be the case. A method should be clear on its own.

src/main/java/filter/expression/Qualifier.java
HansNewbie: For the original one:

Input = milestones that can be aliased, sorted by due date
Output = index of `current` milestone within the given milestone list

Logic =

1. If there is no milestones that can be aliased, there is no `current` milestone hence return empty index.
2. Find first open milestone which is either:
  - overdue but still have open issues in it (if there is no open issues, assume the admin forgot to close the milestone), or
  - not overdue yet.
3. If no such open milestone, consider the last milestone in the list as the `current-1` (only open milestone can be current). In this case, return size of milestone list as `current` index (hence `current` returns no result, but `current-1)

HansNewbie: I mispressed my keyboard. Updated the logic of the method above.

src/main/java/filter/expression/Qualifier.java
HansNewbie: It's more like `isOpenMilestoneStillOngoing()`, where we consider `ongoing` as:

- not due yet, or
- overdue but still have open issues in it

src/main/java/filter/expression/Qualifier.java
HansNewbie: `getMilestonesWithinAliasRange()`? We would like to get all milestones that can be aliased with `current-[n]` to `current+[n]`.

src/main/java/filter/expression/Qualifier.java
HansNewbie: I agree with the comments. With `getMilestonesOfRepos`, we can use it to get all the milestones with due date above as well.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,23 +131,61 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
-            boolean relevant = !(overdue && checker.getOpenIssues() == 0);
+        Optional<TurboMilestone> openAndRelevantMilestone = allMilestones.stream()
+                .filter(ms -> ms.isOpen() && isRelevantMilestone(ms))
+                .findFirst();
 
-            if (checker.isOpen() && relevant) {
-                return Optional.of(currentIndex);
+        if (openAndRelevantMilestone.isPresent()) {
+            return Optional.of(allMilestones.indexOf(openAndRelevantMilestone.get()));
+        } else {
```
HansNewbie: The `else` case here seems to cover only when there is 1 open milestone with no due date. If that is the case, we can put this as a case above after `if (allMilestones.isEmpty())`. Something along `if (allMilestones.stream().filter(ms ->ms.isOpen()).count() == 1)`. The problem with this would be an open milestone that is forgotten to be closed (i.e. overdue but has no open issues in it) can be considered `current`.

If you would like to be more specific, since we would like to cover the case where `current` milestone is the only open milestone with no due date, our `if` case can be changed to check if there is any milestone with no due date in the list; if yes, return that as `current`. We can do that __if__ we expect the list to have at most 1 milestone with no due date. This makes the method to be overly specific but it works for our case.

src/main/java/filter/expression/Qualifier.java
damithc: `getAliasableMilestones`? What's the definition of aliasable? It needs to be defined somewhere.

src/main/java/filter/expression/Qualifier.java
damithc: Consider having the following methods in the `TurboMilestone` class. 
```java
boolean isOverdue()
boolean hasOpenIssues()
boolean isOngoing() //the header comment should clearly define what 'ongoing' means
```

src/main/java/filter/expression/Qualifier.java
damithc: A more cohesive method could be like this:
```java
/**
 * Returns a Map where the key indicates offset from 'current' milestone. 
 * i.e. 0 for the 'current' milestone and -1 for the 'current-1' milestone and so on.
 */
Map<Integer, TurboMilestone> addMilestoneAliasIndex(List<TurboMilestone> milestones)
```

1. This method makes sense on its own and doesn't depend on the call site to sort the input first.
2. If sorting is required within the method, it should be done by a static method in the `TurboMilestone` class?
1. I'm not sure if `Map` is the best data structure though. May be there is a better one.
2. The code should still clearly show the high-level logic (i.e. apply SLAP).
3. The header comment must describe the rules used for determining the index.

HansNewbie: Sounds good. We will still need some sort of `getIndexOfCurrentMilestone()` though, to have a base for the indexes of the other milestones.

damithc: The logic can be like this?
1. Sort by due date
2. traverse the list until the the best match for 'current' is found
3. traverse the list again to add milestones to the map.

So, step 2 is where we locate the current milestone. Yes, if it cannot be done in this method without breaking SLAP, it may have to be extracted as a method.

src/main/java/filter/expression/Qualifier.java
raisfathin: Maybe we should define a new term? The only explanation I can find from the user guide does not define any term for such milestones.

> current or curr can be used to refer to an open milestone with earliest due date, or an overdue open milestone with open issues if any. curr-2, curr-1, curr+1, curr+2, etc. (no space before or after -/+) can then be used to refer to milestones before or after current, sorted by due date.

src/main/java/filter/expression/Qualifier.java
raisfathin: Hmm, maybe it's better for now to use the second one and add an assert at the top to check that there is at most 1 milestone without due date and it has to be open. What do you think?

HansNewbie: Let's try to refactor the other things first. We'll look at this again
later.
On Feb 5, 2016 6:32 PM, "Muhammad Rais Fathin Mudzakir" <
notifications@github.com> wrote:

In src/main/java/filter/expression/Qualifier.java
<https://github.com/HubTurbo/HubTurbo/pull/1226#discussion_r52005564>:

>
> -            if (checker.isOpen() && relevant) {
> -                return Optional.of(currentIndex);
> +        if (openAndRelevantMilestone.isPresent()) {
> +            return Optional.of(allMilestones.indexOf(openAndRelevantMilestone.get()));
> +        } else {

Hmm, maybe it's better for now to use the second one and add an assert at
the top to check that there is at most 1 milestone without due date and it
has to be open. What do you think?

—
Reply to this email directly or view it on GitHub
<https://github.com/HubTurbo/HubTurbo/pull/1226/files#r52005564>.

raisfathin:
Fixed some issues mentioned above. But I think I still have some doubts.

`getSortedMilestonesByDueDate` in `TurboMilestone.java` seems inconsistent with `getDueDateComparator` in the same class. The sorting method considers milestones without due date to have an imaginary due date in the far future. This method is only used in `getMilestoneAliasIndex` in Qualifier class.

Any advice on how to approach this? I feel that the method should be renamed, but I haven't found a better name for it (and having milestones without due date to have due date in the far future kind of makes sense, but it contradicts the [existing milestone sorting functionality](https://github.com/HubTurbo/HubTurbo/blob/master/src/main/java/filter/expression/Qualifier.java#L525-L537)).

HansNewbie:
Yes. The sort milestone different behaviours were intentional (the PRs were
made at the same time). We yried to abstract out similar functionalities
and add on the rest for what we need. In the end, for alias, we decided to
ignore milestones without due date then.

In my opinion, sorting milestones should not account for milestones without
due date (since it does not make sense). We can add the milestones without
due date outside of the sorting method or in a method inside the class
using it. @dariusf any opinion on this?

dariusf:
If I understand you correctly, we account for the (lack of) due dates in the comparator because it's the common interface for sorting. That seems a bit cleaner than handling milestone sorting separately. Right now we sort via the traditional `.stream().sorted(comparator)`, without any special cases.

The inconsistency also bothers me a bit, but I'm not sure how to fix it. Is there some way we can unify how milestones are ordered? Milestones have a creation date as well -- could we use that to order undated milestones for aliases?

damithc:
Technically, sorting is not filtering, so the result of sorting should include all milestones. Omitting those without due dates was a shortcut we took at the time. A better approach is to figure out a reasonable algorithm to sort all milestone irrespective of the existence of due dates. May be do as a separate issue?

HansNewbie:
I agree that we should do sorting milestone irrespective of due dates existence in a separate issue.

@dariusf rather than by creation date, maybe sorting by name would be easier? (I doubt dev team remembers/bothers which milestone they created first when searching them)

dariusf:
Sure, I guess we could assume that people use version numbers for milestone names. Using creation date is just a heuristic which assumes that people make milestones in order.

src/test/java/tests/MilestoneAliasTests.java
```diff
@@ -160,6 +161,76 @@ public void replaceMilestoneAliasesTest() {
     }
 
     @Test
+    public void replaceMilestoneAliasesTestNoRelevantMilestone() {
```
HansNewbie: You need to rename all the variable names here. `relevant` is very confusing; why is `irrelevant` can still be the result of a test while `relevant` can be a non-result of a test. Also, it would be good if you can give a comment before the milestones and each `current` alias about what situation the tests are simulating.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -996,42 +1039,26 @@ private Qualifier convertMilestoneAliasQualifier(List<TurboMilestone> allMilesto
...
             if (aliasMatcher.group(2).equals("+")) {
-                newIndex += offset;
+                offset = Integer.parseInt(aliasMatcher.group(3));
             } else {
-                newIndex -= offset;
+                offset = -Integer.parseInt(aliasMatcher.group(3));
             }
         }
 
         // if out of milestone range, don't convert alias
```
HansNewbie: We can change this comment to `if offset is not within milestone alias range, no issues should be shown`.

src/main/java/backend/resource/TurboMilestone.java
```diff
@@ -83,6 +86,59 @@ public TurboMilestone(TurboMilestone milestone) {
...
+    }
+
+    /**
+     * A milestone is ongoing if it is not due yet or it is overdue but still has open issues.
+     */
+    public boolean isOngoing() {
+        return !isOverdue() || hasOpenIssues();
+    }
+
+    public static List<TurboMilestone> getMilestonesOfRepos(List<TurboMilestone> milestones, List<String> repoIds) {
```
HansNewbie: This method is more of `filterMilestonesOfGivenRepoIds` rather than `get`.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -91,21 +92,17 @@ public static FilterExpression replaceMilestoneAliases(IModel model, FilterExpre
             repoIds.add(model.getDefaultRepo().toLowerCase());
         }
 
-        List<TurboMilestone> allMilestones = model.getMilestones().stream()
-                .filter(ms -> ms.getDueDate().isPresent())
-                .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
-                .sorted((a, b) -> getMilestoneDueDateComparator().compare(a, b))
-                .collect(Collectors.toList());
+        List<TurboMilestone> milestonesOfRepos = TurboMilestone.getMilestonesOfRepos(model.getMilestones(), repoIds);
```
HansNewbie: `milestonesOfReposInPanel` might be a better name.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,28 +133,74 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
-            currentIndex++;
+    private static Optional<Integer> getCurrentMilestoneIndex(List<TurboMilestone> milestones) {
+        if (milestones.isEmpty()) {
+            return Optional.empty();
         }
 
-        // if no open milestone, set current as one after last milestone
-        // - this means that no such milestone, which will return no issue
-        return Optional.of(allMilestones.size());
+        long openMilestonesCount = milestones.stream().filter(TurboMilestone::isOpen).count();
```
HansNewbie: Maybe we can use `TurboMilestones.getOpenMilestone()` here, for readability.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,28 +133,74 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+        return IntStream
+                .range(0, sortedMilestones.size())
+                .boxed()
+                .collect(Collectors.toMap(
+                        i -> i - currentMilestoneIndex.get(),
+                        i -> sortedMilestones.get(i)));
+    }
 
-            currentIndex++;
+    private static Optional<Integer> getCurrentMilestoneIndex(List<TurboMilestone> milestones) {
```
HansNewbie: Put a comment we expect the input milestones to be sorted.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,28 +133,74 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+            return IntStream
+                    .range(0, milestones.size())
+                    .boxed()
+                    .filter(i -> milestones.get(i).isOpen())
+                    .findFirst();
+        } else {
+            // Look for the first milestone in the (sorted) list that is both open and ongoing.
+            int currentIndex = 0;
+            for (TurboMilestone checker : milestones) {
+                if (checker.isOpen() && checker.isOngoing()) {
```
HansNewbie: From the way it's used, `isOngoing` only refers to open milestones. How about adding `isOpen` as part of `isOngoing` instead? I don't think any closed milestone will be ongoing. We can then `stream` the milestones and `findFirst`.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,28 +133,74 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+        Optional<Integer> currentMilestoneIndex = getCurrentMilestoneIndex(sortedMilestones);
 
-            if (checker.isOpen() && relevant) {
-                return Optional.of(currentIndex);
-            }
+        return IntStream
+                .range(0, sortedMilestones.size())
+                .boxed()
+                .collect(Collectors.toMap(
+                        i -> i - currentMilestoneIndex.get(),
```
HansNewbie: Will this be safe if `currentMilestoneIndex` is empty? Maybe cover it with a test as well.

raisfathin: Should be safe since currentMilestoneIndex can only be empty if `milestones` is empty. I added an assert to check for it.

raisfathin:
Ready for review.

HansNewbie:
@raisfathin the CI fails because of style violation. Run the check style locally and correct the mistake. Also, update your branch. Meanwhile, I will review the rest of the codes.

src/main/java/backend/resource/TurboMilestone.java
```diff
@@ -83,6 +86,59 @@ public TurboMilestone(TurboMilestone milestone) {
...
+    public boolean isOverdue() {
+        return dueDate.isPresent() && dueDate.get().isBefore(LocalDate.now());
+    }
+
+    public boolean hasOpenIssues() {
+        return openIssues > 0;
+    }
+
+    /**
+     * A milestone is ongoing if it is not due yet or it is overdue but still has open issues.
```
HansNewbie: Update the comment here to reflect that we include the condition that the milestone must be open to be ongoing.

HansNewbie:
The rest of the codes seems fine.

Can you update the documentation as well, on `docs/filters.md`? To reflect that the current alias includes the only open milestone. You can rewrite the docs if you think the phrasing will get more complicated with this new feature.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,28 +133,81 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+                        i -> sortedMilestones.get(i)));
+    }
+
+    /**
+     * This method expects the milestone list to be sorted (by due date)
+     */
+    private static Optional<Integer> getCurrentMilestoneIndex(List<TurboMilestone> milestones) {
```
HansNewbie: Can we have a comment on this method about what `current` milestone refers to?

raisfathin:
Done.

src/main/java/backend/resource/TurboMilestone.java
```diff
@@ -83,6 +86,60 @@ public TurboMilestone(TurboMilestone milestone) {
...
+    }
+
+    /**
+     * A milestone is ongoing if it is open and not due yet or if it is overdue but still open and has open issues.
+     */
+    public boolean isOngoing() {
+        return isOpen() && (!isOverdue() || hasOpenIssues());
+    }
+
+    public static List<TurboMilestone> filterMilestonesOfGivenRepoIds(List<TurboMilestone> milestones,
```
damithc: `getMilestonesOfRepos`?

src/main/java/backend/resource/TurboMilestone.java
```diff
@@ -83,6 +86,60 @@ public TurboMilestone(TurboMilestone milestone) {
...
+        return (a, b) -> {
+            assert a.getDueDate().isPresent();
+            assert b.getDueDate().isPresent();
+            LocalDate aDueDate = a.getDueDate().get();
+            LocalDate bDueDate = b.getDueDate().get();
+            return aDueDate.compareTo(bDueDate);
+        };
+    }
+
+    public static List<TurboMilestone> getSortedMilestonesByDueDate(List<TurboMilestone> milestones) {
```
damithc: `sortByDueDate`?

src/main/java/filter/expression/Qualifier.java
```diff
@@ -91,21 +92,18 @@ public static FilterExpression replaceMilestoneAliases(IModel model, FilterExpre
             repoIds.add(model.getDefaultRepo().toLowerCase());
         }
 
-        List<TurboMilestone> allMilestones = model.getMilestones().stream()
-                .filter(ms -> ms.getDueDate().isPresent())
-                .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
-                .sorted((a, b) -> getMilestoneDueDateComparator().compare(a, b))
-                .collect(Collectors.toList());
+        List<TurboMilestone> milestonesOfReposInPanel = TurboMilestone.filterMilestonesOfGivenRepoIds(
+                model.getMilestones(), repoIds);
```
damithc: This indentation is more readable
```java
List<TurboMilestone> milestonesOfReposInPanel = TurboMilestone.filterMilestonesOfGivenRepoIds(
                                                                         model.getMilestones(), repoIds);
```

src/main/java/backend/resource/TurboMilestone.java
raisfathin: @HansNewbie suggested to change it to `filterMilestonesOfGivenRepoIds` since it's more like "filtering" rather than "getting". Should I change it back?

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,28 +134,85 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
         return exprWithNormalQualifiers.isSatisfiedBy(model, issue, new MetaQualifierInfo(metaQualifiers));
     }
 
-    private static Optional<Integer> getCurrentMilestoneIndex(List<TurboMilestone> allMilestones) {
-        if (allMilestones.isEmpty()) {
-            return Optional.empty();
+    /**
+     * Get all milestones that can be aliased with current-[n] to current+[n]
+     */
+    private static List<TurboMilestone> getMilestonesWithinAliasRange(List<TurboMilestone> milestonesOfReposInPanel) {
```
damithc: `getAliasableMilestones`? Need to define what `Aliasable` means.

damithc: Is the returned list sorted?

src/test/java/tests/MilestoneAliasTests.java
```diff
@@ -160,6 +161,97 @@ public void replaceMilestoneAliasesTest() {
...
+                new ArrayList<>(),
+                new ArrayList<>(),
+                new ArrayList<>(Arrays.asList(msFinished1, msFinished2, msOngoing1, msOngoing2)),
+                new ArrayList<>()
+        ));
+
+        expr = Qualifier.replaceMilestoneAliases(model, Parser.parse("milestone:current"));
+        milestoneQualifiers = expr.find(Qualifier::isMilestoneQualifier);
+        assertEquals(0, milestoneQualifiers.size());
+    }
```
damithc: For each test case, you need to explain (using comments) what situation is being tested.

src/main/java/filter/expression/Qualifier.java
raisfathin: Oh, I forgot about it. Okay, will add the comment before the next push.

The returned list is not sorted.

src/main/java/backend/resource/TurboMilestone.java
damithc: `filterMilestonesOfRepos`?

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,28 +134,88 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+                    .findFirst();
+
+            if (firstOngoingMilestonePosition.isPresent()) {
+                return firstOngoingMilestonePosition;
+            } else {
+                // if no ongoing milestone, set current as one after last milestone
+                // - this means that no such milestone, which will return no issue
+                return Optional.of(milestones.size());
+            }
+        }
```
damithc: When the `if` branch exits, there is no need for `else`.
```java
if (openMilestonesCount == 1) {
    return getFirstOpenMilestonePosition(milestones);
}

Optional<Integer> firstOngoingMilestonePosition = getFirstOngoingMilestonePosition(milestones);

return firstOngoingMilestonePosition.isPresent()
           ? firstOngoingMilestonePosition 
           : Optional.of(milestones.size());
```

The logic is still not the most intuitive though.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,28 +134,88 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+                    .range(0, milestones.size())
+                    .boxed()
+                    .filter(i -> milestones.get(i).isOngoing())
+                    .findFirst();
+
+            if (firstOngoingMilestonePosition.isPresent()) {
+                return firstOngoingMilestonePosition;
+            } else {
+                // if no ongoing milestone, set current as one after last milestone
+                // - this means that no such milestone, which will return no issue
```
damithc: This comment can be make clearer, especially the second line.

HansNewbie:
Comment again if the PR is ready for another review.

docs/filters.md
```diff
@@ -153,7 +153,7 @@ Aliases: `desc`, `description`
...
 
-`current` or `curr` can be used to refer to an open milestone with earliest due date, or an overdue open milestone with open issues if any. `curr-2`, `curr-1`, `curr+1`, `curr+2`, etc. (no space before or after `-`/`+`) can then be used to refer to milestones before or after `current`, sorted by due date. Milestones with no due date will not be considered in resolving `curr`/`current`. If there is no open milestone, `current` will refer to no milestone, while `current-1` will refer to the last closed milestone.
+`current` or `curr` can be used to refer to an open milestone with earliest due date, or an overdue open milestone with open issues if any. However, if there is only one open milestone, it will be considered as the `current` milestone even if it does not have due date. `curr-2`, `curr-1`, `curr+1`, `curr+2`, etc. (no space before or after `-`/`+`) can then be used to refer to milestones before or after `current`, sorted by due date. Milestones with no due date will not be considered in resolving `curr`/`current`. If there is no open milestone, `current` will refer to no milestone, while `current-1` will refer to the last closed milestone.
```
HansNewbie: You can remove the line "Milestones with no due date will not be considered in resolving `curr/``current`".

raisfathin:
Ready for review.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,28 +134,96 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
         return exprWithNormalQualifiers.isSatisfiedBy(model, issue, new MetaQualifierInfo(metaQualifiers));
     }
 
-    private static Optional<Integer> getCurrentMilestoneIndex(List<TurboMilestone> allMilestones) {
-        if (allMilestones.isEmpty()) {
-            return Optional.empty();
+    /**
+     * Get all milestones that can be aliased with current-[n] to current+[n]
+     *
+     * A milestone can be aliased if and only if it is open and has due date or it does not have due date but it is
```
HansNewbie: This comment seems to be out of place. Do you mean `A milestone can be current ...`? Because alias includes closed milestone.

src/main/java/backend/resource/TurboMilestone.java
```diff
@@ -83,6 +86,60 @@ public TurboMilestone(TurboMilestone milestone) {
...
+    }
+
+    public static List<TurboMilestone> filterMilestonesOfRepos(List<TurboMilestone> milestones,
+                                                               List<String> repoIds) {
+        return milestones.stream()
+                .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
+                .collect(Collectors.toList());
+    }
+
+    public static List<TurboMilestone> getOpenMilestones(List<TurboMilestone> milestones) {
```
damithc: `get..` or `filter..`?

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,28 +134,96 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+     * the only open milestone.
+     */
+    private static List<TurboMilestone> getMilestonesWithinAliasRange(List<TurboMilestone> milestonesOfReposInPanel) {
```
HansNewbie: Are we keeping the name or are we changing it to `getAliasableMilestones`? @damithc @raisfathin

src/main/java/filter/expression/Qualifier.java
raisfathin: Oh, the comment is supposed to explain the definition of aliasable milestone. I believe the comment should be `A milestone can be aliased if and only if it has due date or it does not have due date but it is the only open milestone`. Then it's correct, right?

src/main/java/filter/expression/Qualifier.java
damithc: I'm in favor of `getAliasableMilestones` because the method is not about a `range`.

src/main/java/backend/resource/TurboMilestone.java
HansNewbie: `filter` sounds more accurate.

raisfathin: ...I feel that `filter` sounds weird because filter methods usually require some kind of "filter condition" as its parameter. What do you think?

src/main/java/filter/expression/Qualifier.java
HansNewbie: How about:
`Get all milestones which milestone aliases (current+/-[n]) can resolve to. This will henceforth be called aliasable milestones.`
`Aliasable milestones must have a due date. The only exception is if there is only one open milestone in the repo; it will be included in aliasable milestones regardless of its due date status.`

src/main/java/filter/expression/Qualifier.java
HansNewbie: Sure. The comment for the method can be along the way of the comment I made above.

src/main/java/backend/resource/TurboMilestone.java
HansNewbie: That's true. In that case, `get` would be better.

damithc: um, 'open' is the filter condition. :-)

HansNewbie: Haha. Well, that makes sense. Sorry, let's keep it to `filter` then.

src/main/java/backend/resource/TurboMilestone.java
```diff
@@ -83,6 +86,60 @@ public TurboMilestone(TurboMilestone milestone) {
...
+        return (a, b) -> {
+            assert a.getDueDate().isPresent();
+            assert b.getDueDate().isPresent();
+            LocalDate aDueDate = a.getDueDate().get();
+            LocalDate bDueDate = b.getDueDate().get();
+            return aDueDate.compareTo(bDueDate);
+        };
+    }
+
+    public static List<TurboMilestone> sortByDueDate(List<TurboMilestone> milestones) {
```
damithc: As this is a public method, you may want to mention what happens if multiple milestones have the same due date.

src/main/java/filter/expression/Qualifier.java
damithc: Yes, that definition seems to be complete. See if it can be shortened though. e.g.
```
Returns all the 'aliasable' milestones from the given list.
A milestone is aliasable if it has a due date or if it is the only milestone in that repo.
```

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,28 +134,96 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
-        // if no open milestone, set current as one after last milestone
-        // - this means that no such milestone, which will return no issue
-        return Optional.of(allMilestones.size());
+        // Look for the first milestone in the (sorted) list that is ongoing.
+        Optional<Integer> firstOngoingMilestonePosition = getFirstOngoingMilestonePosition(sortedMilestones);
+
+        // if no ongoing milestone, set current as one after last milestone
+        // - this means that there is no "current" milestone, but it is possible
+        // that there is a milestone before the "current" one (i.e. current-[n]
+        // where n >= 1)
```
damithc: This behavior needs to me be mentioned in the header comment. The caller needs to know about it.

damithc: Anyway, I'm a bit uneasy about this behavior. It seems beyond the advertised purpose of this method, which is to find the current milestone. How does the caller know the returned value has this special meaning? Perhaps we should simply return `Optional.empty()` and let the caller figure out what to do?

HansNewbie: That way we will need a one line after calling this method like `currentMilestoneIndex = currentMilestoneIndex.isPresent() ? currentMilestoneIndex : Optional.of(sortedMilestones.size());`.

That makes this method concise, but the line I wrote would need more comments about the behaviour when `sortedMilestones.size()` is 0. At the same time, we can remove the `assert` above on `!currentMilestoneIndex.isPresent() && sortedMilestones.isEmpty()`.

raisfathin:
Ready for review. But I still don't know what to do with `getCurrentMilestoneIndex`. Having the index to be one after the last milestone if there is no open milestone seems to be in the definition of "current" milestone (as reflected in the docs).

src/test/java/tests/MilestoneAliasTests.java
```diff
@@ -160,6 +161,115 @@ public void replaceMilestoneAliasesTest() {
     }
 
     @Test
+    public void replaceMilestoneAliasesTestOngoingMilestoneWithNoDueDate() {
```
damithc: Use the new naming convention for test cases.

damithc:
Regarding `getCurrentMilestoneIndex`, it's fine to return next index. It fits with the idea that we consider the next (not created yet) milestone is going to be the `current` one.

raisfathin:
Ready for review.

damithc:
@raisfathin please change the label accordingly.

raisfathin:
@damithc Done.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -91,21 +92,18 @@ public static FilterExpression replaceMilestoneAliases(IModel model, FilterExpre
...
-        List<TurboMilestone> allMilestones = model.getMilestones().stream()
-                .filter(ms -> ms.getDueDate().isPresent())
-                .filter(ms -> repoIds.contains(ms.getRepoId().toLowerCase()))
-                .sorted((a, b) -> getMilestoneDueDateComparator().compare(a, b))
-                .collect(Collectors.toList());
-
-        Optional<Integer> currentMilestoneIndex = getCurrentMilestoneIndex(allMilestones);
+        List<TurboMilestone> milestonesOfReposInPanel = TurboMilestone.filterMilestonesOfRepos(
+                                                                                    model.getMilestones(), repoIds);
+        List<TurboMilestone> milestonesWithinAliasRange = getAliasableMilestones(milestonesOfReposInPanel);
```
HansNewbie: I forgot to point this out. Change the variable name to `aliasableMilestones` for consistency purpose.

HansNewbie:
@raisfathin looks good. Just one minor point on the variable name so everything is consistent. Mention me again when it's done and I will update this PR status.

raisfathin:
@HansNewbie Done.

HansNewbie:
@damithc I have reviewed the PR.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,28 +134,97 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+                .boxed()
+                .collect(Collectors.toMap(
+                        i -> i - currentMilestoneIndex.get(),
+                        i -> sortedMilestones.get(i)));
+    }
+
+    /**
+     * "Current" milestone is an ongoing milestone with the earliest due date. However, if there
+     * is only one open milestone, it will be considered as the "current" milestone, even if it
+     * does not have due date.
```
damithc: `a due date`

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,28 +134,97 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+     *
+     * If there is no ongoing or open milestone, set "current" as one after last milestone -
```
damithc: `set` --> `sets`

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,28 +134,97 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+        // Look for the first milestone in the (sorted) list that is ongoing.
+        Optional<Integer> firstOngoingMilestonePosition = getFirstOngoingMilestonePosition(sortedMilestones);
+
+        // if no ongoing milestone, set current as one after last milestone
+        return firstOngoingMilestonePosition.isPresent()
+                ? firstOngoingMilestonePosition
+                : Optional.of(sortedMilestones.size());
+    }
+
+    private static Optional<Integer> getFirstOpenMilestonePosition(List<TurboMilestone> milestones) {
```
damithc: Header comment? The milestones should be sorted?

src/main/java/filter/expression/Qualifier.java
```diff
@@ -136,28 +134,97 @@ public static boolean process(IModel model, FilterExpression expr, TurboIssue is
...
+        return IntStream
+                .range(0, milestones.size())
+                .boxed()
+                .filter(i -> milestones.get(i).isOpen())
+                .findFirst();
+    }
+
+    private static Optional<Integer> getFirstOngoingMilestonePosition(List<TurboMilestone> milestones) {
```
damithc: Same comment as method above.

src/test/java/tests/MilestoneAliasTests.java
```diff
@@ -160,6 +161,117 @@ public void replaceMilestoneAliasesTest() {
...
+        });
+
+        // 2 finished milestones with due dates, and 1 open milestone with no due date
+        // the finished milestone should be able to be referred with current-1 and current-2
+        model = TestUtils.singletonModel(new Model(REPO,
+                new ArrayList<>(),
+                new ArrayList<>(),
+                new ArrayList<>(Arrays.asList(msFinished1, msFinished2, msOngoing1)),
+                new ArrayList<>()
+        ));
```
damithc: This code snippet can be extracted into an `Object mother` method (i.e. a utility method that creates objects)? e.g. `createModel(TurboMilestone... milestones)`

damithc:
Some cosmetic changes suggested. Can merge after fixing those. No need for my approval again.

damithc:
Good job by the way. :-)

raisfathin:
@damithc @HansNewbie Done! Should I merge myself?

HansNewbie:
Do you have merge permission?

raisfathin:
@HansNewbie There is a merge pull request button that I think I can click (does this indicate that I have merge access?).

damithc:
Technically, @raisfathin has permission but don't merge yet. @HansNewbie yes, you should merge.

HansNewbie:
Noted. @raisfathin we don't use that button because we would like to set
the merge message. I will merge later.

raisfathin:
@HansNewbie @dariusf I see. Thanks for clearing up my confusion.

PR #1228, opened 2016-02-01
Allow user to set Work Allocation Board's assignees #1095

### LINKED ISSUE #1095: Allow user to set Work Allocation Board's assignees
On creating work allocation board, user can set the assignees of the panels. If possible, make it customizable, e.g. can set to have 3, 4 or 8 users.

m133225:
Fixes #1095.

- [x] Added basic functionality to choose users and generate board
- [ ] Refactor
- [ ] Improvise UI
- [ ] Add test cases

This dialog will be shown upon clicking "Boards" > "Auto-create" > "Work Allocation Board"
It will then allow the user to choose a number (up to 10) of assignees, after which it will then set that number of user-selection drop-downs.

Currently, this is still buggy:
- When "OK" is clicked before assigning a user to each drop-down
- Dialog box title is not defined

Any comments or suggestions are welcome!

### DIFF
--- a/src/main/java/ui/BoardAutoCreator.java
+++ b/src/main/java/ui/BoardAutoCreator.java
@@ -2,8 +2,11 @@ package ui;
 
 
 import backend.resource.TurboUser;
-import javafx.scene.control.Menu;
-import javafx.scene.control.MenuItem;
+import javafx.event.EventHandler;
+import javafx.geometry.Pos;
+import javafx.scene.control.*;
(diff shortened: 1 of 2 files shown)


damithc:
I feel it is too early to do this issue. May be wait till the 'assignee picker' (similar to label picker) UI is done and then reuse it for this? That way, it's less work and the UIs stay consistent.

m133225:
Closed. To be reopened at a later date when 'assignee picker' is done.

PR #1242, opened 2016-02-06
Make warning in Board Name Dialog to be more noticeable #1239

### LINKED ISSUE #1239: Make warning in Board Name Dialog to be more noticeable
As mentioned in #1208, warning and error for board naming in `BoardNameDialog` should be made more noticeable, possibly by setting the font color to red.

Gisonrg:
Fixes #1239 

Hi, new-comer is here :) Please help to correct me if I made mistakes.

#### Changes
Set the color fill of the `errorText` to be red color by defining a new css class `text-red`.
Using `-fx-fill` in css because I tried using `-fx-text-fill` but it seems like it doesn't work on Text object. 

#### Sample screenshot
screen shot 2016-02-07 at 02 17 56
screen shot 2016-02-07 at 02 18 04

### DIFF
--- a/src/main/java/ui/BoardNameDialog.java
+++ b/src/main/java/ui/BoardNameDialog.java
@@ -93,6 +93,7 @@ public final class BoardNameDialog extends Dialog<String> {
         grid.add(nameArea, 0, 0);
         
         errorText = new Text("");
+        errorText.getStyleClass().add("text-red");
         grid.add(errorText, 0, 1);
         
         getDialogPane().setContent(grid);
--- a/src/main/resources/ui/hubturbo.css
+++ b/src/main/resources/ui/hubturbo.css
@@ -241,3 +241,7 @@
     -fx-font-size: 13px;
     -fx-wrap-text: true;
 }
+
+.text-red {
+    -fx-fill: red;
+}


damithc:
I think the `empty name` should not be shown before the user types anything. If we do, all users will start with an error message, which is not nice.

Gisonrg:
The empty name error will not be show before the user types anything. There will be a default name "New Board" pre-populated for the user. The error will only be shown if the user does not enter anything in the text field.

Here is the default dialog if you save an unsaved board.
screen shot 2016-02-07 at 11 41 12

dariusf:
I think the commit message can be a little more specific. It would be helpful to add the *reason* for the change, since the diff already tells us what changed. :stuck_out_tongue: 

Apart from that, this looks good.

Gisonrg:
@dariusf I see, thanks! 
Maybe something like "Set red color for warning text to make it more noticeable" would be better

dariusf:
Yup, that works. Could you amend the commit? I'll merge after that.

(Commit SHA is 4a80abc1be090d040a5a13118f4c521837ca67c6... just in case)

Gisonrg:
@dariusf 
Fixed

PR #1248, opened 2016-02-10
Launch startup board for first time users #1243

### LINKED ISSUE #1243: Launch startup board for first time users
Launch the startup board if the number of repos is 0.

codemanmeet:
Fixes #1243. Also updated sample board filters.

### DIFF
--- a/src/main/java/ui/BoardAutoCreator.java
+++ b/src/main/java/ui/BoardAutoCreator.java
@@ -64,7 +64,7 @@ public class BoardAutoCreator {
     public Menu generateBoardAutoCreateMenu() {
         Menu autoCreate = new Menu("Auto-create");
         MenuItem sample = new MenuItem(SAMPLE_BOARD);
-        sample.setOnAction(e -> createSampleBoard(false));
+        sample.setOnAction(e -> createSampleBoard(true));
         autoCreate.getItems().add(sample);
 
         MenuItem milestone = new MenuItem(MILESTONES);
@@ -131,7 +131,12 @@ public class BoardAutoCreator {
                         "It is saved under the name \"" + boardName + "\".");
     }
 
-    public void createSampleBoard(boolean isCalledOnStartup) {
+    /**
+     * Creates a sample board to showcase HubTurbo's functionality with sample repos in filters.
+     * The confirmation dialog pops up for the creation of the sample board if isDialogShown is set to true.
+     * @param isDialogShown
+     */
+    public void createSampleBoard(boolean isDialogShown) {
         logger.info("Creating " + SAMPLE_BOARD);
 
         panelControl.closeAllPanels();
@@ -148,8 +153,7 @@ public class BoardAutoCreator {
 
         triggerBoardSaveEventSequence(SAMPLE_BOARD);
 
-        //isCalledOnStartup is set to true when called on launch in UI.java
-        if (!isCalledOnStartup){
+        if (isDialogShown){
             DialogMessage.showInformationDialog("Auto-create Board - " + SAMPLE_BOARD, SAMPLE_BOARD_DIALOG);
         }
     }
--- a/src/main/java/ui/BoardAutoCreator.java
+++ b/src/main/java/ui/BoardAutoCreator.java
@@ -66,7 +66,7 @@ public class BoardAutoCreator {
     public Menu generateBoardAutoCreateMenu() {
         Menu autoCreate = new Menu("Auto-create");
         MenuItem sample = new MenuItem(SAMPLE_BOARD);
-        sample.setOnAction(e -> createSampleBoard());
+        sample.setOnAction(e -> createSampleBoard(false));
         autoCreate.getItems().add(sample);
 
         MenuItem milestone = new MenuItem(MILESTONES);
@@ -133,7 +133,7 @@ public class BoardAutoCreator {
                         "It is saved under the name \"" + boardName + "\".");
     }
 
-    private void createSampleBoard() {
+    public void createSampleBoard(boolean isCalledOnStartup) {
         logger.info("Creating " + SAMPLE_BOARD);
 
         panelControl.closeAllPanels();
@@ -149,8 +149,9 @@ public class BoardAutoCreator {
         panelControl.selectPanel(0); // current
 
         triggerBoardSaveEventSequence(SAMPLE_BOARD);
-
-        DialogMessage.showInformationDialog("Auto-create Board - " + SAMPLE_BOARD, SAMPLE_BOARD_DIALOG);
+        if(!isCalledOnStartup){
+            DialogMessage.showInformationDialog("Auto-create Board - " + SAMPLE_BOARD, SAMPLE_BOARD_DIALOG);
+        }
     }
 
     private List<PanelInfo> generatePanelInfoFromTurboUsers(List<TurboUser> users,
--- a/src/main/java/ui/BoardAutoCreator.java
+++ b/src/main/java/ui/BoardAutoCreator.java
@@ -27,30 +27,28 @@ public class BoardAutoCreator {
     private static final int MAX_WORK_ALLOCATION_PANELS = 5;
     public static final String SAMPLE_BOARD = "Sample Board!";
     private static final String FIRST_SAMPLE_REPO_NAME = "HubTurbo/SampleRepo";
-    private static final String SECOND_SAMPLE_REPO_NAME = "HubTurbo/SampleRepo2";
+    private static final String SECOND_SAMPLE_REPO_NAME = "HubTurbo/HubTurbo";
     public static final String SAMPLE_BOARD_DIALOG = String.format("%s has been created and loaded.", SAMPLE_BOARD);
 
     public static final List<String> SAMPLE_PANEL_NAMES =
             Collections.unmodifiableList(Arrays.asList(
+                    "Seven Recent unread updates from my repos",
                     "All issues in my two sample repos",
-                    "Latest 3 urgent open issues",
                     "Open issues assigned to Darius or Manmeet",
                     "Progress of the current milestone",
-                    "Issues awaiting prioritization",
-                    "Recent unread updates from my repos"
+                    "Issues awaiting prioritization"
             ));
 
     public static final List<String> SAMPLE_PANEL_FILTERS =
             Collections.unmodifiableList(Arrays.asList(
+                    String.format("repo:%s;%s count:7 is:unread updated<24 sort:!updated",
+                            FIRST_SAMPLE_REPO_NAME, SECOND_SAMPLE_REPO_NAME),
             String.format("repo:%s;%s is:issue sort:!updated,comments",
                     FIRST_SAMPLE_REPO_NAME, SECOND_SAMPLE_REPO_NAME),
-            String.format("repo:%s count:3 is:issue is:open label:\"urgent\"", FIRST_SAMPLE_REPO_NAME),
             String.format("repo:%s is:issue is:open (assignee:dariusf || assignee:codemanmeet)",
                     FIRST_SAMPLE_REPO_NAME),
             String.format("repo:%s m:curr sort:status", FIRST_SAMPLE_REPO_NAME),
-            String.format("repo:%s is:open is:issue !label:priority.", FIRST_SAMPLE_REPO_NAME),
-            String.format("repo:%s;%s is:unread updated<24 sort:!updated",
-                    FIRST_SAMPLE_REPO_NAME, SECOND_SAMPLE_REPO_NAME)
+            String.format("repo:%s is:open is:issue !label:priority.", FIRST_SAMPLE_REPO_NAME)
             ));
 
     private final UI ui;
@@ -149,7 +147,9 @@ public class BoardAutoCreator {
         panelControl.selectPanel(0); // current
 
         triggerBoardSaveEventSequence(SAMPLE_BOARD);
-        if(!isCalledOnStartup){
+
+        //isCalledOnStartup is set to true when called on launch in UI.java
+        if (!isCalledOnStartup){
             DialogMessage.showInformationDialog("Auto-create Board - " + SAMPLE_BOARD, SAMPLE_BOARD_DIALOG);
         }
     }
--- a/src/main/java/ui/BoardAutoCreator.java
+++ b/src/main/java/ui/BoardAutoCreator.java
@@ -14,10 +14,7 @@ import util.DialogMessage;
 import util.Utility;
 import util.events.BoardSavedEvent;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
 import java.util.stream.Collectors;
 
 public class BoardAutoCreator {
@@ -30,26 +27,24 @@ public class BoardAutoCreator {
     private static final String SECOND_SAMPLE_REPO_NAME = "HubTurbo/HubTurbo";
     public static final String SAMPLE_BOARD_DIALOG = String.format("%s has been created and loaded.", SAMPLE_BOARD);
 
-    public static final List<String> SAMPLE_PANEL_NAMES =
-            Collections.unmodifiableList(Arrays.asList(
-                    "Seven Recent unread updates from my repos",
-                    "All issues in my two sample repos",
-                    "Open issues assigned to Darius or Manmeet",
-                    "Progress of the current milestone",
-                    "Issues awaiting prioritization"
-            ));
+    public static final Map<String, String> getSamplePanelDetails(){
+        Map<String, String> panelMap = new HashMap<>();
+        panelMap.put("Seven Recent unread updates from my repos",
+                String.format("repo:%s;%s count:7 is:unread updated<24 sort:!updated",
+                FIRST_SAMPLE_REPO_NAME, SECOND_SAMPLE_REPO_NAME));
+        panelMap.put("All issues in my two sample repos",
+                String.format("repo:%s;%s is:issue sort:!updated,comments",
+                FIRST_SAMPLE_REPO_NAME, SECOND_SAMPLE_REPO_NAME));
+        panelMap.put("Open issues assigned to Darius or Manmeet",
+                String.format("repo:%s is:issue is:open (assignee:dariusf || assignee:codemanmeet)",
+                FIRST_SAMPLE_REPO_NAME));
+        panelMap.put("Progress of the current milestone",
+                String.format("repo:%s m:curr sort:status", FIRST_SAMPLE_REPO_NAME));
+        panelMap.put("Issues awaiting prioritization",
+                String.format("repo:%s is:open is:issue !label:priority.", FIRST_SAMPLE_REPO_NAME));
+        return Collections.unmodifiableMap(panelMap);
 
-    public static final List<String> SAMPLE_PANEL_FILTERS =
-            Collections.unmodifiableList(Arrays.asList(
-                    String.format("repo:%s;%s count:7 is:unread updated<24 sort:!updated",
-                            FIRST_SAMPLE_REPO_NAME, SECOND_SAMPLE_REPO_NAME),
-            String.format("repo:%s;%s is:issue sort:!updated,comments",
-                    FIRST_SAMPLE_REPO_NAME, SECOND_SAMPLE_REPO_NAME),
-            String.format("repo:%s is:issue is:open (assignee:dariusf || assignee:codemanmeet)",
-                    FIRST_SAMPLE_REPO_NAME),
-            String.format("repo:%s m:curr sort:status", FIRST_SAMPLE_REPO_NAME),
-            String.format("repo:%s is:open is:issue !label:priority.", FIRST_SAMPLE_REPO_NAME)
-            ));
+    }
 
     private final UI ui;
     private final PanelControl panelControl;
@@ -133,7 +128,8 @@ public class BoardAutoCreator {
 
     /**
      * Creates a sample board to showcase HubTurbo's functionality with sample repos in filters.
-     * The confirmation dialog pops up for the creation of the sample board if isDialogShown is set to true.
+     * The dialog box shows a confirmation message for the successful
+     * creation and loading of the sample board if isDialogShown is set to true.
      * @param isDialogShown
      */
     public void createSampleBoard(boolean isDialogShown) {
@@ -143,8 +139,8 @@ public class BoardAutoCreator {
 
         List<PanelInfo> panelData = new ArrayList<>();
 
-        for (int i = 0; i < SAMPLE_PANEL_NAMES.size(); i++) {
-            panelData.add(new PanelInfo(SAMPLE_PANEL_NAMES.get(i), SAMPLE_PANEL_FILTERS.get(i)));
+        for (Map.Entry<String, String> entry : getSamplePanelDetails().entrySet()) {
+            panelData.add(new PanelInfo(entry.getKey(), entry.getValue()));
         }
 
         createBoard(panelData, SAMPLE_BOARD);
--- a/src/main/java/ui/TestController.java
+++ b/src/main/java/ui/TestController.java
@@ -100,7 +100,7 @@ public final class TestController {
      * so don't need boards created on startup.
      */
     public static boolean shouldOpenSampleBoard() {
-        return !TestController.isTestMode() || TestController.shouldTestStartupBoard();
+        return !isTestMode() || shouldTestStartupBoard();
     }
 
     /**
--- a/src/main/java/ui/TestController.java
+++ b/src/main/java/ui/TestController.java
@@ -58,12 +58,21 @@ public final class TestController {
                         isTestJSONEnabled() ||
                         isTestChromeDriver() ||
                         isTestGlobalConfig() ||
+                        isTestStartupBoard() ||
                         isCloseOnQuit());
     }
 
     public static boolean isTestGlobalConfig() {
         return hasUI() && commandLineArgs.getOrDefault("testconfig", "false").equalsIgnoreCase("true");
     }
+
+    /** Used for testing startup board on launch (i.e. when the number of stored repos is 0)
+     * StartBoardLauncherTest makes use of this argument to test the launch of the board for first time users
+     */
+    public static boolean isTestStartupBoard() {
+        return hasUI() && commandLineArgs.getOrDefault("startupboard", "false").equalsIgnoreCase("true");
+    }
+
 
     // When --bypasslogin=true is passed as an argument, the username and password
     // are empty strings.
--- a/src/main/java/ui/TestController.java
+++ b/src/main/java/ui/TestController.java
@@ -58,7 +58,7 @@ public final class TestController {
                         isTestJSONEnabled() ||
                         isTestChromeDriver() ||
                         isTestGlobalConfig() ||
-                        isTestStartupBoard() ||
+                        shouldTestStartupBoard() ||
                         isCloseOnQuit());
     }
 
@@ -66,10 +66,10 @@ public final class TestController {
         return hasUI() && commandLineArgs.getOrDefault("testconfig", "false").equalsIgnoreCase("true");
     }
 
-    /** Used for testing startup board on launch (i.e. when the number of stored repos is 0)
-     * StartBoardLauncherTest makes use of this argument to test the launch of the board for first time users
+    /**
+     * Determines whether the startup board should be tested using command line arguments.
      */
-    public static boolean isTestStartupBoard() {
+    public static boolean shouldTestStartupBoard() {
         return hasUI() && commandLineArgs.getOrDefault("startupboard", "false").equalsIgnoreCase("true");
     }
 
@@ -92,6 +92,15 @@ public final class TestController {
     // tests to fail).
     public static boolean isCloseOnQuit() {
         return hasUI() && commandLineArgs.getOrDefault("closeonquit", "false").equalsIgnoreCase("true");
+    }
+
+    /**
+     * Returns true if HubTurbo is not being run on Test mode or if startup board creation is being tested.
+     * As for the other tests, they start on a clean state with dummy repos established
+     * so don't need boards created on startup.
+     */
+    public static boolean shouldOpenSampleBoard() {
+        return !TestController.isTestMode() || TestController.shouldTestStartupBoard();
     }
 
     /**
--- a/src/main/java/ui/UI.java
+++ b/src/main/java/ui/UI.java
@@ -143,14 +143,12 @@ public class UI extends Application implements EventDispatcher {
         getMainWindowHandle(mainStage.getTitle());
     }
 
-    private void startupBoardInitiator(boolean shouldCreateSampleBoard){
-        if(shouldCreateSampleBoard){
-            BoardAutoCreator boardCreator = new BoardAutoCreator(this, panels, prefs);
-            if(prefs.getAllBoardNames().contains(boardCreator.SAMPLE_BOARD)){
-                prefs.removeBoard(boardCreator.SAMPLE_BOARD);
-            }
-            boardCreator.createSampleBoard(true);
+    private void launchSampleBoard(){
+        BoardAutoCreator boardCreator = new BoardAutoCreator(this, panels, prefs);
+        if (prefs.getAllBoardNames().contains(boardCreator.SAMPLE_BOARD)){
+            prefs.removeBoard(boardCreator.SAMPLE_BOARD);
         }
+        boardCreator.createSampleBoard(true);
     }
 
     private void disableUI(boolean disable) {
@@ -160,7 +158,7 @@ public class UI extends Application implements EventDispatcher {
     }
 
     private void showMainWindow(String repoId) {
-        boolean shouldCreateSampleBoardAfterWindowInit = logic.getStoredRepos().size() == 0;
+        boolean shouldCreateSampleBoard = logic.getStoredRepos().size() == 0;
         logic.openPrimaryRepository(repoId);
         logic.setDefaultRepo(repoId);
         repoSelector.setText(repoId);
@@ -185,8 +183,8 @@ public class UI extends Application implements EventDispatcher {
         // Should only be called after panels have been initialized
         ensureSelectedPanelHasFocus();
         initialisePickers();
-        if (!TestController.isTestMode()){
-            startupBoardInitiator(shouldCreateSampleBoardAfterWindowInit);
+        if (shouldCreateSampleBoard && (!TestController.isTestMode() || TestController.isTestStartupBoard())){
+            launchSampleBoard();
         }
     }
 
--- a/src/main/java/ui/UI.java
+++ b/src/main/java/ui/UI.java
@@ -143,6 +143,16 @@ public class UI extends Application implements EventDispatcher {
         getMainWindowHandle(mainStage.getTitle());
     }
 
+    private void startupBoardInitiator(boolean shouldCreateSampleBoard){
+        if(shouldCreateSampleBoard){
+            BoardAutoCreator boardCreator = new BoardAutoCreator(this, panels, prefs);
+            if(prefs.getAllBoardNames().contains(boardCreator.SAMPLE_BOARD)){
+                prefs.removeBoard(boardCreator.SAMPLE_BOARD);
+            }
+            boardCreator.createSampleBoard(true);
+        }
+    }
+
     private void disableUI(boolean disable) {
         mainStage.setResizable(!disable);
         menuBar.setDisable(disable);
@@ -150,6 +160,7 @@ public class UI extends Application implements EventDispatcher {
     }
 
     private void showMainWindow(String repoId) {
+        boolean shouldCreateSampleBoardAfterWindowInit = logic.getStoredRepos().size() == 0;
         logic.openPrimaryRepository(repoId);
         logic.setDefaultRepo(repoId);
         repoSelector.setText(repoId);
@@ -174,6 +185,9 @@ public class UI extends Application implements EventDispatcher {
         // Should only be called after panels have been initialized
         ensureSelectedPanelHasFocus();
         initialisePickers();
+        if (!TestController.isTestMode()){
+            startupBoardInitiator(shouldCreateSampleBoardAfterWindowInit);
+        }
     }
 
     private void initialisePickers() {
--- a/src/main/java/ui/UI.java
+++ b/src/main/java/ui/UI.java
@@ -155,6 +155,8 @@ public class UI extends Application implements EventDispatcher {
     }
 
     private void showMainWindow(String repoId) {
+        //We infer this is the first time HT is being used if there are no repo data stored at the start up.
+        //This check needs to be done at the very beginning of the startup, before HT downloads any repo data.
         boolean isAFirstTimeUser = logic.getStoredRepos().isEmpty();
         logic.openPrimaryRepository(repoId);
         logic.setDefaultRepo(repoId);
--- a/src/main/java/ui/UI.java
+++ b/src/main/java/ui/UI.java
@@ -143,7 +143,7 @@ public class UI extends Application implements EventDispatcher {
         getMainWindowHandle(mainStage.getTitle());
     }
 
-    private void openSampleBoard(){
+    private void createAndLoadSampleBoard(){
         BoardAutoCreator boardCreator = new BoardAutoCreator(this, panels, prefs);
         boardCreator.createSampleBoard(false);
     }
@@ -155,8 +155,7 @@ public class UI extends Application implements EventDispatcher {
     }
 
     private void showMainWindow(String repoId) {
-        boolean isFirstTimeUser = logic.getStoredRepos().isEmpty();
-
+        boolean isAFirstTimeUser = logic.getStoredRepos().isEmpty();
         logic.openPrimaryRepository(repoId);
         logic.setDefaultRepo(repoId);
         repoSelector.setText(repoId);
@@ -181,8 +180,8 @@ public class UI extends Application implements EventDispatcher {
         // Should only be called after panels have been initialized
         ensureSelectedPanelHasFocus();
         initialisePickers();
-        if (isFirstTimeUser && TestController.shouldOpenSampleBoard()){
-            openSampleBoard();
+        if (isAFirstTimeUser && TestController.shouldOpenSampleBoard()){
+            createAndLoadSampleBoard();
         }
     }
 
--- a/src/main/java/ui/UI.java
+++ b/src/main/java/ui/UI.java
@@ -143,12 +143,9 @@ public class UI extends Application implements EventDispatcher {
         getMainWindowHandle(mainStage.getTitle());
     }
 
-    private void launchSampleBoard(){
+    private void openSampleBoard(){
         BoardAutoCreator boardCreator = new BoardAutoCreator(this, panels, prefs);
-        if (prefs.getAllBoardNames().contains(boardCreator.SAMPLE_BOARD)){
-            prefs.removeBoard(boardCreator.SAMPLE_BOARD);
-        }
-        boardCreator.createSampleBoard(true);
+        boardCreator.createSampleBoard(false);
     }
 
     private void disableUI(boolean disable) {
@@ -158,7 +155,8 @@ public class UI extends Application implements EventDispatcher {
     }
 
     private void showMainWindow(String repoId) {
-        boolean shouldCreateSampleBoard = logic.getStoredRepos().size() == 0;
+        boolean isFirstTimeUser = logic.getStoredRepos().isEmpty();
+
         logic.openPrimaryRepository(repoId);
         logic.setDefaultRepo(repoId);
         repoSelector.setText(repoId);
@@ -183,8 +181,8 @@ public class UI extends Application implements EventDispatcher {
         // Should only be called after panels have been initialized
         ensureSelectedPanelHasFocus();
         initialisePickers();
-        if (shouldCreateSampleBoard && (!TestController.isTestMode() || TestController.isTestStartupBoard())){
-            launchSampleBoard();
+        if (isFirstTimeUser && TestController.shouldOpenSampleBoard()){
+            openSampleBoard();
         }
     }
 
--- a/src/test/java/guitests/BoardAutoCreatorTest.java
+++ b/src/test/java/guitests/BoardAutoCreatorTest.java
@@ -108,8 +108,7 @@ public class BoardAutoCreatorTest extends UITest {
     }
 
     /**
-     * Tests panelControl properties associated with panels.
-     * PanelDetails obtained from Boards are tested against current Panel infos as well.
+     * Confirms the currently displayed board consists the set of panels specified in panelDetails
      */
     public static void verifyBoard(PanelControl pc, Map<String, String> panelDetails){
         List<PanelInfo> panelInfos = pc.getCurrentPanelInfos();
--- a/src/test/java/guitests/BoardAutoCreatorTest.java
+++ b/src/test/java/guitests/BoardAutoCreatorTest.java
@@ -5,6 +5,7 @@ import static org.loadui.testfx.Assertions.assertNodeExists;
 import static org.loadui.testfx.controls.Commons.hasText;
 
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 
 import org.junit.Before;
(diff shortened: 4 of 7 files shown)


src/main/java/ui/TestController.java
```diff
@@ -65,6 +66,12 @@ public static boolean isTestGlobalConfig() {
         return hasUI() && commandLineArgs.getOrDefault("testconfig", "false").equalsIgnoreCase("true");
     }
 
+    // used for testing the creation of startup board on startup (a.k.a when the number of repos is 0 initially)
```
damithc: Write proper javadoc comments.

src/main/java/ui/UI.java
```diff
@@ -143,6 +143,16 @@ private void login(boolean isBypassLogin) {
         getMainWindowHandle(mainStage.getTitle());
     }
 
+    private void startupBoardInitiator(boolean shouldCreateSampleBoard){
```
damithc: Instead of passing this boolean, simply not call this method if it is false?

src/main/java/ui/TestController.java
codemanmeet: Sure!

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -133,7 +131,7 @@ private void createWorkAllocationBoard() {
                         "It is saved under the name \"" + boardName + "\".");
     }
 
-    private void createSampleBoard() {
+    public void createSampleBoard(boolean isCalledOnStartup) {
```
dariusf: Needs a javadoc comment. Also, the parameter can be named after what it does instead of how it's used. suppressDialog?

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -150,7 +148,10 @@ private void createSampleBoard() {
 
         triggerBoardSaveEventSequence(SAMPLE_BOARD);
 
-        DialogMessage.showInformationDialog("Auto-create Board - " + SAMPLE_BOARD, SAMPLE_BOARD_DIALOG);
+        //isCalledOnStartup is set to true when called on launch in UI.java
```
dariusf: We shouldn't make reference to how methods are used in comments.

src/main/java/ui/TestController.java
```diff
@@ -65,6 +66,14 @@ public static boolean isTestGlobalConfig() {
         return hasUI() && commandLineArgs.getOrDefault("testconfig", "false").equalsIgnoreCase("true");
     }
 
+    /** Used for testing startup board on launch (i.e. when the number of stored repos is 0)
+     * StartBoardLauncherTest makes use of this argument to test the launch of the board for first time users
+     */
+    public static boolean isTestStartupBoard() {
```
dariusf: Again, we shouldn't refer to how methods are used. The description should start on the second line, as a matter of style.

Also, the method name isn't very clear, even though I understand what it does from the method body. `shouldTestStartupBoard`?

src/main/java/ui/UI.java
```diff
@@ -143,6 +143,14 @@ private void login(boolean isBypassLogin) {
         getMainWindowHandle(mainStage.getTitle());
     }
 
+    private void launchSampleBoard(){
+        BoardAutoCreator boardCreator = new BoardAutoCreator(this, panels, prefs);
+        if (prefs.getAllBoardNames().contains(boardCreator.SAMPLE_BOARD)){
+            prefs.removeBoard(boardCreator.SAMPLE_BOARD);
```
dariusf: Maybe you could explain why this part is needed with an example in a comment? Is there a case where it could remove something legitimate?

src/main/java/ui/UI.java
```diff
@@ -150,6 +158,7 @@ private void disableUI(boolean disable) {
     }
 
     private void showMainWindow(String repoId) {
+        boolean shouldCreateSampleBoard = logic.getStoredRepos().size() == 0;
```
dariusf: `logic.getStoredRepos().isEmpty()`

src/main/java/ui/UI.java
```diff
@@ -174,6 +183,9 @@ private void showMainWindow(String repoId) {
         // Should only be called after panels have been initialized
         ensureSelectedPanelHasFocus();
         initialisePickers();
+        if (shouldCreateSampleBoard && (!TestController.isTestMode() || TestController.isTestStartupBoard())){
```
dariusf: Can the other parts of the conditional be part of `shouldCreateSampleBoard`?

It might be helpful to explain why we don't want to open the sample board when in test mode, but without the flag.

Also, we should move that variable down here, since it's only used here.

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,62 @@
...
+    @Test
+    public void testSampleBoardOnLaunch(){
+        selectAll();
+        type("dummy").push(KeyCode.TAB);
+        type("dummy").push(KeyCode.TAB);
+        type("test").push(KeyCode.TAB);
+        type("test");
+        click("Sign in");
+
+        PlatformEx.waitOnFxThread();
```
dariusf: Why is this needed?

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+        click("File");
+        click("Logout");
```
dariusf: There's a UITest method for traversing menus which would help here.

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+        assertEquals(panelControl.getPanelCount(), SAMPLE_PANEL_NAMES.size());
```
dariusf: Maybe we should check the panel names as well? Since they're static. Just to ensure that they end up being used in the board that's open.

src/main/java/ui/UI.java
codemanmeet: I was thinking maybe if we wanted to update the sample board in the future the user could get a new version of the sample board. But now that I think, the user himself would then not be a first time user and he certainly would not delete repo files from the store folder himself. The `Auto-create` itself would enable the creation of a new sample board if need be! I guess this if statement is not necessary. I'll get rid of it.

codemanmeet: I was also thinking of possibility of duplicates, which is also not possible since this is a first time user we are talking about!

src/main/java/ui/UI.java
codemanmeet: Startup board testing is required only on app launch. Other tests go with the usual assumption of bypassing login and testing HubTurbo features assuming a repo has been created. Thus, we need a flag to test the startup board feature. Makes it neater. Yup, it's helpful to explain that. I have added a comment about it too.

`logic.openPrimaryRepository`/`logic.setDefaultRepo` change `logic.getStoredRepos().isEmpty()` so this needs to be tested before those two. The number of stored repos will be 0 the first time a user logs in (i.e.) if the store folder is empty. So the check is needed at that place.

src/test/java/guitests/StartupBoardLauncherTest.java
codemanmeet: Will make use of `clickMenu`.

src/test/java/guitests/StartupBoardLauncherTest.java
codemanmeet: Will do that!

codemanmeet:
@dariusf Have addressed the comments! Let me know if it can be merged.

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -150,7 +152,9 @@ private void createSampleBoard() {
 
         triggerBoardSaveEventSequence(SAMPLE_BOARD);
 
-        DialogMessage.showInformationDialog("Auto-create Board - " + SAMPLE_BOARD, SAMPLE_BOARD_DIALOG);
+        if (!suppressDialog){
```
damithc: As the boolean is inverted here, why not use `isShowDialog` in the first place?

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -133,7 +131,11 @@ private void createWorkAllocationBoard() {
                         "It is saved under the name \"" + boardName + "\".");
     }
 
-    private void createSampleBoard() {
+    /**
+     * Creates a sample board to showcase HubTurbo's functionality with sample repos in filters.
+     * @param suppressDialog
+     */
+    public void createSampleBoard(boolean suppressDialog) {
```
damithc: This is not how we name booleans :-|

src/main/java/ui/UI.java
```diff
@@ -174,6 +180,12 @@ private void showMainWindow(String repoId) {
         // Should only be called after panels have been initialized
         ensureSelectedPanelHasFocus();
         initialisePickers();
+        //Startup board testing required only on app launch.
+        //Checks if the number of repos is 0 indicative of first time users of the app.
+        if (shouldCreateSampleBoardOnAppLaunch && (!TestController.isTestMode()
+                || TestController.shouldTestStartupBoard())){
```
damithc: The line wrapping should be at the highest level possible, in this case, at `&&`. The current line wrapping has less readability.

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -150,7 +152,9 @@ private void createSampleBoard() {
 
         triggerBoardSaveEventSequence(SAMPLE_BOARD);
 
-        DialogMessage.showInformationDialog("Auto-create Board - " + SAMPLE_BOARD, SAMPLE_BOARD_DIALOG);
+        if (isShowDialog){
```
dariusf: Maybe this could be `isDialogShown`?

src/main/java/ui/UI.java
```diff
@@ -174,6 +180,12 @@ private void showMainWindow(String repoId) {
         // Should only be called after panels have been initialized
         ensureSelectedPanelHasFocus();
         initialisePickers();
+        //Startup board testing required only on app launch.
+        //Checks if the number of repos is 0 indicative of first time users of the app.
+        if (shouldCreateSampleBoardOnAppLaunch &&
+                (!TestController.isTestMode() || TestController.shouldTestStartupBoard())){
```
dariusf: Could this be factored out into a variable and named? Would be a bit easier to read.

src/main/java/ui/UI.java
```diff
@@ -174,6 +181,12 @@ private void showMainWindow(String repoId) {
         // Should only be called after panels have been initialized
         ensureSelectedPanelHasFocus();
         initialisePickers();
+        //Startup board testing required only on app launch.
+        boolean shouldBeMainLaunchOrStartupBoardTesting =
+                !TestController.isTestMode() || TestController.shouldTestStartupBoard();
+        if (shouldCreateSampleBoardOnAppLaunch && shouldBeMainLaunchOrStartupBoardTesting){
```
dariusf: I feel like these names are at too low a level of abstraction to help readability. The first comment is unnecessary if we use it as the variable name.

```java
boolean isFirstTimeUser = logic.getStoredRepos().isEmpty();
```

I get that giving a succinct name to the `not test mode || has flag` condition is difficult, so we can just give it a general name, and explain why we need it with a header comment (all tests start with a clean slate and we don't want the sample board then, etc.).

```java
if (isFirstTimeUser && shouldOpenSampleBoard()) {
    // ...
}

boolean shouldOpenSampleBoard() {
    return !TestController.isTestMode() || TestController.shouldTestStartupBoard();
}
```

Exact names can be improved, but the level of abstraction should be similar.

src/main/java/ui/UI.java
```diff
@@ -174,6 +181,12 @@ private void showMainWindow(String repoId) {
...
+            launchSampleBoard();
```
dariusf: openSampleBoard maybe? More consistent with the terms we use in the UI.

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -150,7 +152,9 @@ private void createSampleBoard() {
 
         triggerBoardSaveEventSequence(SAMPLE_BOARD);
 
-        DialogMessage.showInformationDialog("Auto-create Board - " + SAMPLE_BOARD, SAMPLE_BOARD_DIALOG);
+        if (isDialogShown){
+            DialogMessage.showInformationDialog("Auto-create Board - " + SAMPLE_BOARD, SAMPLE_BOARD_DIALOG);
+        }
```
damithc: This behavior is not documented in the header comment.

src/main/java/ui/TestController.java
```diff
@@ -65,6 +66,14 @@ public static boolean isTestGlobalConfig() {
         return hasUI() && commandLineArgs.getOrDefault("testconfig", "false").equalsIgnoreCase("true");
     }
 
+    /**
+     * Tests the launch of the sample board on first time app startup.
+     */
```
damithc: The comment doesn't match the method. The method doesn't test anything.

src/main/java/ui/UI.java
```diff
@@ -143,13 +143,29 @@ private void login(boolean isBypassLogin) {
...
     }
 
+    /**
+     * Returns true if HubTurbo is not being run on Test mode or if startup board creation is being tested.
+     * As for the other tests, they start on a clean state with dummy repos established
+     * so don't need boards created on startup.
+     */
+    private boolean shouldOpenSampleBoard() {
+        return !TestController.isTestMode() || TestController.shouldTestStartupBoard();
+    }
```
damithc: Any way to avoid/minimize polluting production code with test-specific code like this?

src/main/java/ui/UI.java
```diff
@@ -143,13 +143,29 @@ private void login(boolean isBypassLogin) {
...
+
     private void showMainWindow(String repoId) {
+        boolean isFirstTimeUser = logic.getStoredRepos().isEmpty();
```
damithc: Declare variable as close to where it is needed as possible.

codemanmeet: This is the closest to where it can be. This has to be here since `logic.getStoredRepos()` gets altered in the next line by `logic.openPrimaryRepository`.

src/main/java/ui/UI.java
codemanmeet: Yup moved it out to `TestController.java`

src/main/java/ui/BoardAutoCreator.java
codemanmeet: Added it

src/main/java/ui/TestController.java
codemanmeet: changed.

codemanmeet:
@dariusf @damithc Made all the relevant refactoring changes! Can I go ahead and merge now?

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -27,30 +27,28 @@
...
                     FIRST_SAMPLE_REPO_NAME, SECOND_SAMPLE_REPO_NAME),
-            String.format("repo:%s count:3 is:issue is:open label:\"urgent\"", FIRST_SAMPLE_REPO_NAME),
             String.format("repo:%s is:issue is:open (assignee:dariusf || assignee:codemanmeet)",
                     FIRST_SAMPLE_REPO_NAME),
             String.format("repo:%s m:curr sort:status", FIRST_SAMPLE_REPO_NAME),
-            String.format("repo:%s is:open is:issue !label:priority.", FIRST_SAMPLE_REPO_NAME),
-            String.format("repo:%s;%s is:unread updated<24 sort:!updated",
-                    FIRST_SAMPLE_REPO_NAME, SECOND_SAMPLE_REPO_NAME)
+            String.format("repo:%s is:open is:issue !label:priority.", FIRST_SAMPLE_REPO_NAME)
             ));
```
damithc: Instead of two parallel `List` objects, isn't it better to use one `Map` object?

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -133,7 +131,12 @@ private void createWorkAllocationBoard() {
                         "It is saved under the name \"" + boardName + "\".");
     }
 
-    private void createSampleBoard() {
+    /**
+     * Creates a sample board to showcase HubTurbo's functionality with sample repos in filters.
+     * The confirmation dialog pops up for the creation of the sample board if isDialogShown is set to true.
```
damithc: What's the purpose of the dialog box?

codemanmeet: The dialog box confirms the creation of the sample board and tells the user it's been saved.

codemanmeet:
@damithc I have changed the two lists for sample panel names and sample panel filters to a hashmap and updated header comments. Let me know if the PR is good now.

src/main/java/ui/TestController.java
```diff
@@ -86,6 +95,15 @@ public static boolean isCloseOnQuit() {
     }
 
     /**
+     * Returns true if HubTurbo is not being run on Test mode or if startup board creation is being tested.
+     * As for the other tests, they start on a clean state with dummy repos established
+     * so don't need boards created on startup.
+     */
+    public static boolean shouldOpenSampleBoard() {
+        return !TestController.isTestMode() || TestController.shouldTestStartupBoard();
```
dariusf: We no longer need to reference TestController here.

codemanmeet: Will remove the reference!

src/main/java/ui/BoardAutoCreator.java
damithc: Why do we need a dialog here? New users are going to see this dialog?

src/main/java/ui/UI.java
```diff
@@ -143,6 +143,11 @@ private void login(boolean isBypassLogin) {
         getMainWindowHandle(mainStage.getTitle());
     }
 
+    private void openSampleBoard(){
```
damithc: `open` or `create`?

src/main/java/ui/UI.java
```diff
@@ -150,6 +155,8 @@ private void disableUI(boolean disable) {
     }
 
     private void showMainWindow(String repoId) {
+        boolean isFirstTimeUser = logic.getStoredRepos().isEmpty();
```
damithc: Move this near to where the variable is used (i.e. line 184).

src/main/java/ui/UI.java
```diff
@@ -174,6 +181,9 @@ private void showMainWindow(String repoId) {
         // Should only be called after panels have been initialized
         ensureSelectedPanelHasFocus();
         initialisePickers();
+        if (isFirstTimeUser && TestController.shouldOpenSampleBoard()){
+            openSampleBoard();
+        }
```
damithc: May be this problem goes beyond this PR, but the dependency from this class to `TestController` should be in the other direction.

src/test/java/guitests/BoardAutoCreatorTest.java
```diff
@@ -105,14 +104,21 @@ public void sampleBoardAutoCreationTest() {
...
+        int i = 0;
+        for (String panelName : BoardAutoCreator.getSamplePanelDetails().keySet()) {
+            assertEquals(panelInfos.get(i).getPanelName(), panelName);
+            i++;
+        }
+
+        int j = 0;
+        for (String panelFilter : BoardAutoCreator.getSamplePanelDetails().values()) {
+            assertEquals(panelInfos.get(j).getPanelFilter(), panelFilter);
+            j++;
```
damithc: What is being tested here? Can you add a comment to explain?

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,71 @@
...
+        for (String panelName : BoardAutoCreator.getSamplePanelDetails().keySet()) {
+            assertEquals(panelInfos.get(i).getPanelName(), panelName);
+            i++;
+        }
+
+        int j = 0;
+        for (String panelFilter : BoardAutoCreator.getSamplePanelDetails().values()) {
+            assertEquals(panelInfos.get(j).getPanelFilter(), panelFilter);
+            j++;
+        }
```
damithc: This code block seems like a copy-paste. Can reuse instead?

src/main/java/ui/UI.java
codemanmeet: The sample board is opened as well as saved in the config file (when it is created and opened). Let's go ahead with `openAndSaveSampleBoard`.

src/main/java/ui/UI.java
codemanmeet: As mentioned before in the review, this has to be on this line as `logic.getStoredRepos()` is altered by `logic.openPrimaryRepository` which is just after this line. We want to capture the boolean value right on this line.

src/main/java/ui/UI.java
codemanmeet: Yup, this problem transcends beyond this PR. TestController is used throughout `UI.java`. Would have to create a separate issue for this. Will go ahead and do that.

src/test/java/guitests/BoardAutoCreatorTest.java
codemanmeet: This was a refactoring done to convert those two `Lists` to one `Map`. Will add a comment for testing here.

src/test/java/guitests/StartupBoardLauncherTest.java
codemanmeet: Yup, we can reuse :+1:

src/main/java/ui/BoardAutoCreator.java
codemanmeet: The dialog is needed when the sample board is created using `Auto Create` menu. The dialog is not seen by new users.

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -133,15 +126,20 @@ private void createWorkAllocationBoard() {
                         "It is saved under the name \"" + boardName + "\".");
     }
 
-    private void createSampleBoard() {
+    /**
+     * Creates a sample board to showcase HubTurbo's functionality with sample repos in filters.
+     * The dialog box shows a confirmation message for the saving of the sample board if isDialogShown is set to true.
```
damithc: The header comment says `saving of the sample board` but the dialog is not about saving the sample board right?

src/main/java/ui/UI.java
```diff
@@ -143,6 +143,11 @@ private void login(boolean isBypassLogin) {
         getMainWindowHandle(mainStage.getTitle());
     }
 
+    private void openAndSaveSampleBoard(){
```
damithc: Is this the correct name for this method? The code is all about 'creating' the board, not opening or saving.

src/test/java/guitests/BoardAutoCreatorTest.java
damithc: I don't think you need two loops to iterate a map.
On a related note, I think we need classes to represent information of a board (`BoardInfo`?), and a panel (`PanelInfo`?) The current code is getting messy due to lack of right abstractions.

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Test
+    public void testSampleBoardOnLaunch(){
+        selectAll();
+        type("dummy").push(KeyCode.TAB);
+        type("dummy").push(KeyCode.TAB);
+        type("test").push(KeyCode.TAB);
+        type("test");
+        click("Sign in");
+        clickMenu("File", "Logout");
```
damithc: Let me ask again: can lines 42-47 be written at a higher level? e.g. 
```
login();
logout();
```

src/main/java/ui/BoardAutoCreator.java
damithc: Got it. Showing the dialog is an additional step that you do at the end, so an alternative implementation could be like this:
```
void createSampleBoard(){...}

void createSampleBoardAndInformUser(){
  createSampleBoard();
  //show dialog
}
```
Using boolens to control behaviors is not wrong, but not pretty either.
It's ok to keep the current one.

codemanmeet: Okay, got it.

src/main/java/ui/BoardAutoCreator.java
codemanmeet: yup, the dialog is about creating and loading sample boards. have changed the message to that.

src/main/java/ui/UI.java
codemanmeet: Yup, same here. it's `createAndLoadSampleBoard`.

src/test/java/guitests/BoardAutoCreatorTest.java
codemanmeet: We already have `panelInfo` class created. I'm making use of it as well. `BoardInfo` would include a list of panelInfos. This would result in quite a bit refactoring for the existing code on boards, which is beyond the scope of this PR. 

I am now using one loop to iterate through the map.

src/test/java/guitests/StartupBoardLauncherTest.java
codemanmeet: I have abstracted it within `StartupBoardLauncherTest`.

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    private void dummyLogin(){
+        selectAll();
+        type("dummy").push(KeyCode.TAB);
+        type("dummy").push(KeyCode.TAB);
+        type("test").push(KeyCode.TAB);
+        type("test");
+        click("Sign in");
+    }
```
damithc: Generalize it into `void login(owner, repo, password)` and pushed to parent class for reuse by others??

damithc: BTW, name methods to sound like actions. e.g. `loginUsingDummyData`

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,69 @@
...
+        UI ui = TestController.getUI();
+        panelControl = ui.getPanelControl();
+        testPref = UI.prefs;
+    }
+
+    @Test
+    public void testSampleBoardOnLaunch(){
+        dummyLogin();
+        logout();
+
```
damithc: Why do we do a login and a logout? What happens after that?

codemanmeet: We check the stored sample board and its corresponding panel details after that.

codemanmeet:
Prof., made all the relevant changes. Let me know if `mergeApproved` can be applied now!

src/test/java/guitests/StartupBoardLauncherTest.java
damithc: Whoever reading the code needs to be guided through the code using comments. For starters, as a reader, I'm puzzled why a login and a logout is done at the beginning.

After that, I should see something like `assertEquals(expectedBoard, actualBoard)` but unfortunately we don't have the right asbtractions to do that, so we end up spraying assert statements all over the place.

codemanmeet: Yup, as discussed before in the previous review, we need to incorporate `BoardInfo` class in our code. It requires quite a bit of refactoring for the existing board code. 

I have added comments now to explain the test scenario!

src/main/java/ui/UI.java
```diff
@@ -150,6 +155,7 @@ private void disableUI(boolean disable) {
     }
 
     private void showMainWindow(String repoId) {
+        boolean isAFirstTimeUser = logic.getStoredRepos().isEmpty();
```
damithc: I thought I asked this line to be moved down? There is no point setting this variable until we need it, which happens at line 183.

codemanmeet: I replied that this line cannot be moved down. `logic.openPrimaryRepository` changes `logic.getStoredRepos()` and hence the check has to be made exactly at this point. The boolean value obtained here is then used to decide whether the sample board should be opened.

src/test/java/guitests/BoardAutoCreatorTest.java
```diff
@@ -105,14 +104,22 @@ public void sampleBoardAutoCreationTest() {
...
+    /**
+     * Tests whether the sample panels along with their filters are rightly created and present in the UI
+     */
+    public static void testSamplePanelInfos(PanelControl pc){
+        List<PanelInfo> panelInfos = pc.getCurrentPanelInfos();
+        int i = 0;
+        for (String panelName : BoardAutoCreator.getSamplePanelDetails().keySet()) {
+            assertEquals(panelInfos.get(i).getPanelName(), panelName);
+            assertEquals(panelInfos.get(i).getPanelFilter(), BoardAutoCreator.getSamplePanelDetails().get(panelName));
+            i++;
```
damithc: This method has high coupling because it takes more than it needs. Comparing boards doesn't need `PanelControl` or `BoardAutoCreator`. Can you do this instead? `verifySampeBoard(pc.getCurrentPanelInfos(), BoardAutoCreator.getSamplePanelDetails())`?

src/test/java/guitests/UITest.java
```diff
@@ -361,6 +361,15 @@ private void awaitCondition(Callable<Boolean> condition, int timeoutInSeconds) {
         }
     }
 
+    public void loginUsingDummyData(String owner, String repoName, String username, String password){
```
damithc: Why is it called `DaummyData`?

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,61 @@
...
+        assertEquals(panelControl.getCurrentlySelectedPanel(), Optional.of(0));
+        assertEquals(panelControl.getNumberOfSavedBoards(), 1);
+
+        assertEquals(testPref.getAllBoardNames().get(0), SAMPLE_BOARD);
+        testSamplePanelInfos(panelControl);
+    }
+
+    private void logout(){
+        clickMenu("File", "Logout");
+    }
```
damithc: push to parent too?

src/test/java/guitests/UITest.java
damithc: This is a public method meant to be reused. Give a good header comment.

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,61 @@
...
+        UI ui = TestController.getUI();
+        panelControl = ui.getPanelControl();
+        testPref = UI.prefs;
+    }
+
+    @Test
+    public void testSampleBoardOnLaunch(){
+        // First we login and logout for first time users.
+        loginUsingDummyData("dummy", "dummy", "test", "test");
+        logout();
```
damithc: Why logout?
How does the test ensure this is simulates a first time user?

codemanmeet: It ensures that using `FXTestUtils.launchApp(TestUI.class, "--startupboard=true");`

src/test/java/guitests/StartupBoardLauncherTest.java
codemanmeet: Okay.

src/main/java/ui/UI.java
damithc: OK, that makes sense. No harm adding a comment to that effect.

src/test/java/guitests/StartupBoardLauncherTest.java
codemanmeet: We could do away with `logout` but I think it's always a good practice to tear down tests upon completion.

src/test/java/guitests/UITest.java
codemanmeet: Okay

src/main/java/ui/UI.java
codemanmeet: Will do :+1:

src/main/java/ui/UI.java
```diff
@@ -150,6 +155,8 @@ private void disableUI(boolean disable) {
     }
 
     private void showMainWindow(String repoId) {
+        //Check whether there are existing stored repos. If not, it's indicative of first time users of Hubturbo.
+        boolean isAFirstTimeUser = logic.getStoredRepos().isEmpty();
```
damithc: The comment simply repeats what the statement says. What's missing is WHY this statement has to be top of the method. That was the confusion I had and that's what I'm trying to avoid in future readers.

src/test/java/guitests/StartupBoardLauncherTest.java
damithc: If logout is part of teardown, then it should be put in the teardown method.

src/test/java/guitests/BoardAutoCreatorTest.java
```diff
@@ -105,14 +105,22 @@ public void sampleBoardAutoCreationTest() {
...
+    }
 
-        List<PanelInfo> panelInfos = panelControl.getCurrentPanelInfos();
-        for (int i = 0; i < SAMPLE_PANEL_NAMES.size(); i++){
-            assertEquals(panelInfos.get(i).getPanelFilter(), SAMPLE_PANEL_FILTERS.get(i));
-            assertEquals(panelInfos.get(i).getPanelName(), SAMPLE_PANEL_NAMES.get(i));
+    /**
+     * Tests whether the sample panels along with their filters are rightly created and present in the UI
+     */
+    public static void verifySampleBoard(PanelControl pc, Map<String, String> samplePanelDetails){
```
damithc: 1. Why does the method take `PanelControl` as the parameter instead of `pc.getCurrentPanelInfos()`?
2. If you change it to take `pc.getCurrentPanelInfos()`, the method becomes more general, so the name/comment should not mention 'sample board'

Alternatively, you could feed both `PanelControl` and `BoardAutoCreator` into this method in which case all asserts should be done inside method. At the moment, some asserts are done outside and some inside.

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,54 @@
...
+    @Before
+    public void init(){
+        UI ui = TestController.getUI();
+        panelControl = ui.getPanelControl();
+        testPref = UI.prefs;
+    }
+
+    @Test
+    public void testSampleBoardOnLaunch(){
+        //First we login and then logout for first time users.
```
damithc: This comment is redundant. Remember, we should comment about WHY, now HOW. The HOW should be obvious from the code.

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,54 @@
...
+        login("dummy", "dummy", "test", "test");
+        logout();
```
damithc: Move the logout to teardown()?

src/test/java/guitests/UITest.java
```diff
@@ -362,6 +362,29 @@ private void awaitCondition(Callable<Boolean> condition, int timeoutInSeconds) {
...
 
     /**
+     * Performs logout from File -> Logout on HubTurbo's pView.
+     */
+    public void logout(){
+        clickMenu("File", "Logout");
+    }
+
+    /**
+     * Performs UI login on the dialog box upon startup.
```
damithc: Does it have to be `upon startup` or it can be anytime the login dialog is displayed?

src/main/java/ui/UI.java
```diff
@@ -150,6 +155,8 @@ private void disableUI(boolean disable) {
     }
 
     private void showMainWindow(String repoId) {
+        //logic.openPrimaryRepository changes logic.getStoredRepos so the check for first time users is done here.
```
damithc: What do you mean `changes logic.getStoredRepos`?
```
//We infer this is the first time HT is being used if there are no repo data stored at the start up.
//This check needs to be done at the very beginning of the startup, before HT downloads any repo data.
```

src/test/java/guitests/BoardAutoCreatorTest.java
```diff
@@ -105,14 +105,21 @@ public void sampleBoardAutoCreationTest() {
...
         assertEquals(panelControl.getNumberOfSavedBoards(), 1);
+        verifyBoard(panelControl.getCurrentPanelInfos(), BoardAutoCreator.getSamplePanelDetails());
+    }
 
-        List<PanelInfo> panelInfos = panelControl.getCurrentPanelInfos();
-        for (int i = 0; i < SAMPLE_PANEL_NAMES.size(); i++){
-            assertEquals(panelInfos.get(i).getPanelFilter(), SAMPLE_PANEL_FILTERS.get(i));
-            assertEquals(panelInfos.get(i).getPanelName(), SAMPLE_PANEL_NAMES.get(i));
+    /**
+     * Tests whether the panels along with their filters are rightly created and present in the UI
```
damithc: No it doesn't. The current implementation checks if the panels specified in one parameter is a subset of the other.

src/test/java/guitests/BoardAutoCreatorTest.java
```diff
@@ -105,14 +105,21 @@ public void sampleBoardAutoCreationTest() {
         waitUntilNodeAppears(SAMPLE_BOARD_DIALOG);
         click("OK");
 
-        assertEquals(panelControl.getPanelCount(), SAMPLE_PANEL_NAMES.size());
+        assertEquals(panelControl.getPanelCount(), BoardAutoCreator.getSamplePanelDetails().size());
```
damithc: This can be done inside the `verifyBoard` method right?

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,53 @@
...
+    public void testSampleBoardOnLaunch(){
+        login("dummy", "dummy", "test", "test");
+        waitUntilNodeAppears(panelControl.getPanel(0));
+        assertEquals(panelControl.getPanelCount(), BoardAutoCreator.getSamplePanelDetails().size());
+        assertEquals(panelControl.getCurrentlySelectedPanel(), Optional.of(0));
+        assertEquals(panelControl.getNumberOfSavedBoards(), 1);
+
+        assertEquals(testPref.getAllBoardNames().get(0), SAMPLE_BOARD);
+        verifyBoard(panelControl.getCurrentPanelInfos(), BoardAutoCreator.getSamplePanelDetails());
+    }
```
damithc: Doesn't this test make the one in `BoardAutoCreatorTest.java` redundant?

src/test/java/guitests/UITest.java
```diff
@@ -362,6 +362,29 @@ private void awaitCondition(Callable<Boolean> condition, int timeoutInSeconds) {
...
+    public void logout(){
+        clickMenu("File", "Logout");
+    }
+
+    /**
+     * Performs UI login on the login dialog box.
+     * @param owner
+     * @param repoName
+     * @param username
+     * @param password
```
damithc: Side question: can you check if these 4 lines add any value when the IDE generates popup documentation for method calls?

src/test/java/guitests/StartupBoardLauncherTest.java
codemanmeet: Both are needed as that one checks whether the board was created correctly through the `Auto Create` menu. This one is for new users.

src/test/java/guitests/BoardAutoCreatorTest.java
codemanmeet: Okay, will do

src/test/java/guitests/UITest.java
codemanmeet: Yup, it is useful. The @param is actually auto-generated by Intellij (which performs this according to Javadoc standards). The documentation appears like this - 
``java
guitests.UITest
public void login(String owner,
                  String repoName,
                  String username,
                  String password)
Performs UI login on the login dialog box.
Parameters:
owner - The owner of the repo.
repoName - The repository name
username - The Github/dummy username
password - The Github/dummy password
``

src/test/java/guitests/BoardAutoCreatorTest.java
```diff
@@ -104,15 +104,23 @@ public void sampleBoardAutoCreationTest() {
...
-        assertEquals(panelControl.getCurrentlySelectedPanel(), Optional.of(0));
-        assertEquals(panelControl.getNumberOfSavedBoards(), 1);
-
-        List<PanelInfo> panelInfos = panelControl.getCurrentPanelInfos();
-        for (int i = 0; i < SAMPLE_PANEL_NAMES.size(); i++){
-            assertEquals(panelInfos.get(i).getPanelFilter(), SAMPLE_PANEL_FILTERS.get(i));
-            assertEquals(panelInfos.get(i).getPanelName(), SAMPLE_PANEL_NAMES.get(i));
+    /**
+     * Tests panelControl properties associated with panels.
+     * panelDetails obtained from Boards are also tested as a subset of the panels in PanelControl.
```
damithc: Why a subset? I said 'subset' earlier because that was what the method was doing. Is this method checking for subset or exact match?

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,47 @@
...
+    public void init(){
+        UI ui = TestController.getUI();
+        panelControl = ui.getPanelControl();
+        testPref = UI.prefs;
+    }
+
+    @Test
+    public void boardCreation_firstTimeUser_sampleBoardCreated(){
+        login("dummy", "dummy", "test", "test");
+        teardown();
```
damithc: Why call teardown here? teardown is supposed to happen at the end of the test. Instead of calling teardown, use a junit annotation for the teardown method.

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,47 @@
...
+        //Ensures that only 1 board was created and it was the sample board
+        assertEquals(testPref.getAllBoardNames().size(), 1);
+        assertEquals(testPref.getAllBoardNames().get(0), SAMPLE_BOARD);
+        //Verifies the details of the sample board created.
+        verifyBoard(panelControl, BoardAutoCreator.getSamplePanelDetails());
```
damithc: Add blank lines between logical blocks e.g.
```java
login("dummy", "dummy", "test", "test");

//Ensures that only 1 board was created and it was the sample board
assertEquals(testPref.getAllBoardNames().size(), 1);
assertEquals(testPref.getAllBoardNames().get(0), SAMPLE_BOARD);
 
//Verifies the details of the sample board created.
verifyBoard(panelControl, BoardAutoCreator.getSamplePanelDetails());
```

src/test/java/guitests/UITest.java
```diff
@@ -362,6 +362,29 @@ private void awaitCondition(Callable<Boolean> condition, int timeoutInSeconds) {
...
 
     /**
+     * Performs logout from File -> Logout on HubTurbo's pView.
+     */
+    public void logout(){
+        clickMenu("File", "Logout");
+    }
+
+    /**
+     * Performs UI login on the login dialog box. Assumes the cursor is on the login dialog box but nothing selected.
```
damithc: Is this condition really necessary ` box but nothing selected`?

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,47 @@
...
+import ui.issuepanel.PanelControl;
+import static guitests.BoardAutoCreatorTest.verifyBoard;
+
+import static org.junit.Assert.assertEquals;
+import static ui.BoardAutoCreator.SAMPLE_BOARD;
+
+public class StartupBoardLauncherTest extends UITest{
+
+    private PanelControl panelControl;
+    private Preferences testPref;
```
damithc: Any reason why this is called `testPref`?

src/test/java/guitests/BoardAutoCreatorTest.java
codemanmeet: It is testing the exact match.

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,47 @@
...
+        testPref = UI.prefs;
+    }
+
+    @Test
+    public void boardCreation_firstTimeUser_sampleBoardCreated(){
+        login("dummy", "dummy", "test", "test");
+        teardown();
+        //Ensures that only 1 board was created and it was the sample board
+        assertEquals(testPref.getAllBoardNames().size(), 1);
+        assertEquals(testPref.getAllBoardNames().get(0), SAMPLE_BOARD);
```
damithc: See if there is a way to do this two things in one shot. I have a feeling there is a more elegant way to check the list of boards contains only sample board. I could be wrong.

src/test/java/guitests/UITest.java
codemanmeet: Nope it's not necessary. If nothing is selected, `selectAll` will select the whole filter text anyways.

src/test/java/guitests/StartupBoardLauncherTest.java
codemanmeet: It's a private preferences variable called in the test class, hence `testPref`. It's called the same in a number of other test classes like `RepositorySelectorTest`, `MenuControlTest` etc.

src/test/java/guitests/BoardAutoCreatorTest.java
```diff
@@ -104,15 +104,23 @@ public void sampleBoardAutoCreationTest() {
...
-        assertEquals(panelControl.getCurrentlySelectedPanel(), Optional.of(0));
-        assertEquals(panelControl.getNumberOfSavedBoards(), 1);
-
-        List<PanelInfo> panelInfos = panelControl.getCurrentPanelInfos();
-        for (int i = 0; i < SAMPLE_PANEL_NAMES.size(); i++){
-            assertEquals(panelInfos.get(i).getPanelFilter(), SAMPLE_PANEL_FILTERS.get(i));
-            assertEquals(panelInfos.get(i).getPanelName(), SAMPLE_PANEL_NAMES.get(i));
+    /**
+     * Tests panelControl properties associated with panels.
+     * PanelDetails obtained from Boards are tested against current Panel infos as well.
```
damithc: I didn't think writing a proper comment for a method takes this many iterations :-p
```
Confirms the currently displayed board consists the set of panels specified in panelDetails
```
May need to rename the method to match the description `verifyCurrentBoard`

src/test/java/guitests/StartupBoardLauncherTest.java
damithc: By the same logic, the previous variable should be called `testPanelControl`? Is it a preference object specifically created for testing? If that is the case, it can be called `testPref`. If it is a regular preference object that happens to be used during testing, then I don't see why it is prefixed with `test` when other variables are not.

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,47 @@
...
+    public void launchApp(){
+        FXTestUtils.launchApp(TestUI.class, "--startupboard=true");
+    }
+
+    @Before
+    public void init(){
+        UI ui = TestController.getUI();
+        panelControl = ui.getPanelControl();
+        testPref = UI.prefs;
+    }
```
damithc: looks like you don't need these three variables at all. 
Just use this at line 41?
```java
List<String> boardNames = TestController.getUI().getPanelControl().getAllBoardNames();
```
You can probably follow it up with this (not sure)
```java
assertEquals(boardNames, Arrays.asList(new String[]{SAMPLE_BOARD})
```

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Test
+    public void boardCreation_firstTimeUser_sampleBoardCreated(){
+        login("dummy", "dummy", "test", "test");
+        // Workaround since we are unable to synchronize board creation on Travis post login through the dialog box.
```
damithc: I don't think anybody who didn't listen to your conversation with darius will be able to understand this phrase `unable to synchronize board creation on Travis post login through the dialog box` 
Comments are not 'notes to self'. Someone else should be able to make sense out of them.
@dariusf why can't logout be at line 46 instead of here?

dariusf: I'm not entirely sure, but I'm working with Manmeet on the issue. Our conversation earlier was on a slightly different problem.

codemanmeet: After discussing with @dariusf , it's been rectified through `countPanelsShown` which takes a `FutureTask<Integer>` and performs a `PlatformEx.runAndWait` on it. `TestUtils.awaitCondition` is then run on it until all the panels are shown on the UI. Thanks @dariusf !

src/test/java/guitests/StartupBoardLauncherTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Before
+    public void init(){
+        panelControl = TestController.getUI().getPanelControl();
+    }
+
+    @Test
+    public void boardCreation_firstTimeUser_sampleBoardCreated(){
+        login("dummy", "dummy", "test", "test");
+        TestUtils.awaitCondition(() -> BoardAutoCreator.getSamplePanelDetails().size() == countPanelsShown());
```
dariusf: @damithc we've removed the `logout()` by synchronising properly when checking UI state. I'm hoping to add an abstraction for this similar in spirit to [this](https://github.com/HubTurbo/HubTurbo/blob/952-nondeterministic-tests/src/test/java/guitests/UITest.java#L360-L366), but haven't worked out details and best practices yet. Could we leave this here for now (with a comment?) and abstract it away when we know what exactly is required?

damithc: Sure. This line shows we are waiting for something to complete, which is not surprising in UI tests and is unlikely to confuse the reader. An unexpected logout with an incomprehensible comment is not.

dariusf:
@codemanmeet this is ready. Could you squash the unnecessary commits first? Also coverage isn't working for some reason.

Issue #1230, opened 2016-02-01
New Updater - Create a new process that will update JAR and restart the application


HansNewbie:
The new updater should be able to restart the application once download of the new JAR finishes successfully (and replace the old JAR with the new JAR in the background).

HansNewbie:
Issue should have been closed in #1304. Not sure why it wasn't.

PR #1249, opened 2016-02-13
New Updater - Able to download file #1229

### LINKED ISSUE #1229: New Updater - Able to download file
This issue is part of the addition of new updater inside HubTurbo itself. The updater should be able to download file given a URL and it can show a download progress window.

HansNewbie:
Fixes #1229

### DIFF
--- a/src/main/java/ui/MenuControl.java
+++ b/src/main/java/ui/MenuControl.java
@@ -63,7 +63,9 @@ public class MenuControl extends MenuBar {
                 createRefreshMenuItem(),
                 createDocumentationMenuItem());
 
-        getMenus().addAll(file, newMenu, panels, boards, repos, view);
+        Menu update = createUpdateMenu();
+
+        getMenus().addAll(file, newMenu, panels, boards, repos, view, update);
     }
 
     private Menu createFileMenu() {
@@ -229,7 +231,7 @@ public class MenuControl extends MenuBar {
             delete.getItems().clear();
 
             Map<String, List<PanelInfo>> boards = prefs.getAllBoards();
-            
+
             for (Map.Entry<String, List<PanelInfo>> entry : boards.entrySet()) {
                 final String boardName = entry.getKey();
                 final List<PanelInfo> panelSet = entry.getValue();
@@ -345,5 +347,19 @@ public class MenuControl extends MenuBar {
 
     private void onRepoRemove(String repoId) {
         ui.logic.removeStoredRepository(repoId).thenRun(() -> ui.triggerEvent(new UnusedStoredReposChangedEvent()));
+    }
+
+    private Menu createUpdateMenu() {
+        Menu update = new Menu("Update");
+
+        MenuItem checkProgress = new MenuItem("Check progress...");
+
+        checkProgress.setOnAction(e -> {
+            ui.showUpdateProgressWindow();
+        });
+
+        update.getItems().addAll(checkProgress);
+
+        return update;
     }
 }
--- a/src/main/java/ui/TestController.java
+++ b/src/main/java/ui/TestController.java
@@ -7,6 +7,8 @@ import backend.json.JSONStoreStub;
 import backend.stub.DummySource;
 import javafx.application.Application;
 import prefs.Preferences;
+import updater.UpdateManager;
+import updater.UpdateManagerStub;
 
 import java.util.HashMap;
 import java.util.Optional;
@@ -156,5 +158,18 @@ public final class TestController {
     public static RepoIO createTestingRepoIO(Optional<JSONStore> jsonStoreToBeUsed) {
         return new RepoIO(Optional.of(new DummySource()), jsonStoreToBeUsed,
                           Optional.of(RepoStore.TEST_DIRECTORY));
+    }
+
+    /**
+     * Creates update manager if not in test mode, its stub otherwise.
+     * @return update manager to be used
+     */
+    public static UpdateManager createUpdateManager() {
+        if (isTestMode()) {
+            return new UpdateManagerStub();
+        } else {
+            UpdateProgressWindow updateProgressWindow = new UpdateProgressWindow();
+            return new UpdateManager(updateProgressWindow);
+        }
     }
 }
--- a/src/main/java/ui/UI.java
+++ b/src/main/java/ui/UI.java
@@ -34,6 +34,8 @@ import ui.components.StatusUI;
 import ui.components.pickers.LabelPicker;
 import ui.issuepanel.PanelControl;
 import undo.UndoController;
+import updater.UpdateManager;
+import updater.UpdateManagerStub;
 import util.*;
 import util.events.*;
 import util.events.Event;
@@ -83,6 +85,7 @@ public class UI extends Application implements EventDispatcher {
     public GUIController guiController;
     private NotificationController notificationController;
     public UndoController undoController;
+    public UpdateManager updateManager;
 
 
     // Main UI elements
@@ -217,6 +220,9 @@ public class UI extends Application implements EventDispatcher {
 
         uiManager = new UIManager(this);
         status = new HTStatusBar(this);
+
+        updateManager = TestController.createUpdateManager();
+        updateManager.run();
     }
 
     private void initApplicationState() {
@@ -305,7 +311,7 @@ public class UI extends Application implements EventDispatcher {
         scene.getStylesheets().add(css);
     }
 
-    public static void loadFonts(){
+    public static void loadFonts() {
         Font.loadFont(UI.class.getResource("octicons/octicons-local.ttf").toExternalForm(), 32);
     }
 
@@ -605,5 +611,9 @@ public class UI extends Application implements EventDispatcher {
 
     public boolean isWindowFocused() {
         return mainStage.isFocused();
+    }
+    
+    public void showUpdateProgressWindow() {
+        updateManager.showUpdateProgressWindow();
     }
 }
--- a/src/main/java/ui/UpdateProgressWindow.java
+++ b/src/main/java/ui/UpdateProgressWindow.java
@@ -1,0 +1,142 @@
+package ui;
+
+import javafx.application.Platform;
+import javafx.geometry.Insets;
+import javafx.geometry.Pos;
+import javafx.scene.Group;
+import javafx.scene.Scene;
+import javafx.scene.control.Label;
+import javafx.scene.control.ProgressBar;
+import javafx.scene.layout.VBox;
+import javafx.stage.Stage;
+import updater.LabeledDownloadProgressBar;
+
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Represents the window showing the progress of the downloading of the new update
+ */
+public class UpdateProgressWindow {
+    private static final String WINDOW_TITLE = "Update Progress";
+    private static final String LABEL_NO_DOWNLOAD = "There is no update being downloaded.";
+
+    private Stage window;
+    private VBox windowMainLayout;
+    private final List<LabeledDownloadProgressBar> downloads;
+
+    public UpdateProgressWindow() {
+        this.downloads = new ArrayList<>();
+        this.initProgressWindow();
+    }
+
+    public void showWindow() {
+        Platform.runLater(() -> {
+                    window.show();
+                    window.toFront();
+                }
+        );
+    }
+
+    public void hideWindow() {
+        Platform.runLater(window::hide);
+    }
+
+    private void initProgressWindow() {
+        Stage stage = new Stage();
+        stage.setTitle(WINDOW_TITLE);
+
+        windowMainLayout = new VBox();
+
+        Group root = new Group();
+        Scene scene = new Scene(root);
+        stage.setScene(scene);
+
+        scene.setRoot(windowMainLayout);
+
+        window = stage;
+
+        reloadWindowLayout();
+    }
+
+    /**
+     * Creates a new download progress bar
+     *
+     * @param downloadUri uri of download as identifier to the progress bar
+     * @param downloadLabel label of progress bar
+     * @return the download progress bar created
+     */
+    public LabeledDownloadProgressBar createNewDownloadProgressBar(URI downloadUri, String downloadLabel) {
+        ProgressBar progressBar = new ProgressBar(-1.0);
+        progressBar.setPrefWidth(400);
+        LabeledDownloadProgressBar downloadProgressBar =
+                new LabeledDownloadProgressBar(downloadUri, downloadLabel, progressBar);
+
+        downloads.add(downloadProgressBar);
+
+        reloadWindowLayout();
+
+        return downloadProgressBar;
+    }
+
+    /**
+     * Removes a download progress bar from update progress window
+     *
+     * This should be called after a download is completed
+     *
+     * @param downloadUri download URI to be used as download identifier
+     */
+    public void removeDownloadProgressBar(URI downloadUri) {
+        downloads.removeIf(dpt -> dpt.getUriIdentifier().equals(downloadUri));
+
+        reloadWindowLayout();
+    }
+
+    /**
+     * Updates the update progress window's layout if there is any changes to download progress to be shown
+     */
+    private void reloadWindowLayout() {
+        assert window != null && windowMainLayout != null;
+
+        Platform.runLater(() -> {
+            windowMainLayout.getChildren().clear();
+
+            if (downloads.isEmpty()) {
+                Label noDownloadLabel = createNoDownloadLabel();
+
+                windowMainLayout.getChildren().add(noDownloadLabel);
+            } else {
+                for (LabeledDownloadProgressBar progressTracker : downloads) {
+                    VBox downloadProgressItem = createDownloadProgressItem(progressTracker);
+
+                    windowMainLayout.getChildren().add(downloadProgressItem);
+                }
+            }
+        });
+    }
+
+    private VBox createDownloadProgressItem(LabeledDownloadProgressBar progressTracker) {
+        Label downloadLabel = new Label();
+        downloadLabel.setText(progressTracker.getDownloadLabel());
+
+        ProgressBar progressBar = progressTracker.getProgressBar();
+
+        VBox downloadProgressItem = new VBox();
+        downloadProgressItem.setSpacing(20);
+        downloadProgressItem.setPadding(new Insets(20));
+        downloadProgressItem.setAlignment(Pos.CENTER_LEFT);
+
+        downloadProgressItem.getChildren().addAll(downloadLabel, progressBar);
+
+        return downloadProgressItem;
+    }
+
+    private Label createNoDownloadLabel() {
+        Label noDownloadLabel = new Label();
+        noDownloadLabel.setText(LABEL_NO_DOWNLOAD);
+        noDownloadLabel.setPadding(new Insets(50));
+
+        return noDownloadLabel;
+    }
+}
--- a/src/main/java/updater/DownloadCountingOutputStream.java
+++ b/src/main/java/updater/DownloadCountingOutputStream.java
@@ -1,0 +1,48 @@
+package updater;
+
+import javafx.beans.property.DoubleProperty;
+import javafx.beans.property.SimpleDoubleProperty;
+import javafx.beans.value.ObservableValue;
+import org.apache.commons.io.output.CountingOutputStream;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.function.Consumer;
+
+/**
+ * This class counts the bytes that have been downloaded and tell its listener(s) about current percentage downloaded
+ */
+public class DownloadCountingOutputStream extends CountingOutputStream {
+    private final long totalFileSizeInBytes;
+    private final DoubleProperty percentageDownloaded; // 0.0 to 1.0
+
+    /**
+     *
+     * @param out output stream
+     * @param totalFileSizeInBytes total file size to be downloaded
+     */
+    public DownloadCountingOutputStream(OutputStream out, long totalFileSizeInBytes) {
+        super(out);
+
+        this.totalFileSizeInBytes = totalFileSizeInBytes;
+        percentageDownloaded = new SimpleDoubleProperty(0);
+    }
+
+    public void addListener(Consumer<Double> listener) {
+        percentageDownloaded.addListener(
+                (ObservableValue<? extends Number> observable, Number oldValue, Number newValue) ->
+                        listener.accept(newValue.doubleValue()));
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * Updates percentage downloaded
+     */
+    @Override
+    protected void afterWrite(int n) throws IOException {
+        super.afterWrite(n);
+
+        percentageDownloaded.set(1.0 * this.getByteCount() / this.totalFileSizeInBytes);
+    }
+}
--- a/src/main/java/updater/FileDownloader.java
+++ b/src/main/java/updater/FileDownloader.java
@@ -1,0 +1,104 @@
+package updater;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.io.*;
+import java.net.URI;
+import java.net.URLConnection;
+import java.util.function.Consumer;
+
+/**
+ * A class in charge of downloading file.
+ */
+public class FileDownloader {
+    private static final Logger logger = LogManager.getLogger(FileDownloader.class.getName());
+
+    public static final int CONNECTION_TIMEOUT = 15000;
+    public static final int READ_CONNECTION_TIMEOUT = 30000;
+
+    private final URI source;
+    private final File dest;
+    private long totalFileSizeInBytes;
+
+    private final Consumer<Double> onDownload;
+
+    /**
+     * @param source URI of file source, expect to be a valid URL
+     * @param dest local destination file
+     * @param onDownload callback that listen to download stream
+     */
+    public FileDownloader(URI source, File dest, Consumer<Double> onDownload) {
+        this.source = source;
+        this.dest = dest;
+        this.onDownload = onDownload;
+    }
+
+    /**
+     * Starts download from source to destination as specified in constructor
+     *
+     * @return true if download is successful, else false
+     */
+    public boolean download() {
+        URLConnection sourceConnection = null;
+
+        try {
+            sourceConnection = setupConnection(source);
+        } catch (IOException e) {
+            logger.error("URI of source file is not a well-formed URL");
+            return false;
+        }
+
+        assert sourceConnection != null;
+
+        totalFileSizeInBytes = sourceConnection.getContentLengthLong();
+
+        try (
+                InputStream inputStream = setupStreamFromSource(sourceConnection);
+                OutputStream outputStream = setupStreamToDest(dest);
+        ) {
+            downloadStream(inputStream, outputStream);
+        } catch (IOException e) {
+            logger.error("Failed to create streams for download or socket timeout", e);
+            return false;
+        }
+
+        return true;
+    }
+
+    private URLConnection setupConnection(URI source) throws IOException {
+        URLConnection connection = source.toURL().openConnection();
+        connection.setConnectTimeout(CONNECTION_TIMEOUT);
+        connection.setReadTimeout(READ_CONNECTION_TIMEOUT);
+
+        return connection;
+    }
+
+    private InputStream setupStreamFromSource(URLConnection source) throws IOException {
+        InputStream inputStream = source.getInputStream();
+
+        return inputStream;
+    }
+
+    private OutputStream setupStreamToDest(File dest) throws IOException {
+        if (dest.exists() && !dest.delete()) {
+            throw new IOException("Failed to delete old download file");
+        }
+
+        if (dest.createNewFile()) {
+            return new FileOutputStream(dest);
+        } else {
+            throw new IOException("Failed to create new download file");
+        }
+    }
+
+    private int downloadStream(InputStream inputStream, OutputStream outputStream) throws IOException {
+        DownloadCountingOutputStream countingOutputStream =
+                new DownloadCountingOutputStream(outputStream, totalFileSizeInBytes);
+
+        countingOutputStream.addListener(onDownload);
+
+        return IOUtils.copy(inputStream, countingOutputStream);
+    }
+}
--- a/src/main/java/updater/UpdateManager.java
+++ b/src/main/java/updater/UpdateManager.java
@@ -1,0 +1,152 @@
+package updater;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import ui.UpdateProgressWindow;
+
+import java.io.File;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+/**
+ * The class that will handle updating of HubTurbo application
+ */
+public class UpdateManager {
+    private static final Logger logger = LogManager.getLogger(UpdateManager.class.getName());
+
+    // Error messages
+    private static final String ERROR_INIT_UPDATE = "Failed to initialize update";
+    private static final String ERROR_DOWNLOAD_UPDATE_DATA = "Failed to download update data";
+    private static final String ERROR_DOWNLOAD_UPDATE_APP = "Failed to download updated application";
+
+    // Directories and file names
+    private static final String UPDATE_DIRECTORY = "updates";
+    private static final String UPDATE_SERVER_DATA_NAME =
+            "https://raw.githubusercontent.com/HubTurbo/AutoUpdater/master/HubTurbo.xml";
+    private static final String UPDATE_LOCAL_DATA_NAME = "HubTurbo.json";
+    private static final String UPDATE_APP_NAME = "HubTurbo.jar";
+
+    // Class member variables
+    private final UpdateProgressWindow updateProgressWindow;
+
+    private final ExecutorService pool = Executors.newSingleThreadExecutor();
+
+    public UpdateManager(UpdateProgressWindow updateProgressWindow) {
+        this.updateProgressWindow = updateProgressWindow;
+    }
+
+    /**
+     * Driver method to trigger UpdateManager to run. Update will be run on another thread.
+     *
+     * - Run is not automatic upon instancing the class in case there would like to be conditions on when to run update,
+     *   e.g. only if user is logged in
+     */
+    public void run() {
+        pool.execute(() -> runUpdate());
+    }
+
+    /**
+     * Runs update sequence.
+     */
+    private void runUpdate() {
+        // Fail if folder cannot be created
+        if (!initUpdate()) {
+            logger.error(ERROR_INIT_UPDATE);
+            return;
+        }
+
+        if (!downloadUpdateData()) {
+            logger.error(ERROR_DOWNLOAD_UPDATE_DATA);
+            return;
+        }
+
+        // TODO check if there is a new update since last update
+        // - if there isn't, check if any update has not been applied
+        // - if there is, download update according to user preference, i.e. auto or prompted
+
+        if (!downloadUpdateForApplication()) {
+            logger.error(ERROR_DOWNLOAD_UPDATE_APP);
+            return;
+        }
+
+        // TODO prompt user for restarting application to apply update
+        // If yes, quit application and run new process that will:
+        // - replace JAR
+        // - start the new JAR
+
+    }
+
+    /**
+     * Initializes system for updates
+     * - Creates directory(ies) for updates
+     */
+    private boolean initUpdate() {
+        File updateDir = new File(UPDATE_DIRECTORY);
+
+        if (!updateDir.exists() && !updateDir.mkdirs()) {
+            logger.error("Failed to create update directories");
+            return false;
+        }
+
+        return true;
+    }
+
+    /**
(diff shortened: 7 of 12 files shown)


src/main/java/updater/FileDownloader.java
```diff
@@ -0,0 +1,123 @@
...
+import util.events.Event;
+import util.events.UpdateManagerProgressWindowEvent;
+
+import java.io.*;
+import java.net.URI;
+import java.net.URLConnection;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+public class FileDownloader {
```
damithc: Add header comments to classes and methods.

HansNewbie: Yes I will.

damithc: Sorry, didn't realize it's still ongoing.

src/main/java/updater/UpdateManager.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /**
+     * Driver method to trigger UpdateManager to run. Update will be run on another thread.
+     *
+     * - Run is not automatic upon instancing the class in case there would like to be conditions on when to run update,
+     *   e.g. only if user is logged in
+     */
+    public void run() {
+        if (!TestController.isTestMode()) {
```
HansNewbie: @dariusf I added this here so that update manager won't run in test mode (since each test will try to download update and it may send repeated requests to GitHub server). Is this okay or do you have other suggestion in mind?

HansNewbie:
@dariusf can help me to review this?

src/main/java/updater/DownloadCountingOutputStream.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Optional;
+
+/**
+ * This class is the subject in observer pattern where the observer is a progress tracker.
+ */
+public class DownloadCountingOutputStream extends CountingOutputStream {
+    private final ProgressTracker progressTracker;
```
dariusf: Rather than take a concrete UI element to mutate, maybe this class could publish download progress in some other way (a callback, or an observable property, JavaFX-style), which another class could then make use of. Makes things a bit more decoupled.

src/main/java/updater/FileDownloader.java
```diff
@@ -0,0 +1,161 @@
...
+        progressTracker = new ProgressTracker();
+    }
+
+    /**
+     * Set if current download can have its progress shown
+     *
+     * @param enableProgressWindow
+     */
+    public void setEnableProgressWindow(boolean enableProgressWindow) {
+        UI.events.triggerEvent(new UpdateManagerProgressWindowEvent(enableProgressWindow));
```
dariusf: Same thing here -- if the class knows about the update manager's concrete event, we can't stub it out, and it's implicitly coupled to whatever is triggering the event.

src/main/java/updater/FileDownloader.java
```diff
@@ -0,0 +1,161 @@
...
+        this.progressTracker.setProgressWindowDetails(windowTitle, windowMessage);
+    }
+
+    /**
+     * Download file from a URL to a local destination file
+     *
+     * @param source URI of file source, expect to be a valid URL
+     * @param dest local destination file
+     */
+    public void downloadFileFromUrlToFile(URI source, File dest) {
```
dariusf: Can this class download multiple files simultaneously, or be used multiple times to download a file? If not, maybe these parameters should be taken in the constructor, so each FileDownloader maps to one download, one file, and (eventually) one progress bar. Then this method could take no parameters and just start the download.

dariusf:
Currently, FileDownloader maintains a reference to a UI element (ProgressTracker), and passes it to the the DownloadCountingOutputStream it creates. Both update the UI element directly.

I'm thinking FileDownloader could represent a file that's being downloaded instead (DownloadFile?). It would still internally use the DownloadCountingOutputStream to know download progress, but notify other classes through some other means. The ProgressTracker would then hook into it and display how it's progressing, and would be managed by the UI.

This is less coupled and can be composed a bit better. For example, if we wanted to download a bunch of things concurrently, that could be done with many DownloadFiles and a single window which observes all of them.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -97,6 +97,8 @@ public static FilterExpression replaceMilestoneAliases(IModel model, FilterExpre
                 .sorted((a, b) -> getMilestoneDueDateComparator().compare(a, b))
                 .collect(Collectors.toList());
 
+        model.getMilestones().stream().filter(ms ->ms.isOpen()).count();
```
dariusf: Leftover from something else?

src/main/java/updater/FileDownloader.java
```diff
@@ -0,0 +1,161 @@
...
+        InputStream inputStream = null;
+        OutputStream outputStream = null;
+        progressTracker.setProgressValue(-1);
+
+        URLConnection sourceConnection = null;
+
+        try {
+            sourceConnection = setupDownloadConnection(source);
+        } catch (IOException e) {
+            logger.error("URI of source file is not a well-formed URL");
```
dariusf: Should we `return` here if there is a problem? Seeing as there's an assertion right after that would be tripped if this failed.

src/main/java/updater/FileDownloader.java
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        assert sourceConnection != null;
+
+        totalFileSizeInBytes = sourceConnection.getContentLengthLong();
+
+        try {
+            inputStream = setupStreamFromSource(sourceConnection);
+            outputStream = setupStreamToDest(dest);
```
dariusf: Could we use [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) here? Then there's no need for `finally`. closeIOStreams wouldn't be needed any more.

src/main/java/updater/FileDownloader.java
```diff
@@ -0,0 +1,161 @@
...
+
+            assert inputStream != null && outputStream != null;
+
+            download(inputStream, outputStream);
+        } catch (Exception e) {
+            e.printStackTrace();
```
dariusf: We should be specific about which exceptions we expect here. Also it would be good to log through the usual channel.

src/main/java/updater/ProgressTracker.java
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * A class that track progress of download. It acts as a observer to a download output stream.
+ */
+public class ProgressTracker {
+    private ProgressBar progressBar;
+    private String progressWindowTitle;
+    private String progressWindowMessage;
+
+    private Stage progressWindow;
```
dariusf: We can use an Optional for this, so we can't forget to check if it's there.

src/main/java/updater/UpdateManager.java
dariusf: Checking for test mode is fine, but maybe we should do it when this method is called, rather than in here -- the environment it's run in doesn't seem like a concern of this class.

src/main/java/updater/UpdateManager.java
```diff
@@ -0,0 +1,151 @@
...
+     * 1. Initialize system for update, e.g. make directory
+     * 2. Check if there is any new update
+     *   2.a. If there is no new update but the previously downloaded update is not applied yet,
+     *        prompt user for updating and restarting application
+     * 3. Download update according to user preference
+     * 4. Once update is done, prompt user for updating and restarting application
+     */
+    private void runUpdate() {
+        if (!initUpdate()) {
+            logger.error("Failed to initialize update");
```
dariusf: Should we `return` here?

src/main/java/updater/UpdateManager.java
```diff
@@ -0,0 +1,151 @@
...
+
+    private void downloadUpdateData() {
+        try {
+            fileDownloader.setProgressTrackerWindowDescription("Checking for update", "Checking for update...");
+            fileDownloader.setEnableProgressWindow(false);
+            fileDownloader.downloadFileFromUrlToFile(
+                    new URI(UPDATE_SERVER_DATA_NAME),
+                    new File(UPDATE_DIRECTORY + File.separator + UPDATE_LOCAL_DATA_NAME));
+        } catch (URISyntaxException e) {
+            e.printStackTrace();
```
dariusf: We should log using a logger, so the output goes into the same place. Same for below.

HansNewbie: Sorry, forgot about this.

src/main/java/updater/UpdateManager.java
HansNewbie: Yes. Missed it.

src/main/java/filter/expression/Qualifier.java
HansNewbie: Yes. My bad.

src/main/java/updater/FileDownloader.java
HansNewbie: Yes.

src/main/java/updater/FileDownloader.java
HansNewbie: Sure.

src/main/java/updater/FileDownloader.java
HansNewbie: This class only download 1 file. Will change this.

HansNewbie:
@dariusf I tried my best to follow your idea. Can you take a look?

I separated the window from update manager. I use listeners for the progress. What happens now is that UpdateManager can request the update window for a progress display, which will then be passed (optionally) as a parameter to `DownloadFile`. For the update window, I made it possible to have multiple download tracker at the same time if needed.

I have not updated the comments in case there is more design changes required.

HansNewbie:
Updated comments.

HansNewbie:
@dariusf can take a look at the design?

HansNewbie:
I will clean up the commit afterwards.

src/main/java/ui/MenuControl.java
```diff
@@ -346,4 +348,22 @@ private void updateRepoRemoveList(Menu remove) {
     private void onRepoRemove(String repoId) {
         ui.logic.removeStoredRepository(repoId).thenRun(() -> ui.triggerEvent(new UnusedStoredReposChangedEvent()));
     }
+
+    private Menu createUpdateMenu() {
+        Menu update = new Menu("Update");
+
+        MenuItem checkProgress = new MenuItem("Check progress...");
+
+        if (ui.updateManager == null) {
```
dariusf: When does this happen?

src/main/java/ui/UpdateProgressWindow.java
```diff
@@ -0,0 +1,144 @@
...
+                    window.toFront();
+                }
+        );
+    }
+
+    /**
+     * hides update download progress window
+     */
+    public void hideWindow() {
+        if (window != null) {
```
dariusf: Will this ever be called when the window is not yet created? Seems like that might be a bug if it happens.

Would be it be possible to initialise the stage on construction, just not show it? Then we don't have to check for null everywhere. And sometimes the null checks may not communicate the right intent, if something cannot possibly be null at that point.

src/main/java/updater/DownloadCountingOutputStream.java
```diff
@@ -0,0 +1,51 @@
...
+     * @param totalFileSizeInBytes total file size to be downloaded
+     */
+    public DownloadCountingOutputStream(OutputStream out, long totalFileSizeInBytes) {
+        super(out);
+
+        this.totalFileSizeInBytes = totalFileSizeInBytes;
+        percentageDownloaded = new SimpleDoubleProperty(0);
+    }
+
+    public void addMethodThatListen(Consumer<Double> method) {
```
dariusf: Name can be a bit more succinct. `addListener`?

src/main/java/updater/DownloadFile.java
```diff
@@ -0,0 +1,129 @@
...
+    private static final Logger logger = LogManager.getLogger(DownloadFile.class.getName());
+
+    public static final int CONNECTION_TIMEOUT = 15000;
+    public static final int READ_CONNECTION_TIMEOUT = 30000;
+
+    private final URI source;
+    private final File dest;
+    private long totalFileSizeInBytes;
+
+    private final Optional<Consumer<Double>> methodToCallOnDownload;
```
dariusf: This too. `onDownload`? Also, I don't think it needs to be optional. Callers can just pass a no-op `a -> {}`, or we can do that ourselves in a constructor overload.

src/main/java/updater/DownloadCountingOutputStream.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * {@inheritDoc}
+     *
+     * Update percentage downloaded
+     */
+    @Override
+    protected void afterWrite(int n) throws IOException {
+        super.afterWrite(n);
+
+        percentageDownloaded.set(1.0 * this.getByteCount() / this.totalFileSizeInBytes);
```
dariusf: I'm not sure this is the right place for Platform.runLater. Without context, we won't know why it's needed, because this class could be used to update the progress of anything, not necessarily JavaFX nodes. We can put it in at the point where we actually modify the UI maybe?

src/main/java/updater/DownloadFile.java
```diff
@@ -0,0 +1,129 @@
...
+
+        totalFileSizeInBytes = sourceConnection.getContentLengthLong();
+
+        try (
+                InputStream inputStream = setupStreamFromSource(sourceConnection);
+                OutputStream outputStream = setupStreamToDest(dest);
+        ) {
+            if (inputStream == null || outputStream == null) {
+                logger.error("Failed to create streams for download");
+                return false;
```
dariusf: Rather than check for these here, we should let the IOExceptions propagate up from the stream setup methods, so they'll be caught below.

src/main/java/updater/DownloadProgressTracker.java
```diff
@@ -0,0 +1,40 @@
...
+import java.net.URL;
+import java.util.Optional;
+
+/**
+ * This class is used to track download if a UI element (i.e. progress bar) is needed to show the download progress
+ */
+public class DownloadProgressTracker {
+    private final String downloadLabel;
+    private final URL urlIdentifier;
+    private Optional<ProgressBar> progressBar;
```
dariusf: Could we pass the progress bar in on construction? It's a bit awkward to construct these objects piecemeal.

If we want to be more OO, this is really a progress bar augmented with a label and url. Maybe we could name it something which reflects that?

dariusf:
I think overall design is fine, just not completely clear on the details yet.

src/main/java/ui/MenuControl.java
HansNewbie: When we are on test mode, since ui.updateManager won't be created.

src/main/java/ui/UpdateProgressWindow.java
HansNewbie: Supposedly this will never be called when window is not yet created. I will try to initialise the stage on construction.

src/main/java/updater/DownloadCountingOutputStream.java
HansNewbie: Noted.

src/main/java/updater/DownloadCountingOutputStream.java
HansNewbie: Okay.

src/main/java/ui/MenuControl.java
dariusf: Is it possible to stub the update manager out in test mode instead of not creating it?

HansNewbie: So both update window and update manager will be stubbed?

dariusf: Yes, if necessary.

src/main/java/ui/UI.java
```diff
@@ -217,6 +221,16 @@ private void initPreApplicationState() {
 
         uiManager = new UIManager(this);
         status = new HTStatusBar(this);
+
+        updateProgressWindow = new UpdateProgressWindow();
+
+        if (!TestController.isTestMode()) {
+            updateManager = new UpdateManager(updateProgressWindow);
+        } else {
+            updateManager = new UpdateManagerStub();
```
HansNewbie: Something like this? Or should I abstract this out to a method somewhere (not sure which class should the instancing belong to)?

dariusf: TestController. There are a few methods to conditionally create similar things already in there.

Also you may be able to use mockito to stub it, then you don't need a new class.

HansNewbie: I tried mockito to mock it but currently mockito is only referenced in tests and not in the main application. I don't think introducing mocking in TestController would be a good idea.

dariusf: Okay, never mind then.

src/main/java/ui/TestController.java
```diff
@@ -157,4 +159,30 @@ public static RepoIO createTestingRepoIO(Optional<JSONStore> jsonStoreToBeUsed)
...
     }
+
+    /**
+     * Create an update progress window if not in test mode.
+     *
+     * @return update progress window
+     */
+    public static UpdateProgressWindow createUpdateProgressWindow() {
+        if (isTestMode()) {
+            return null;
```
HansNewbie: I wanted to create stub for this but javafx stage initialization happens in the constructor which can't be changed in the stub, hence I use `null`. Do you think this is fine?

A problem I encounter is that some guitests seems to be bypassing `isTestMode()` which ends up calling update progress window and giving error.

HansNewbie: The test class in problem is `PlatformExTest` because it does not set the test mode to true.

src/test/java/guitests/PlatformExTest.java
```diff
@@ -12,11 +12,6 @@
 
 public class PlatformExTest extends UITest {
 
-    @Override
-    public void launchApp() {
-        FXTestUtils.launchApp(TestUI.class);
```
HansNewbie: I need to remove override so that it is in test mode; otherwise the test cannot find whether HT main window is the stage or updater is the stage. What do you think?

src/main/java/ui/TestController.java
dariusf: Do we need to stub out this window in other places other than the method below? If not, a simple solution would be to just instantiate the window there.

Can those other tests be easily fixed, and when they are, does that solve your problem?

src/test/java/guitests/PlatformExTest.java
dariusf: I don't completely remember if that's needed, but it could be that it was mistakenly added. If the test still passes after it's removed, sure.

HansNewbie: The test still passes.

src/main/java/ui/TestController.java
HansNewbie: If I instantiate the window in the method below, `UI` will have no direct reference to the window. I will need a reference to the window in `MenuControl` to show update window. Would it make sense to have `ui.updateManager.showUpdateWindow`?

After changing the caller in `PlatformEx`, all the tests seem to be fine.

dariusf: I think that's okay. Maybe there could be a wrapper in UI for that which the menu can call. The update manager already has the responsibility of managing the window, so it doesn't seem out of place. What do you think?

HansNewbie: Makes sense. I will do that, then.

HansNewbie:
@dariusf can take a look?

src/main/java/ui/UpdateProgressWindow.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /**
+     * Create a new download progress bar
+     *
+     * @param downloadUrl url of download as identifier to the progress bar
+     * @param downloadLabel label of progress bar
+     * @return the method to set the progress to the progress bar of the download progress tracker
+     */
+    public Consumer<Double> createNewDownloadProgressBar(URL downloadUrl, String downloadLabel) {
```
dariusf: This method is doing two things. Maybe we could just have this method return the progress bar, and pass the consumer at the call site instead?

src/main/java/ui/UpdateProgressWindow.java
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * Remove a download progress bar from update progress window
+     *
+     * This should be called after a download is completed
+     *
+     * @param downloadUrl download URL to be used as download identifier
+     */
+    public void removeDownloadProgressBar(URL downloadUrl) {
+        downloads.removeIf(dpt -> dpt.getUrlIdentifier() == downloadUrl);
```
dariusf: Is URL safe to compare with `==`?

src/main/java/ui/UpdateProgressWindow.java
```diff
@@ -0,0 +1,139 @@
...
+                    Label downloadLabel = new Label();
+                    downloadLabel.setText(progressTracker.getDownloadLabel());
+
+                    ProgressBar progressBar = progressTracker.getProgressBar();
+
+                    VBox downloadProgressItem = new VBox();
+                    downloadProgressItem.setSpacing(20);
+                    downloadProgressItem.setPadding(new Insets(20));
+                    downloadProgressItem.setAlignment(Pos.CENTER_LEFT);
+                    downloadProgressItem.getChildren().addAll(downloadLabel, progressBar);
```
dariusf: Maybe we could factor out the parts which create JavaFX nodes here? A bit more SLAP.

src/main/java/updater/DownloadFile.java
```diff
@@ -0,0 +1,106 @@
...
+
+        assert sourceConnection != null;
+
+        totalFileSizeInBytes = sourceConnection.getContentLengthLong();
+
+        try (
+                InputStream inputStream = setupStreamFromSource(sourceConnection);
+                OutputStream outputStream = setupStreamToDest(dest);
+        ) {
+            if (download(inputStream, outputStream) != totalFileSizeInBytes) {
```
dariusf: When does this happen, if the download is cut off halfway?

src/main/java/updater/UpdateManager.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    /**
+     * Method that will run update sequence.
+     * 1. Initialize system for update, e.g. make directory
+     * 2. Check if there is any new update
+     *   2.a. If there is no new update but the previously downloaded update is not applied yet,
+     *        prompt user for updating and restarting application
+     * 3. Download update according to user preference
+     * 4. Once update is done, prompt user for updating and restarting application
```
dariusf: I'm not sure if we need to go into such detail in the comment. The body of this method looks quite nice already. Maybe we could write the parts which aren't mentioned (confirmation dialogs, etc.) as inline comments?

dariusf:
Looks good, mostly just cosmetic changes.

src/main/java/ui/UpdateProgressWindow.java
HansNewbie: Sure. I was thinking about the same thing but was hesitant to change.

src/main/java/ui/UpdateProgressWindow.java
HansNewbie: Will change. Apparently `URL.equals()` is broken, so I will use `URI` instead.

src/main/java/updater/DownloadFile.java
HansNewbie: This happens when there is timeout (e.g. server closes connection suddenly).

src/main/java/updater/UpdateManager.java
HansNewbie: Okay. Since right now the other parts is still TODO from another issue, I will create a succinct inline comment for now.

HansNewbie:
I have made the relevant changes and I added a simple test for download counting output stream.

src/main/java/ui/UpdateProgressWindow.java
```diff
@@ -0,0 +1,145 @@
...
+    private void reloadWindowLayout() {
+        assert window != null && windowMainLayout != null;
+
+        Platform.runLater(() -> {
+            windowMainLayout.getChildren().clear();
+
+            if (downloads.isEmpty()) {
+                Label noDownloadLabel = new Label();
+                noDownloadLabel.setText(LABEL_NO_DOWNLOAD);
+                noDownloadLabel.setPadding(new Insets(50));
```
dariusf: This part could also be made a bit more SLAP.

src/main/java/updater/UpdateManager.java
dariusf: Maybe we could check this inside `download` and throw an exception there instead?

Also, it would be useful if the three methods below this had comments describing the circumstances in which they fail.

HansNewbie: I didn't quite get your comment. What check should be done in `download`?

src/main/java/updater/DownloadFile.java
dariusf: If the result of IOUtils.copy is equal to totalFileSizeInBytes.

HansNewbie: Hmmm, the thing is I can't reproduce the error I am trying to capture here manually. What I saw last time was IOUtils printed a message for socket time out (I think Github server closed the connection), but it was not caught. In this case, the only way to check if download ended successfully is by checking file size. I agree the case does not seems possible, but I have yet to encounter the same error.

HansNewbie: I just tried and it seems the socket timeout exception will be caught. Should I remove this line of IOUtils.copy is equal to totalFileSizeInBytes then?

dariusf: Oops, think I replied in the wrong thread haha. Yeah, okay.

damithc:
What is this PR supposed to add?

HansNewbie:
@damithc Downloading file capability and a window to show the progress of such download if needed.

damithc:
How does it change the HT behavior? Or there is no change to the behavior yet?

HansNewbie:
Right now there is no change in behaviour yet except that there will be a background download. What is being downloaded is still a stub. After this PR is merged, #1271 and #1230 can be finished, which then will make HT to download latest version and auto update itself.

I added a new menu to check download update progress.
screen shot 2016-02-27 at 12 05 10 pm
screen shot 2016-02-27 at 12 05 19 pm

damithc:
I see. The `update` menu is present during the download only?
Is this a releasable PR? We cannot merge a partial (not releasable) functionality into the master.

HansNewbie:
The `update` menu is always present, but when it is not downloading, the download window will show `No update being downloaded` instead of a progress bar. The function to download is complete, but if we are taking into account what we would like this holistic feature to work, then this is not releasable.

Should I merge the other issues into this PR branch as well once each of them is done? i.e.:
1. Create a PR for each issue (which has partial functionality)
2. Once each of them has been reviewed, open a new PR that fixes all the issues with reviewed contents, and request for another review.

damithc:
Either you break the feature into releasable increments and keep merging to master, OR open a separate feature branch and keep merging to that until the full feature is ready. First option is preferable.
By releasable, I mean it should not be confusing or misleading to end users at any time.

HansNewbie:
@damithc I don't think it's possible to break this issue and the 2 other mentioned issues into releasable increments. I will create a new branch to do all these 3 issues (#1229, #1271, #1230) together, then.

damithc:
In that case, create a new `updater` branch and PR against that. You should still do one issue at a time but instead of merging to master, you should merge to the feature branch `updater` until it is releasable.

HansNewbie:
@damithc noted.

damithc:
I'll continue to review in this PR, but when it is approved, rebase onto the feature branch instead of merging to master.

src/main/java/ui/TestController.java
```diff
@@ -157,4 +159,17 @@ public static RepoIO createTestingRepoIO(Optional<JSONStore> jsonStoreToBeUsed)
         return new RepoIO(Optional.of(new DummySource()), jsonStoreToBeUsed,
                           Optional.of(RepoStore.TEST_DIRECTORY));
     }
+
+    /**
+     * Create update manager if not in test mode, its stub otherwise.
```
damithc: `Creates...`

src/main/java/ui/UI.java
```diff
@@ -606,4 +612,11 @@ public boolean isWindowMinimized() {
     public boolean isWindowFocused() {
         return mainStage.isFocused();
     }
+
+    /**
+     * Show update progress window
+     */
```
damithc: The header comment doesn't add value. Can be omitted.

src/main/java/ui/UpdateProgressWindow.java
```diff
@@ -0,0 +1,151 @@
...
+import javafx.scene.layout.VBox;
+import javafx.stage.Stage;
+import updater.LabeledDownloadProgressBar;
+
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * UpdateProgressWindow is the class that will display update download progresses
```
damithc: `The class that represents the Window showing the progress of the downloading of the new update`

src/main/java/ui/UpdateProgressWindow.java
```diff
@@ -0,0 +1,151 @@
...
+    private final List<LabeledDownloadProgressBar> downloads;
+
+    public UpdateProgressWindow() {
+        this.downloads = new ArrayList<>();
+        this.initProgressWindow();
+    }
+
+    /**
+     * shows update download progress window
+     */
```
damithc: Get rid of header comments that doesn't add value.

src/main/java/ui/UpdateProgressWindow.java
```diff
@@ -0,0 +1,151 @@
...
+
+        scene.setRoot(windowMainLayout);
+
+        window = stage;
+
+        reloadWindowLayout();
+    }
+
+    /**
+     * Create a new download progress bar
```
damithc: `Creates ...`

src/main/java/updater/DownloadFile.java
```diff
@@ -0,0 +1,104 @@
...
+
+import java.io.*;
+import java.net.URI;
+import java.net.URLConnection;
+import java.util.function.Consumer;
+
+/**
+ * A class in charge of downloading file.
+ */
+public class DownloadFile {
```
damithc: `DownloadFile` sounds like a method name rather than a class name.

HansNewbie: How about `FileDownloader`?

src/main/java/ui/UpdateProgressWindow.java
HansNewbie: Alright.

src/main/java/updater/DownloadFile.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        assert sourceConnection != null;
+
+        totalFileSizeInBytes = sourceConnection.getContentLengthLong();
+
+        try (
+                InputStream inputStream = setupStreamFromSource(sourceConnection);
+                OutputStream outputStream = setupStreamToDest(dest);
+        ) {
```
damithc: Is this a new syntax for `try` block? I've never seen it before.

src/main/java/updater/DownloadFile.java
```diff
@@ -0,0 +1,104 @@
...
+        this.dest = dest;
+        this.onDownload = onDownload;
+    }
+
+    /**
+     * Start download from source to destination as specified in constructor
+     *
+     * @return true if download is successful, else false
+     */
+    public boolean startDownload() {
```
damithc: It simply starts the download or does it do the full download?

src/main/java/updater/DownloadFile.java
```diff
@@ -0,0 +1,104 @@
...
+            download(inputStream, outputStream);
+        } catch (IOException e) {
+            logger.error("Failed to create streams for download or socket timeout", e);
+            return false;
+        }
+
+        return true;
+    }
+
+    private URLConnection setupDownloadConnection(URI source) throws IOException {
```
damithc: Does this method specific to downloads or it's simply setting up a connection? If it is the latter, the method name is more specific than the method.

src/main/java/updater/DownloadFile.java
HansNewbie: It does the full download, but it will check the download parameters before doing the real download.

src/main/java/updater/DownloadFile.java
HansNewbie: Yes, [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html). With this new syntax, there is no need for `finally` to close the streams in case of exceptions..

src/main/java/updater/UpdateManager.java
```diff
@@ -0,0 +1,159 @@
...
+
+import java.io.File;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+/**
+ * UpdateManager is the class that will handle updating of HubTurbo application
```
damithc: Minimize mentioning class/method names in comments. The more you mention names, the higher the chance of comments getting outdated due to an automated refactoring.

src/main/java/updater/DownloadFile.java
HansNewbie: Only setting up a connection. Will change the name.

src/main/java/updater/UpdateManager.java
HansNewbie: Noted.

src/main/java/updater/UpdateManager.java
```diff
@@ -0,0 +1,159 @@
...
+     * - Run is not automatic upon instancing the class in case there would like to be conditions on when to run update,
+     *   e.g. only if user is logged in
+     */
+    public void run() {
+        pool.execute(() -> runUpdate());
+    }
+
+    /**
+     * Method that will run update sequence.
+     */
```
damithc: Be consistent in how you start the comment. e.g. ~~`Method that will run`~~ `Runs the update ...`

src/test/java/tests/DownloadCountingOutputStreamTest.java
```diff
@@ -0,0 +1,39 @@
...
+        DownloadCountingOutputStream dcos = new DownloadCountingOutputStream(baos, totalByteSize);
+        dcos.addListener(val -> listenerValue.setValue(val));
+
+        try {
+            for (int i = 0; i < totalByteSize; i++) {
+                dcos.write(i);
+            }
+        } catch (IOException e) {
+            fail("IOException occurs on writing");
+        }
```
damithc: The try-catch is redundant? An exception will fail the test anyway.

src/test/java/tests/DownloadCountingOutputStreamTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+/**
+ * Test DownloadCountingOutputStream
+ */
+public class DownloadCountingOutputStreamTest {
+    @Test
+    public void downloadCountingOS_downloading_updateListener() {
```
damithc: At first I thought `OS` meant Operating System :-p

damithc:
Some comments added. Mostly cosmetic.

src/main/java/updater/DownloadFile.java
damithc: Yes, that's better.

src/main/java/updater/DownloadFile.java
damithc: In that case, the method name is misleading?

src/test/java/tests/DownloadCountingOutputStreamTest.java
HansNewbie: Yes, I was afraid of the confusion, but afraid of name being too long. Will spell out `OS`.

src/main/java/updater/DownloadFile.java
damithc: I see. nice.

src/main/java/updater/UpdateManager.java
HansNewbie: Noted.

src/test/java/tests/DownloadCountingOutputStreamTest.java
HansNewbie: Alright.

src/main/java/updater/DownloadFile.java
HansNewbie: Yeah. Would renaming it to `download` be better? I will change the private function that actually runs the download to something else.

src/test/java/tests/DownloadCountingOutputStreamTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+/**
+ * Test DownloadCountingOutputStream
+ */
```
damithc: Redundant header comment

damithc:
Can go ahead and merge to the feature branch.

src/test/java/tests/DownloadCountingOutputStreamTest.java
HansNewbie: Will remove.

HansNewbie:
@damithc noted.

HansNewbie:
I have merged this PR into `updater` branch. Will close this PR and the issue manually.

PR #1251, opened 2016-02-15
The board name shown does not match the board #1102

### LINKED ISSUE #1102: The board name shown does not match the board
If the current board is modified (e.g. by adding a panel), the board name should reflect the change. e.g. adding an '*' to the end of the name to indicate it has been modified.

m133225:
Fixes #1102 

This indicator will be displayed when filter or panel has been modified, removed when board is saved or a board is reopened.


m133225:
Closed since #1255 is trying to implement a neater way to solve this issue.

PR #1255, opened 2016-02-15
The board name shown does not match the board #1102

### LINKED ISSUE #1102: The board name shown does not match the board

whattokingu:
Fixes #1102 

I know PR [#1251](https://github.com/HubTurbo/HubTurbo/pull/1251) solves the same problem. 
But my solution is more general. As requested by @damithc , the solution compares the saved panels from GlobalConfig and the current panels, to determine if the current panels requires saving.

### DIFF
--- a/src/main/java/ui/issuepanel/PanelControl.java
+++ b/src/main/java/ui/issuepanel/PanelControl.java
@@ -177,7 +177,7 @@ public class PanelControl extends HBox {
         updatePanelIndices();
         ((AbstractPanel) child).close();
         updateFocus(index);
-
+        ui.updateTitle(this.hasUnsavedPanels());
         UI.events.triggerEvent(new UsedReposChangedEvent());
     }
 
@@ -190,10 +190,12 @@ public class PanelControl extends HBox {
 
     public void createNewPanelAtStart() {
         addPanelAt(0).filterTextField.requestFocus();
+        ui.updateTitle(this.hasUnsavedPanels());
     }
 
     public void createNewPanelAtEnd() {
         addPanel().filterTextField.requestFocus();
+        ui.updateTitle(this.hasUnsavedPanels());
     }
 
     public void swapPanels(int panelIndex, int panelIndex2) {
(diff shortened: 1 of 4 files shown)


src/main/java/ui/issuepanel/PanelControl.java
```diff
@@ -336,4 +339,31 @@ private void setHvalue(double val) {
...
+        for (int i = 0; i < savedPanelInfos.size(); i++) {
+            PanelInfo savedPanel = savedPanelInfos.get(i);
+            PanelInfo currentPanel = currentPanelInfos.get(i);
+            if (!savedPanel.getPanelName().equals(currentPanel.getPanelName())
+                    || !savedPanel.getPanelFilter().equals(currentPanel.getPanelFilter())) {
+                return true;
+            }
+        }
+        return false;
+    }
```
damithc: Can we distill this method into something like `return openBoard.toString().equals(/*string from settings*/)`?
@dariusf do we have a `Board` class that we can use here?

whattokingu:
I investigated the failed tests. The 2 tests that failed are:
guitests > BoardAutoCreatorTest > [milestoneBoardAutoCreationTest](https://github.com/HubTurbo/HubTurbo/blob/master/src/test/java/guitests/BoardAutoCreatorTest.java#L39)
guitests > BoardAutoCreatorTest > [workAllocationBoardAutoCreationTest](https://github.com/HubTurbo/HubTurbo/blob/master/src/test/java/guitests/BoardAutoCreatorTest.java#L70)

When I run the failed tests individually, the tests passed.
It turns out the method I committed, `hasUnsavedPanels` calls [`prefs.getLastOpenBoard`](https://github.com/HubTurbo/HubTurbo/blob/ce677ede4f19fe9f1b72db8230ba21ae1c81f658/src/main/java/ui/issuepanel/PanelControl.java#L352), which still holds the value of the previous test. 

In this case, I argue that the [`cleanUpBoards`](https://github.com/HubTurbo/HubTurbo/blob/master/src/test/java/guitests/BoardAutoCreatorTest.java#L29) which is run before the tests should be updated to call [`testPref.clearLastOpenBoard`](https://github.com/HubTurbo/HubTurbo/blob/ce677ede4f19fe9f1b72db8230ba21ae1c81f658/src/test/java/guitests/BoardAutoCreatorTest.java#L36).

I have added a commit that does that.

dariusf:
There have been quite a few changes to things that this PR depends on. @garbanzos is working on refactoring the board-related parts. Maybe this could be integrated into her work? Otherwise this PR may no longer be relevant.

PR #1258, opened 2016-02-15
Recover from failed labels replace #1246

### LINKED ISSUE #1246: Recover when failed to update issue's labels on GitHub.

ndt93:
Fixes #1246.

* Mitigate some concurrency issues by synchronising various methods in `Model`, `MultiModel` and `TurboIssue` and using issue lookup in models.
>The root cause for this is that `MultiModel` is not properly synchronised i.e. although its methods are synchronised, its internal states can be retrieved and read/modified somewhere else, technically changing the `MultiModel`'s state without synchronisation. An example is that we can assign labels to a dangling `TurboIssue` out of nowhere and the models' state and `Ui` is changed. The `queue` helps to solve this by making the end-to-end operation on `MultiModel` mutually execlusive. However the operation includes incremental steps of downloading and local models update, which cause other operations to be laggy. So a solution that doesn't trades off speed or correctness involves seperating the end-to-end download and end-to-end local updates, which would requires major changes and planning.

* Reconciling labels updates from GitHub and local changes using physical time timestamp. This requires all mutating actions on labels to update the timestamp. Making the labels getter return a new list leads to regression so this is probably violated somewhere by codes that get the labels list and modify it directly.

* Display error message and reverting to previous labels if failed to update on GitHub. This is also achieved through the labels modification timestamps. I was intending to have a `Retry` dialog box but that needs changes up at the `UI` level which probably conflicts with the current labels picker refactoring.

### DIFF
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -20,6 +20,7 @@ import util.events.RepoOpeningEvent;
 import util.events.testevents.ClearLogicModelEvent;
 import util.events.testevents.ClearLogicModelEventHandler;
 
+import java.time.LocalDateTime;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
@@ -39,10 +40,10 @@ public class Logic {
     public LoginController loginController;
     public UpdateController updateController;
 
-    public Logic(UIManager uiManager, Preferences prefs) {
+    public Logic(UIManager uiManager, Preferences prefs, Optional<MultiModel> models) {
         this.uiManager = uiManager;
         this.prefs = prefs;
-        models = new MultiModel(prefs);
+        this.models = models.orElse(new MultiModel(prefs));
 
         loginController = new LoginController(this);
         updateController = new UpdateController(this);
@@ -197,20 +198,51 @@ public class Logic {
      * @return true if label replacement on GitHub was a success, false otherwise.
      */
     public CompletableFuture<Boolean> replaceIssueLabels(TurboIssue issue, List<String> newLabels) {
+        List<String> originalLabels = issue.getLabels();
+
         logger.info("Changing labels for " + issue + " on UI");
-        if (!models.replaceIssueLabels(issue, newLabels)) {
-           return CompletableFuture.completedFuture(false);
+        Optional<TurboIssue> replaceResult = models.replaceIssueLabels(issue.getRepoId(), issue.getId(), newLabels);
+        if (!replaceResult.isPresent()) {
+            return CompletableFuture.completedFuture(false);
         }
         refreshUI();
 
-        return updateIssueLabelsOnRepo(issue, newLabels);
+        return updateIssueLabelsOnRepo(issue, newLabels)
+                .thenApply((isUpdateSuccessful) -> {
+                    if (isUpdateSuccessful) {
+                        return true;
+                    }
+                    logger.error("Unable to update model on server");
+                    revertLocalLabelsReplace(replaceResult.get(), originalLabels);
+                    return false;
+                });
+    }
+
+    private Optional<TurboIssue> lookUpIssue(String repoId, int issueId) {
+        Optional<Model> modelLookUpResult = models.getModelById(repoId);
+        if (!modelLookUpResult.isPresent()) {
+            logger.error("Model " + repoId + " not found in models");
+            return Optional.empty();
+        }
+        return modelLookUpResult.get().getIssueById(issueId);
     }
 
     private CompletableFuture<Boolean> updateIssueLabelsOnRepo(TurboIssue issue, List<String> newLabels) {
         logger.info("Changing labels for " + issue + " on GitHub");
-        return repoOpControl.replaceIssueLabels(issue, newLabels)
-                .thenApply(labels -> newLabels.containsAll(labels))
-                .exceptionally(Futures.withResult(false));
+        return repoOpControl.replaceIssueLabels(issue, newLabels);
+    }
+
+    private void revertLocalLabelsReplace(TurboIssue modifiedIssue, List<String> originalLabels) {
+        TurboIssue currentIssue = lookUpIssue(modifiedIssue.getRepoId(), modifiedIssue.getId()).orElse(modifiedIssue);
+        LocalDateTime originalLabelsModifiedAt = modifiedIssue.getLabelsLastModifiedAt();
+        LocalDateTime currentLabelsAssignedAt = currentIssue.getLabelsLastModifiedAt();
+        boolean isCurrentLabelsModifiedFromOriginalLabels = originalLabelsModifiedAt.isEqual(currentLabelsAssignedAt);
+
+        if (isCurrentLabelsModifiedFromOriginalLabels) {
+            logger.info("Reverting labels for issue " + currentIssue);
+            models.replaceIssueLabels(currentIssue.getRepoId(), currentIssue.getId(), originalLabels);
+            refreshUI();
+        }
     }
 
     /**
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -20,10 +20,7 @@ import util.events.RepoOpeningEvent;
 import util.events.testevents.ClearLogicModelEvent;
 import util.events.testevents.ClearLogicModelEventHandler;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
 
@@ -201,7 +198,9 @@ public class Logic {
      */
     public CompletableFuture<Boolean> replaceIssueLabels(TurboIssue issue, List<String> newLabels) {
         logger.info("Changing labels for " + issue + " on UI");
-        issue.setLabels(newLabels);
+        if (!models.replaceIssueLabels(issue, newLabels)) {
+           return CompletableFuture.completedFuture(false);
+        }
         refreshUI();
 
         return updateIssueLabelsOnRepo(issue, newLabels);
@@ -210,7 +209,7 @@ public class Logic {
     private CompletableFuture<Boolean> updateIssueLabelsOnRepo(TurboIssue issue, List<String> newLabels) {
         logger.info("Changing labels for " + issue + " on GitHub");
         return repoOpControl.replaceIssueLabels(issue, newLabels)
-                .thenApply(labels -> labels.containsAll(newLabels))
+                .thenApply(labels -> newLabels.containsAll(labels))
                 .exceptionally(Futures.withResult(false));
     }
 
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -201,39 +201,65 @@ public class Logic {
         List<String> originalLabels = issue.getLabels();
 
         logger.info("Changing labels for " + issue + " on UI");
-        Optional<TurboIssue> replaceResult = models.replaceIssueLabels(issue.getRepoId(), issue.getId(), newLabels);
-        if (!replaceResult.isPresent()) {
+        /* Calls models to replace the issue's labels locally since the the reference to the issue here
+           could be invalidated by changes to the models elsewhere */
+        Optional<TurboIssue> localReplaceResult =
+                models.replaceIssueLabels(issue.getRepoId(), issue.getId(), newLabels);
+        if (!localReplaceResult.isPresent()) {
             return CompletableFuture.completedFuture(false);
         }
         refreshUI();
 
-        return updateIssueLabelsOnRepo(issue, newLabels)
-                .thenApply((isUpdateSuccessful) -> {
-                    if (isUpdateSuccessful) {
-                        return true;
-                    }
-                    logger.error("Unable to update model on server");
-                    revertLocalLabelsReplace(replaceResult.get(), originalLabels);
-                    return false;
-                });
+        return updateIssueLabelsOnServer(issue, newLabels)
+                .thenApply((isUpdateSuccessful) -> handleIssueLabelsUpdateOnServerResult(
+                            isUpdateSuccessful, localReplaceResult.get(), originalLabels));
     }
 
-    private Optional<TurboIssue> lookUpIssue(String repoId, int issueId) {
+    /**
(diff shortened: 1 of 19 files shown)


ndt93:
@dariusf Ready for review.

src/main/java/backend/Logic.java
```diff
@@ -200,18 +198,51 @@ public Model getRepo(String repoId) {
...
+                    if (isUpdateSuccessful) {
+                        return true;
+                    }
+                    logger.error("Unable to update model on server");
+                    revertLocalLabelsReplace(replaceResult.get(), originalLabels);
+                    return false;
+                });
+    }
+
+    private Optional<TurboIssue> lookUpIssue(String repoId, int issueId) {
```
damithc: `getIssue`?

src/main/java/backend/Logic.java
```diff
@@ -200,18 +198,51 @@ public Model getRepo(String repoId) {
...
+        Optional<Model> modelLookUpResult = models.getModelById(repoId);
```
damithc: Feels like this should be a parameter to this method

src/main/java/backend/Logic.java
```diff
@@ -200,18 +198,51 @@ public Model getRepo(String repoId) {
...
 
     private CompletableFuture<Boolean> updateIssueLabelsOnRepo(TurboIssue issue, List<String> newLabels) {
         logger.info("Changing labels for " + issue + " on GitHub");
-        return repoOpControl.replaceIssueLabels(issue, newLabels)
-                .thenApply(labels -> labels.containsAll(newLabels))
-                .exceptionally(Futures.withResult(false));
+        return repoOpControl.replaceIssueLabels(issue, newLabels);
+    }
+
+    private void revertLocalLabelsReplace(TurboIssue modifiedIssue, List<String> originalLabels) {
```
damithc: header comment?

src/main/java/backend/resource/Model.java
```diff
@@ -186,6 +190,22 @@ private void ______OPERATIONS_____() {}
             .collect(Collectors.toList());
     }
 
+    /**
+     * Assigns {@code labels} to issue {@code issueId}
+     * @param issueId
+     * @param labels
+     * @return the modified TurboIssue if successful
+     */
+    public synchronized Optional<TurboIssue> replaceIssueLabels(int issueId, List<String> labels) {
```
damithc: Method name says `replace` but the comment says `assigns`

src/main/java/backend/Logic.java
```diff
@@ -200,18 +198,51 @@ public Model getRepo(String repoId) {
...
      */
     public CompletableFuture<Boolean> replaceIssueLabels(TurboIssue issue, List<String> newLabels) {
+        List<String> originalLabels = issue.getLabels();
+
         logger.info("Changing labels for " + issue + " on UI");
-        issue.setLabels(newLabels);
+        Optional<TurboIssue> replaceResult = models.replaceIssueLabels(issue.getRepoId(), issue.getId(), newLabels);
+        if (!replaceResult.isPresent()) {
+            return CompletableFuture.completedFuture(false);
+        }
```
damithc: These 4 lines looks like a high-level step. Can extract or at least add a comment to state what the step is?

src/main/java/backend/Logic.java
```diff
@@ -200,18 +198,51 @@ public Model getRepo(String repoId) {
...
         refreshUI();
 
-        return updateIssueLabelsOnRepo(issue, newLabels);
+        return updateIssueLabelsOnRepo(issue, newLabels)
+                .thenApply((isUpdateSuccessful) -> {
+                    if (isUpdateSuccessful) {
+                        return true;
+                    }
+                    logger.error("Unable to update model on server");
+                    revertLocalLabelsReplace(replaceResult.get(), originalLabels);
```
damithc: Any changes to the header comment to match these changes to the implementation?

src/main/java/backend/Logic.java
```diff
@@ -200,18 +198,51 @@ public Model getRepo(String repoId) {
...
+                    return false;
+                });
```
damithc: See if you can improve SLAP lines 210-219. Not sure if possible though.

src/main/java/backend/Logic.java
dariusf: Does returning a completed future have the semantics we want? Futures chained onto this by callers don't run if it's already completed. Just a sanity check.

src/main/java/backend/Logic.java
```diff
@@ -200,18 +198,51 @@ public Model getRepo(String repoId) {
...
+    }
+
+    private Optional<TurboIssue> lookUpIssue(String repoId, int issueId) {
+        Optional<Model> modelLookUpResult = models.getModelById(repoId);
+        if (!modelLookUpResult.isPresent()) {
+            logger.error("Model " + repoId + " not found in models");
```
dariusf: Maybe we could factor out this pattern of mapping over the optional, but logging if it's empty? Could be a utility method, since I think Optional::map just ignores the empty case.

src/test/java/tests/LogicTests.java
```diff
@@ -33,67 +41,115 @@ public LogicTests() throws NoSuchFieldException, IllegalAccessException {
...
     @Test
-    public void testReplaceIssueLabelsInconsistent() throws ExecutionException, InterruptedException {
-        TurboIssue issue = createIssueWithLabels(Arrays.asList("label1", "label2", "label3"));
-        mockRepoIOReplaceIssueLabelsResult(Arrays.asList("label3", "label4"));
+    public void testReplaceIssueLabelsNoRevert() throws ExecutionException, InterruptedException {
+        List<String> originalLabels = Arrays.asList("label1", "label2");
+        List<String> newLabels = Arrays.asList("label3", "label4");
+
+        TurboIssue issue = createIssueWithLabels(1, originalLabels);
+        Thread.sleep(10);
```
dariusf: Maybe the sleep could be encapsulated in a method for creating a 'modified' issue, with a date that can be taken to be after the current time? That way we don't have to repeat it everywhere.

src/main/java/backend/resource/Model.java
```diff
@@ -99,7 +103,7 @@ public UpdateSignature getUpdateSignature() {
     }
 
     @Override
-    public List<TurboIssue> getIssues() {
+    public synchronized List<TurboIssue> getIssues() {
```
dariusf: Why is this needed? Is it safe to synchronise only this method and not others which access `issues`?

dariusf:
I'm wondering if we should move towards the big change of separating the whole stream of updates into two. While this is a nice solution to the problem given the current tradeoffs, it's quite complex to extend correctly, and will get even more so as requirements change.

src/main/java/backend/Logic.java
ndt93: Good point. I will need to double check this part.

src/main/java/backend/Logic.java
ndt93: Ok noted. I will update the header and try to refactor more.

src/main/java/backend/Logic.java
ndt93: Sure, that should be helpful.

src/main/java/backend/resource/Model.java
ndt93: The `UpdateModelTask` calls `getIssues` on a `Model` to do a reconciliation, while at the same time, we can also do `replaceIssueLabels` on that `Model`. So this would help to some extent in that case. The thing that breaks this is that after the `getIssues`, we modify the returned list of issues directly without the synchronisation protection from `Model`.

ndt93: And to deal with that, we synchronise the `reconcile`, `getLabels`, `setLabels` methods on the `TurboIssue` level also :-P.

ndt93: But the ultimate wall is that after a new `Model` is created from update data but hasn't been put into the `MultiModel` yet, methods outside that try to modify the issues has no way to access the updated `Model`, meanwhile the updated `Model` has no knowledge of the outside changes and is just gonna squash the changes into the `MultiModel`. This is like a scaled up version of the classic load -> increment -> store mutex problem.

ndt93:
I thought so too. The download is the bottleneck. The models update time is unnoticeable.

src/main/java/backend/Logic.java
ndt93: That would make this method less useful though since if we already have a `Model`, we can call `Model`.getIssueById` to get the issue.

src/main/java/backend/Logic.java
ndt93: Seems like a the chained Futures still run even if the previous Future has completed before the chaining. Did a simple experiment using this:
```java
CompletableFuture<Integer> future = CompletableFuture.completedFuture(1);
future.thenApply((x) -> {
    System.out.println(x);
    return x;
});
```

dariusf: Interesting, maybe I tested it wrongly last time. Guess Futures.unit is useless then. :stuck_out_tongue: We can probably replace instances of that with completedFuture.

src/main/java/backend/resource/Model.java
dariusf: :sob: We have to eventually synchronise how updates come in.

ndt93:
@dariusf @damithc Ready for review again.

src/main/java/backend/Logic.java
```diff
@@ -200,18 +198,77 @@ public Model getRepo(String repoId) {
...
+                            isUpdateSuccessful, localReplaceResult.get(), originalLabels));
+    }
+
+    /**
+     * Gets the issue identified by {@code repoId} and {@code issueId} in {@link Logic#models}
+     * @param repoId
+     * @param issueId
+     * @return
+     */
+    private Optional<TurboIssue> getIssue(String repoId, int issueId) {
```
dariusf: Should this be a MultiModel method instead? Seeing as it only operates on `models`.

Also, would a caller ever expect to handle the empty case (can the issue be missing in some legitimate case)? If not, maybe it should throw a runtime exception upon failure, since that probably indicates something seriously wrong. I think this is a problem with other MultiModel methods as well.

src/main/java/util/Utility.java
```diff
@@ -328,5 +324,23 @@ public static String getNameClosestToDesiredName(String desiredName, List<String
...
 
+    /**
+     * If a value is present in the optional, applies mapping function to it and return the result,
+     * otherwise executes the ifEmpty function and returns an empty optional
+     * @param optional
+     * @param mapper
+     * @param ifEmpty
+     * @return
+     */
+    public static <T, U> Optional<U> flatMapOptional(Optional<T> optional,
```
dariusf: We may need to give this a more specific name to distinguish it from Optional::flatMap. Maybe something with 'log' or 'safe' somewhere? Not sure there's a succinct name...

src/test/java/tests/TestUtils.java
```diff
@@ -129,4 +130,28 @@ public static HttpRequest createMockServerRequest(String method, int page,
...
     }
+
+    /**
+     * Delays for a specified time then applies a function and returns its result
+     * @param delay time to be delayed in milliseconds
+     * @param supplier
+     * @return
+     * @throws InterruptedException
+     */
+    public static <T> T delayThenApply(long delay, Supplier<T> supplier) throws InterruptedException {
```
dariusf: `apply` is the name of the method for `Function`. Maybe use `delayThenGet` here?

src/main/java/backend/Logic.java
ndt93: By right, it should be something seriously wrong but for our codes at the moment, there is a slim chance that it might happen. For example, during update reconciliation, an issue is removed from a `Model`'s issues list but before the replacement can be put in, the program is context switched to replacing labels which does an issue look up as above but won't find the issues.

src/main/java/util/Utility.java
ndt93: I think can be `safeFlatMapOptional` since it's general enough here to perform actions other than logging

src/test/java/tests/TestUtils.java
ndt93: Sure.

ndt93:
@dariusf I've made some changes. You can review again.

src/main/java/backend/Logic.java
dariusf: Okay, guess the interfaces can only be changed after we synchronise properly. We should put this in MultiModel, together with all the other similar methods.

src/main/java/backend/resource/TurboIssue.java
```diff
@@ -216,23 +246,26 @@ public String toString() {
...
+    public static List<TurboIssue> reconcile(List<TurboIssue> existing, List<TurboIssue> changed) {
         List<TurboIssue> existingCopy = new ArrayList<>(existing);
         for (TurboIssue issue : changed) {
             int id = issue.getId();
 
-            // TODO O(n^2), fix by preprocessing and copying into a map
-            Optional<Integer> corresponding = findIssueWithId(existingCopy, id);
-            if (corresponding.isPresent()) {
+            Optional<Integer> correspondingIssueIndex = findIssueWithId(existingCopy, id);
+            if (correspondingIssueIndex.isPresent()) {
```
damithc: The code is easier to understand if you deal with the shorter branch first.

src/test/java/tests/LogicTests.java
```diff
@@ -33,67 +41,115 @@ public LogicTests() throws NoSuchFieldException, IllegalAccessException {
...
-     * Tests that the returned status is false when the returned result from RepoIO is inconsistent
-     * with the given labels argument
+     * Tests that no revert is taken place if the issue's labels are modified elsewhere after
+     * {@link Logic#replaceIssueLabels(TurboIssue, List)} is called
      */
     @Test
-    public void testReplaceIssueLabelsInconsistent() throws ExecutionException, InterruptedException {
-        TurboIssue issue = createIssueWithLabels(Arrays.asList("label1", "label2", "label3"));
-        mockRepoIOReplaceIssueLabelsResult(Arrays.asList("label3", "label4"));
+    public void testReplaceIssueLabelsNoRevert() throws ExecutionException, InterruptedException {
```
damithc: Can use the new naming convention for test methods?

damithc:
Two optional improvements suggested. Can merge after that.

PR #1259, opened 2016-02-15
User can give an alias to a repo #784

### LINKED ISSUE #784: User can give an alias to a repo
so that there is no need to type of the full repo identifier in filters.
This will probably need a new 'Projects' menu item, just like Panels and Borads.

nicholaslum444:
Fixes #784 

Proposed design
------
- A list of RepoInfo objects in the user-defined preferences file

- RepoInfo stores repo-specific information, i.e. repo ID, repo alias, etc.

- User should be able to define aliases for repos before HubTurbo downloads those repos into the backend.

- Additional "middle-man" code to intercept any reference to repos, convert any aliases to repo ID

- The backend/model never knows the existence of aliases; it is purely a front-end/UX feature

Proposed scope for this Issue/PR
------
- User manually writes the repo aliases in the user-defined preferences file

-  Aliases may be used in HubTurbo once they are defined - only in places where repoIds are used formally

- No editing of aliases in the GUI; again, user has to manually edit the aliases in the file.

Further work
------
- **Question:** do we provide gui for editing aliases only, or for editing all the information in each RepoInfo object?

### DIFF
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -14,8 +14,7 @@ import ui.TestController;
 import ui.UI;
 import util.Futures;
 import util.HTLog;
-import util.RepoAliasMap;
-import util.Utility;
+import util.RepoConfig;
 import util.events.RepoOpenedEvent;
 import util.events.RepoOpeningEvent;
 import util.events.testevents.ClearLogicModelEvent;
@@ -112,7 +111,7 @@ public class Logic {
      */
     private CompletableFuture<Boolean> openRepository(String repoIdOrAlias, boolean isPrimaryRepository) {
         // First resolves the given string to a repo id.
-        final String repoId = prefs.getRepoAliasMap().resolveRepoId(repoIdOrAlias);
+        final String repoId = prefs.getRepoConfig().resolveRepoId(repoIdOrAlias);
 
         if (isPrimaryRepository) prefs.setLastViewedRepository(repoId);
         if (isAlreadyOpen(repoId) || models.isRepositoryPending(repoId)) {
@@ -161,8 +160,8 @@ public class Logic {
      */
     public Set<String> getStoredReposWithAlias() {
         return repoIO.getStoredRepos().stream().map(repoId -> {
-            RepoAliasMap repoAliasMap = prefs.getRepoAliasMap();
-            return repoAliasMap.hasAlias(repoId) ? repoAliasMap.getAlias(repoId) : repoId;
+            RepoConfig repoConfig = prefs.getRepoConfig();
+            return repoConfig.hasRepoAlias(repoId) ? repoConfig.getRepoAlias(repoId) : repoId;
         }).collect(Collectors.toSet());
     }
 
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -104,6 +104,12 @@ public class Logic {
         return openRepository(repoId, false);
     }
 
+    /**
+     * Opens the repository represented by a repo id or alias.
+     * @param repoIdOrAlias A string that is either a repo id or a repo alias
+     * @param isPrimaryRepository True if the repo in repoIdOrAlias is the primary repository
+     * @return A CompletableFuture (Boolean) indicating if the opening succeeded
+     */
     private CompletableFuture<Boolean> openRepository(String repoIdOrAlias, boolean isPrimaryRepository) {
         // First resolves the given string to a repo id.
         RepoAliasMap repoAliasMap = RepoAliasMap.getInstance();
@@ -153,16 +159,11 @@ public class Logic {
 
     /**
      * Gets the repo ids of the stored repos, but replaces the repo id with its alias if it has one
-     * @return
      */
     public Set<String> getStoredReposWithAlias() {
         return repoIO.getStoredRepos().stream().map(repoId -> {
             RepoAliasMap repoAliasMap = RepoAliasMap.getInstance();
-            if (repoAliasMap.hasAlias(repoId)) {
-                return repoAliasMap.getAlias(repoId);
-            } else {
-                return repoId;
-            }
+            return repoAliasMap.hasAlias(repoId) ? repoAliasMap.getAlias(repoId) : repoId;
         }).collect(Collectors.toSet());
     }
 
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -151,6 +151,21 @@ public class Logic {
         return repoIO.getStoredRepos().stream().collect(Collectors.toSet());
     }
 
+    /**
+     * Gets the repo ids of the stored repos, but replaces the repo id with its alias if it has one
+     * @return
+     */
+    public Set<String> getStoredReposWithAlias() {
+        return repoIO.getStoredRepos().stream().map(repoId -> {
+            RepoAliasMap repoAliasMap = RepoAliasMap.getInstance();
+            if (repoAliasMap.hasAlias(repoId)) {
+                return repoAliasMap.getAlias(repoId);
+            } else {
+                return repoId;
+            }
+        }).collect(Collectors.toSet());
+    }
+
     public boolean isAlreadyOpen(String repoId) {
         return getOpenRepositories().contains(repoId.toLowerCase());
     }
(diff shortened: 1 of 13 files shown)


nicholaslum444:
@dariusf Hi, I'm ready for review.

I'm failing three unstable tests though, they are: 
`LabelPickerTests. addAndRemoveLabelTest`
`LabelPickerTests. undoLabelChangeTest`
`UpdateIssuesTest. updateIssues`

But all the other test cases pass.

src/main/java/backend/Logic.java
```diff
@@ -103,8 +104,11 @@ public void refresh() {
         return openRepository(repoId, false);
     }
 
-    private CompletableFuture<Boolean> openRepository(String repoId, boolean isPrimaryRepository) {
-        assert Utility.isWellFormedRepoId(repoId);
+    private CompletableFuture<Boolean> openRepository(String repoIdOrAlias, boolean isPrimaryRepository) {
```
damithc: Header comment?

src/main/java/backend/Logic.java
```diff
@@ -147,6 +151,21 @@ public void refresh() {
         return repoIO.getStoredRepos().stream().collect(Collectors.toSet());
     }
 
+    /**
+     * Gets the repo ids of the stored repos, but replaces the repo id with its alias if it has one
+     * @return
+     */
```
damithc: Can be shortened like this:
```java
/**
     * Returns the repo ids of the stored repos, but replaces the repo id with its alias if it has one
     */
```

src/main/java/backend/Logic.java
```diff
@@ -147,6 +151,21 @@ public void refresh() {
...
+    public Set<String> getStoredReposWithAlias() {
+        return repoIO.getStoredRepos().stream().map(repoId -> {
+            RepoAliasMap repoAliasMap = RepoAliasMap.getInstance();
+            if (repoAliasMap.hasAlias(repoId)) {
+                return repoAliasMap.getAlias(repoId);
+            } else {
+                return repoId;
+            }
```
damithc: Use ternary operator ` ? : `?

src/main/java/backend/UpdateController.java
```diff
@@ -7,6 +7,7 @@
 import filter.expression.Qualifier;
 import org.apache.logging.log4j.Logger;
 import filter.expression.QualifierType;
+import org.apache.logging.log4j.core.Filter;
```
damithc: Is this correct?

src/main/java/filter/expression/Qualifier.java
```diff
@@ -113,6 +115,35 @@ public static FilterExpression replaceMilestoneAliases(IModel model, FilterExpre
...
+     * @param q The qualifier to check
+     * @return True if the qualifier is a repo qualifier
+     */
+    private static boolean isRepoQualifier(Qualifier q) {
+        switch (q.getType()) {
+            case REPO:
+                return true;
+            default:
+                return false;
+        }
```
damithc: just a `return q.getType() == REPO` will do?

src/main/java/filter/expression/Qualifier.java
```diff
@@ -113,6 +115,35 @@ public static FilterExpression replaceMilestoneAliases(IModel model, FilterExpre
...
+            } else {
+                return q;
+            }
+        });
+    }
+
+    /**
+     * Returns true if the qualifier is a repo qualifier
+     * @param q The qualifier to check
+     * @return True if the qualifier is a repo qualifier
```
damithc: 3rd comment line is redundant (it's same as 1st line)

src/main/java/util/RepoAliasMap.java
```diff
@@ -0,0 +1,260 @@
...
+            String[][] array = gson.fromJson(reader, String[][].class);
+            reader.close();
+            if (array == null) {
+                return new String[0][2];
+            } else {
+                return array;
+            }
+
+        } catch (IOException e) {
+            e.printStackTrace();
```
damithc: ?

src/main/java/util/RepoAliasMap.java
```diff
@@ -0,0 +1,260 @@
...
+ * The mapping is a bijection, i.e. a repo id can only have one alias, and an alias can only be mapped to one repo id.
+ *
+ * This is a singleton class. Use the getInstance method to retrieve the instance.
+ * One can also construct a test instance. However this is only meant for test usage.
+ *
+ * The mapping of aliases are pulled from a mapping file in the settings directory.
+ * There is currently no functionality to update the mapping file from the code.
+ * The file has to be manually inserted
+ */
+public final class RepoAliasMap {
```
damithc: @dariusf do we have a class to represent a `Repo`? The aliases should be an attribute of Repo class. 
@nicholaslum444 As for storing aliases on hard disk, I don't think we need an extra json file to store the mapping. aliases should be stored wherever the repo details are stored.

src/main/java/util/RepoAliasMap.java
```diff
@@ -0,0 +1,260 @@
...
+    //   Key = repoID, Value = alias.
+    private final HashBiMap<String, String> aliasMap;
+
+
+    // Static attributes
+    //------------------
+    private static final Logger logger = LogManager.getLogger(RepoAliasMap.class.getName());
+
+    public static final String MAPPING_FILE_NAME_DEFAULT = "repo_alias_mapping.json";
+    public static final String TEST_MAPPING_FILE_NAME_DEFAULT = "test_repo_alias_mapping.json";
```
damithc: Minimize putting test related code inside functional code.

src/main/java/util/RepoAliasMap.java
dariusf: `Model`, but that contains the data of a repository, which can come from anywhere and may not have these configuration details when created. We've been storing these in a `Preferences` class instead and looking them up at runtime.

I agree that we should reuse the existing mechanisms for storing preferences, and build on them if they are inadequate.

nicholaslum444: @damithc From what I see, the file that stores the repo details contains only the json data that is pulled directly from github. There's a utility that reads the entire file and converts it into a Repo class automatically. I think we shouldn't mix our own data with the github data. The separate Map class keeps the aliasing separate from the actual repo data.

Also, if it is a separate file, the user can easily assign aliases to any repositories that have not yet been downloaded. This is important for now as we don't have functionality to modify the mappings from inside the app.

nicholaslum444: @dariusf I considered placing it inside the preferences file. However, it would not be easy for a user to supply his own standalone alias mapping, as he would need to modify the actual preferences file and accidental data corruption might occur.

dariusf: Yes, session and configuration data need to be separated, but a specialised class for a particular type of configuration isn't the answer. A lot of existing functionality has to be reinvented (serialisation, synchronisation, I/O, where to put the singleton, how to manage it in tests...) and will be even more work to merge eventually.

nicholaslum444: @dariusf I see your point. So in this case, we'll place the alias mapping in the preferences file (global.json). The user will then have to edit the preferences file directly to change his mappings.

dariusf: Yeah, at least for now. Future work can be done to separate the files into one that can be safely edited, checked into version control, etc. and one that should be left alone.

nicholaslum444:
@dariusf Hi, I have updated the code to make use of Preferences instead.

Please let me know if there are further improvements I can make

damithc:
I'd rather have a List of `Repo` objects or a `Repos` class than a class called `RepoAliasMap`. The former is more OOP. Class design should follow concepts in the domain rather than data structures catered for specific features.

nicholaslum444:
@damithc Do you mean I should just rename and generalise the RepoAliasMap to `Repos`? I assume you are looking for a general class that deals with repository meta-information specific to hubturbo (i.e. not the normal issues, labels etc).

damithc:
Yes, that's right. A class to represent HT-specific meta information of a Repo.

dariusf:
Guess this is blocked until #1272 is done.

nicholaslum444:
Will be opening another PR since this is quite outdated

Issue #1181, opened 2016-01-14
Fix KeyboardShortcutTest running on small screens


m133225:
KeyboardShortcutTest fails when it tries to click the last issue, which cannot be seen in the panel

dariusf:
We might have to scroll the issue card into view before trying to click it.

xinan:
However, `KeyboardShortcutTest` clicks the last issue but hitting the `END` key instead of using the cursor. (As this test is testing for all the keyboard shortcuts) Guess the problem is not that the issue is not in view, it is that the key is not working.

PR #1260, opened 2016-02-15
for PRs, show both author and assignee in pView #691

### LINKED ISSUE #691: for PRs, show both author and assignee in pView
Assignee is usually the reviewer while the author is the person who created the PR. I think both are important.

m133225:
Fixes #691 

The current implementation shows both the author & assignee (if there is) for all issues.
1. Should it still be narrowed down to PRs only?
2. Are there any suggestions for a better layout?
3. Still has a problem where it doesn't work for bigger repos (HubTurbo & Atom)


damithc:
1. Yes, I don't think author is that important for issues. Most issues have been created long ago by past devs.
2. something like `author --> assignee`
3. how does the repo size matter?

m133225:
After some digging through the backend, I realised that the reason it doesn't work for some repos is due to permissions. By default, HubTurbo attempts to get a list of **collaborators** which is not publicly visible (even to contributors). Therefore, loading a repo that the user does not have collaboration rights in will result in no author/assignee to appear in HubTurbo.

Also, in context of this PR, **collaborators** are also unable to see a PR's author/assignee if the author/assignee is not a **collaborator**, since they are not part of the list of users loaded into HubTurbo.

I believe this is not the intended behaviour from the beginning, and all users (**collaborators** or contributors) should be able to see the assignees of issues/PRs. Should this be considered an issue and be dealt with separately, which will probably put this PR on hold?

acruis:
While the `TurboUser` object will not be inside the model, like you said, doesn't TurboIssue contain the String field `assignee` containing the userID of the assignee? I think at least the name of the assignee can be displayed. 

If you put in a check to show the user profile picture if it exists in the model, that would be even better. It would ensure that changes to the backend to download the missing TurboUsers (on a different PR, perhaps) can integrate seamlessly.

src/main/java/ui/listpanel/ListPanelCard.java
```diff
@@ -228,23 +236,55 @@ private void updateDetails() {
...
-            ImageView avatar = new ImageView();
-            if (assignee.getAvatarURL().length() != 0) {
-                Image image = assignee.getAvatar();
-                assert image != null;
-                avatar.setImage(image);
+        if (issue.isPullRequest()) {
+            HBox authorBox = createDisplayUserBox(guiElement.getAuthor(), issue.getCreator());
+            authorAssigneeBox.getChildren().add(authorBox);
+            if (issue.getAssignee().isPresent()) {
+                authorAssigneeBox.getChildren().add(new Label("--->"));
```
damithc: This should be a proper arrow, not an ascii arrow.

damithc: Remember to post a screenshot with the revised UI.

m133225:
Ready for review!

This is the revised UI when the avatars cannot be loaded:

acruis:
Are you checking for the presence of author and assignee labels on the HubTurbo window during integration testing?

m133225:
No, I haven't added the tests yet. Will probably be adding them to IssuePanelTests soon.

m133225:
@acruis Ready for review after having added the tests. However, this requires extra dummy issues, as only 1 existing dummy issue was a PR, and its author was assigned as assignee. Adding dummy issues affected many other tests due to the the issue count, which I tried to make them relative to a global variable in DummyRepoState.

src/main/java/ui/listpanel/ListPanelCard.java
```diff
@@ -191,8 +196,12 @@ private void setupIssueDetailsBox() {
         issueDetails.setPrefWrapLength(CARD_WIDTH);
         issueDetails.setHgap(3);
         issueDetails.setVgap(3);
+    }
 
-        updateDetails();
+    private void setupAuthorAssigneeBox() {
+        authorAssigneeBox.setPrefWidth(CARD_WIDTH);
+        authorAssigneeBox.setPadding(new Insets(0, 0, 1, 0));
+        //authorAssigneeBox.setSpacing(1);
```
acruis: Remove this comment?

src/test/java/guitests/KeyboardShortcutsTest.java
```diff
@@ -58,14 +59,14 @@ public void keyboardShortcutsTest() {
...
         press(JUMP_TO_FILTER_BOX);
-        assertEquals(10, selectedIssueId);
+        assertEquals(DummyRepoState.noOfDummyIssues, selectedIssueId);
         clearSelectedIssueId();
 
         // jump to first issue using number key(1) or ENTER
         push(KeyCode.ESCAPE);
         press(JUMP_TO_NTH_ISSUE_KEYS.get(1));
         PlatformEx.waitOnFxThread();
-        assertEquals(10, selectedIssueId);
```
acruis: This entire block of `JUMP_TO_NTH_ISSUE_KEYS` can be put into a while-loop. Much cleaner!

src/test/java/guitests/IssuePanelTests.java
```diff
@@ -160,4 +165,30 @@ public void testCreateLabelUpdateEventNodesForNonExistentLabel() {
...
     }
+
+    @Test
+    public void showAuthorAssignee_assignedPullRequest_authorAssigneeShown() {
+        click("#dummy/dummy_col0_filterTextField");
+        selectAll();
+        type("id:11");
+        push(KeyCode.ENTER);
+        assertEquals(true, GuiTest.exists("User 11"));
+        assertEquals(true, GuiTest.exists("\uf03e"));
```
acruis: Do put a comment to explain what this means.

acruis:
That's alright, it helps DummyRepoState become more scalable (Y)

src/test/java/guitests/IssuePanelTests.java
```diff
@@ -21,6 +22,7 @@
 
 import org.eclipse.egit.github.core.Comment;
 import org.eclipse.egit.github.core.User;
+import org.junit.Ignore;
```
acruis: This is an unused import.

acruis:
Okay, no other problems from me.

m133225:
Actually, I modified the test in KeyboardShortcutsTest, line 137 & 140 to click on a node that appears on my (small) screen, so that it would pass the test locally . This would fix #1181.
Should I revert this change? :fearful:

src/main/java/ui/listpanel/ListPanelCard.java
```diff
@@ -228,23 +236,57 @@ private void updateDetails() {
...
+        HBox authorBox = new HBox();
+        authorBox.setAlignment(Pos.BASELINE_CENTER);
+        return authorBox;
+    }
 
-            HBox assigneeBox = new HBox();
-            assigneeBox.setAlignment(Pos.BASELINE_CENTER);
-            assigneeBox.getChildren().addAll(avatar, assigneeNameLabel);
-            issueDetails.getChildren().add(assigneeBox);
+    private ImageView getAuthorAvatar(TurboUser user) {
```
damithc: Header comment for this method? I asked because the method seems to be doing more than a simple `get`. Also see if some functionality can be pushed to `TurboUser` class.

acruis:
Ready for review?

m133225:
Yes, ready for review.

acruis:
OK, no comments from me.

src/main/java/ui/listpanel/ListPanelCard.java
```diff
@@ -228,23 +236,64 @@ private void updateDetails() {
...
+        return userBox;
+    }
+
+    /**
+     * Attempts to get the TurboUser's avatar
+     *
+     * Returns an empty ImageView if avatar image obtained from TurboUser is null
+     *
+     * @param user
+     * @return ImageView that contains the avatar image if it exists
```
damithc: The header comment is a bit repetitive. May be just the `@return ImageView that contains the avatar image if it exists, an empty ImageView if it doesn't exist` is enough?

damithc:
Just a really minor change requested. Can merge even without it. No further approval needed.

src/main/java/ui/listpanel/ListPanelCard.java
m133225: Seems a bit long though, thought it would look weird. But I'll go ahead and change it then.

acruis:
@damithc As mentioned by the comments above, merging this PR will cause Travis tests to fail since #1181 isn't fixed yet. Is it OK to merge this then?

damithc:
We can wait till that issue is fixed.
@dariusf what do you think?

dariusf:
We shouldn't merge a broken test. If the test is fragile, we should try to rewrite it or work around it (with some explanation) pending a future fix.

Issue #316, opened 2015-01-06
Allow HT to stretch full height of a portrait monitor


damithc:
I have two monitors. When I maximize HT on my secondary monitor (portrait), its height is limited to the height of my primary monitor (landscape)

PR #1261, opened 2016-02-15
Stop repositioning the browser when there is insufficient space #1218

### LINKED ISSUE #1218: Stop repositioning the browser when there is insufficient space
If there is too little space to show the browser, we should maximise it and not bring it to the front, letting the user manually switch to it.

acruis:
Fixes #1218 

There are a couple of things I want to discuss regarding this PR:

* **Tests:** I don't think I can write any for the resizing logic, since this change only takes effect upon generating a chrome window, unless I refactor the dimension checks out into a public method and then perform an unit test on that?
* **How little is little?** The bView will be set to screen size if the pView width is equal to the screen width. Is this OK, or should we specify a threshold? (maximise bView only when pView is larger than 90% of the screen size, for example)
* **ui.getWidth() vs ui.getDimensions():** I defined `ui.getScreenWidth()` for `BrowserComponent` to call, in order to get the screen width. However, `ui.getScreenWidth()` does not have a lot in its logic. The alternative is just to make `ui.getDimensions()` public. Which one would be more preferable?


dariusf:
1. Sounds good.
2. Yes, we discussed having a threshold. Are you accounting for vertical space as well as horizontal?
3. Wouldn't you need the extra logic of getDimensions to get the right screen size? I think it might be better to share the method if possible.

acruis:
1. Alright, will do.
2. I'm not sure if there will be an use case for placing the bView below the pView, since the current GitHub UI doesn't look very useful when squashed vertically. @damithc any thoughts on this?
3. Yup, if you look at getScreenWidth code it's just an extra call after triggering getDimensions. I'll make the method public, then.

dariusf:
I didn't really mean placing the browser below the main window, just accounting for window height as well if you're changing the methods which compute dimensions.

acruis:
Ah, I see what you mean.

Instead of this:
screen shot 2016-02-16 at 1 03 15 pm
we should do this?
screen shot 2016-02-16 at 1 05 55 pm

dariusf:
I was actually thinking of #316. :stuck_out_tongue: I mean, if you're changing those methods, the fix may be easy. But it's not important for this one.

acruis:
Ah oops, I didn't account for #316. I don't mind doing this on the current PR, but I might have to look for a portrait monitor LOL.

acruis:
Should I implement the solution for #316 here, or do we close this PR and implement the solution in a new one?

dariusf:
If it's closely related, I guess so, otherwise never mind.

acruis:
OK, fixed #316. Do try this fix out if you have a dual monitor setup!

I'm not sure how to approach testing this fix, though.
* UI testing would make sense on a multiple monitor setup, or if it's possible to change the screen resolution on the go, but I'm not sure if either of the above are possible/practical on Travis' xvfb.
* For unit testing, the Stage's methods to set the boundaries (`setMaxHeight` etc.) are `final`, and Mockito [doesn't handle that](https://github.com/mockito/mockito/wiki/FAQ#what-are-the-limitations-of-mockito).

dariusf:
I think it should be enough just to test the underlying logic. What the UI does with those calculations is less important if we are sure they are correct.

acruis:
Alright, ready for review.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -123,12 +130,25 @@ private ChromeDriverEx createChromeDriver() {
         ChromeDriverEx driver = new ChromeDriverEx(options, isTestChromeDriver);
         WebDriver.Options manage = driver.manage();
         if (!isTestChromeDriver) {
+            setWindowSize(manage);
```
dariusf: `manage` may not be the most intuitive name. Seems like this is a configuration object of some kind?

`manage` as a method name is also rather vague.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -123,12 +130,25 @@ private ChromeDriverEx createChromeDriver() {
...
+            initialiseJNA();
+        }
+        return driver;
+    }
+
+    public void setWindowSize(WebDriver.Options manage) {
```
dariusf: This should probably have a comment. It positions the browser based on the main window, right? The UI is like an implicit parameter.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -123,12 +130,25 @@ private ChromeDriverEx createChromeDriver() {
...
+        Rectangle screenDimensions = ui.getDimensions();
+
+        if (availableDimensions.getWidth() > maximiseThreshold * screenDimensions.getWidth()) {
             manage.window().setPosition(new Point((int) ui.getCollapsedX(), 0));
             manage.window().setSize(new Dimension((int) ui.getAvailableDimensions().getWidth(),
                     (int) ui.getAvailableDimensions().getHeight()));
-            initialiseJNA();
+        } else {
+            manage.window().setPosition(new Point(0, 0));
+            manage.window().setSize(new Dimension((int) ui.getDimensions().getWidth(),
```
dariusf: The indentation of the `setSize` part could be improved a bit, to line up the parameters. Helps us see how the two lines are different.

src/main/java/ui/UI.java
```diff
@@ -331,6 +332,84 @@ private void setupGlobalKeyboardShortcuts(Scene scene) {
...
+
+    /**
+     * Calculates a stage's boundaries based on the screen information given as argument. If the list of screens
+     * is empty, then the boundaries will have the default value of Region.USE_COMPUTED_SIZE (-1.0).
+     *
+     * @param intersectingScreenBounds A list of screens, which should be non-empty
+     * @return A Region object containing relevant information in the maxHeight, minHeight and maxWidth fields.
+     */
+    public static Region getStageBounds(List<Rectangle2D> intersectingScreenBounds) {
+        List<Double> occupyingHeights = intersectingScreenBounds.stream()
```
dariusf: Since we never use this again, it could be a `Stream<Double>`, so we don't need to build the intermediate list.

src/main/java/ui/UI.java
```diff
@@ -427,6 +492,18 @@ public BrowserComponent getBrowserComponent() {
...
     /**
+     * Returns the maximum usable size of the stage, or to the screen size if this fails.
+     */
+    public Rectangle getDimensions() {
+        Optional<Rectangle> dimensions = Utility.getUsableScreenDimensions();
+        if (dimensions.isPresent()) {
+            return dimensions.get();
+        } else {
+            return Utility.getScreenDimensions();
+        }
```
dariusf: Maybe we could put all of the screen-related methods in some kind of utility class? They're spread all over now.

src/main/java/ui/UI.java
acruis: We need to reuse the list of heights to calculate the max and min.

src/main/java/ui/UI.java
acruis: Something like a `ScreenUtility`?

src/main/java/browserview/BrowserComponent.java
acruis: Does it make sense to make this a static method and then get `UI` passed in explicitly as a parameter?

src/main/java/browserview/BrowserComponent.java
acruis: Agreed, does `driverOptions` sound clearer? `manage` is taken from the chromedriver API, but if we're OK with changing the API then we can change it to `getDriverOptions`.

src/main/java/ui/UI.java
dariusf: Yup, I meant to leave it as a stream so we could do `occupyingHeights.max(...)` and `occupyingHeights.min(...)`. But it seems reusing streams is not possible (ugh). As an alternative, we have this API:

```java
DoubleStream x = DoubleStream.of(1.0, 2.0);
DoubleSummaryStatistics dss = x.summaryStatistics();
System.out.println(dss.getMax());
System.out.println(dss.getMin());
```

... but it's pretty verbose.

src/main/java/browserview/BrowserComponent.java
dariusf: Hmm, guess we can just have

```java
WebDriver.Options driverOptions = driver.manage();
```

src/main/java/browserview/BrowserComponent.java
dariusf: Nah, it's okay. We can just mention it in the comment maybe, with a high-level description.

src/main/java/ui/UI.java
acruis: Actually, using DoubleSummaryStatistics would be OK since we are already using mapToDouble. I'll change the code to use DoubleSummaryStatistics if you're fine with that :+1:

dariusf: Yeah, okay.

acruis:
Finally! Ready for review again.

dariusf:
So what was the problem with the test earlier? Is there a way to synchronise explicitly without waitOnFxThread?

acruis:
The surface problem is that clicking the filter text field does not cause it to be focused. I didn't confirm whether the main pView itself gained focus at all when the login dialog box was dismissed, but the failures seem to be because the pView was not presented as the first window. (as the login dialog box was presented first in the failing tests)

dariusf:
Would that work? If it does we could put it in some helper method that we use for setting the contents of fields.

src/main/java/ui/ScreenManager.java
```diff
@@ -0,0 +1,184 @@
...
+import util.PlatformSpecific;
+
+import javax.swing.*;
+import java.awt.*;
+import java.awt.Dimension;
+import java.util.*;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ScreenManager {
```
damithc: Header comment?

src/main/java/ui/ScreenManager.java
```diff
@@ -0,0 +1,184 @@
...
+
+    private static final Logger logger = LogManager.getLogger(ScreenManager.class.getName());
+
+    // The minimum fraction of available width to total screen width for the browser to be
+    // resized to available width.
```
damithc: This can be formatted as a javadoc comment.

damithc: BTW, it's not easy to understand what that comment means.

src/main/java/ui/ScreenManager.java
```diff
@@ -0,0 +1,184 @@
...
+        stage.setMinHeight(dimensions.getHeight());
+        stage.setMaxHeight(dimensions.getHeight());
+        stage.setX(0);
+        stage.setY(0);
+        // Then calculate sizes using screen information.
+        setBoundsFromScreen(stage);
+    }
+
+    /**
+     * Configure setBoundsFromScreen to trigger whenever the x-coordinate of the given stage changes.
```
damithc: `Configures`

src/main/java/ui/ScreenManager.java
```diff
@@ -0,0 +1,184 @@
...
+                UIManager.setLookAndFeel(
+                        UIManager.getSystemLookAndFeelClassName());
+            }
+
+            GraphicsEnvironment ge = GraphicsEnvironment.getLocalGraphicsEnvironment();
+            return Optional.of(ge.getMaximumWindowBounds());
+        } catch (Exception e) {
+            logger.error(e.getLocalizedMessage(), e);
+        }
+        return Optional.empty();
```
damithc: This `return` should be inside the `catch` block?

damithc:
Some cosmetic changes requested. Comment when done.

acruis:
Alright, added changes.

PR #1275, opened 2016-02-20
Create utility method to set TextField for UITest #1274

### LINKED ISSUE #1274: Create utility method for modifying text fields in tests
It should be a UITest method.

```java
((TextField) find(fieldName)).setText(text);
```

becomes

```java
setTextField(fieldName, text)
```

jinified:
Fixes #1274

### DIFF
--- a/src/test/java/guitests/UITest.java
+++ b/src/test/java/guitests/UITest.java
@@ -8,7 +8,6 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.*;
-import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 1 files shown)


src/test/java/guitests/UITest.java
```diff
@@ -395,6 +399,16 @@ public void clickMenu(String... menuNames) {
...
     }
 
+    /**
+     * Sets a text field with given text. Does not simulate clicking and typing 
+     * in the text field.
+     * @param fieldName
+     * @param text
+     */
+    public void setTextField(String fieldName, String text) {
+        ((TextField) find(fieldName)).setText(text);
```
dariusf: `fieldId` might be more precise. Also it would be useful to `waitUntilNodeAppears` on the field before setting it.

jinified: Was going to use `fieldId` :sob: but was afraid that it might be too specific since it is more like a JavaFX term.

dariusf:
Could you update to the latest master? I'll merge this after that.

dariusf:
Thanks!

PR #1262, opened 2016-02-16
Nondeterministic tests #952

dariusf:
Fixes #952.

On hindsight, accumulating test output in an issue didn't really do anything to fix the problems.

This fixes a few unstable tests and lays the groundwork for fixing the remaining ones. I'll make issues for each of the latter group. In future, we should probably make issues for individual unstable tests and stable ones that we see failing randomly.

Most of them weren't being synchronised properly. The remaining unstable ones have deeper issues than just lack of synchronisation.

PR #1277, opened 2016-02-21
Replace manual menu navigation in tests with clickMenu() #1094

### LINKED ISSUE #1094: Replace manual menu navigation in tests with clickMenu()
Menu tests that use `push(KeyCode.DOWN)` to traverse the menu manually should be replaced with `clickMenu(String... menuNames)` as menu manual traversal will fail in the future if the menu is rearranged.

bsmmoon:
closes #1094 

I replaced `push(KeyCode)` with `clickMenu()`.

There are some cases that needs manual traverse.
1. click() inside clickMenu() cannot handle duplicate menu names (ex. `New` and `Boards / New`)
2. click() will move mouse in straight line, so it may go over another menu and trigger it. It usually happens for diagonal move towards submenu.

These cases are left to use manual traverse.

One more remark..
From line 168, the comment says it deletes `empty` and then `Board 1`, but it actually was deleting `Board 1` and then `empty`. This branch follows the comment and deletes `empty` first and then `Board 1`.

### DIFF
--- a/src/test/java/guitests/BoardAutoCreatorTest.java
+++ b/src/test/java/guitests/BoardAutoCreatorTest.java
@@ -40,7 +40,7 @@ public class BoardAutoCreatorTest extends UITest {
 
         assertEquals(panelControl.getNumberOfSavedBoards(), 0);
 
-        clickMenu("Boards", "Auto-create", "Milestones");
+        traverseMenu("Boards", "Auto-create", "Milestones");
 
         PlatformEx.waitOnFxThread();
         assertNodeExists(hasText("Milestones board has been created and loaded.\n\n" +
@@ -70,7 +70,7 @@ public class BoardAutoCreatorTest extends UITest {
     public void workAllocationBoardAutoCreationTest() {
         assertEquals(panelControl.getNumberOfSavedBoards(), 0);
 
-        clickMenu("Boards", "Auto-create", "Work Allocation");
+        traverseMenu("Boards", "Auto-create", "Work Allocation");
 
         PlatformEx.waitOnFxThread();
         assertNodeExists(hasText("Work Allocation board has been created and loaded.\n\n" +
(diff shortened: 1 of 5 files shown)


src/test/java/unstable/MenuControlTest.java
```diff
@@ -73,8 +72,7 @@ public void menuControlTest() {
         assertEquals(false, testPref.getLastOpenBoard().isPresent());
         
         // Testing board save as
-        click("Boards");
-        push(KeyCode.DOWN).push(KeyCode.DOWN).push(KeyCode.ENTER);
+        clickMenu("Boards", "Save");
         ((TextField) find("#boardnameinput")).setText("Board 1");
```
damithc: Abstract this into a `setTextField` method?

src/test/java/unstable/MenuControlTest.java
```diff
@@ -73,8 +72,7 @@ public void menuControlTest() {
...
         click("OK");
         PlatformEx.waitOnFxThread();
```
damithc: Abstract this line into a method with a simple name?

damithc:
Let's expand this PR to replace repetitive code snippets in all UI tests into higher-level method calls.

src/test/java/unstable/MenuControlTest.java
dariusf: This is already being done in #1275.

dariusf:
This is a rather old issue, and `clickMenu` has since been subsumed by `traverseMenu`, which was added in #1262. @bsmmoon could you try that instead and see if it works better? Arrow-key traversal is evil.

bsmmoon:
@dariusf Most of the tests in MenuControlTest is removed in the master now. Shall I make similar test cases that it used to have using the new methods?
@damithc OK, I'll work on the rest of the UI tests.

dariusf:
It's been moved to BoardTests (the unstable version). No need to make new ones.

bsmmoon:
@dariusf OK, I think I pretty much changed most of the arrow-key in the guitests.

By the way, I see lots of `assertEquals(expected, actual)` has expected and actual parameters swapped. Shall I correct it or just leave it as it is?

damithc:
Post a new issue regarding swapped expected/actual and note down which places. It can be fixed as a separate PR.

bsmmoon:
@damithc Made. #1281

bsmmoon:
@dariusf I think it is ready for review.

dariusf:
All this seems like it can be just one change. Could you squash? Tell me after that and I'll merge.

bsmmoon:
@dariusf OK, done!

dariusf:
Thanks!

Issue #1233, opened 2016-02-02
Show loading indicator when there is a change of board


damithc:
When there is a change of board, panels start out as empty. Instead, we should show the loading indicator.

damithc:
When a board is loaded, all panels are created at that point. Can we show the loading indicator in new panels by default, until the panel is filled with the issue list?

damithc:
@m133225 may be you can take a look at this when you are done with #1234 ?

PR #1280, opened 2016-02-22
Show loading indicator when a filter is changed #1234

### LINKED ISSUE #1234: Show loading indicator when a filter is changed
When a filter is changed, the issues shown may not match the filter for a brief period. We should show the loading indicator during that period.

m133225:
Fixes #1234

Created 2 new events to handle setting the loading indicator when confirm is pressed on the filter.


src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -210,6 +217,14 @@ private void unindicatePanelLoading(RepoOpenedEvent e) {
         }
     }
 
+    private void indicatePanelLoading() {
+        addPanelLoadingIndication();
+    }
+
+    private void unindicatePanelLoading() {
```
damithc: better names (to avoid the non-word `unindicate`): `showPanelLoadingIndicator`, `hidePanelLodingIndicator`

damithc:
By the way, is it possible to show the indicator in front of the existing cards rather than replace the cards with the indicator?

m133225:
Possible. Tried to do it by reducing the opacity of the issues while the loading indicator is shown. However, should this also be the behaviour when loading a new repo?
Might need to refactor the way this is implemented since the issues are removed on adding the indicator, and the method responsible for this is called by the parent class FilterPanel, which does not have knowledge of the list of issues.

damithc:
Yes, it is desirable in all cases to show in front rather than in place of issues.

damithc:
Any chance of a screenshot/gif?

ndt93:
@m133225 Would be helpful to make a comment for reviewing or to remind the reviewer of something if you want faster response time, since there is no notification for changing labels or pushing commits.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -125,7 +125,9 @@ private void setUpEventHandler() {
     private Node createFilterBox() {
         filterTextField = new FilterTextField("")
                 .setOnConfirm((text) -> {
+                    Platform.runLater(() -> ui.triggerEvent(new PanelReloadingEvent()));
                     applyStringFilter(text);
+                    Platform.runLater(() -> ui.triggerEvent(new PanelReloadedEvent()));
```
ndt93: Does this take in to account if some of the `repo`s are changed in the filter? Since opening a repo happens asynchronously with `applyStringFilter(text)` and takes quite some time, so the loading indicator is most useful for it.

m133225: It doesn't, currently, since it only shows the loading gif if the panel is loading issues from the repo currently selected in the RepositorySelector.

However, a simple change to remove that check should be able to produce this behaviour with some problems. The typical execution (locally) would be PanelReloadingEvent :arrow_right:  PanelReloadedEvent :arrow_right: RepoOpeningEvent :arrow_right: RepoOpenedEvent, which will cause the spinning gif to appear and disappear twice. However, it is possible that the RepoOpeningEvent is adding to the event queue before PanelReloadedEvent (due to asynchronous execution), which may cause the gif to end prematurely.

Would the above be sufficient for this PR, or it is preferable to solve the concurrency issue now?

m133225:
Noted. So it is preferable if we comment AND change the PR labels?

ndt93:
Yup, for PRs we should both comment and change labels.

There are also a number of other concurrency issues if we want to take into account the repos opening, especially when there are multiple repos in the filter. @nicholaslum444 ran into those issues previously and put them on hold at the moment, which is the reason for opening this issue.

@damithc The loading gif in the PR doesn't take into account the repos opening i.e. the `repo`s clauses in the filter. So is the feature implemented here sufficient?

m133225:
On a side note, this is a sample screenshot of the loading indicator:

damithc:
Ideally, the indicator should show between (a) user pressing enter after changing the filer and (b) panel becomes consistent with the new filter. Yes, it should keep showing while a repo is being loaded.
If this PR is a partial solution, then remember to create another issue for the remainder.

src/main/java/util/events/PanelReloadedEventHandler.java
```diff
@@ -0,0 +1,9 @@
+package util.events;
+
+import com.google.common.eventbus.Subscribe;
+
+@FunctionalInterface
+public interface PanelReloadedEventHandler extends EventHandler {
```
damithc: Header comment?

src/main/java/util/events/PanelReloadedEvent.java
```diff
@@ -0,0 +1,4 @@
+package util.events;
+
+public class PanelReloadedEvent extends Event {
```
damithc: Header comment?

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -383,22 +386,49 @@ private void changeLabels() {
...
+    protected void showRepoOpeningIndicator(RepoOpeningEvent e) {
+        setTranslucentCellFactory();
+        setLoadingIndicator();
+    }
+
+    @Override
+    protected void hideRepoOpeningIndicator(RepoOpenedEvent e) {
+        removeLoadingIndicator();
+    }
+
```
ndt93: These two methods are very similar to the ones below, so better to merge them? We can probably use the methods' names below since they are more general.

m133225: Hmm, I was thinking that there might be a different way of handling these two events in the future.

m133225: Should they still be merged?

ndt93: I would prefer merging, also because these methods are within `ListPanel`, `RepoOpenning..` seems to be at a higher level abstraction level i.e. possibly apply to multiple panels. Those who want to implement a different effect for repository opening can do it in the future.

But either way it's up to you. Keeping this could be nice as a template to extend more effects for repo opening in the future.

m133225:
Ready for review.

These are the summarised changes:
***Changes to event:***
- Addition
 - ApplyingFilterEvent & AppliedFilterEvent
 - PrimaryRepoOpeningEvent & PrimaryRepoOpenedEvent
 - FilterRepoOpeningEvent & FilterRepoOpenedEvent
- Removal
 - RepoOpeningEvent & RepoOpenedEvent

This is to account for the different possible scenarios for the loading indicator (with faded issue cards):
- Opening a repo due to a change in Repository Selector
 - Requires loading indicator for all panels that do not specify a repo
- Opening a repo due to a change in filter (New)
 - Requires loading indicator for the changed panel
- (Simply) Applying a filter (New)
 - Requires loading indicator for the changed panel

***Changes to when the filtering is done:***
- Filtering is now done only after all the opening of repositories are completed
 - Previously, filtering was done immediately after calling the asynchronous openRepositoriesInFilters,
which caused the issues to disappear immediately, before the repos have been opened

**Issue**
There is one problem, however. For opening a repo through the filters, the loading indicator will stop after the opening of repo is completed. However, it is possible that the sorting is still incomplete at this point. This may cause the loading indicator to end slightly prematurely.

src/main/java/backend/Logic.java
```diff
@@ -118,17 +135,26 @@ public void refresh() {
 
             logger.info("Opening " + repoId);
             UI.status.displayMessage("Opening " + repoId);
-            Platform.runLater(() -> UI.events.triggerEvent(new RepoOpeningEvent(repoId, isPrimaryRepository)));
+
+            Platform.runLater(() -> UI.events.triggerEvent(new FilterRepoOpeningEvent()));
+            if (isPrimaryRepository) {
+                Platform.runLater(() -> UI.events.triggerEvent(new PrimaryRepoOpeningEvent()));
+            }
```
ndt93: I think this part can be extracted into a method.

src/main/java/backend/Logic.java
```diff
@@ -118,17 +135,26 @@ public void refresh() {
...
-                                    UI.events.triggerEvent(new RepoOpenedEvent(repoId, isPrimaryRepository))
-                            )
-                    )
+                                    UI.events.triggerEvent(new PrimaryRepoOpenedEvent()));
+                            return;
+                        }
+
+                        Platform.runLater(() ->
+                                UI.events.triggerEvent(new AppliedFilterEvent(panel.get())));
+                    })
```
ndt93: This part (the `lambda`) can be extracted into a method too. It's a bit nested here and doesn't seem to flow well with the rest of the `Future`s in the chain.

src/main/java/backend/UIManager.java
```diff
@@ -35,5 +36,9 @@ public void updateRateLimits(ImmutablePair<Integer, Long> rateLimits) {
     public List<FilterExpression> getAllFilters() {
         return ui.guiController.getAllFilters();
     }
+
+    public List<FilterPanel> getAllPanels() {
```
ndt93: Comments for this method? (since it's a public API).

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -383,22 +383,69 @@ private void changeLabels() {
...
+    /**
+     * Determines whether the loading indicator should show
+     *
+     */
+    private boolean isIndicatorApplicable() {
+        HashSet<String> allReposInFilterExpr =
+                Qualifier.getMetaQualifierContent(getCurrentFilterExpression(), QualifierType.REPO);
+        return allReposInFilterExpr.isEmpty();
+    }
```
ndt93: Can give clear behaviour explanation for this method in the comment? i.e. why is the indicator only applicable when there are no repos in filter.

src/main/java/util/events/FilterRepoOpenedEvent.java
```diff
@@ -0,0 +1,4 @@
+package util.events;
+
+public class FilterRepoOpenedEvent extends UnusedStoredReposChangedEvent {
+}
```
ndt93: Header comment for this?

src/main/java/util/events/FilterRepoOpeningEvent.java
```diff
@@ -0,0 +1,4 @@
+package util.events;
+
+public class FilterRepoOpeningEvent extends Event {
+}
```
ndt93: Header comment is missing.

m133225:
@ndt93 Hmm. Actually I'm more concerned about the concurrency parts for now. Do they seem to be managed correctly?

ndt93:
The events are triggered in the `open...` methods which is one of the reason for the loading indicator ending prematurely since multiple repos are concurrently with each other.

The issues filtering (sorting etc.) should not have a very high chance of causing concurrency problems as it would take tens of millions of issues to be sorted through to reach a noticeable time. So that leave us with the time for rendering the issue on the UI (which I think is the duration that the indicator is showing when no repo is needed to be opened) and the repo opening time which takes a few seconds due to disk IO and JSON parsing.

damithc:
A premature hiding of the indicator is fine if the gap is small and it doesn't happen often.

m133225:
On a side note, this innately fixes #1233 to some extent (thought it would be a full fix :angry:). However, it doesn't work as intended when there are multiple panels that specify the same repo - only 1 panel will show the indicator, since the other panels will not know when the repo is done opening.

m133225:
Ready for review.

src/main/java/backend/Logic.java
```diff
@@ -93,21 +95,54 @@ public void refresh() {
...
-        return openRepository(repoId, true);
+        return openRepository(repoId, true, Optional.empty());
     }
 
-    public CompletableFuture<Boolean> openRepositoryFromFilter(String repoId) {
-        return openRepository(repoId, false);
+    /**
+     * Opens the repoId if it isn't already open, else simply refreshes the UI
+     * After opening the repo, it will trigger a FilterOpeningEvent with panel as argument
+     *
```
ndt93: Is this comment still relevant? I couldn't find `FilterOpeningEvent`.

src/main/java/backend/Logic.java
```diff
@@ -93,21 +95,54 @@ public void refresh() {
...
+        return openRepository(repoId, false, Optional.of(panel));
     }
 
-    private CompletableFuture<Boolean> openRepository(String repoId, boolean isPrimaryRepository) {
+    /**
+     * Opens the repoId if it isn't already open, else simply refreshes the UI
+     *
+     * After opening the repo, it will trigger a PrimaryRepoOpenedEvent or AppliedFilterEvent depending on the
+     * caller origin determined by isPrimaryRepository
+     *
```
ndt93: Explaining in terms `isPrimaryRepository` is sufficient I suppose. The caller method is kind of vague and doesn't really matter from this method's viewpoint.

src/main/java/backend/Logic.java
```diff
@@ -93,21 +95,54 @@ public void refresh() {
...
+     * @param repoId
+     * @param isPrimaryRepository false if it was due to repo being specified in a panel's filter
+     * @param panel should be Optional.empty() if isPrimaryRepository is true, else it should contain
+     *              the panel whose filter specified the repo
```
ndt93: `isPrimaryRepository` and `panel` parameters seem to be very closely coupled i.e. the value of one parameter requires another value for the other. Is it better to use just 1 parameter for this? Then you can declare a variable `isPrimaryRepository` inside the method based on `panel`'s value to keep things readable.

src/main/java/backend/Logic.java
```diff
@@ -137,6 +167,36 @@ public void refresh() {
         });
     }
 
+    /**
+     * Triggers the relevant event based on the caller source determined by isPrimaryRepository
+     * @param isPrimaryRepository true if called by repository selector, else false
```
ndt93: Similar to above, mentioning the caller source can be confusing.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -211,37 +212,28 @@ private void applyFilterExpression(FilterExpression filter) {
...
-    }
+    protected abstract void stopLoadingAnimationIfApplicable(PrimaryRepoOpenedEvent e);
 
-    private boolean isIndicatorApplicable(boolean isPrimaryRepo) {
-        HashSet<String> allReposInFilterExpr =
-                Qualifier.getMetaQualifierContent(getCurrentFilterExpression(), QualifierType.REPO);
+    protected abstract void startLoadingAnimationIfApplicable(ApplyingFilterEvent e);
 
-        boolean isPrimaryRepoChanged = isPrimaryRepo && allReposInFilterExpr.isEmpty();
+    protected abstract void stopLoadingAnimationIfApplicable(AppliedFilterEvent e);
```
ndt93: Should be better to comment on what its means by `Applicable` here. If more details can be provided based on the concrete method, comments should be added there too.

m133225:
Ready for review.

ndt93:
No more comments from me. @damithc

src/main/java/backend/Logic.java
```diff
@@ -118,25 +154,49 @@ public void refresh() {
...
+    /**
+     * Triggers opening repo event based on isPrimaryRepository
+     * @param isPrimaryRepository triggers PrimaryRepoOpeningEvent if true, FilterRepoOpeningEvent otherwise
+     */
+    private void notifyRepoOpening(boolean isPrimaryRepository) {
+        if (isPrimaryRepository) {
+            Platform.runLater(() -> UI.events.triggerEvent(new PrimaryRepoOpeningEvent()));
+        } else {
+            Platform.runLater(() -> UI.events.triggerEvent(new FilterRepoOpeningEvent()));
+        }
```
damithc: Can use the ` ? : ` operator here?

src/main/java/backend/UpdateController.java
```diff
@@ -31,23 +32,32 @@ public UpdateController(Logic logic) {
     }
 
     /**
-     * Given a list of filter expressions, dispatch metadata update if needed and then process them to return
+     * Given a list of panels, opens the repositories specified in the panels' filters.
+     *
+     * After which, dispatch metadata update if needed and then process them to return
```
damithc: `dispatches`? `processes`?

damithc:
Couple of minor cosmetic changes suggested.

dariusf:
Coverage is [here](https://coveralls.io/builds/5320390). You might want to squash the noisy commits. Apart from that this looks good.

PR #1284, opened 2016-02-24
Supports repo as criteria to id filter #1238

### LINKED ISSUE #1238: Support specifying repo for id filters
e.g. `id:hubturbo/hubturbo/2431`
This will allow us to create watch lists using one id filter e.g.  `id:hubturbo/hubturbo/2431;teammates/repo/3489`

jinified:
Fixes #1238 

## Summary
1. Added new rule to match "#" in `Lexer`
2. Added parsing of `CompoundId` in `parseQualifierContent`

## Tested
1. Supports `id:hubturbo/hubturbo#123 ; test/test#12`

## Todo
- [x] determine edge cases for matching of "#"
- [x] write unit tests

### DIFF
--- a/src/main/java/filter/Parser.java
+++ b/src/main/java/filter/Parser.java
@@ -204,11 +204,6 @@ public final class Parser {
         }
     }
 
-    /**
-     * Checks if this is an enhanced id lookup i.e id:test/test#111
-     * @param type
-     * @return
-     */
     private boolean isCompoundId(QualifierType type, Token token) {
         return type == QualifierType.ID && token.getType() == TokenType.COMPOUND_ID;
     }
@@ -295,7 +290,7 @@ public final class Parser {
 
     /**
      * Parses filter for conjunction lookup of repo and id i.e id:test/test#id
-     * @return 
+     * @return filter expression
      */
     private FilterExpression parseCompoundId() {
         String repoId = consume().getValue();
--- a/src/main/java/filter/Parser.java
+++ b/src/main/java/filter/Parser.java
@@ -181,6 +181,8 @@ public final class Parser {
     private FilterExpression parseQualifierContent(QualifierType type) {
         if (type == QualifierType.SORT) {
             return parseSortKeys();
+        } else if (isCompoundId(type, lookAhead())) {
+            return parseCompoundId();
         } else if (isRangeOperatorToken(lookAhead())) {
             // < > <= >= [number range | date range]
             return parseRangeOperator(type, lookAhead());
@@ -200,6 +202,15 @@ public final class Parser {
             throw new ParseException(String.format("Invalid content for qualifier %s: %s",
                 type, lookAhead()));
         }
+    }
+
+    /**
+     * Checks if this is an enhanced id lookup i.e id:test/test#111
+     * @param type
+     * @return
+     */
+    private boolean isCompoundId(QualifierType type, Token token) {
+        return type == QualifierType.ID && token.getType() == TokenType.COMPOUND_ID;
     }
 
     private boolean isKeywordToken(Token token) {
@@ -280,6 +291,18 @@ public final class Parser {
         }
         assert false : "Should never reach here";
         return null;
+    }
+
+    /**
+     * Parses filter for conjunction lookup of repo and id i.e id:test/test#id
+     * @return 
+     */
+    private FilterExpression parseCompoundId() {
+        String repoId = consume().getValue();
+        FilterExpression left = new Qualifier(
+            QualifierType.REPO, repoId.substring(0, repoId.length() - 1));
+
+        return new Conjunction(left, parseQualifierContent(QualifierType.ID));
     }
 
     private FilterExpression parseDateOrDateRange(QualifierType type) {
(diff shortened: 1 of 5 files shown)


dariusf:
Hmm, I was actually hoping for this to be done in the lexer, so the entire `owner/name#id` chunk would be a single token that we could check.

Extending lookAhead like this leads to cases where it can return the wrong result given an offset outside the token stream.

jinified:
The wrong result are you referring to case like `id:hello/test##nothing` will yield an error since the content to id would yield an invalid input ?

dariusf:
Hmm, I was referring to cases where you look ahead 2 tokens but the token stream only contains 1, but now that I think about it, the last token in the stream will be EOF, so it's not that severe -- the EOF will lead to some kind of error or error-handling.

jinified:
@dariusf  Do I think I should continue with this approach ?

dariusf:
Maybe keep it in a branch temporarily. I think the lexer approach will be a lot more straightforward.

jinified:
@dariusf ready for review

src/main/java/filter/Parser.java
```diff
@@ -181,6 +181,8 @@ private FilterExpression parseSeparatedContent(Supplier<FilterExpression> conten
     private FilterExpression parseQualifierContent(QualifierType type) {
         if (type == QualifierType.SORT) {
             return parseSortKeys();
+        } else if (isCompoundId(type, lookAhead())) {
```
dariusf: Checking for failure here leads to a non-id qualifier being used with a compound id being a parse error. Could we follow the other `parse*` functions and let it fail later, as a semantic error? If for some reason we want to use compound ids with other qualifiers.

src/main/java/filter/lexer/Lexer.java
```diff
@@ -21,6 +21,7 @@
             // These have higher priority than Symbol
             new Rule("\\d{4}-\\d{1,2}-\\d{1,2}", TokenType.DATE), // YYYY-MM?-DD?
             new Rule("[A-Za-z]+(-[A-Za-z]+)*\\s*:", TokenType.QUALIFIER),
+            new Rule("[A-Za-z0-9#][/A-Za-z0-9.'+-]+#", TokenType.COMPOUND_ID),
```
dariusf: I'm not sure about this, because it doesn't check that a number actually follows. This may lead to the error appearing at the `id` qualifier that is inserted when parsing, instead of at the `id` that contains the whole compound id.

Could we use something like `[a-z]+/[a-z]+#[0-9]+` (simple example) to tokenise it instead, then extract the id as well in `parseCompoundId`? A simple `value.split("/|#")` would work to get them out, since we've confirmed that it matches the pattern.

src/main/java/filter/Parser.java
jinified: Ok noted

src/main/java/filter/lexer/Lexer.java
jinified: Do we want to enforce the pattern preceding before `#` to fits a repoId structure i.e `test/test` ? I think it is better to allow any valid symbol but I do agree that we can enforce the `id` after the `#`.

dariusf: Sure, we should do that.

So you want to allow something like `id:owner/repo#>=3`? This would be quite an elegant way to support that. In that case maybe we should do it this way. @damithc

damithc: I'm ok with that. We don't need to support those extra things at the moment though. It can be added later.

src/main/java/filter/Parser.java
jinified: I realized there is a problem for cases where the `Qualifier` accepts any string as input i.e `title` if we don't check the `QualifierType`.
Take this expression, `title:something/something#`, will result in a filter expression that try to match `(repo:something/something && id:1)` instead of the title

dariusf: Okay then. Could we check the type outside, like with `sort`? So it's more explicit which qualifiers we do this for.

src/main/java/filter/lexer/Lexer.java
dariusf: So, the regex should check that it fits syntax of a repo id. We get the range support for free, so I think we can do it this way. The range thing needs tests.

Maybe the token type could be named `COMPOUND_ID_PREFIX`?

src/test/java/tests/FilterLexerTests.java
```diff
@@ -1,55 +1,57 @@
...
+import static org.junit.Assert.assertFalse;
 
 public class FilterLexerTests {
+
     @Test
-    public void lexer() {
-        assertEquals(new Lexer("").lex(), Arrays.asList(
+    public void lex_generalInputs() {
+
+        assertEquals(getTokens(""), Arrays.asList(
```
dariusf: Maybe `tokenise`? `getTokens` is a bit general for what it does.

jinified:
@dariusf  ready for review

PR #1285, opened 2016-02-24
Fix KeyboardShortcutTest running on small screens #1181

### LINKED ISSUE #1181: Fix KeyboardShortcutTest running on small screens

m133225:
Fixes #1181.

The main reason for the test failing is that the test is trying to click on an issue outside of the small screen. 

Since the purpose of clicking on an issue is to gain focus of the panel for keyboard shortcut testing, we can simply click the first issue (id 10) which will be the least likely to fail on machines with small screens.

EDIT: It seems to pass waitUntilNodeAppears for id 1-9, but not for 10. Will look into this.


m133225:
Ready for review! A simple change of id from 1 to 9 and added some comments to explain the rationale for not choosing issue 10.
As discussed with @dariusf, there is a more complex problem in UI testing with locating the first issue card by css, which should be dealt in a separate issue.

damithc:
@m133225 PRs are to be assigned to the reviewer.

m133225:
Oops, sorry, I was trying to assign the relevant issue.

dariusf:
@acruis could you help to review this?

src/test/java/guitests/KeyboardShortcutsTest.java
```diff
@@ -169,10 +169,12 @@ public void keyboardShortcutsTest() {
         // test shortcut on repo dropdown
         doubleClick(repoSelectorComboBox);
         pushKeys(SWITCH_DEFAULT_REPO);
-        waitUntilNodeAppears("#dummy/dummy_col1_1");
+        // wait for issue 9 to appear then click on it
+        // issue 9 is chosen instead of issue 10 due to a problem with finding issue 10's node
```
acruis: What is the problem?

m133225: Not too sure, but as of now, in waitUntilNodeAppears, when it tries to find issue 10's node the second time (first time being unable to find, so it sleeps and checks again), it fails with an illegal argument exception. The code on our end seems to be fine, but when testfx's findByCssSelector is triggered, it throws an exception at line 443.

Should these details be in the comment?

acruis: Yes, at least mention that issue 10 cannot be found due to it being the first card in the panel.

m133225:
@acruis Modified the comments and updated the branch!
Hopefully this can be merged by today so #1260 can proceed! :smile:

src/test/java/guitests/KeyboardShortcutsTest.java
```diff
@@ -169,10 +169,13 @@ public void keyboardShortcutsTest() {
         // test shortcut on repo dropdown
         doubleClick(repoSelectorComboBox);
         pushKeys(SWITCH_DEFAULT_REPO);
-        waitUntilNodeAppears("#dummy/dummy_col1_1");
+        // wait for issue 9 to appear then click on it
+        // issue 9 is chosen instead of issue 10
+        // as there is a problem with finding issue 10's node due to it being the first card in the panel
+        waitUntilNodeAppears("#dummy/dummy_col1_9");
```
damithc: 'Works for 9 but not for 10' can often be traced to a short-sighted algorithm that did not cater for double-digit numbers. Sure it's not the case here?
'Doesn't work for the first one' can often be traced to a 'starts from 0 instead of 1' indexing problem.

m133225: Highly certain this is not a case of mishandling double-digit numbers.
In the context of #1260, where the number of dummy issues have been increased from 10 to 12, it still fails when trying to find issue 12, while still being successful when the issue to check for is changed from 12 to 11.

The following debug statements also suggest that this is not due to the indexing problems. The node with the id `dummy/dummy_col1_12` is produced, but cannot be found.
screen shot 2016-02-28 at 3 34 02 pm

acruis: Yup, from how you tested, it looks like it's just a problem with selecting the first line.

acruis:
No further comments from me.

m133225:
Merged (incorrectly :cry: ).

PR #1289, opened 2016-02-27
User can right-click on an issue card and 'mark all below as unread/read'

### LINKED ISSUE #1256: User can right-click on an issue card and 'mark all below as unread/read'
... so that user doesn't have to mark each one individually.

codemanmeet:
Fixes #1256. Also, if the user right clicks on an issue, it is not loaded in the bView.

### DIFF
--- a/src/main/java/ui/components/NavigableListView.java
+++ b/src/main/java/ui/components/NavigableListView.java
@@ -216,6 +216,9 @@ public abstract class NavigableListView<T> extends ScrollableListView<T> {
         return selectedIndex.map(getItems()::get);
     }
 
+    /**
+     * @return The integer value of the selected index in the list panel
+     */
     public int getSelectedIndex() {
         return selectedIndex.get();
     }
--- a/src/main/java/ui/components/NavigableListView.java
+++ b/src/main/java/ui/components/NavigableListView.java
@@ -216,4 +216,8 @@ public abstract class NavigableListView<T> extends ScrollableListView<T> {
         return selectedIndex.map(getItems()::get);
     }
 
+    public int getSelectedIndex() {
+        return selectedIndex.get();
+    }
+
 }
--- a/src/main/java/ui/listpanel/ListPanel.java
+++ b/src/main/java/ui/listpanel/ListPanel.java
@@ -49,8 +49,12 @@ public class ListPanel extends FilterPanel {
     private final ContextMenu contextMenu = new ContextMenu();
 
     private final MenuItem markAsReadUnreadMenuItem = new MenuItem();
+    private final MenuItem markAllBelowAsRead = new MenuItem();
+    private final MenuItem markAllBelowAsUnread = new MenuItem();
     private static final String markAsReadMenuItemText = "Mark as read (E)";
     private static final String markAsUnreadMenuItemText = "Mark as unread (U)";
+    private static final String markAllAsUnreadMenuItemText = "Mark all below as unread";
+    private static final String markAllAsReadMenuItemText = "Mark all below as read";
 
     private static final MenuItem changeLabelsMenuItem = new MenuItem();
     private static final String changeLabelsMenuItemText = "Change labels (L)";
@@ -295,7 +299,18 @@ public class ListPanel extends FilterPanel {
             changeLabels();
         });
 
-        contextMenu.getItems().addAll(markAsReadUnreadMenuItem, changeLabelsMenuItem);
+        markAllBelowAsRead.setText(markAllAsReadMenuItemText);
+        markAllBelowAsRead.setOnAction(e -> {
+            markAllBelowAsRead();
+        });
+
+        markAllBelowAsUnread.setText(markAllAsUnreadMenuItemText);
+        markAllBelowAsUnread.setOnAction(e -> {
+            markAllBelowAsUnread();
+        });
+
+        contextMenu.getItems().addAll(markAsReadUnreadMenuItem, markAllBelowAsRead,
+                markAllBelowAsUnread, changeLabelsMenuItem);
         contextMenu.setOnShowing(e -> updateContextMenu(contextMenu));
         listView.setContextMenu(contextMenu);
 
@@ -360,6 +375,28 @@ public class ListPanel extends FilterPanel {
 
             parentPanelControl.refresh();
             listView.selectNextItem();
+        }
+    }
+
+    private void markAllBelowAsUnread() {
+        Optional<GuiElement> item = listView.getSelectedItem();
+        if (item.isPresent()) {
+            for(int i = listView.getSelectedIndex(); i < listView.getItems().size(); i++){
+                TurboIssue issue = listView.getItems().get(i).getIssue();
+                issue.markAsUnread(UI.prefs);
+            }
+            parentPanelControl.refresh();
+        }
+    }
+
+    private void markAllBelowAsRead() {
+        Optional<GuiElement> item = listView.getSelectedItem();
+        if (item.isPresent()) {
+            for(int i = listView.getSelectedIndex(); i < listView.getItems().size(); i++){
+                TurboIssue issue = listView.getItems().get(i).getIssue();
+                issue.markAsRead(UI.prefs);
+            }
+            parentPanelControl.refresh();
         }
     }
 
--- a/src/main/java/ui/listpanel/ListPanel.java
+++ b/src/main/java/ui/listpanel/ListPanel.java
@@ -49,8 +49,8 @@ public class ListPanel extends FilterPanel {
     private final ContextMenu contextMenu = new ContextMenu();
 
     private final MenuItem markAsReadUnreadMenuItem = new MenuItem();
-    private final MenuItem markAllBelowAsRead = new MenuItem();
-    private final MenuItem markAllBelowAsUnread = new MenuItem();
+    private final MenuItem markAllBelowAsReadMenuItem = new MenuItem();
+    private final MenuItem markAllBelowAsUnreadMenuItem = new MenuItem();
     private static final String markAsReadMenuItemText = "Mark as read (E)";
     private static final String markAsUnreadMenuItemText = "Mark as unread (U)";
     private static final String markAllAsUnreadMenuItemText = "Mark all below as unread";
@@ -299,18 +299,18 @@ public class ListPanel extends FilterPanel {
             changeLabels();
         });
 
-        markAllBelowAsRead.setText(markAllAsReadMenuItemText);
-        markAllBelowAsRead.setOnAction(e -> {
+        markAllBelowAsReadMenuItem.setText(markAllAsReadMenuItemText);
+        markAllBelowAsReadMenuItem.setOnAction(e -> {
             markAllBelowAsRead();
         });
 
-        markAllBelowAsUnread.setText(markAllAsUnreadMenuItemText);
-        markAllBelowAsUnread.setOnAction(e -> {
+        markAllBelowAsUnreadMenuItem.setText(markAllAsUnreadMenuItemText);
+        markAllBelowAsUnreadMenuItem.setOnAction(e -> {
             markAllBelowAsUnread();
         });
 
-        contextMenu.getItems().addAll(markAsReadUnreadMenuItem, markAllBelowAsRead,
-                markAllBelowAsUnread, changeLabelsMenuItem);
+        contextMenu.getItems().addAll(markAsReadUnreadMenuItem, markAllBelowAsReadMenuItem,
+                markAllBelowAsUnreadMenuItem, changeLabelsMenuItem);
         contextMenu.setOnShowing(e -> updateContextMenu(contextMenu));
         listView.setContextMenu(contextMenu);
 
@@ -343,6 +343,8 @@ public class ListPanel extends FilterPanel {
         Optional<GuiElement> item = listView.getSelectedItem();
         if (item.isPresent()) {
             markAsReadUnreadMenuItem.setDisable(false);
+            markAllBelowAsReadMenuItem.setDisable(false);
+            markAllBelowAsUnreadMenuItem.setDisable(false);
             TurboIssue selectedIssue = item.get().getIssue();
 
             if (selectedIssue.isCurrentlyRead()) {
@@ -352,6 +354,8 @@ public class ListPanel extends FilterPanel {
             }
         } else {
             markAsReadUnreadMenuItem.setDisable(true);
+            markAllBelowAsReadMenuItem.setDisable(true);
+            markAllBelowAsUnreadMenuItem.setDisable(true);
         }
 
         return markAsReadUnreadMenuItem;
--- a/src/test/java/guitests/ContextMenuTests.java
+++ b/src/test/java/guitests/ContextMenuTests.java
@@ -50,11 +50,9 @@ public class ContextMenuTests extends UITest {
         sleep(EVENT_DELAY);
 
         ContextMenu contextMenu = issuePanel.getContextMenu();
-        MenuItem readUnreadItem = contextMenu.getItems().get(0);
-        MenuItem changeLabelsItem = contextMenu.getItems().get(1);
-
-        assertTrue(readUnreadItem.isDisable());
-        assertTrue(changeLabelsItem.isDisable());
+        for(MenuItem menuItem : contextMenu.getItems()){
+            assertTrue(menuItem.isDisable());
+        }
     }
 
     /**
@@ -78,6 +76,28 @@ public class ContextMenuTests extends UITest {
         click("Mark as unread (U)");
         sleep(EVENT_DELAY);
         assertFalse(listPanelCell.getIssue().isCurrentlyRead());
+    }
+
+    /**
+     * Tests selecting "Mark all below as read" and "Mark all below as unread" context menu items
+     */
+    @Test
+    public void markAllBelowAsReadUnread() {
+        click("#dummy/dummy_col0_9");
+        rightClick("#dummy/dummy_col0_9");
+        click("Mark all below as read");
+        for(int i=9; i >= 1; i--){
+            ListPanelCell listPanelCell = find("#dummy/dummy_col0_" + String.valueOf(i));
+            assertTrue(listPanelCell.getIssue().isCurrentlyRead());
+        }
+
+        click("#dummy/dummy_col0_9");
+        rightClick("#dummy/dummy_col0_9");
+        click("Mark all below as unread");
+        for(int i=9; i >= 1; i--){
+            ListPanelCell listPanelCell = find("#dummy/dummy_col0_" + String.valueOf(i));
+            assertFalse(listPanelCell.getIssue().isCurrentlyRead());
+        }
     }
 
     /**


src/main/java/ui/components/NavigableListView.java
```diff
@@ -216,4 +216,11 @@ public void selectNextItem() {
         return selectedIndex.map(getItems()::get);
     }
 
+    /**
+     * @return The integer value of the selected index in the list panel
+     */
+    public int getSelectedIndex() {
+        return selectedIndex.get();
```
dariusf: This will fail if the optional is empty.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -153,6 +163,22 @@ private void setupListView() {
         });
     }
 
+    private void setupIssueSelectedTrigger(){
+        // filter right mouse button out. Choose only the left mouse button for triggering IssueSelectedEvent
+        listView.addEventFilter(MouseEvent.MOUSE_PRESSED, e -> {
+            if (e.isSecondaryButtonDown()) {
+                shouldTriggerIssueSelectedEvent = false;
```
dariusf: A better way to do this is to modify setOnItemSelected to pick up which key is being pressed, then pass that to the callback above via the parameter.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -359,10 +400,33 @@ private void markAsRead() {
             issue.markAsRead(UI.prefs);
 
             parentPanelControl.refresh();
+            shouldTriggerIssueSelectedEvent = true;
             listView.selectNextItem();
         }
     }
 
+    //Mark all issues below and including the selected issue as unread.
```
dariusf: You should use a proper javadoc comment.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -359,10 +400,33 @@ private void markAsRead() {
...
+    private void markAllBelowAsUnread() {
+        if (listView.getSelectedIndex() >= 0) {
```
dariusf: When is this negative?

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -359,10 +400,33 @@ private void markAsRead() {
...
+            for (int i = listView.getSelectedIndex(); i < listView.getItems().size(); i++){
+                TurboIssue issue = listView.getItems().get(i).getIssue();
+                issue.markAsUnread(UI.prefs);
+            }
+            parentPanelControl.refresh();
```
dariusf: Why is there a need for this?

src/test/java/guitests/ContextMenuTests.java
```diff
@@ -97,4 +105,23 @@ public void test3() {
...
 
+    private void markAllReadUnreadHelper(boolean isTestingRead){
+        click("#dummy/dummy_col0_9");
+        rightClick("#dummy/dummy_col0_9");
+        if (isTestingRead){
+            click("Mark all below as read");
+        } else {
+            click("Mark all below as unread");
+        }
+        PlatformEx.waitOnFxThread();
```
dariusf: Why is this needed?

src/main/java/ui/listpanel/ListPanel.java
codemanmeet: This is necessary. The bold text for unread issues will not appear unless we do a quick refresh.

src/test/java/guitests/ContextMenuTests.java
codemanmeet: Removed this. I just added this to see why the test was failing on the CI but passing every time locally. Still have to debug this further.

src/main/java/ui/listpanel/ListPanel.java
codemanmeet: Yup, using `BiConsumer` now!

src/main/java/ui/components/NavigableListView.java
codemanmeet: Added an `isPresent` check for it.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -363,6 +383,34 @@ private void markAsRead() {
         }
     }
 
+    /**
+     * Mark all issues below and including the selected issue as unread.
+     */
+    private void markAllBelowAsUnread() {
+        if (listView.getSelectedIndex() >= 0) {
```
damithc: Reverse the condition and eliminate the nesting.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -363,6 +383,34 @@ private void markAsRead() {
...
+
+            parentPanelControl.refresh();
+        }
+    }
+
+    /**
+     * Mark all issues below and including the selected issue as read.
+     */
+    private void markAllBelowAsRead() {
+        if (listView.getSelectedIndex() >= 0) {
```
damithc: same here

src/test/java/guitests/ContextMenuTests.java
```diff
@@ -97,4 +105,22 @@ public void test3() {
         sleep(EVENT_DELAY);
     }
 
+    private void markAllReadUnreadHelper(boolean isTestingRead){
```
damithc: Is this a method name or a class name?

damithc: Add a header comment.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -216,4 +222,11 @@ public void selectNextItem() {
         return selectedIndex.map(getItems()::get);
     }
 
+    /**
+     * @return The integer value of the selected index in the list panel or -1 if it is null.
```
damithc: The code mentions `null` anywhere.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -363,6 +383,34 @@ private void markAsRead() {
...
+            for (int i = listView.getSelectedIndex(); i < listView.getItems().size(); i++) {
+                TurboIssue issue = listView.getItems().get(i).getIssue();
+                issue.markAsRead(UI.prefs);
+            }
+
+            parentPanelControl.refresh();
+        }
+    }
```
damithc: The two methods look very similar. Can reuse?

src/main/java/ui/components/NavigableListView.java
```diff
@@ -216,4 +222,11 @@ public void selectNextItem() {
...
+     */
+    public int getSelectedIndex() {
+        return selectedIndex.isPresent() ? selectedIndex.get() : -1;
+    }
```
damithc: I'm not sure you need this method at all.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -143,7 +149,7 @@ private void setupKeyEvents() {
                 e.consume();
                 if (selectedIndex.isPresent()) {
                     logger.info("Enter key selection on item " + selectedIndex.get());
-                    onItemSelected.accept(selectedIndex.get());
+                    onItemSelected.accept(selectedIndex.get(), SHOULD_LOAD_BVIEW_ISSUE);
```
damithc: This line is repeated in a few places. Can abstract into a method with an explanatory name?

src/main/java/ui/listpanel/ListPanel.java
codemanmeet: You need this to figure out if selected index was present.

src/main/java/ui/components/NavigableListView.java
codemanmeet: You need to find out the index to know where to start iterating over the items that need to be marked read/unread.

damithc: This method is basically replaces null with -1 and then you check for -1 later. Why not just return the `selectedIndex` instead and let the caller deals with null?

src/main/java/ui/listpanel/ListPanel.java
damithc:
```java
if (!listView.getSelectedIndex().isPresent()) { return; }
//rest of the code, no nesting
```

src/main/java/ui/components/NavigableListView.java
codemanmeet: Yup, got it! Thanks!

src/main/java/ui/components/NavigableListView.java
```diff
@@ -45,7 +46,9 @@
     // (when nothing is selected, both should be no-ops)
     private boolean saveSelectionCalled = false;
 
-    private IntConsumer onItemSelected = i -> {};
+    public static final boolean SHOULD_LOAD_BVIEW_ISSUE = true;
```
dariusf: Details about what this field is going to be used for shouldn't leak into this class. This just indicates whether or not the selection was caused by a right-click, right?

Actually, is there a case where we want a right click to trigger a selection event? If not, maybe we don't even need to change this.

codemanmeet: Yup, we want a selection event to occur on a right click. The issue should be selected on a right click but the bView should not open. The relevant context menu opens up for the issue on a right click.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -45,7 +46,9 @@
...
+
+    private BiConsumer<Integer, Boolean> onItemSelected = (i, e) -> {};
```
dariusf: Since this is the definition site, maybe name the parameters here better, so we take the chance to explain them.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -111,25 +114,36 @@ public void restoreSelection() {
...
             }
         }
     }
 
+    /**
+     * Passes the relevant index and boolean value for whether the bview should be loaded
+     * to the callback variable onItemSelected
+     */
+    private void setItemSelectedParams(int index, boolean shouldLoadBViewIssue){
+        onItemSelected.accept(index, shouldLoadBViewIssue);
```
dariusf: I think the interface can be improved so this doesn't just delegate. How about something like `itemSelected(index)` and `itemSelectedSecondary(index)`? Exact names should be improved, turned into verbs, etc. but now we have two methods for dealing with the two cases, and it's clear from call sites how the selection callback is being invoked. We shouldn't need the constant with this.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -111,25 +114,36 @@ public void restoreSelection() {
...
-
             // The currently-selected index is sometimes -1 when an issue is clicked.
             // When this happens we ignore this event.
-
             if (currentlySelected != -1) {
                 selectedIndex = Optional.of(currentlySelected);
 
                 logger.info("Mouse click on item index " + selectedIndex.get());
-                onItemSelected.accept(selectedIndex.get());
+                // whenever the right mouse button is clicked, do not load issue on bView.
```
dariusf: I don't think this comment is necessary. Same reason as above: this class doesn't know about the browser or how callers are going to use it.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -135,14 +139,15 @@ private void setupListView() {
         setupKeyboardShortcuts();
         setupContextMenu();
 
-        listView.setOnItemSelected(i -> {
+        listView.setOnItemSelected((i, e) -> {
```
dariusf: The `e` parameter can have a better name. `wasRightClick` or something? Then we can do `if (wasRightClick) { ... }` below.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -295,7 +300,18 @@ private ContextMenu setupContextMenu() {
             changeLabels();
         });
 
-        contextMenu.getItems().addAll(markAsReadUnreadMenuItem, changeLabelsMenuItem);
+        markAllBelowAsReadMenuItem.setText(markAllAsReadMenuItemText);
+        markAllBelowAsReadMenuItem.setOnAction(e -> {
+            markAllItemsBelow(IS_READ);
```
dariusf: This reads slightly better without the `IS_` prefix to me.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -111,25 +114,43 @@ public void restoreSelection() {
...
                 // The next item will be considered selected
-                onItemSelected.accept(nextIndex);
+                setItemSelectedParamsForPrimaryKey(nextIndex);
             }
         }
     }
 
+    /**
+     * Passes the relevant index to the onItemSelected callback variable.
+     * This method is called for primary keys
```
damithc: What's this 'primary key'?

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -363,6 +383,27 @@ private void markAsRead() {
         }
     }
 
+    /**
+     * Mark all items on and below the selected item in the list view as read/unread
+     * @param isRead isRead set to true mark all items on and below the selected issue as read.
```
damithc: This comment line doesn't read well

src/test/java/guitests/ContextMenuTests.java
```diff
@@ -97,4 +105,27 @@ public void test3() {
         sleep(EVENT_DELAY);
     }
 
+    /**
+     * Tests marked issues from Issue #9 in the UI right to the end of the list as read/unread
+     * @param isTestingRead If true, tests whether issues on/below the selected issue are being correctly marked read
+     *                      If false, tests whether issues on/below the selected are being correctly marked unread
+     */
+    private void markAllReadOrUnread(boolean isTestingRead){
```
damithc: Method name doesn't match the intent of the method. 

In fact, I'm not sure if this method is a good unit of abstraction. Methods are not simply containers for a chunk of statements. They are abstractions. Here are some better choices for methods:
`markIssuesBelow(int index, boolean isMarkAsRead)`
`verifyReadStatusOfIssuesBelow(int index, boolean isExpectedStatusRead)`

Using these abstractions, you can do a better job of testing. e.g.
```
take a panel of 10 issues.
mark all below 5.
check if all below 5 are marked as read.
mark all below 5 as read (again).
ensure they are still marked as read.
...
```

src/main/java/ui/components/NavigableListView.java
```diff
@@ -111,25 +114,43 @@ public void restoreSelection() {
...
+                setItemSelectedParamsForPrimaryKey(nextIndex);
             }
         }
     }
 
+    /**
+     * Passes the relevant index to the onItemSelected callback variable.
+     * This method is called for primary keys e.g. keyboard keys and the left mouse key
+     */
+    private void setItemSelectedParamsForPrimaryKey(int index){
```
damithc: The name is hard to understand. I can't think of a better name though. @darius?

src/test/java/guitests/ContextMenuTests.java
```diff
@@ -97,4 +106,46 @@ public void test3() {
...
+
+    /**
+     * Brings the {index} issue and the issues below that in view for a list panel.
+     */
+    private void bringSelectedAndIssuesBelowInFocus(int index) {
+        click("#dummy/dummy_col0_" + index);
+        for (int i = 0; i < 15; i++) {
+            push(KeyCode.DOWN);
+        }
+    }
```
damithc: How can you have multiple issues in focus at the same time? Isn't this method same as moving focus to the last issue in the panel?

src/test/java/guitests/ContextMenuTests.java
```diff
@@ -97,4 +106,46 @@ public void test3() {
...
+    /**
+     * Tests whether a list panel cell corresponding to a particular index is marked as read/unread
+     */
+    private void verifyReadStatusOfIssue(int index, boolean isExpectedStatusRead){
+        ListPanelCell listPanelCell = find("#dummy/dummy_col0_" + index);
+        if (isExpectedStatusRead){
+            assertTrue(listPanelCell.getIssue().isCurrentlyRead());
+        } else {
+            assertFalse(listPanelCell.getIssue().isCurrentlyRead());
+        }
```
damithc: Instead of the `if-else`, you can use `assertEquals` and pass the `isExpectedStatusRead` as the expected?

src/test/java/guitests/ContextMenuTests.java
```diff
@@ -97,4 +106,46 @@ public void test3() {
...
     }
 
+    /**
+     * Marks and tests issues in a panel of 10 issues
+     * from Issue #{index} in the UI right to the end of the list as read/unread
+     * @param isMarkAsRead If true, tests whether issues on/below the selected issue are being correctly marked read
+     *                      If false, tests whether issues on/below the selected are being correctly marked unread
+     * @param index The issue number in the panel
+     */
+    private void markIssuesBelow(int index, boolean isMarkAsRead){
```
damithc: This method is doing more than just marking. --> the method name is not good.

src/test/java/guitests/ContextMenuTests.java
```diff
@@ -81,6 +80,16 @@ public void test2() {
     }
 
     /**
+     * Tests selecting "Mark all below as read" and "Mark all below as unread" context menu items
+     */
+    @Test
+    public void markAllBelowAsReadUnread_tenIssuesInListView_markIssue9andBelowReadUnread() {
+        bringSelectedAndIssuesBelowInFocus(7);
+        markIssuesBelow(7, true);
+        markIssuesBelow(7, false);
```
damithc: As mentioned in my previous review, this test can be stronger. Here you are testing two typical cases only. How do you know if calling the `mark as read` twice will not toggle the state back to unread? The test cases don't cover that.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -45,7 +46,9 @@
     // (when nothing is selected, both should be no-ops)
     private boolean saveSelectionCalled = false;
 
-    private IntConsumer onItemSelected = i -> {};
+    public static final boolean IS_PRIMARY_KEY = true;
+
+    private BiConsumer<Integer, Boolean> onItemSelected = (index, wasRightClick) -> {};
```
dariusf: The name `wasRightClick` is kind of inconsistent with how it works (the value is true only if it was *not* a right click).

src/main/java/ui/components/NavigableListView.java
dariusf: `triggerItemSelected` (reasonable default, hence shorter name) and `triggerItemSelectedSecondary`?

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -135,14 +139,15 @@ private void setupListView() {
...
         setupContextMenu();
 
-        listView.setOnItemSelected(i -> {
+        listView.setOnItemSelected((index, wasPrimaryKey) -> {
             updateContextMenu(contextMenu);
 
-            TurboIssue issue = listView.getItems().get(i).getIssue();
-            ui.triggerEvent(
+            TurboIssue issue = listView.getItems().get(index).getIssue();
+            if (wasPrimaryKey.booleanValue()) {
```
dariusf: Explicit unboxing is unnecessary.

src/main/java/ui/components/NavigableListView.java
damithc: Sounds good.

src/main/java/ui/components/NavigableListView.java
codemanmeet: oops, it should be called `wasPrimaryKey`

src/test/java/guitests/ContextMenuTests.java
codemanmeet: Ok, I'll call mark issues as read as well as mark issues as unread `twice`then.

src/test/java/guitests/ContextMenuTests.java
codemanmeet: By focus I mean in the view of the panel. I want to scroll down so that issue `index` and all issues below can be seen on the UI of the panel. Basically, move towards the end of the panel.

src/test/java/guitests/ContextMenuTests.java
damithc: It's not a matter of blindly following my comments. The current test is not strong enough. The case I mentioned is just an example. Try to figure out the corner cases and come up with a stronger set of test cases. 
BTW, are you testing the 'no loading when right-click' behavior?

src/test/java/guitests/ContextMenuTests.java
damithc: 1. Name the method accordingly. `bringIssuesBelowIntoView`? 
2. The success depends on the number of issues and the index. In some cases, the panel many not be long enough to keep in view all issues in concern.
3. Why do you want to do this anyway?

codemanmeet: We want to obtain the `listPanelCell` using the `find` method and check whether all those issues contained in those `listPanelCell`s are marked correctly as read/unread. The `find` method works only if those issues can be `seen` in the panel. Hence the need to scroll down towards the end.

src/test/java/guitests/ContextMenuTests.java
codemanmeet: I have added tests for checking whether `IssueSelectedEvent` is called upon a left click and a key press and not called upon a right click. I am also checking whether all context menu items are disabled for the empty list case so corner case is being checked. `Mark all below as read` and `Mark all below as unread` are also being checked thoroughly now so I think we are good now! :+1:

src/main/java/ui/components/NavigableListView.java
```diff
@@ -111,25 +114,43 @@ public void restoreSelection() {
...
 
                 // The next item will be considered selected
-                onItemSelected.accept(nextIndex);
+                triggerItemSelected(nextIndex);
             }
         }
     }
 
+    /**
+     * Passes the relevant index to the onItemSelected callback variable.
```
damithc: Can this be rephrased to make it easier to understand? Similar to how we changed the method name?

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -135,14 +139,15 @@ private void setupListView() {
         setupKeyboardShortcuts();
         setupContextMenu();
 
-        listView.setOnItemSelected(i -> {
+        listView.setOnItemSelected((index, wasRightKey) -> {
```
damithc: `RightKey`?

src/test/java/guitests/ContextMenuTests.java
```diff
@@ -81,6 +81,22 @@ public void test2() {
     }
 
     /**
+     * Tests selecting "Mark all below as read" and "Mark all below as unread" context menu items
+     */
+    @Test
+    public void markAllBelowAsReadUnread_twelveIssuesInListView_markIssue7andBelowReadUnread() {
```
damithc: This name doesn't look like it follows the convention. The third part should be the expected result?

src/test/java/guitests/ContextMenuTests.java
```diff
@@ -81,6 +81,22 @@ public void test2() {
...
+        scrollToTheEndOfThePanel();
+        for (int i = 0; i < 2; i++) {
+            markAndVerifyIssuesBelow(7, true);
+        }
+        for (int i = 0; i < 2; i++) {
+            markAndVerifyIssuesBelow(7, false);
+        }
+        markAndVerifyIssuesBelow(1, true);
+        markAndVerifyIssuesBelow(1, false);
```
damithc: Add comments to explain what exactly you are testing. Why 1? Why 7?

src/test/java/guitests/ContextMenuTests.java
```diff
@@ -97,4 +113,42 @@ public void test3() {
...
+    /**
+     * Marks and tests issues in a panel of 12 issues
+     * from Issue #{index} in the UI right to the end of the list as read/unread
+     * @param isMarkAsRead If true, tests whether issues on/below the selected issue are being correctly marked read
+     *                      If false, tests whether issues on/below the selected are being correctly marked unread
+     * @param index The issue number in the panel
+     */
+    private void markAndVerifyIssuesBelow(int index, boolean isMarkAsRead){
+        click("#dummy/dummy_col0_" + index);
+        rightClick("#dummy/dummy_col0_" + index);
```
damithc: I don't like how `"#dummy/dummy_col0_"` is being repeated a lot, all over the place. Is there a way to avoid it?

src/main/java/ui/components/NavigableListView.java
codemanmeet: Sure!!

src/test/java/guitests/ContextMenuTests.java
codemanmeet: We need this because the Test FX makes use of UI element identifiers to essentially `find` elements on the UI and performs testing there onwards. I have extracted this to a constant String to make it neater.

src/test/java/guitests/ContextMenuTests.java
codemanmeet: 1 - to test for the corner case where there are no issues `below` but `on` is satisfied
7 - to account for the general case where we pick an issue in the middle or anywhere on the list and check whether the issues on/below have been marked

Empty list has been checked with context menu items disabled in the `Disabled` tests.

codemanmeet:
@dariusf Ready for review! I've moved the unstable context menu tests to the unstable folder.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -45,7 +46,9 @@
     // (when nothing is selected, both should be no-ops)
     private boolean saveSelectionCalled = false;
 
-    private IntConsumer onItemSelected = i -> {};
+    public static final boolean IS_SECONDARY_KEY = true;
+
+    private BiConsumer<Integer, Boolean> onItemSelected = (index, rightKey) -> {};
```
dariusf: We should decide on the term we like better ('secondary key' or 'right mouse button'?) and be consistent everywhere.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -111,25 +114,43 @@ public void restoreSelection() {
...
-                onItemSelected.accept(nextIndex);
+                triggerItemSelected(nextIndex);
             }
         }
     }
 
+    /**
+     * Passes the relevant index to the onItemSelected callback variable as well as "false" value as the second
+     * parameter indicating that this method is called for triggering callbacks for left clicks, key presses etc.
+     */
```
dariusf: This description is kind of operational -- we can more or less see that from the code. Maybe you could describe how this works at a higher level of abstraction?

src/main/java/ui/components/NavigableListView.java
```diff
@@ -111,25 +114,43 @@ public void restoreSelection() {
...
-
             if (currentlySelected != -1) {
                 selectedIndex = Optional.of(currentlySelected);
 
                 logger.info("Mouse click on item index " + selectedIndex.get());
-                onItemSelected.accept(selectedIndex.get());
+                if (e.getButton().equals(MouseButton.SECONDARY)){
+                    triggerItemSelectedSecondary(selectedIndex.get());
+                } else {
+                    triggerItemSelected(selectedIndex.get());
```
dariusf: We can just use currentlySelected here.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -401,4 +442,11 @@ protected void removePanelLoadingIndication() {
         logger.info("Preparing to remove panel loading indication");
         listView.getStyleClass().removeIf(cssClass -> cssClass.equals("listview-loading"));
     }
+
+    /**
+     * @return listView object associated with this list panel.
+     */
+    public IssueListView getListView(){
```
dariusf: Instead of adding this just for tests, you can use reflection.

src/main/java/ui/components/NavigableListView.java
codemanmeet: Alright! Let's stick with right click then!

src/main/java/ui/components/NavigableListView.java
```diff
@@ -111,25 +114,46 @@ public void restoreSelection() {
...
 
                 // The next item will be considered selected
-                onItemSelected.accept(nextIndex);
+                triggerItemSelected(nextIndex);
             }
         }
     }
 
+    /**
+     * This method triggers the callback for onItemSelected with the index and "false" parameter indicating that
```
damithc: `This method` is redundant. Be consistent with how we usually comment methods. 'Triggers the ..`

src/main/java/ui/components/NavigableListView.java
```diff
@@ -111,25 +114,46 @@ public void restoreSelection() {
...
+     * it is called for left clicks, key presses etc but not right clicks.
+     * The boolean parameter enables us to perform different actions
+     * (e.g. deciding whether the bView should be loaded with that issue) based on different kinds of inputs
+     */
```
damithc: This comment sounds like you are deliberately trying to perplex the reader. :-|
It's a simple method that needs a simple explanation.
`Triggers the 'selected' action on the card at index.`

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -363,6 +383,27 @@ private void markAsRead() {
         }
     }
 
+    /**
+     * Mark all items on and below the selected item in the list view as read/unread
```
damithc: `Marks the selected item and items below it ...`

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -295,7 +300,18 @@ private ContextMenu setupContextMenu() {
...
+            markAllItemsBelow(READ);
+        });
+
+        markAllBelowAsUnreadMenuItem.setText(markAllAsUnreadMenuItemText);
+        markAllBelowAsUnreadMenuItem.setOnAction(e -> {
+            markAllItemsBelow(!READ);
+        });
+
+        contextMenu.getItems().addAll(markAsReadUnreadMenuItem, markAllBelowAsReadMenuItem,
+                markAllBelowAsUnreadMenuItem, changeLabelsMenuItem);
```
damithc: Indent like this:
```java
contextMenu.getItems().addAll(markAsReadUnreadMenuItem, markAllBelowAsReadMenuItem,
                              markAllBelowAsUnreadMenuItem, changeLabelsMenuItem);
```

src/test/java/unstable/ContextMenuUnstableTests.java
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * Marks and tests issues in a panel of 12 issues
+     * from Issue #{index} in the UI right to the end of the list as read/unread
+     * @param isMarkAsRead If true, tests whether issues on/below the selected issue are being correctly marked read
+     *                      If false, tests whether issues on/below the selected are being correctly marked unread
+     * @param index The issue number in the panel
+     */
+    private void markAndVerifyIssuesBelow(int index, boolean isMarkAsRead){
+        ListPanel issuePanel = find(PANEL_IDENTIFIER);
+        click(PANEL_IDENTIFIER + "_" + index);
```
damithc: Can't we have a method `clickItemAt(index)` somewhere? Not sure where to put it though. @dariusf ?

src/test/java/unstable/ContextMenuUnstableTests.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * Marks and tests issues in a panel of 12 issues
+     * from Issue #{index} in the UI right to the end of the list as read/unread
+     * @param isMarkAsRead If true, tests whether issues on/below the selected issue are being correctly marked read
+     *                      If false, tests whether issues on/below the selected are being correctly marked unread
+     * @param index The issue number in the panel
+     */
+    private void markAndVerifyIssuesBelow(int index, boolean isMarkAsRead){
```
damithc: @dariusf I remember we now have a general way to navigate a menu. May be we can have a general way to click on a menu time in the context menu? It can be a separate issue.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -111,25 +114,41 @@ public void restoreSelection() {
...
 
                 // The next item will be considered selected
-                onItemSelected.accept(nextIndex);
+                triggerItemSelected(nextIndex);
             }
         }
     }
 
+    /**
+     * Triggers the selected item on the card at the specified {index} for actions that are not right clicks
```
damithc: Is `{index}` some standard notation used in Javadocs?

src/test/java/guitests/IssueSelectedEventTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Before
+    public void setup() {
+        UI.events.registerEvent((IssueSelectedEventHandler) e -> IssueSelectedEventTest.increaseEventTestCount());
+    }
+
+    /**
+     * Tests whether right click doesn't trigger IssueSelectedEvent
+     */
```
damithc: Doesn't look like this comment says anything more than the method name

src/test/java/unstable/ContextMenuUnstableTests.java
```diff
@@ -0,0 +1,100 @@
...
+            listViewField.setAccessible(true);
+            IssueListView listViewValue = (IssueListView) listViewField.get(issuePanel);
+
+            // scrolls to the end of the panel
+            listViewValue.scrollAndShow(12);
+        } catch (NoSuchFieldException e) {
+            throw new RuntimeException(e);
+        } catch (IllegalAccessException e) {
+            throw new RuntimeException(e);
+        }
```
damithc: These two catch blocks can be merged. In any case, catching and rethrowing exceptions inside tests is meaningless. Just remove the try clause.

src/test/java/unstable/ContextMenuUnstableTests.java
```diff
@@ -0,0 +1,100 @@
...
+
+
+        //checking for issue #7 and below twice to make sure issues are marked correctly for read/unread cases
+        for (int i = 0; i < 2; i++) {
+            markAndVerifyIssuesBelow(7, true);
+        }
```
damithc: You used a `for` loop to do something twice? Just repeat the line twice.

src/test/java/unstable/ContextMenuUnstableTests.java
```diff
@@ -0,0 +1,100 @@
...
+            markAndVerifyIssuesBelow(7, false);
+        }
+
+        //checking for the last issue to ensure correct marking of issues on/below as read/unread when no issues below
+        markAndVerifyIssuesBelow(1, true);
+        markAndVerifyIssuesBelow(1, false);
+    }
+
+    /**
+     * Marks and tests issues in a panel of 12 issues
```
damithc: Should the comment mention number 12? How is that relevant to the method?

src/test/java/unstable/ContextMenuUnstableTests.java
```diff
@@ -0,0 +1,100 @@
...
+     * from Issue #{index} in the UI right to the end of the list as read/unread
+     * @param isMarkAsRead If true, tests whether issues on/below the selected issue are being correctly marked read
+     *                      If false, tests whether issues on/below the selected are being correctly marked unread
+     * @param index The issue number in the panel
+     */
+    private void markAndVerifyIssuesBelow(int index, boolean isMarkAsRead){
+        ListPanel issuePanel = find(PANEL_IDENTIFIER);
```
damithc: I think this should be a parameter to this method. The method should be able to work with any panel, no?

src/test/java/unstable/ContextMenuUnstableTests.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+        if (isMarkAsRead){
+            click("Mark all below as read");
+        } else {
+            click("Mark all below as unread");
+        }
+        for (int i = index; i >= 1; i--){
+            verifyReadStatusOfIssue(i, isMarkAsRead);
+        }
+    }
```
damithc: Just because two things always happen together doesn't mean they should be put in one method. A method should do only one thing. The methods that should be here are:
`verifyReadStatusOfIssuesBelow(panel, index, expectedStatus)` 
`markReadStatusOfIssuesBelow(panel, index, statusToMark)`

src/test/java/unstable/ContextMenuUnstableTests.java
codemanmeet: The try clause is compulsory - Error:(41, 67) java: unreported exception java.lang.NoSuchFieldException; must be caught or declared to be thrown.
Error:(43, 76) java: unreported exception java.lang.IllegalAccessException; must be caught or declared to be thrown

damithc: Add a `throws` clause lah. That's what I meant :-)

codemanmeet: haha sure :)

src/test/java/unstable/ContextMenuUnstableTests.java
dariusf: Yes, that would be nice to have. Maybe we could also factor out the way these id strings are generated, so test code doesn't need to have such constants in them. This can probably go in UITest if it's generic enough to click multiple kinds of things.

src/test/java/unstable/ContextMenuUnstableTests.java
dariusf: Yes, that's a good idea.

src/test/java/unstable/ContextMenuUnstableTests.java
```diff
@@ -0,0 +1,97 @@
...
+        verifyReadStatusOfIssuesBelow(issuePanel, 1, false);
+    }
+
+    /**
+     * Marks issues in the panel from Issue index in the UI right to the end of the list as read/unread
+     * @param isMarkAsRead If true, marks the selected issue and below as read
+     *                     If false, marks the selected issue and below as unread
+     * @param index The issue number in the panel
+     */
+    private void markAndVerifyIssuesBelow(ListPanel issuePanel, int index, boolean isMarkAsRead){
```
damithc: Method name ???

damithc: Use `click` in the method name.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -49,9 +49,13 @@
     private final ContextMenu contextMenu = new ContextMenu();
 
     private final MenuItem markAsReadUnreadMenuItem = new MenuItem();
+    private final MenuItem markAllBelowAsReadMenuItem = new MenuItem();
+    private final MenuItem markAllBelowAsUnreadMenuItem = new MenuItem();
     private static final String markAsReadMenuItemText = "Mark as read (E)";
     private static final String markAsUnreadMenuItemText = "Mark as unread (U)";
-
+    private static final String markAllAsUnreadMenuItemText = "Mark all below as unread";
+    private static final String markAllAsReadMenuItemText = "Mark all below as read";
```
damithc: Shouldn't these constants be named in ALL_CAPS?

src/test/java/unstable/ContextMenuUnstableTests.java
```diff
@@ -0,0 +1,97 @@
...
+
+        click(PANEL_IDENTIFIER + "_filterTextField");
+        push(KeyCode.ENTER);
+    }
+
+    /**
+     * Tests selecting "Mark all below as read" and "Mark all below as unread" context menu items
+     */
+    @Test
+    public void markAllBelowAsReadUnread_twelveIssuesInListView_issuesCorrectlyMarkedReadUnread()
```
damithc: I don't think `twelve` is important enough to be mentioned in the method name. `_multipleIssuesInPanel_issuesCorrectlyMarked`?

src/test/java/unstable/ContextMenuUnstableTests.java
```diff
@@ -0,0 +1,97 @@
...
+            throws NoSuchFieldException, IllegalAccessException {
+
+        ListPanel issuePanel = find(PANEL_IDENTIFIER);
+        Field listViewField = ListPanel.class.getDeclaredField("listView");
+        listViewField.setAccessible(true);
+        IssueListView listViewValue = (IssueListView) listViewField.get(issuePanel);
+
+        // scrolls to the end of the panel
+        listViewValue.scrollAndShow(12);
```
damithc: If you want this to be the last issue, may be link it to size of list rather than hard code it?

src/test/java/unstable/ContextMenuUnstableTests.java
```diff
@@ -0,0 +1,97 @@
...
+
+
+        //checking for issue #7 and below twice to make sure issues are marked correctly for read/unread cases
```
damithc: Break the comment into 4 and insert at the right places.

src/test/java/unstable/ContextMenuUnstableTests.java
```diff
@@ -0,0 +1,97 @@
...
+        rightClick("#" + issuePanel.getId() + "_" + index);
+        ContextMenu contextMenu = issuePanel.getContextMenu();
+        for (MenuItem menuItem : contextMenu.getItems()){
+            awaitCondition(menuItem::isVisible);
+        }
+        if (isMarkAsRead){
+            click("Mark all below as read");
+        } else {
+            click("Mark all below as unread");
+        }
```
damithc: 1. Can use the Ternary operator?
2. Use constants from the relevant class instead of using magic strings (need to make them public first)?

src/main/java/ui/listpanel/ListPanel.java
codemanmeet: Sure, will create a first timer issue for this.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -111,25 +114,41 @@ public void restoreSelection() {
...
 
+    /**
+     * Triggers the selected item on the card at the specified index for actions that are not right clicks
+     */
+    private void triggerItemSelected(int index){
+        onItemSelected.accept(index, !IS_RIGHT_CLICK);
+    }
+
+    /**
+     * Triggers the selected item on the card at the specified {index} for actions that are right clicks
```
dariusf: We've removed the `{index}` above, so this should probably be changed as well.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -111,25 +114,41 @@ public void restoreSelection() {
...
+     */
+    private void triggerItemSelectedSecondary(int index){
```
dariusf: We aren't using the term 'secondary' any more, right? Not sure if we should change this to RightClick.

src/test/java/guitests/IssueSelectedEventTest.java
```diff
@@ -0,0 +1,68 @@
...
+import ui.UI;
+import ui.listpanel.ListPanel;
+import util.events.IssueSelectedEventHandler;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class IssueSelectedEventTest extends UITest{
+
+    static int eventTestCount;
```
dariusf: Instead of doing this (which isn't thread-safe without synchronisation; `++` is not atomic, and the field isn't private), you can use an AtomicInteger and make it local to each test method. There are examples of this in the PlatformEx tests.

src/test/java/unstable/ContextMenuUnstableTests.java
```diff
@@ -0,0 +1,103 @@
...
+import org.loadui.testfx.GuiTest;
+import ui.components.FilterTextField;
+import ui.components.IssueListView;
+import ui.listpanel.ListPanel;
+import ui.listpanel.ListPanelCell;
+import java.lang.reflect.*;
+
+import static org.junit.Assert.assertEquals;
+
+public class ContextMenuUnstableTests extends UITest {
```
dariusf: I think this can have a more specific name. It's testing read/unread for the issue list view, so you might want those to feature in there somewhere.

src/test/java/unstable/ContextMenuUnstableTests.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private static final String PANEL_IDENTIFIER = "#dummy/dummy_col0";
+
+    @Before
+    public void setup() {
+        Platform.runLater(stage::show);
+        Platform.runLater(stage::requestFocus);
```
dariusf: You probably want `runAndWait` here, and just one call, with both method calls inside.

src/test/java/unstable/ContextMenuUnstableTests.java
```diff
@@ -0,0 +1,103 @@
...
+
+        FilterTextField filterTextField = find(PANEL_IDENTIFIER + "_filterTextField");
+        filterTextField.setText("");
+        Platform.runLater(filterTextField::requestFocus);
```
dariusf: Is this needed? There's a click right after. If this was what you tried to get it to pass on Travis, there should probably be a comment.

src/main/java/ui/components/NavigableListView.java
damithc: @codemanmeet I think the syntax you are looking for is `{@code index}` or something like that (can't remember the exact syntax, but should be easy to find if you look up javadoc documentation).

codemanmeet: Oh yup, that's what I was looking for! Thanks! But researched this further and realized it's only needed for code blocks to be used in javadoc comments above methods, which is not applicable in my case.

src/main/java/ui/components/NavigableListView.java
codemanmeet: Sure, will change this to rightClick too!

codemanmeet:
I made an error in rebasing and lost a couple of my changes (through `git push -f` :( . I managed to recover my changes but had to create a new branch. I have created a fresh pull request for the issue!

PR #1297, opened 2016-02-29
Create a Milestone Picker that allows the user to add/remove/modify the milestone of an issue #1286

### LINKED ISSUE #1286: Create a Milestone Picker that allows the user to add/remove/modify the milestone of an issue

m133225:
Fixes #1286.

This milestone picker works similarly to the label picker, but simplified since only 1 milestone can be assigned at a time.


m133225:
Just some updates on the UI:

There has been a slight rework of the milestone picker layout away from the label picker's one.
Instead of showing all possible milestones, the list of milestones shown will only be up to 5, according to the query, if there is any. Due date and progress of the milestones are also included.

This is a preview. The left box is the currently assigned milestone, and the right box is the new milestone to be assigned.


There are still problems with this to be fixed/features to be introduced
- One of which is that the user is unable to click a milestone that does not appear. The planned solution for this is to allow a toggle of the full list of milestones (using some shortcut).
- Node colour, probably. It now uses the GitHub's open/closed status colour, but does not appear to be suited to this background.
- Auto filling in the current milestone when the milestone picker is first opened, so that the user will have an easier time assigning the current milestone to a list of issues.

Suggestions and comments (for the UI) are requested! :smile:

damithc:
I think the UI is fine for the first version. We'll tweak it further as separate PRs. Good job getting here. Let's focus on getting it merged.
Some suggested optimizations (can be done as separate issues) to make this fast for power users:
`L` `Enter` : assigns `cur` milestone
`L` `right arrow` `Enter` : assigns `cur+1` open milestone
`L` `left arrow` `Enter` : assigns `cur-1` milestone
`L` `backspace/del` `Enter` : removes the existing milestone without assigning a new one

m133225:
Ready for review. Will be attempting to add more tests meanwhile.

damithc:
Looks like you are not using HT to apply labels. This PR has two labels from an exclusive group :-p

m133225:
Sorry, replacing this with a new PR #1352, since there is an unidentified problem with building this PR on Travis.

PR #1304, opened 2016-03-02
New Updater - Create a new process that will update JAR and restart the application #1230

### LINKED ISSUE #1230: New Updater - Create a new process that will update JAR and restart the application

HansNewbie:
Fixes #1230


HansNewbie:
@dariusf Travis is not triggered. Since the branch is temporary, should I not set it to run and just test locally or should I change the travis configuration to run it? (I am not sure if setting .travis.yml in this PR will change the setting, though)

dariusf:
You can add `updater` to `.travis.yml`, at least for now while you're not merging into master. That will allow us to have the builds trigger and show up in PRs.

HansNewbie:
@dariusf can take a look?

src/main/java/updater/JarReplaceAndExecute.java
```diff
@@ -0,0 +1,193 @@
...
+import java.nio.file.StandardCopyOption;
+import java.util.HashMap;
+
+/**
+ * Note: This class will be compiled into a JAR on its own
+ *
+ * Replaces a JAR file and start that JAR
+ */
+public class JarReplaceAndExecute extends Application{
+    private static final String BACKUP_FILENAME_APPENDIX = "_backup";
```
dariusf: SUFFIX might be a better word.

src/main/java/updater/JarReplaceAndExecute.java
```diff
@@ -0,0 +1,193 @@
...
+
+    @SuppressWarnings("PMD")
```
dariusf: What's the warning being suppressed? Should be in a comment since it's not always obvious to readers.

src/main/java/updater/JarReplaceAndExecute.java
```diff
@@ -0,0 +1,193 @@
...
+        log("Moving source to target");
+        if (!moveFile(sourceJarFile.toPath(), targetJarFile.toPath())) {
+            log("Failed to move update.");
+            restoreBackup(targetJarFile);
+            quit();
+        }
+
+        // start target
+        // - if target can't be started, rollback backup
+        if (!executeJar(targetJarPath, "")) {
```
dariusf: It might be clearer to negate this, following the comment above.

src/main/java/updater/JarReplaceAndExecute.java
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * Runs a command to execute a JAR file with the given arguments and options
+     * @param jarPath path (in string) to JAR file to be executed
+     * @param argsAndOptions raw arguments and options to run the JAR
+     * @return true if JAR can be executed, false otherwise
+     */
+    private boolean executeJar(String jarPath, String argsAndOptions) {
+        String command = "java -jar " + jarPath + " " + argsAndOptions;
+        command = command.trim();
```
dariusf: Can this method be static?

This line would be a bit nicer to read with String.format(...).trim().

src/main/java/updater/JarReplaceAndExecute.java
```diff
@@ -0,0 +1,193 @@
...
+        log("Starting JAR with command: " + command);
+
+        Process process = null;
+
+        try {
+            process = Runtime.getRuntime().exec(command);
+        } catch (IOException e) {
+            e.printStackTrace();
```
dariusf: We should log something here.

src/main/java/updater/UpdateManager.java
```diff
@@ -90,6 +103,17 @@ private boolean initUpdate() {
             return false;
         }
 
+        File replaceAndExecuteJar = new File(UPDATE_DIRECTORY + File.separator + UPDATE_RESTARTER_APP_NAME);
+        if (!replaceAndExecuteJar.exists()) {
```
dariusf: Could we invert this? So if it exists, we return true, then outside the block we can do all the rest.

src/main/java/updater/UpdateManager.java
```diff
@@ -149,4 +173,29 @@ public void showUpdateProgressWindow() {
...
         updateProgressWindow.hideWindow();
     }
+
+    private boolean runReplaceAndExecuteJar() {
+        String restarterAppPath = UPDATE_DIRECTORY + File.separator + UPDATE_RESTARTER_APP_NAME;
+        String cmdArg = "--source=" + UPDATE_DIRECTORY + File.separator + UPDATE_APP_NAME +
+                " --target=" + UPDATE_APP_NAME;
+
+        String command = "java -jar " + restarterAppPath + " " + cmdArg;
+        logger.info("Executing JAR of restarter with command: " + command);
```
dariusf: We can also use String.format here.

HansNewbie:
@dariusf I made the changes. I will clean up the commits once it's acceptable.

src/main/java/updater/JarReplaceAndExecute.java
```diff
@@ -0,0 +1,193 @@
...
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.util.HashMap;
+
+/**
+ * Note: This class will be compiled into a JAR on its own
+ *
+ * Replaces a JAR file and start that JAR
+ */
+public class JarReplaceAndExecute extends Application{
```
damithc: Try not to name class to sound like a method

src/main/java/updater/JarReplaceAndExecute.java
```diff
@@ -0,0 +1,193 @@
...
+    private static final String BACKUP_FILENAME_SUFFIX = "_backup";
+
+    @SuppressWarnings("PMD")
+    // For "Consider using varargs for methods or constructors which take an array the last parameter."
+    public static void main(String[] args) {
```
dariusf: I see. Maybe use a `// PMD` comment instead.

HansNewbie:
@dariusf managed to make it work when compiled as JAR (shadowJar has some artifact, I guess).

src/main/java/updater/JarReplaceAndExecute.java
HansNewbie: I renamed it to `JarReplacerAndExecutor`.

dariusf:
So shadowJar works?

HansNewbie:
So what happened was the Jar that I put in resources seems to be ignored by shadowJar, so I can't extract the Jar that will restart the application. What I did was I remove the `.jar` from the filename in the resource. I add the `.jar` back upon extracting.

src/main/java/updater/JarReplacerAndExecutor.java
```diff
@@ -0,0 +1,231 @@
...
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.util.HashMap;
+
+/**
+ * Note: This class will be compiled into a JAR on its own
+ *
+ * Replaces a JAR file and start that JAR
+ */
+public class JarReplacerAndExecutor extends Application{
```
damithc: `And` in the class name is a sign of violating Single Responsibility Principle. :-)

HansNewbie: How about `JarReplacer`, then execution of the JAR is an option in the command line?

HansNewbie:
@dariusf ready for review

src/main/java/updater/JarUpdater.java
```diff
@@ -0,0 +1,255 @@
...
+public class JarUpdater extends Application {
+    private final ExecutorService pool = Executors.newSingleThreadExecutor();
+
+    private static final int MAX_RETRY = 10;
+    private static final int WAIT_TIME = 2000;
+    private static final String BACKUP_FILENAME_SUFFIX = "_backup";
+
+    @SuppressWarnings("PMD")
+    // PMD - "Consider using varargs for methods or constructors which take an array the last parameter."
+    public static void main(String[] args) {
```
dariusf: What was I meant was using `// PMD` after the opening brace. That way we don't need to suppress warnings.

src/main/java/updater/JarUpdater.java
```diff
@@ -0,0 +1,255 @@
...
+    /**
+     * Creates directories for target file if they do not exist.
+     * Otherwise, makes backup if target file exists.
+     */
+    private void prepareTargetJarFile(File targetJarFile) {
+        if (targetJarFile.getParentFile() != null &&
+                !createDirectories(targetJarFile.getParentFile().getAbsoluteFile())) {
+            log("Failed to make directories of target file.");
+            quit();
+        } else if (targetJarFile.exists() && !makeJarBackup(targetJarFile)) {
```
dariusf: Do we need to notify the user in this case?

src/main/java/updater/JarUpdater.java
```diff
@@ -0,0 +1,255 @@
...
+            System.setOut(out);
+        } catch (FileNotFoundException e) {
+            log("File not found, will not create logger");
+        } catch (UnsupportedEncodingException e) {
+            log("Encoding not supported, will not create logger");
+        }
+    }
+
+    private static void log(String message) {
+        System.out.println(message);
```
dariusf: Maybe we should use a logger here, and print to a different file? In case there are issues with updating and we need to diagnose.

damithc:
Shouldn't this be in `ongoing` state rather than `toMerge`?

dariusf:
True, I guess the non-cosmetic changes would require another round.

src/main/java/updater/JarUpdater.java
HansNewbie: Because I am compiling this on its own, all the libraries and dependencies are not inside, so logger is not included, which is why I created the `log` method. I tried to use logger but I didn't manage to make the gradle task to include it. Should I try again?

HansNewbie: I tried again, but the `logger` class could not be loaded, although I have tinkered around with the gradle build script.

That was the original reason why I create `log` method. I set the `System.out` to another file.

src/main/java/updater/JarUpdater.java
HansNewbie: I tried
```
  // PMD
  ...
```

and
```
public static void main(String[] args) { // PMD
  ...
```

but both didn't suppress the PMD warning. The only one that can suppress with comment would be `// NOPMD`. Which one would you like me to use?

HansNewbie:
@dariusf 
- I added a critical error possibility for the one you commented on warning user.
- I tried tinkering with gradle but I just can't seem to include apache logger into the JAR. Currently, the log will be written to a separate file by setting `System.out` to a file output.
- I tried the `// PMD` but didn't work. Unless if you meant `// NOPMD`.

src/main/java/updater/JarUpdater.java
dariusf: Oops, yeah, NOPMD.

src/main/java/updater/JarUpdater.java
```diff
@@ -0,0 +1,275 @@
...
+            quit();
+        }
+
+        // Optionally start target jar file
+        if (executeJarOption != null && executeJarOption.equalsIgnoreCase("y")) {
+            if (executeJar(targetJarPath, "")) {
+                removeBackup(targetJarFile);
+            } else { // if target can't be started, rollback backup
+                restoreBackup(targetJarFile);
+                executeJar(targetJarPath, "");
```
dariusf: Is it okay to ignore the return value of this? Maybe instead of having error-handling everywhere, we can use an exception and handle it higher up? I know this won't work for everything, but if all these stages should return/quit/abort on failure, an exception to stop execution there might work better.

src/main/java/updater/JarUpdater.java
```diff
@@ -0,0 +1,275 @@
...
+            }
+        } else {
+            removeBackup(targetJarFile);
+        }
+
+        quit();
+    }
+
+    private boolean moveFile(Path source, Path dest) {
```
dariusf: This can be static.

I'm not sure how this method compares to FileUtils.moveFile. Maybe it can be moved into a utility class of some kind?

src/main/java/updater/JarUpdater.java
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    /**
+     * Runs a command to execute a JAR file with the given arguments and options
+     * @param jarPath path (in string) to JAR file to be executed
+     * @param argsAndOptions raw arguments and options to run the JAR
+     * @return true if JAR can be executed, false otherwise
+     */
+    private static boolean executeJar(String jarPath, String argsAndOptions) {
+        String command = String.format("java -jar %1$s %2$s", jarPath, argsAndOptions).trim();
```
dariusf: We don't need the positional arguments here. Same for the other format strings.

src/main/java/updater/JarUpdater.java
dariusf: Okay, that works.

src/main/java/updater/JarUpdater.java
HansNewbie: `FileUtils.moveFile` won't replace existing file while this method will. I will put into `IOUtilities` class in util.

HansNewbie: @dariusf Actually, if I move it out to another class, I will still need another method to call it since I need to handle the exception. Will it be fine to keep it here and make it static instead?

src/main/java/updater/JarUpdater.java
HansNewbie: Are you saying that we should do something like:
```
pool.execute(() -> {
  try {
    run();
  } catch (Exception e) {
    quit();
  }
});
```
at the very top?

For this particular line, it's okay to ignore because it is executing the backup, which is expected to run properly. If it does not, we can't do anything else since the backup is also not working.

dariusf: Yes, except catching IOException instead of everything, and logging it before quitting. Each place that throws the exception provides the message.

src/main/java/updater/JarUpdater.java
dariusf: Yeah, okay. Alternatively, it could throw the IOException. Whichever you feel works better.

HansNewbie:
@dariusf I made the changes.

src/main/java/updater/JarUpdater.java
```diff
@@ -0,0 +1,278 @@
...
+        if (!prepareTargetJarFile(targetJarFile)) {
+            showCriticalErrorOnUpdating();
+            return;
+        }
+
+        log("Moving source to target");
+        if (!moveFile(sourceJarFile.toPath(), targetJarFile.toPath())) {
+            restoreBackup(targetJarFile);
+            showCriticalErrorOnUpdating();
+            return;
```
dariusf: I don't think we need the `return`s here, seeing as execution ends with these methods.

src/main/java/updater/JarUpdater.java
```diff
@@ -0,0 +1,278 @@
...
+        }
+
+        // Optionally start target jar file
+        if (executeJarOption != null && executeJarOption.equalsIgnoreCase("y")) {
+            if (executeJar(targetJarPath, "")) {
```
dariusf: Maybe we can add an overload for this?

src/main/java/updater/JarUpdater.java
```diff
@@ -0,0 +1,278 @@
...
+     * Creates directories for target file if they do not exist.
+     * Otherwise, makes backup if target file exists.
+     * @return true if preparation successful, false otherwise
+     */
+    private boolean prepareTargetJarFile(File targetJarFile) {
+        if (targetJarFile.getParentFile() != null &&
+                !createDirectories(targetJarFile.getParentFile().getAbsoluteFile())) {
+            log("Failed to make directories of target file.");
+            return false;
+        } else if (targetJarFile.exists() && !makeJarBackup(targetJarFile)) {
```
dariusf: If might be better to separate these, as the logic goes linearly anyway.

dariusf:
Mostly just cosmetic changes.

src/main/java/updater/JarUpdater.java
HansNewbie: We need the return because the `showCriticalErrorOnUpdating` is in the UI thread. If I don't `return` here, the code continues to run.

src/main/java/updater/JarUpdater.java
HansNewbie: Okay.

src/main/java/updater/JarUpdater.java
dariusf: Okay, never mind.

src/main/java/updater/JarUpdater.java
```diff
@@ -0,0 +1,284 @@
...
+            } else { // if target can't be started, rollback backup
+                restoreBackup(targetJarFile);
+                executeJar(targetJarPath);
+            }
+        } else {
+            removeBackup(targetJarFile);
+        }
+    }
+
+    private static boolean moveFile(Path source, Path dest) {
```
damithc: Header comment?

damithc:
1. Add header comments to methods, unless the method signature says it all.
2. Make sure to push file handling work to a utility class e.g. `FileHelper`?

HansNewbie:
@damithc on file handling, I didn't take it out from this class because of compilation issue. Since `JarUpdater` class is compiled separately, I can't include the Apache logger library as its dependency (I tried tinkering with gradle to no avail). As such, this class uses its own logger instead of `logger.log`.

I can include `FileHelper` utility class but since I can't use Apache logger, either the methods I use don't log anything or the methods throw the exception back to the caller, which in that case I will still need another method in this class for moving file with exception handling. Would that be preferable?

damithc:
If you can't make the logger work in this PR, at least create a new issue (later) to fix that problem.

HansNewbie:
@damithc noted. In that case, should that issue also includes moving file handling to standard `FileHelper` class? Currently some file handling methods already exists in `Utility` class.

damithc:
Yes, that too.

HansNewbie:
I added the header comments.

src/main/java/updater/JarUpdater.java
```diff
@@ -0,0 +1,302 @@
...
+        } else {
+            log("source: " + sourceJarPath);
+            log("target: " + targetJarPath);
+        }
+
+        File sourceJarFile = new File(sourceJarPath);
+        File targetJarFile = new File(targetJarPath);
+
+        if (!sourceJarFile.exists()) {
+            throw new IOException("Such source file does not exist, aborting.");
```
damithc: `Source file [file name] does not exist. Aborting...`?

src/main/java/updater/JarUpdater.java
```diff
@@ -0,0 +1,302 @@
...
+            return false;
+        }
+
+        if (targetJarFile.exists() && !makeJarBackup(targetJarFile)) {
+            log("Failed to make backup.");
+            return false;
+        }
+
+        return true;
+    }
```
damithc: 1. I think this method is doing two things. Should be 2 methods?
2. May be use exceptions instead of `false` to indicate problems? Most of these `false` cases are 'exceptional' situations.

src/main/java/updater/JarUpdater.java
```diff
@@ -0,0 +1,302 @@
...
+    /**
+     * Makes back up of a JAR file.
+     *
+     * In some platforms (Windows in particular), JAR file cannot be modified if it was executeded and
+     * the process has not ended yet. As such, we will make several tries with wait in making backup.
+     *
+     * @param jarFile Jar file to be backed up
+     * @return true if backup is created, false otherwise
+     */
+    private boolean makeJarBackup(File jarFile) {
```
damithc: This method is not specific to jar files, right?. Write it in a more general way?

src/main/java/updater/JarUpdater.java
```diff
@@ -0,0 +1,302 @@
...
+        log("Jar file cannot be backed up. Most likely is in use by another process.");
+        return false;
+    }
+
+    /**
+     * Moves file to its backup file (the original file will be removed)
+     * @param file file to be moved to its backup
+     * @return true if move is successful, false otherwise
+     */
+    private boolean moveFileToBackup(File file) {
```
damithc: This method is a general 'move file' operation. Write it in a more general way.

damithc: Never mind. Just realized it is a wrapper for a more general method.

src/main/java/updater/JarUpdater.java
```diff
@@ -0,0 +1,302 @@
...
+     * Removes the backup of a file
+     * @param file the file which backup should be removed
+     */
+    private void removeBackup(File file) {
+        log("Removing backup");
+        String backupFilename = getBackupFilename(file.getName());
+        File backup = new File(file.getParent(), backupFilename);
+        if (backup.exists() && !backup.delete()) {
+            log("Failed to remove backup");
+        }
```
damithc: I know this code works, but it is written in a misleading way. Make the logic more straight-forward, even if it makes the code more verbose. Can also extract a more general `deleteFile` method.

damithc: In fact consider extracting a `AutoUpdateFileHelper` which can be merged with `FileHelper` later on.

src/main/java/updater/UpdateManager.java
```diff
@@ -90,6 +96,30 @@ private boolean initUpdate() {
...
+            }
+        }
+
+        try (InputStream in = UpdateManager.class.getClassLoader().getResourceAsStream("updater/replaceAndExecute");
+             OutputStream out = new FileOutputStream(replaceAndExecuteJar)) {
+            IOUtils.copy(in, out);
+        } catch (IOException e) {
+            logger.error("Can't copy replacer-and-executor JAR", e);
+            return false;
+        }
```
damithc: Same comment as before. 1. Method doing multiple things. 2. Consider using exceptions instead of boolean returns.

src/main/java/updater/UpdateManager.java
```diff
@@ -149,4 +182,45 @@ public void showUpdateProgressWindow() {
     public void hideUpdateProgressWindow() {
         updateProgressWindow.hideWindow();
     }
+
+    private boolean runJarUpdaterWithExecute() {
+        return runJarUpdater(true);
+    }
+
+    private boolean runJarUpdater(boolean executeJar) {
```
damithc: Parameter not named as a boolean

damithc:
Some changes requested.

src/main/java/updater/JarUpdater.java
HansNewbie: This method is specific to jar files. There is a more general way `makeBackup(file)`.

HansNewbie:
I made the requested changes.
- I removed most boolean return and uses exception instead; some boolean return is used where it fits (e.g. for program sequence).
- Created JarUpdaterFileHandler.

src/main/java/updater/UpdateManager.java
HansNewbie: I didn't make any changes to this because this method does one thing, which is copying a file from the resource to a local file. I take that part of copying file to a target file includes creating such target file if it does not exist yet.

I kept the boolean returns because this is an initialisation sequence. I can make it throw exception and handle the exception in the caller, but it seems better to know if extracting the resource to a local file was successful or not rather than throw exception.

HansNewbie:
@damithc is this alright?

damithc:
You may want to describe how the auto-updater works in our design docs.

HansNewbie:
@damithc Yes I will.

PR #1305, opened 2016-03-02
Warn user for invalid username #1219

### LINKED ISSUE #1219: Warn users if an invalid username is used
`assignee` and `author` can check that the user referenced is a valid one.

We may need some support for displaying warnings, which don't propagate via an exception and show the error graphics, can be viewed in a list, etc.

raisfathin:
Fixes #1219 

~~TODO: add tests~~

Sample screenshot of warning:
screen shot 2016-03-03 at 01 55 29


raisfathin:
@dariusf It turned out that if we reuse the existing `SemanticException`, some tests will fail since they are expecting `SemanticException` to be thrown while in this case, we catch all `SemanticException` to be shown as a warning.

dariusf:
You mean in tests? How come semantic exceptions are shown as warnings?

raisfathin:
Yes, there are tests which use this [method](https://github.com/HubTurbo/HubTurbo/blob/master/src/test/java/tests/FilterEvalTests.java#L839-L845), therefore, we can't reuse the `SemanticException` to do the warning thing.

If I remember correctly, you said that we can reuse the existing `SemanticException` as a replacement for the old `WarningException` (and treating the warnings as error). When I tried to implement it, those tests which expect `SemanticException` to be thrown would fail. Not completely sure, though. Seems to be a misunderstanding on my part.

dariusf:
They're failing because the exception message is different, right? We can rename that method so it's only used for tests which verify exceptions related to qualifier types.

raisfathin:
@dariusf Ah, true. So, should I create a field in the exception to store whether it's related to qualifier types or just match the message with some regex (say, "\\w+ expects \\w+")?

I think the second one is worse since if the message changes in the future, the test should be fixed. As for the first one, I don't know whether it's a good practice to put fields like that inside an Exception.

dariusf:
The issue with the first method is that we would then need a field for every use of the exception. I think it's fine to check the message here. We can use a constant to ensure that it doesn't get out of date.

raisfathin:
New warning:
screen shot 2016-03-07 at 13 14 13

Currently, it can only notify the user for up to one invalid user. It also halts filter processing whenever it encounters an invalid user.

raisfathin:
@dariusf Should I remove codes that handle multiple warnings and just support one warning?

dariusf:
Yes, I think that's fine. We need to change a lot more code to support multiple warnings in general (not just at branching points of the syntax tree).

raisfathin:
Done! Ready for review.

src/main/java/backend/Logic.java
```diff
@@ -330,9 +330,10 @@ private boolean insertMetadata(Map<Integer, IssueMetadata> metadata, String repo
     }
 
     /**
-     * Carries the current set of GUI elements, as well as the current list of users in the model, to the GUI.
+     * Carries the current set of GUI elements including warning, as well as
+     * the current list of users in the model, to the GUI.
      */
-    public void updateUI(Map<FilterExpression, List<GuiElement>> elementsToShow) {
+    public void updateUI(Map<FilterExpression, ImmutablePair<List<GuiElement>, Optional<String>>> elementsToShow) {
```
dariusf: Tuples are kind of ugly in Java, unfortunately. We should probably make a new class for this.

src/main/java/backend/UpdateController.java
```diff
@@ -87,7 +89,13 @@ private void openRepositoriesInFilters(List<FilterExpression> filterExprs) {
...
                 .filter(Qualifier::hasUpdatedQualifier)
                 .flatMap(filterExpr -> allModelIssues.stream()
-                        .filter(issue -> Qualifier.process(models, filterExpr, issue)))
+                        .filter(issue -> {
+                            try {
+                                return Qualifier.process(models, filterExpr, issue);
+                            } catch (SemanticException e) {
+                                return false;
+                            }
+                        }))
```
dariusf: Hmm, this is going to conflict with #1323... It might be best to wait until Martin is done. We can use a similar approach to carry the warnings to the UI.

I feel like we shouldn't try to distinguish warnings and errors at this point, since only one warning is returned (same as an error), and we will want a list of errors in future anyway. Can we just treat a missing assignee as an error, and change errors so that they all show in a tooltip (to use the new UI features implemented here, and because we want to move them out of the status bar in future anyway)? @damithc

We can work towards being able to collect multiple warnings/errors in future.

damithc: Yes, I'm ok with not distinguishing between errors and warnings.
If @raisfathin can finish the repo picker soon, we can use the area currently occupied by the repo dropdown to show feedback/errors/warnings (as we discussed last week). The status bar below can be used to show status of automatic actions such as refreshes.

raisfathin: @dariusf Okay, that approach seems to be better. Btw, this warning also shows all errors related to `SemanticException`. Is this intended/OK?

screen shot 2016-03-08 at 19 08 16

dariusf: Yup, that's fine. We'll just do something simple here, pending something more complex after the repo picker/multiple errors and warnings.

raisfathin:
Waiting for #1323 to be merged.

raisfathin:
I incorporated @m133225's changes, so now the errors are displayed at the status bar instead. I noticed that the status bar is too short (by default) to display the errors. Examples:
screen shot 2016-03-17 at 13 13 15
screen shot 2016-03-17 at 13 13 22

raisfathin:
Ready for review.

src/main/java/backend/resource/MultiModel.java
```diff
@@ -209,6 +209,16 @@ public synchronized void setDefaultRepo(String repoId) {
     }
 
     @Override
+    public List<TurboUser> getUsersOfRepo(String repoId) {
+        Optional<Model> model = getModelById(repoId);
+        if (model.isPresent()) {
+            return model.get().getUsers();
+        } else {
+            return new ArrayList<>();
+        }
```
damithc: See if this can be done in one line. `... ? ... : ...`

src/test/java/tests/FilterEvalTests.java
```diff
@@ -824,11 +827,61 @@ public void filterLabelMatching() {
...
+
     /**
      * Tests the filter string in the context of an empty model
      */
     public boolean matches(String filterExpr, TurboIssue issue) {
-        return Qualifier.process(empty, Parser.parse(filterExpr), issue);
+        return getQualifierProcessResult(empty, Parser.parse(filterExpr), issue);
+    }
+
+    private boolean getQualifierProcessResult(IModel model, FilterExpression expr, TurboIssue issue) {
```
damithc: The return value is boolean but the method name is not.

raisfathin:
Fixed!

src/test/java/tests/FilterEvalTests.java
```diff
@@ -824,6 +825,40 @@ public void filterLabelMatching() {
         assertFalse(Qualifier.labelMatches(".", "p.high"));
     }
 
+    @Test
+    public void QualifierProcess_useInvalidUsername_getUserWarning() {
```
damithc: Why start with upper case letter?

src/test/java/tests/FilterEvalTests.java
```diff
@@ -824,6 +825,40 @@ public void filterLabelMatching() {
...
+        TurboUser user = new TurboUser(REPO, "fox", "charlie");
+        IModel model = TestUtils.singletonModel(new Model(REPO,
+                new ArrayList<>(),
+                new ArrayList<>(),
+                new ArrayList<>(),
+                new ArrayList<>(Arrays.asList(user))));
```
damithc: 8 space indentation is a last resort. In this case, the following may be better?
```java
IModel model = TestUtils.singletonModel(new Model(REPO,
                                                  new ArrayList<>(),
                                                  new ArrayList<>(),
                                                  new ArrayList<>(),
                                                  new ArrayList<>(Arrays.asList(user))));
```
Alternatively,
```java
IModel model = TestUtils.singletonModel(new Model(REPO, new ArrayList<>(), new ArrayList<>(),
                                                  new ArrayList<>(), new ArrayList<>(Arrays.asList(user))));
```

damithc:
@ndt93 a review for this?

ndt93:

src/test/java/tests/FilterEvalTests.java
raisfathin: It used to call `Qualifier.process` directly. Will change that.

Btw, In a case where a test method calls `Class.method()`, what is the correct test method name? Is it `classMethod_...`?

damithc: Typically the test will be in a `Test[ClassName]` class (i.e. each class should have a corresponding `Test*` class), so the test method need not mention the class name.

src/main/java/backend/Logic.java
```diff
@@ -398,7 +398,8 @@ private boolean insertMetadata(Map<Integer, IssueMetadata> metadata, String repo
     }
 
     /**
-     * Carries the current set of GUI elements, as well as the current list of users in the model, to the GUI.
+     * Carries the current set of GUI elements including warning, as well as
```
dariusf: This no longer needs to be changed I think.

src/main/java/backend/UpdateController.java
```diff
@@ -4,6 +4,7 @@
 import backend.resource.MultiModel;
 import backend.resource.TurboIssue;
 import filter.FilterException;
+import filter.SemanticException;
```
dariusf: This too, it's no longer needed.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -858,6 +855,17 @@ private boolean assigneeSatisfies(IModel model, TurboIssue issue) {
 
         if (!assignee.isPresent()) return false;
 
+        List<TurboUser> usersOfRepo = model.getUsersOfRepo(issue.getRepoId());
```
dariusf: This bit can probably be factored out into MultiModel, and reused below.

src/test/java/tests/FilterEvalTests.java
```diff
@@ -31,6 +31,8 @@
     private final IModel empty;
     public static final String REPO = "test/test";
 
+    private static final String USER_WARNING_ERROR_FORMAT = "Cannot find username containing %s in %s";
```
dariusf: Could this be made a constant in Qualifier, then reused here?

src/main/java/filter/expression/Qualifier.java
raisfathin: Are you referring to `model.getUsersOfRepo`?

dariusf: I mean the part which determines `shouldWarnUser`. Both look quite similar.

raisfathin: So, something like `isUserInRepo`?

src/test/java/tests/FilterEvalTests.java
raisfathin: Sure, will add that to Qualifier.

src/test/java/tests/FilterEvalTests.java
raisfathin: Yes. Will change that.

src/main/java/filter/expression/Qualifier.java
dariusf: Yup.

raisfathin:
Ready for review!

dariusf:
Looks good. It'd be nice to squash the unnecessary commits, like those from review, refactoring, fixing style violations, etc.

raisfathin:
Done.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -853,16 +855,21 @@ private boolean assigneeSatisfies(IModel model, TurboIssue issue) {
...
     }
 
-    private boolean authorSatisfies(TurboIssue issue) {
+    private boolean authorSatisfies(IModel model, TurboIssue issue) {
         if (!content.isPresent()) return false;
 
+        boolean shouldWarnUser = !model.isUserInRepo(issue.getRepoId(), content.get());
+        if (shouldWarnUser) {
+            throw new SemanticException(String.format(USER_WARNING_ERROR_FORMAT, content.get(), issue.getRepoId()));
+        }
```
damithc: This code block seems similar to another block above. If they are, may be extract into a method?

damithc:
Minor change requested

raisfathin:
Done.

damithc:
Looks like now I cannot filter by assignees/authors who are not members of the project. Was that caused by this PR?

damithc:
I should be able to filter by any user name, even if he/she is not a project member. But I don't mind getting a warning in those cases.

raisfathin:
@damithc Yes. Hmm, there was a discussion about this being an error somewhere at the top. Should I undo the merge?

raisfathin:
@damithc Ok, undoing merge is not a good idea since it will break the history :|

Currently, I can think of 2 alternatives:
 - Remove `enforceUserInRepo` condition, which essentially undo the change, or
 - Fix the issue by adding every user in all issues of a repo to the list of users in that repo.

Any other better alternative?

damithc:
Sorry, I guess I missed that discussion. 
The fix can be done as a separate issue, as long as it is done quickly.
The ideal behavior is if the user is not involved in the repo, display a warning. The search should still be performed and an empty result should be shown. The warning will help the user to catch inadvertent typos.

raisfathin:
Alright. Creating a separate issue.

PR #1310, opened 2016-03-03
More obvious sign of filter syntax problem #1301

### LINKED ISSUE #1301: More obvious sign of filter syntax problem
Should we make it more obvious when the syntax for the filter is wrong? Something like the bottom half of this:

image

(example from Wireshark)

garbanzos:
Fixes #1301. 

The background colour of the filter text field will change to red when the filter is invalid. Otherwise, the background colour is white. 

The check whether to change the background color happens when 
 - the user presses enter 
 - the filter in the text field changes  

Screenshot
image

### DIFF
--- a/src/main/java/ui/components/FilterTextField.java
+++ b/src/main/java/ui/components/FilterTextField.java
@@ -26,6 +26,11 @@ public class FilterTextField extends TextField {
     // Specialised to HubTurbo's filter syntax.
     private static final String WORD_BOUNDARY_REGEX = "[ (:)]";
 
+    // Background colours of FilterTextField.
+    // The background colour set depends on whether the text in the textfield is a valid filter.
+    private static final String VALID_FILTER_STYLE = "-fx-control-inner-background: white";
+    private static final String INVALID_FILTER_STYLE = "-fx-control-inner-background: #EE8993";
+
     // Callback functions
     private Runnable cancel = () -> {};
     private Function<String, String> confirm = (s) -> s;
@@ -51,8 +56,10 @@ public class FilterTextField extends TextField {
             boolean wasError = false;
             try {
                 Parser.parse(getText());
+                setStyleForValidFilter();
             } catch (ParseException e) {
                 wasError = true;
+                setStyleForInvalidFilter();
             }
             return ValidationResult.fromErrorIf(this, "Parse error", wasError);
         });
@@ -91,6 +98,20 @@ public class FilterTextField extends TextField {
                 }
             }
         });
+    }
+
+    /**
+     * Sets the style of FilterTextField to the style for valid filter
+     */
+    public final void setStyleForValidFilter() {
+        this.setStyle(VALID_FILTER_STYLE);
+    }
+
+    /**
+     * Sets the style of FilterTextField to the style for invalid filter
+     */
+    public final void setStyleForInvalidFilter() {
+        this.setStyle(INVALID_FILTER_STYLE);
     }
 
     /**
--- a/src/main/java/ui/issuepanel/FilterPanel.java
+++ b/src/main/java/ui/issuepanel/FilterPanel.java
@@ -177,11 +177,13 @@ public abstract class FilterPanel extends AbstractPanel {
             FilterExpression filter = Parser.parse(filterString);
             if (filter != null) {
                 this.applyFilterExpression(filter);
+                filterTextField.setStyleForValidFilter();
             } else {
                 this.applyFilterExpression(Qualifier.EMPTY);
             }
         } catch (FilterException ex) {
             this.applyFilterExpression(Qualifier.EMPTY);
+            filterTextField.setStyleForInvalidFilter();
             // Overrides message in status bar
             UI.status.displayMessage(getUniquePanelName(
                 panelMenuBar.getPanelName()) + ": " + ex.getMessage());


acruis:
Do you have a screenshot for this?

damithc:
@garbanzos In the first comment, did you mean 'red when the filter is **in**valid'? That's how it should be.

src/main/java/ui/components/FilterTextField.java
```diff
@@ -94,6 +97,18 @@ private void setup() {
     }
 
     /**
+     * Set the background of FilterTextField to green if there is no syntax errors.
```
acruis: `green`?

damithc: I suggested not to use green because the only case that needs to be highlighted is errors. If there is no error, we don't have to do anything special.

garbanzos:
@damithc oops yeah i will edit it

src/main/java/ui/components/FilterTextField.java
```diff
@@ -48,13 +48,14 @@ public FilterTextField(String initialText) {
     private void setup() {
         setPrefColumnCount(30);
         validationSupport.registerValidator(this, (c, newValue) -> {
-            boolean wasError = false;
+            boolean isValidFilter = true;
             try {
                 Parser.parse(getText());
             } catch (ParseException e) {
```
garbanzos: Is there a reason why we are catching ParseException and not catching FilterException (which in includes both ParseException and SemanticExpection)?

garbanzos:
Ready for review.

damithc:
Change the label to reflect the status.

ndt93:
@acruis Can you help review this? Thanks.

src/main/java/ui/components/FilterTextField.java
```diff
@@ -94,6 +95,20 @@ private void setup() {
...
+     * Set the background of FilterTextField to white if there is no syntax errors.
+     * Otherwise, background is set to red
+     *
+     * @param isValidFilter
+     */
+    public void setTextFieldBackground(boolean isValidFilter){
+        if (isValidFilter) {
+            this.setStyle("-fx-control-inner-background: white");
+        } else {
+            this.setStyle("-fx-control-inner-background: #EE8993");
```
damithc: May be declare these two magic strings as constants? If you do that, you don't even need this method. It can be `thisSetStyle(isValidFilter ? VALID_FILTER_STYLE : INVALID_FILTER_STYLE)`

src/main/java/ui/components/FilterTextField.java
```diff
@@ -94,6 +95,20 @@ private void setup() {
     }
 
     /**
+     * Set the background of FilterTextField to white if there is no syntax errors.
```
damithc: `Sets ...`

src/main/java/ui/components/FilterTextField.java
garbanzos: setTextFieldBackground is also used in FilterPanel to change the background colour of the field text after the user presses enter.

damithc: In that case, you can keep the method but still use constants.

src/main/java/ui/components/FilterTextField.java
```diff
@@ -94,6 +100,20 @@ private void setup() {
     }
 
     /**
+     * Sets the background of FilterTextField to white if there is no syntax errors.
```
damithc: The color is no longer under the control of this method. Just say the appropriate style will be applied.

src/main/java/ui/components/FilterTextField.java
```diff
@@ -94,6 +100,20 @@ private void setup() {
...
+     * Otherwise, background is set to red
+     *
+     * @param isValidFilter
+     */
+    public void setTextFieldBackground(boolean isValidFilter){
+        if (isValidFilter) {
+            this.setStyle(VALID_FILTER_STYLE);
+        } else {
+            this.setStyle(INVALID_FILTER_STYLE);
+        }
```
damithc: The `... ? ... : ...` operator doesn't work here?

src/main/java/ui/components/FilterTextField.java
damithc: Another thing: this method doesn't even know about syntax errors. Its behavior is controlled by the parameter. Write the comment accordingly. i.e. The comment should reflect ONLY what the method does, not how it is used.

src/main/java/ui/components/FilterTextField.java
```diff
@@ -94,6 +100,16 @@ private void setup() {
     }
 
     /**
+     * Sets the background of FilterTextField to the appropriate color. 
+     * The color set depends on isValidFilter.
```
damithc: `color` --> `style`?
Also need to mention how the parameter affects the behavior. i.e. if true, ... otherwise ...

damithc: Alternatively (preferred), you can have two methods. `setStyleForValidFilter()` and `setStyleForInvalidFilter()`.
The call site will be more readable that way.

src/main/java/ui/components/FilterTextField.java
```diff
@@ -48,13 +53,15 @@ public FilterTextField(String initialText) {
     private void setup() {
         setPrefColumnCount(30);
         validationSupport.registerValidator(this, (c, newValue) -> {
-            boolean wasError = false;
+            boolean isValidFilter = true;
```
damithc: What was the reason for this change?

garbanzos: Previously, `setTextFieldBackground` required a param `isValidFilter` and was called in `setup`. 

I thought that `setTextFieldBackground (isValidFilter)` was more reader friendly than `setTextFieldBackground(!wasError) so I changed it. 

But since the setting of the style of the background no longer require an argument. I think I will revert the variable back to `wasError`.

damithc: Understood. Yup, better to revert back. In the current state it looks like an unnecessary change.

acruis:
I'm not sure if we have any tests to check this behaviour yet in the FilterTextField. I think it's possible to do unit tests for this with Mockito. @damithc does this change warrant more tests to be written?

damithc:
Yes, it would be nice to cover the behavior in tests. But can be done as a separate issue as long as @garbanzos takes that issue immediately after this.

garbanzos:
Created issue #1311.

acruis:
OK, no other comments.

dariusf:
@garbanzos you can [merge](https://github.com/HubTurbo/HubTurbo/blob/master/docs/process.md#merging-a-pull-request) this. Do squash the noisy commits and ensure coverage passes before that.

dariusf:
Not sure why it isn't showing here, but it seems [coverage has increased](https://coveralls.io/builds/5315059). Think this is good to go.

PR #1320, opened 2016-03-07
Warn users who don't have all dependencies #1082

### LINKED ISSUE #1082: Warn users who don't have all dependencies
A dialog should open on startup to warn users (and/or give the option to exit) if they don't have all the dependencies. This can be implemented in the updater.

- JRE 1.8.0_40
- Chrome: warn those who don't have it installed, and those who have an outdated version (unlikely, also do we need a specific version?)

HansNewbie:
Fixes #1082


HansNewbie:
@dariusf I am not sure if I put the check in a good place. Do you think it's okay?

dariusf:
I think the placement is fine.

We should represent the version number as something other than a string. Also, we should parse it with a regex. :stuck_out_tongue: And possibly parameterise the method with it, since it's hard to test without stubbing out the constant.

A class with fromString/toString methods should be useful. There might already be something like this in the standard library, not sure.

HansNewbie:
@dariusf I create a `JavaVersion` class that can regex and format back from/to string. Is this what you were talking about? I didn't find any standard library for version number.

dariusf:
Yes on the class, but the parsing should probably be more robust than splitting the string. There are changes coming in [JDK 9](https://blogs.oracle.com/java-platform-group/entry/a_new_jdk_9_version) which would break with this (existing conventions [here](http://www.oracle.com/technetwork/java/javase/versioning-naming-139433.html)). I'm not sure if we have to implement the whole thing right now. Probably something that works, and is extensible for JDK 9 in future.

HansNewbie:
@dariusf ready for review

src/main/java/util/JavaVersion.java
```diff
@@ -0,0 +1,93 @@
...
+
+        if (javaVersionMatcher.find()) {
+            return new JavaVersion(Integer.parseInt(javaVersionMatcher.group(1)),
+                    Integer.parseInt(javaVersionMatcher.group(2)),
+                    Integer.parseInt(javaVersionMatcher.group(3)),
+                    Integer.parseInt(javaVersionMatcher.group(4)),
+                    Integer.parseInt((javaVersionMatcher.group(6) != null ? javaVersionMatcher.group(6) : "0")));
+        } else {
+            logger.error("Java version not valid");
+            return null;
```
dariusf: Instead of returning null, could we throw an exception? We have to handle the case of finding an unknown Java version as well. Right now there will be an NPE when trying to do the comparison.

src/main/java/util/JavaVersion.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+
+    public String toString() {
+        return String.format("%1$d.%2$d.%3$d_%4$d-b%5$d", discard, major, minor, update, build);
```
dariusf: Could we use `%d` instead? We don't need the positional parts if we're using all the fields in order.

src/main/java/util/JavaVersion.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    public int compareTo(JavaVersion other) {
```
dariusf: We should explicitly `@Override` here.

src/main/java/util/JavaVersion.java
```diff
@@ -0,0 +1,93 @@
...
+        if (this.equals(other)) {
+            return 0;
+        }
+
+        return this.discard < other.discard ? -1 :
```
dariusf: Maybe we could define equality (the more specific concept) in terms of Comparable (the more general concept) instead? A bit more DRY. So this would be:

```java
return this.major != other.major ? this.major - other.major :
    this.minor != other.minor ? this.minor - other.minor : 0;
```

and `equals` would be `this.compareTo(other) == 0` (along with null checks, reference equality, etc.).

src/main/java/util/JavaVersion.java
```diff
@@ -0,0 +1,93 @@
...
+
+        return Integer.parseInt(hash);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
+        if (!JavaVersion.class.isAssignableFrom(obj.getClass())) {
```
dariusf: Is this subtype equality needed?

src/test/java/guitests/JavaVersionWarningTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+        Platform.runLater(() -> {
+            try {
+                methodCheckJavaDependency.invoke(ui, runtime, required);
+            } catch (IllegalAccessException | InvocationTargetException e) {
+                fail("Failed to call method");
+            }
+        });
+
+        String message = "Your Java version is older than HubTurbo's requirement. " +
```
dariusf: Could we have this in a string constant in JavaVersion, then reuse it here (with String.format)?

src/test/java/tests/JavaVersionTest.java
```diff
@@ -0,0 +1,74 @@
...
+        assertEquals(javaVersion, null);
+    }
+
+    @Test
+    public void javaVersionComparable_Java8Version_HashCodesAndEqualAreCorrect() {
+        String version = "1.8.0_60";
+        JavaVersion expectedVersion = new JavaVersion(1, 8, 0, 60, 0);
+        JavaVersion javaVersion = JavaVersion.fromString(version);
+        assertEquals(expectedVersion.hashCode(), javaVersion.hashCode());
+        assertEquals(true, expectedVersion.equals(javaVersion));
```
dariusf: This can use assertTrue.

src/main/java/ui/UI.java
```diff
@@ -540,4 +541,22 @@ public boolean isWindowMinimized() {
     public boolean isWindowFocused() {
         return mainStage.isFocused();
     }
+
+    private void checkJavaDependency() {
```
damithc: Add header comments

src/main/java/ui/UI.java
```diff
@@ -540,4 +541,22 @@ public boolean isWindowMinimized() {
...
+        JavaVersion javaRuntimeVersion = JavaVersion.fromString(System.getProperty("java.runtime.version"));
+        JavaVersion javaRequiredVersion = JavaVersion.fromString(JavaVersion.REQUIRED_VERSION);
+
+        checkJavaDependency(javaRuntimeVersion, javaRequiredVersion);
+    }
+
+    private void checkJavaDependency(JavaVersion javaRuntimeVersion, JavaVersion javaRequiredVersion) {
+        if (javaRuntimeVersion.compareTo(javaRequiredVersion) < 0) {
```
damithc: `If` blocks should capture odd cases. The normal case should not be nested. Reverse the condition and use an early return.

src/main/java/ui/UI.java
```diff
@@ -540,4 +541,22 @@ public boolean isWindowMinimized() {
...
     }
+
+    private void checkJavaDependency() {
+        JavaVersion javaRuntimeVersion = JavaVersion.fromString(System.getProperty("java.runtime.version"));
+        JavaVersion javaRequiredVersion = JavaVersion.fromString(JavaVersion.REQUIRED_VERSION);
+
+        checkJavaDependency(javaRuntimeVersion, javaRequiredVersion);
+    }
+
+    private void checkJavaDependency(JavaVersion javaRuntimeVersion, JavaVersion javaRequiredVersion) {
```
damithc: Method name doesn't exactly match what the method does.

src/main/java/ui/UI.java
```diff
@@ -540,4 +541,22 @@ public boolean isWindowMinimized() {
     public boolean isWindowFocused() {
         return mainStage.isFocused();
     }
+
+    private void checkJavaDependency() {
+        JavaVersion javaRuntimeVersion = JavaVersion.fromString(System.getProperty("java.runtime.version"));
+        JavaVersion javaRequiredVersion = JavaVersion.fromString(JavaVersion.REQUIRED_VERSION);
+
+        checkJavaDependency(javaRuntimeVersion, javaRequiredVersion);
```
damithc: Level of abstraction is odd here. I'd rather see something like this:
```java
if(isJavaVersionTooLow(..., ...)){
    showJavaVersionWarning();
}
```

src/main/java/util/JavaVersion.java
```diff
@@ -0,0 +1,93 @@
...
+ */
+public class JavaVersion implements Comparable<JavaVersion> {
+    private static final Logger logger = LogManager.getLogger(JavaVersion.class.getName());
+
+    private static final Pattern JAVA_8_VERSION_PATTERN =
+            Pattern.compile("(\\d+)\\.(\\d+)\\.(\\d+)\\_(\\d+)(-b(\\d+))?");
+
+    public static final String REQUIRED_VERSION = "1.8.0_60";
+
+    private final int discard;
```
damithc: Explain these variables

src/main/java/util/JavaVersion.java
HansNewbie: No, but don't we need to check the type of the object?

HansNewbie:
@dariusf ready for review.

src/main/java/util/JavaVersion.java
dariusf: Yes, but that's usually done with `instanceof`. This is a bit more permissive in that it will return true for subtypes, which is usually not what we want.

src/main/java/util/JavaVersion.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    public String toString() {
+        return String.format("%d.%d.%d_%d-b%d", discard, major, minor, update, build);
+    }
+
+    @Override
+    public int compareTo(JavaVersion other) {
+        if (this.equals(other)) {
+            return 0;
```
dariusf: We shouldn't still need this.

src/main/java/util/JavaVersion.java
```diff
@@ -0,0 +1,97 @@
...
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
+        if (!JavaVersion.class.isAssignableFrom(obj.getClass())) {
+            return false;
+        }
+        final JavaVersion other = (JavaVersion) obj;
+
+        return this.discard == other.discard &&
```
dariusf: Instead of this, we can check `compareTo(other) == 0`.

src/main/java/util/JavaVersion.java
```diff
@@ -0,0 +1,97 @@
...
+                this.major == other.major &&
+                this.minor == other.minor &&
+                this.update == other.update &&
+                this.build == other.build;
+    }
+
+    public static boolean isJavaVersionTooLow(JavaVersion runtime, JavaVersion required) {
+        return runtime.compareTo(required) < 0 ? true : false;
```
dariusf: This can be simplified to `runtime.compareTo(required) < 0`.

src/main/java/ui/UI.java
```diff
@@ -540,4 +541,27 @@ public boolean isWindowMinimized() {
...
         return mainStage.isFocused();
     }
+
+    /**
+     * Warns user if the Java runtime version is lower than HT's requirement
+     */
+    private void warnIfJavaVersionOutdated() {
+        try {
+            JavaVersion javaRuntimeVersion = JavaVersion.fromString(System.getProperty("java.runtime.version"));
+            JavaVersion javaRequiredVersion = JavaVersion.fromString(JavaVersion.REQUIRED_VERSION);
```
dariusf: Seems like the required version should be a property of the application (UI) instead.

This is also a bit verbose. Maybe we could drop some of the prefixes for the local variables (which are less likely to be ambiguous)?

src/main/java/ui/UI.java
```diff
@@ -540,4 +544,27 @@ public boolean isWindowMinimized() {
...
+    private void warnIfJavaVersionOutdated() {
+        try {
+            JavaVersion runtimeVersion = JavaVersion.fromString(System.getProperty("java.runtime.version"));
+            JavaVersion requiredVersion = JavaVersion.fromString(REQUIRED_VERSION);
+
+            if (JavaVersion.isJavaVersionTooLow(runtimeVersion, requiredVersion)) {
+                showJavaVersionOutdatedWarning(runtimeVersion, requiredVersion);
+            }
+        } catch (IllegalArgumentException e) {
+            logger.error("Java Version string failed to be parsed.", e);
```
damithc: I'm not sure if this is the best way to handle this error. Shouldn't there be a more visible indication if this happens?

src/main/java/ui/UI.java
```diff
@@ -540,4 +544,27 @@ public boolean isWindowMinimized() {
...
+        }
+    }
+
+    private void showJavaVersionOutdatedWarning(JavaVersion runtimeVersion, JavaVersion requiredVersion) {
+        String message = String.format(JavaVersion.OUTDATED_JAVA_VERSION_MESSAGE,
+                requiredVersion.toString(), runtimeVersion.toString());
+        DialogMessage.showInformationDialog("Update your Java version",
+                message);
```
damithc: Is this line break necessary?

src/main/java/ui/UI.java
```diff
@@ -540,4 +544,27 @@ public boolean isWindowMinimized() {
...
+                showJavaVersionOutdatedWarning(runtimeVersion, requiredVersion);
+            }
+        } catch (IllegalArgumentException e) {
+            logger.error("Java Version string failed to be parsed.", e);
+        }
+    }
+
+    private void showJavaVersionOutdatedWarning(JavaVersion runtimeVersion, JavaVersion requiredVersion) {
+        String message = String.format(JavaVersion.OUTDATED_JAVA_VERSION_MESSAGE,
+                requiredVersion.toString(), runtimeVersion.toString());
```
damithc: 8 space indentation is a last resort. See if there is a more readable choice. e.g. 
```java
String message = String.format(JavaVersion.OUTDATED_JAVA_VERSION_MESSAGE,
                               requiredVersion.toString(), runtimeVersion.toString());
```

src/main/java/util/JavaVersion.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public static final String OUTDATED_JAVA_VERSION_MESSAGE =
+            "Your Java version is older than HubTurbo's requirement. " +
+            "Use it at your own risk.%n%n" +
+            "Required version\t: %s%n" +
+            "Installed version\t: %s";
+
+    // Java version in String is formatted as "<discard>.<major>.<minor>_<update>" with optional suffix of
+    // "-b<build>". Discard is not used because all Java versions (up to Java 8) will start with "1.".
+    // This might change in Java 9.
```
damithc: Use a proper javadoc comment?

src/main/java/util/JavaVersion.java
```diff
@@ -0,0 +1,86 @@
...
+
+        if (javaVersionMatcher.find()) {
+            return new JavaVersion(Integer.parseInt(javaVersionMatcher.group(1)),
+                    Integer.parseInt(javaVersionMatcher.group(2)),
+                    Integer.parseInt(javaVersionMatcher.group(3)),
+                    Integer.parseInt(javaVersionMatcher.group(4)),
+                    Integer.parseInt((javaVersionMatcher.group(6) != null ? javaVersionMatcher.group(6) : "0")));
+        } else {
+            throw new IllegalArgumentException("String is not a valid Java Version. " + javaVersion);
+        }
```
damithc: Use a guard clause so that the main body of the method is not nested. e.g.
```
if() throw exception;
//normal body
```

src/main/java/util/JavaVersion.java
```diff
@@ -0,0 +1,86 @@
...
+        if (!(obj instanceof JavaVersion)) {
+            return false;
+        }
+        final JavaVersion other = (JavaVersion) obj;
+
+        return compareTo(other) == 0;
+    }
+
+    public static boolean isJavaVersionTooLow(JavaVersion runtime, JavaVersion required) {
```
damithc: Actually, isn't this is a case of being `lower` rather than `tooLow`? Rename method accordingly.

src/test/java/tests/JavaVersionTest.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        version = "1.0.7";
+        try {
+            JavaVersion.fromString(version);
+            fail("Invalid Java version string is parsed with no exception returned.");
+        } catch (IllegalArgumentException e) {
+            // pass
+        }
+    }
```
damithc: Isn't ther a more elegant way to check for exceptions in JUnit?

src/test/java/tests/JavaVersionTest.java
```diff
@@ -0,0 +1,135 @@
...
+import util.JavaVersion;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.fail;
+
+public class JavaVersionTest {
+    @Test
+    public void javaVersionParsing_Java8VersionNoBuild_ParsedVersionCorrectly() {
```
damithc: In `_Java8VersionNoBuild_` is the `Java8` part worthy of mentioning?

src/test/java/tests/JavaVersionTest.java
```diff
@@ -0,0 +1,135 @@
...
+        String version = "10.80.100_60";
+        JavaVersion expectedVersion = new JavaVersion(10, 80, 100, 60, 0);
+        JavaVersion javaVersion = JavaVersion.fromString(version);
+        assertEquals(expectedVersion, javaVersion);
+
+        version = "010.080.100_60";
+        expectedVersion = new JavaVersion(10, 80, 100, 60, 0);
+        javaVersion = JavaVersion.fromString(version);
+        assertEquals(expectedVersion, javaVersion);
+    }
```
damithc: Looks like you can merge the three tests above into one `javaVersion_acceptableVerionString` or something like that. Furthermore, you can extract a utility method `verifyVersionParsedCorrectly(String, int, int, int, int, int)`

src/test/java/guitests/JavaVersionWarningTest.java
```diff
@@ -0,0 +1,50 @@
...
+        Platform.runLater(() -> {
+            try {
+                methodShowJavaVersionOutdatedWarning.invoke(ui, runtime, required);
+            } catch (IllegalAccessException | InvocationTargetException e) {
+                fail("Failed to call method");
+            }
+        });
+
+        String message = String.format(JavaVersion.OUTDATED_JAVA_VERSION_MESSAGE,
+                required.toString(), runtime.toString());
```
damithc: Line wrapping can be better. Refer to coding standard for guidelines.

src/test/java/guitests/JavaVersionWarningTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+        waitUntilNodeAppears(hasText(message));
+        click("OK");
+    }
```
damithc: Do you have test cases for when the version is same, and when the version is higher?

src/main/java/ui/UI.java
```diff
@@ -51,6 +51,9 @@
     public static final int VERSION_MINOR = 22;
     public static final int VERSION_PATCH = 0;
 
+    // HT Required Java Version
+    public static final String REQUIRED_VERSION = "1.8.0_60";
```
damithc: Write proper javadoc comments

damithc:

src/test/java/guitests/JavaVersionWarningTest.java
HansNewbie: This test case is testing that the warning dialog will show up. The version difference is calculated outside this method. I will rename the method expected behavior accordingly. Consequently, the version difference will be tested by testing the method `JavaVersion.isJavaVersionLower`.

src/test/java/tests/JavaVersionTest.java
HansNewbie: Not for now, but there is a proposed change for Java 9 where the version format will change. Should I remove `8` for now?

damithc: Yes, remove it for now.

src/main/java/ui/UI.java
HansNewbie: I will rework this. So:
- `requiredVersion` will be covered by a test case, hence there should be no way that it fails parsing. I will just log an error saying `should have been covered by test case`.
- if `runtimeVersion` cannot be parsed, will show an error dialog saying that the Java version is not compatible with HubTurbo.
- Otherwise, show the current warning.

src/main/java/ui/UI.java
```diff
@@ -540,4 +546,42 @@ public boolean isWindowMinimized() {
...
+    private void warnIfJavaVersionOutdated() {
+        JavaVersion requiredVersion, runtimeVersion;
+        String javaRuntimeVersionString = System.getProperty("java.runtime.version");
+
+        try {
+            requiredVersion = JavaVersion.fromString(REQUIRED_JAVA_VERSION);
+        } catch (IllegalArgumentException e) {
+            logger.error("Required Java Version string cannot be parsed. This should have been covered by test.");
+            return;
+        }
```
damithc: This should be an assertion failure?

src/main/java/ui/UI.java
```diff
@@ -540,4 +546,42 @@ public boolean isWindowMinimized() {
...
+
+        if (JavaVersion.isJavaVersionLower(runtimeVersion, requiredVersion)) {
+            showJavaVersionOutdatedWarning(runtimeVersion, requiredVersion);
+        }
+    }
+
+    private void showJavaVersionOutdatedWarning(JavaVersion runtimeVersion, JavaVersion requiredVersion) {
+        String message = String.format(JavaVersion.WARNING_MSG_OUTDATED_JAVA_VERSION,
+                                       requiredVersion.toString(), runtimeVersion.toString());
+        DialogMessage.showInformationDialog("Update your Java version", message);
```
damithc: This is the message you want to show to the user ?????

src/main/java/util/JavaVersion.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Represents Java version
+ */
+public class JavaVersion implements Comparable<JavaVersion> {
+
+    private static final Pattern JAVA_8_VERSION_PATTERN =
+            Pattern.compile("(\\d+)\\.(\\d+)\\.(\\d+)\\_(\\d+)(-b(\\d+))?");
+
+    public static final String WARNING_MSG_OUTDATED_JAVA_VERSION =
```
damithc: I don't think this belongs in this class. JavaVersion is a general utility class. It should not contain a HT-specific message shown to the user.

src/main/java/ui/UI.java
HansNewbie: But what if the user is on Java 9, but he has not updated his HT? Assertion failure would mean he needs to download the latest version that accounts for Java 9 manually.

src/main/java/util/JavaVersion.java
HansNewbie: Okay, will move it.

src/main/java/util/JavaVersion.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    public int compareTo(JavaVersion other) {
+        return this.discard != other.discard ? this.discard - other.discard :
+                this.major != other.major ? this.major - other.major :
+                this.minor != other.minor ? this.minor - other.minor :
+                this.update != other.update ? this.update - other.update :
+                this.build != other.build ? this.build - other.build : 0;
+    }
```
damithc: This statement has lot of branches. Do you have test cases to cover all of them?

HansNewbie: Oops. Forgot to test this method.

damithc:
A few more minor issues pending.

HansNewbie:
@damithc I have tested `compareTo` method and move the warning messages to `UI`. I left the `requiredVersion` error as it is since assertion failure will block the use of HT completely.

src/main/java/ui/UI.java
```diff
@@ -51,17 +51,32 @@
...
     public static final int VERSION_PATCH = 0;
 
+    private static final Logger logger = LogManager.getLogger(UI.class.getName());
+    private static HWND mainWindowHandle;
+    private final GlobalHotkey globalHotkey = new GlobalHotkey(this);
+
+    private static final int REFRESH_PERIOD = 60;
+
+    /**
+     * HT Required Java Version
```
damithc: `Minimum Java version required by HT`

src/main/java/ui/UI.java
damithc: This is not the user's java version. There is no way this exception happen right? If it did happen, we want to catch it during testing.

src/main/java/ui/UI.java
```diff
@@ -540,4 +556,42 @@ public boolean isWindowMinimized() {
     public boolean isWindowFocused() {
         return mainStage.isFocused();
     }
+
+    /**
+     * Warns user if the Java runtime version is lower than HT's requirement
+     */
+    private void warnIfJavaVersionOutdated() {
+        JavaVersion requiredVersion, runtimeVersion;
+        String javaRuntimeVersionString = System.getProperty("java.runtime.version");
```
damithc: Declare the variable close to where it is used.

damithc:
Can merge after addressing comments.

src/main/java/ui/UI.java
HansNewbie: Yes, this should never happen. I have a test that check this. Should I put
an assertion failure if it ever happens?

HansNewbie:
Changes made: assertion failure and requested cosmetic changes.

PR #1323, opened 2016-03-07
Exception thrown from asynchronous filtering in UpdateController is not caught #1322

### LINKED ISSUE #1322: Exception thrown from asynchronous filtering in UpdateController is not caught
In particular, UpdateController's processFilter does its opening of repositories and filtering of issues in a separate thread. When the filter parsing throws an Exception, the main thread doesn't know about it.

This causes errors in filters to not appear as well as other weird interactions (?).

This can be solved either by:
1. Checking whether the filter is valid before doing the filtering. (Currently it only checks the syntax before)
or 
2. The asynchronous filtering triggering an (ParseError?)Event to tell the panel that a parsing error has occured.

m133225:
Fixes #1322.

Created a new event FilterExceptionEvent which is triggered by processPanels (possibly multiple times, once for each erroneous panel) when there is an error in filtering.

### DIFF
--- a/src/main/java/backend/UpdateController.java
+++ b/src/main/java/backend/UpdateController.java
@@ -59,7 +59,7 @@ public class UpdateController {
 
             if (toUpdate.isEmpty()) {
                 // If no issues requiring metadata update, just run the filter and sort.
-                logic.updateUI(processPanels(filterPanels));
+                logic.updateUI(processFilters(filterExprs));
                 return;
             }
 
@@ -74,7 +74,7 @@ public class UpdateController {
                             + results.size() + " repos"))
                     .thenCompose(n -> logic.getRateLimitResetTime())
                     .thenApply(logic::updateRemainingRate)
-                    .thenRun(() -> logic.updateUI(processPanels(filterPanels))); // Then filter the second time.
+                    .thenRun(() -> logic.updateUI(processFilters(filterExprs))); // Then filter the second time.
         });
     }
 
@@ -116,20 +116,19 @@ public class UpdateController {
     }
 
     /**
-     * Filters, sorts and counts issues within the model according to the given panels' filter expressions.
+     * Filters, sorts and counts issues within the model according to the given filter expressions
      * In here, "processed" is equivalent to "filtered, sorted and counted".
      *
-     * @param filterPanels Filter panels to process.
+     * @param filterExprs Filter expressions
      * @return Filter expressions and their corresponding issues after filtering, sorting and counting.
      */
-    private Map<FilterExpression, List<GuiElement>> processPanels(List<FilterPanel> filterPanels) {
+    private Map<FilterExpression, List<GuiElement>> processFilters(List<FilterExpression> filterExprs) {
         MultiModel models = logic.getModels();
         List<TurboIssue> allModelIssues = models.getIssues();
 
         Map<FilterExpression, List<GuiElement>> processed = new HashMap<>();
 
-        filterPanels.stream().distinct().forEach(filterPanel -> {
-            FilterExpression filterExpr = filterPanel.getCurrentFilterExpression();
+        filterExprs.stream().distinct().forEach(filterExpr -> {
             boolean hasUpdatedQualifier = Qualifier.hasUpdatedQualifier(filterExpr);
 
             try {
@@ -145,7 +144,7 @@ public class UpdateController {
 
                 processed.put(filterExpr, processedElements);
             } catch (FilterException e) {
-                Platform.runLater(() -> UI.events.triggerEvent(new FilterExceptionEvent(filterPanel, e.getMessage())));
+                Platform.runLater(() -> UI.events.triggerEvent(new FilterExceptionEvent(filterExpr, e.getMessage())));
             }
         });
 
--- a/src/main/java/backend/UpdateController.java
+++ b/src/main/java/backend/UpdateController.java
@@ -49,7 +49,7 @@ public class UpdateController {
 
         // Filter and sort the issues first even if the metadata is not yet available so that criteria not
         // based on metadata can have immediate effect.
-        logic.updateUI(processFilter(filterExprs));
+        logic.updateUI(processFilters(filterExprs));
 
         // Open specified repos
         openRepositoriesInFilters(filterPanels)
--- a/src/main/java/backend/UpdateController.java
+++ b/src/main/java/backend/UpdateController.java
@@ -3,14 +3,18 @@ package backend;
 import backend.resource.Model;
 import backend.resource.MultiModel;
 import backend.resource.TurboIssue;
+import filter.FilterException;
 import filter.expression.FilterExpression;
 import filter.expression.Qualifier;
+import javafx.application.Platform;
 import org.apache.logging.log4j.Logger;
 import filter.expression.QualifierType;
 import ui.GuiElement;
+import ui.UI;
 import ui.issuepanel.FilterPanel;
 import util.Futures;
 import util.HTLog;
+import util.events.FilterExceptionEvent;
 
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
@@ -55,7 +59,7 @@ public class UpdateController {
 
             if (toUpdate.isEmpty()) {
                 // If no issues requiring metadata update, just run the filter and sort.
-                logic.updateUI(processFilter(filterExprs));
+                logic.updateUI(processPanels(filterPanels));
                 return;
             }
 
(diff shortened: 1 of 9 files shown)


m133225:
Ready for review.

src/main/java/backend/UpdateController.java
```diff
@@ -107,32 +111,37 @@ public void processAndRefresh(List<FilterPanel> filterPanels) {
...
-     * Filters, sorts and counts issues within the model according to the given filter expressions.
+     * Filters, sorts and counts issues within the model according to the given panels' filter expressions.
      * In here, "processed" is equivalent to "filtered, sorted and counted".
      *
-     * @param filterExprs Filter expressions to process.
+     * @param filterPanels Filter panels to process.
      * @return Filter expressions and their corresponding issues after filtering, sorting and counting.
      */
-    private Map<FilterExpression, List<GuiElement>> processFilter(List<FilterExpression> filterExprs) {
+    private Map<FilterExpression, List<GuiElement>> processPanels(List<FilterPanel> filterPanels) {
```
dariusf: Actually, do we need to change this? This makes it so we filter twice if there are two panels with the same filter.

The filter expression is all we need to uniquely identify a panel, so we can pass it in the exception and compare it with the panels'.

src/main/java/backend/UpdateController.java
```diff
@@ -107,32 +111,37 @@ public void processAndRefresh(List<FilterPanel> filterPanels) {
...
         List<TurboIssue> allModelIssues = models.getIssues();
 
         Map<FilterExpression, List<GuiElement>> processed = new HashMap<>();
 
-        filterExprs.stream().distinct().forEach(filterExpr -> {
-            boolean hasUpdatedQualifier = Qualifier.hasUpdatedQualifier(filterExpr);
+        filterPanels.stream().distinct().forEach(filterPanel -> {
+            try {
+                FilterExpression filterExpr = filterPanel.getCurrentFilterExpression();
+                boolean hasUpdatedQualifier = Qualifier.hasUpdatedQualifier(filterExpr);
```
dariusf: Do these two lines need to be in the try block? We should try to minimise its scope.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -230,6 +231,16 @@ private void applyFilterExpression(FilterExpression filter) {
 
     protected abstract void stopLoadingAnimationIfApplicable(AppliedFilterEvent e);
 
+    private void showErrorStatus(FilterExceptionEvent e) {
+        if (e.panel != this) return;
```
dariusf: We can break this into two methods:
- One handles the event and checks if it is applicable to this panel. It also gets the error string from the exception.
- The other takes a string and displays the error, makes the field red, etc.

The error-handling upon catching a FilterException (applyStringFilter) can then reuse the second method.

src/main/java/util/events/FilterExceptionEvent.java
```diff
@@ -0,0 +1,17 @@
...
+
+import filter.FilterException;
+import ui.issuepanel.FilterPanel;
+
+/**
+ * This class is meant to indicate when there is an exception thrown during the filtering of issues
+ */
+public class FilterExceptionEvent extends Event {
+    public FilterPanel panel;
+    public FilterException exception;
```
dariusf: Do we need the exception here? Would the error message suffice?

dariusf:
The exception event should also trigger in `tallyMetadataUpdate`.

src/main/java/backend/UpdateController.java
m133225: Isn't it possible for 2 panels to have the same filter expression? So we can't uniquely identify a panel based on just its expression...

src/main/java/ui/issuepanel/FilterPanel.java
m133225: Actually, is the error-handling in applyStringFilter necessary? I think it wouldn't be used at all. (Not sure if Parser.parse will throw a FilterException.)

src/main/java/backend/UpdateController.java
dariusf: Yup, but if they have the same filter expression, they would have the same contents (sans side effects) and errors.

src/main/java/ui/issuepanel/FilterPanel.java
dariusf: It throws a ParseException on malformed input (and is what is causing our `is:` woes :smiling_imp:).

m133225: Never mind, it is necessary as Parser.parse might throw a ParseException which is a subclass of FilterException.

src/main/java/backend/UpdateController.java
m133225: It is possible for a panel to have a erroneous filter which is not applied yet, when the user just types without confirming (at least that's the behaviour for now).
In that case, applying the same erroneous filter in another panel will cause a change in both panels, which might not be appropriate.

dariusf: But isn't that a good thing? May not be super intuitive if you see something else change, but it makes sense. :stuck_out_tongue:

acruis: I would disagree with renaming it to `panels`. Logic doesn't have to care about what is holding the filters; it just has to process the filters themselves. This is why there is a `GUIController`, which distributes the filters to their respective panels.

m133225: Noted. Have changed the method back to processFilters.

m133225:
Ready for review.

By using `==` to compare the FilterExpression in the ParseExceptionEvent, the behaviour of only the responsible panel handling the event is able to be maintained. 

(At first, thought the filter string might be re-parsed which may produce a different FilterExpression object.)

src/main/java/util/events/FilterExceptionEvent.java
```diff
@@ -0,0 +1,16 @@
+package util.events;
+
+import filter.expression.FilterExpression;
+
+/**
+ * This class is meant to indicate that there is an exception thrown during the filtering of issues
+ */
+public class FilterExceptionEvent extends Event {
+    public FilterExpression filterExpr;
+    public String exceptionMessage;
```
dariusf: These can be final.

dariusf:
How about `tallyMetadataUpdate`? Does the exception also need to be handled there?

I don't know about `==` being used for filter expressions, because they are conceptually values and should be interchangeable if they are `equals`. What you said about the filter string being re-parsed is exactly what I'm referring to: if that behaviour changed someday, this would break.

Is it important that only the panel interacted with changes? If yes, is there some other way we can tell which panel the filter expression came from?

acruis:
We should make an issue to change the filter text field's behavior, so that it will update its internal filter expression to the string currently in the filter (regardless of whether the user has pressed <kbd>Enter</kbd>) whenever the filter text field loses focus.

Also, I think it's bad practice to use events as a means of carrying data from Logic to the GUI. The UIManager and GuiController exist, in the first place, to consolidate information flow from Logic to the GUI. Events are a convenient way to pass data from one place to another without writing a lot of boilerplate, but this is done at the risk of obfuscating the flow of information within the program. You can understand your own code, but it will be difficult for other developers to build onto your work.

I do admit that we're using events somewhat haphazardly in the application right now. It used to be worse, but at least, there's a GuiController now to centralize data flowing from Logic. You can go ahead and merge this PR, but events should probably be avoided (even refactored away) in the future.

m133225:
@acruis Ready for review again. Apologies for the delay.

Had trouble:
1) Getting tests to pass after the previous PR #1280 was merged
- many of them checked for the filter result before the filtering has been completed

Several 'PlatformEx.waitOnFxThread()' have been added to various tests(related to filtering through the UI) to fix this. Maybe this should have been done in a separate issue/PR, but I thought it was rather urgent for this PR to be merged.

2) Identifying new tests to add for the relevant code
Added 3 tests to FilterTests.

dariusf:
How reliably do those tests pass now? waitOnFxThread doesn't guarantee that filtering is actually completed. If we want stronger guarantees, we have to synchronise in some other way (waiting for some condition to be true, for example).

m133225:
Definitely, it is not a guarantee for the tests, but I think it works rather reliably for now. Will restart the build a few times to double check this.

acruis:
Do solve conflicts with master when you're done running the tests.

m133225:
@acruis There weren't conflicts though... Updated the branch.

It seems to be passing rather reliably except for other unrelated tests which happen to fail from time to time.

acruis:
Okay, this is good to go after merging from master.

m133225:
@damithc Sorry, I had merged this PR forgetting to wait for your review. I was thinking that you had already approved this PR before. :sad:

PR #1325, opened 2016-03-08
User can right-click on an issue card and 'mark all below as unread/read' #1256

### LINKED ISSUE #1256: User can right-click on an issue card and 'mark all below as unread/read'

codemanmeet:
Fixes #1256. Also, the bView is not loaded on a right click now.


dariusf:
Needs to be squashed also.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -49,9 +49,13 @@
     private final ContextMenu contextMenu = new ContextMenu();
 
     private final MenuItem markAsReadUnreadMenuItem = new MenuItem();
+    private final MenuItem markAllBelowAsReadMenuItem = new MenuItem();
+    private final MenuItem markAllBelowAsUnreadMenuItem = new MenuItem();
     private static final String markAsReadMenuItemText = "Mark as read (E)";
     private static final String markAsUnreadMenuItemText = "Mark as unread (U)";
-
+    public static final String markAllAsUnreadMenuItemText = "Mark all below as unread";
+    public static final String markAllAsReadMenuItemText = "Mark all below as read";
```
damithc: Why not use ALL_CAPS?

src/test/java/guitests/IssueSelectedEventTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class IssueSelectedEventTest extends UITest{
+
+    private static final String PANEL_IDENTIFIER = "#dummy/dummy_col0";
+
+    @Test
+    public void noTriggerIssueSelectedOnRightClick_IssueInPanelRightClicked_IssueSelectedNotTriggered() {
```
damithc: I think the convention is `feature_scenario_expected` So here it should be `eventTriggerOnIssueSelection_byRightClick_selectionEventNotTriggered` or something like that.

src/test/java/unstable/MarkAllIssuesAsReadUnreadTests.java
```diff
@@ -0,0 +1,107 @@
...
+
+        // scrolls to the end of the panel
+        listViewValue.scrollAndShow(issuePanel.getIssueCount());
+
+
+        //checking for issue #7 and below as marked read
+        clickAndMarkIssuesBelow(issuePanel, 7, true);
+        verifyReadStatusOfIssuesBelow(issuePanel, 7, true);
+
+        //checking for issue #7 and below as marked read
```
damithc: Mention that you are doing it again, and why you are doing it again. The comment is supposed to give the missing info, not repeat the obvious info.

damithc:

codemanmeet:
Yup, addressed them!

dariusf:
Merged in 59b9c8b8b7eebcb79291b67d1fcce9bc73a98a8e.

PR #1328, opened 2016-03-09
Make auto-complete on demand #1293

### LINKED ISSUE #1293: Make auto-complete on-demand
Currently, it is getting in the way more than it is helping. Suggestion:
Do not activate insert the suggested letters unless user requests for auto-complete suggestions e.g using a keyboard shortcut
Optionally, we can show the suggestions automatically, but outside the filter box (e.g. just below cursor). The suggestion will not be inserted unless the user requests.

jinified:
Fixes #1293

## Progress
- [x] add more test for new autocomplete 
- [x] refactor `SuggestionMenu.java`
- [x] checks why test not covering the case when ignoring 
- [x] prevent menu from calling `show` when already visible 
- [x] minor refactoring


jinified:
@damithc  Just want to confirm a few behaviours :

1. Press <kbd>Enter</kbd> will complete current suggestion and load the filter expression.
2. Press <kbd>Esc</kbd> will removes focus from text field 
3. User can choose to click on suggested keywords to complete it 
4. User can also press <kbd>Tab</kbd> to complete current suggestion. Should the user be able to use the <kbd>Up</kbd> and <kbd>Down</kbd> to navigate the suggested word in drop down ?

damithc:
Yes, it is ok to use up/down to navigate suggestions.
From what I remember, the first `ESC` is supposed to revert to the previous filter (i.e. abort the current edit) and one more `ESC` is supposed to remove focus from filter.

jinified:
Ready for review

src/main/java/ui/components/SuggestionMenu.java
```diff
@@ -0,0 +1,80 @@
...
+    
+    // Callback functions
+    private EventHandler<ActionEvent> actionHandler = (event) -> defaultActionHandler(event);
+
+    public SuggestionMenu(int maxEntries) {
+        this.maxEntries = maxEntries;
+    }
+
+    /**
+     * Load search result in menu
```
damithc: Loads

src/main/java/ui/components/SuggestionMenu.java
```diff
@@ -0,0 +1,80 @@
...
+     * @param searchResult
+     */
+    public void loadSuggestions(List<String> searchResult) {
+        getItems().clear();
+        searchResult.stream().limit(maxEntries).forEach(this::addMenuItem);
+        selected = searchResult.stream().findFirst();
+    }
+
+    /**
+     * Get content of selected menu item
```
damithc: Gets

build.gradle
```diff
@@ -71,6 +71,7 @@ allprojects {
         compile "org.ocpsoft.prettytime:prettytime:$prettytimeVersion"
         compile "org.seleniumhq.selenium:selenium-java:$seleniumJavaVersion"
         compile "org.seleniumhq.selenium:selenium-server:$seleniumServerVersion"
+		compile group: 'org.fxmisc.richtext', name: 'richtextfx', version: '0.6.10'
```
dariusf: Indentation looks a bit off here.

src/main/java/ui/components/FilterTextField.java
```diff
@@ -70,15 +80,15 @@ private void setup(Predicate<String> validation) {
...
-                startCompletion(e);
+            
+            char typed = key.charAt(0);
+            
+            if (typed == '\t' && suggestion.isShowing()) {
+                suggestion.hide();
+                suggestion.getSelectedContent().ifPresent(this::completeWord);
+            } else {
+                suggestion.loadSuggestions(getMatchingKeywords(getCurrentWord()));
+                suggestion.show(this, Side.BOTTOM, 0, 0);
```
dariusf: Maybe we could abstract some of the parameters away with an overload?

src/main/java/ui/components/FilterTextField.java
```diff
@@ -89,14 +99,15 @@ private void setup(Predicate<String> validation) {
...
-            } else if (e.getCode() == KeyCode.ESCAPE) {
-                if (getText().equals(previousText)) {
-                    onCancel.run();
-                } else {
-                    revertEdit();
-                }
+
+            if (e.getCode() == KeyCode.ENTER) handleOnEnter();
+
+            if (e.getCode() == KeyCode.ESCAPE) handleOnEscape();
```
dariusf: We could use `else if` here, to make it clear that the effects of the key handlers are mutually exclusive.

src/main/java/ui/components/FilterTextField.java
```diff
@@ -106,6 +117,24 @@ private void setup(Predicate<String> validation) {
...
+        if (isNavigating) {
+            suggestion.hide();
+            suggestion.getSelectedContent().ifPresent(this::completeWord);
+        } else {
+            confirmEdit();
+        }
+    }
+
+    private void handleOnEscape() {
+        if (getText().equals(previousText)) {
```
dariusf: Maybe the first escape key press could dismiss the menu, but not revert the field? So in total there would be 3 possible uses for the key.

src/main/java/ui/components/FilterTextField.java
```diff
@@ -242,6 +215,30 @@ private void confirmEdit() {
         positionCaret(caretPosition);
     }
 
+    // SuggestionMenu 
+    
+    private final SuggestionMenu setupSuggestion() {
```
dariusf: This method is private, so it doesn't have to be final -- it can't be overridden anyway.

src/main/java/ui/components/SuggestionMenu.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+public class SuggestionMenu extends ContextMenu {
+    
+    private final int maxEntries;
+
+    // Selected menu item's content
+    private Optional<String> selected = Optional.empty();
+    
+    // Callback functions
+    private EventHandler<ActionEvent> actionHandler = (event) -> defaultActionHandler(event);
```
dariusf: We can use a method reference here.

src/main/java/ui/issuepanel/FilterPanel.java
```diff
@@ -113,9 +112,6 @@ private void setUpEventHandler() {
 
         // Update keywords
         List<String> all = new ArrayList<>(QualifierType.getCompletionKeywords());
-        all.addAll(e.users.stream()
-            .map(TurboUser::getLoginName)
-            .collect(Collectors.toList()));
```
dariusf: We're no longer showing user suggestions?

build.gradle
jinified: Thanks. Reminds me I should remove it since we are not using it

src/main/java/ui/components/FilterTextField.java
jinified: Are you referring to method signature of `SuggestionMenu` show method ?

dariusf:
I checked this out to try it, and it's working really well.

One minor issue: the suggestion menu moves around as I edit the filter. It doesn't follow the word I'm currently editing though.

test

If we can't avoid this, maybe we could leave the menu stuck to the left side?

Also, are we using the RichTextFX dependency?

src/main/java/ui/components/FilterTextField.java
jinified: Noted

src/main/java/ui/components/FilterTextField.java
jinified: Noted

src/main/java/ui/components/FilterTextField.java
dariusf: Yup, so we can just have `suggestion.show(this)`. The details of where the menu is positioned are a concern of the menu.

src/main/java/ui/issuepanel/FilterPanel.java
jinified: The plan is to show only `Qualifiers` and their aliases

jinified:
For now we will not be using the RichTextFX as it would take significant amount of effort to port it. 
The move around issue can probably be avoided by commenting out the code to `show` the menu on every key type

src/main/java/ui/issuepanel/FilterPanel.java
dariusf: Oh. I kind of liked this feature though. If the concern is the users occupying space in the list when people want to see filter syntax, maybe we could order them last?

dariusf:
Okay, understood. So with RichTextFX, showing the menu on every key would allow us to position it properly?

jinified:
Yes you are right

jinified:
@damithc Is it possible to create another issue to include the aliases in the `SuggestionMenu` as that would require changes in the `QualifierType.java` . @dariusf  Ready for review

damithc:
Sure @jinified . Go ahead. We prefer smaller PRs :-)

dariusf:
I think the code is fine, just need a bit of clarification on the features we're implementing.

- Are we still going to have users show up? If not, will the completion list be entirely static? If this is the case, we don't need to hook into the model update event to update the completion list.
- Not sure about aliases showing up in the completions, since they likely won't be understandable by a new user.

damithc:
We'll probably hook up the assignee picker to autocomplete user names.
The idea is to show aliases in brackets (e.g. `assignee (as)`). That way the auto-complete can help users learn aliases over time.

dariusf:
Okay. In that case, should we take the chance to remove [these parts](https://github.com/jinified/HubTurbo/blob/1293-auto-complete-on-demand/src/main/java/ui/issuepanel/FilterPanel.java#L112-L121), since the completion list will no longer depend on model updates?

Apart from that I have no more comments.

jinified:
Since the `close` method is of type `abstract`. Do we just leave it to be empty for now in `FilterPanel` ?

jinified:
@dariusf  final changes were made

src/main/java/ui/components/SuggestionMenu.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Gets content of selected menu item
+     */
+    public Optional<String> getSelectedContent() {
+        return selected;
+    }
+    
+    /**
+     *  Wraps a label inside a menu item.
```
damithc: The method says `add` but the comments says `Wraps`

src/main/java/ui/components/FilterTextField.java
```diff
@@ -32,6 +36,10 @@
     public static final String VALID_FILTER_STYLE = "-fx-control-inner-background: white";
     public static final String INVALID_FILTER_STYLE = "-fx-control-inner-background: #EE8993";
 
+    // For on-the-fly parsing and checking
+    private final ValidationSupport validationSupport = new ValidationSupport();
+    private final SuggestionMenu suggestion;
```
damithc: Perhaps we should use `Suggestions` everywhere when we refer to something containing multiple suggestions?

jinified: `suggestions` in place of `suggestion` right ? Ok.

jinified:
@damithc  can review the changes that were made

src/main/java/ui/components/FilterTextField.java
```diff
@@ -245,6 +225,30 @@ private void confirmEdit() {
         positionCaret(caretPosition);
     }
 
+    // SuggestionMenu 
+    
+    private SuggestionMenu setupSuggestion() {
```
damithc: SuggestionsMenu?

damithc:
Added minor comment. Waiting for reviewer's greenlight.

PR #1329, opened 2016-03-10
Separate updates download and local model update #1302

### LINKED ISSUE #1302: Separate GitHub data requesting and local models update process
The server data requesting process can be fully concurrently with any other local operations while the models update/modification must be mutually exclusive.

Both processes are currently coupled together and can't to be made mutually exclusive together since retrieving data from server takes a long time. Separate them out could help dealing with synchronisation problems while not trading off speeds since local models update time is unnoticeable.

The sync problems are much more likely to happen when metadata retrievals are required (for `update` filters) since the refreshing window is much wider e.g. #1299

ndt93:
Fixes #1302 

No functionality or significant structural change in this PR. Changes consist mostly of moving code around but are quite tricky not to scramble the original logic, to avoid deadlocks and to manage the dependencies between `RepoIO`, `RepoOpControl` and `MultiModel`.


src/main/java/backend/Logic.java
```diff
@@ -258,18 +258,20 @@ public Model getRepo(String repoId) {
...
+                repoOpControl.replaceIssueLabelsLocally(issue, newLabels);
+        if (localReplaceResult.isCompletedExceptionally() || localReplaceResult.isCancelled()) {
             return CompletableFuture.completedFuture(false);
         }
-        refreshUI();
+        Optional<TurboIssue> locallyModifiedIssue = localReplaceResult.join();
+        if (!locallyModifiedIssue.isPresent()) {
+            return CompletableFuture.completedFuture(false);
+        }
+        localReplaceResult.thenRun(this::refreshUI);
```
dariusf: Hmm, is this the same as doing `this.refreshUI()`? We're refreshing immediately, on this thread.

src/main/java/backend/control/RepoOpControl.java
```diff
@@ -10,23 +12,43 @@
...
+    private RepoOpControl(RepoIO repoIO, MultiModel models) {
         this.repoIO = repoIO;
+        this.models = models;
+    }
+
+    /**
+     * Gets an RepoOpControl instance that has been created previously by {@code createRepoOpControl}
+     * @return
+     */
+    public static RepoOpControl getRepoOpControl() {
```
dariusf: Do we need this to be a singleton? It just seems like a bit inconsistent that it is used in two ways (storing it in an instance variable in Logic sometimes, using the global method sometimes).

src/main/java/backend/control/RepoOpControl.java
```diff
@@ -42,6 +64,26 @@ public RepoOpControl(RepoIO repoIO) {
...
+     * @param updates
+     * @param syncOperation
+     * @return
+     */
+    public CompletableFuture<Model> updateLocalModel(GitHubModelUpdatesData updates,
+                                                     boolean syncOperation) {
+        init(updates.getRepoId());
+        CompletableFuture<Model> result = new CompletableFuture<>();
+        UpdateLocalModelOp op = new UpdateLocalModelOp(models, updates, result);
+        if (syncOperation) {
```
dariusf: Hmm, this feels a bit weird. If we don't need to synchronise it, we can just call the Logic version of this?

src/main/java/backend/github/DownloadIssuesUpdatesTask.java
```diff
@@ -0,0 +1,37 @@
...
+    private final Model model;
+
+    public DownloadIssuesUpdatesTask(TaskRunner taskRunner, Repo repo, Model model) {
+        super(taskRunner, repo);
+        this.model = model;
+    }
+
+    @Override
+    public void run() {
+        ImmutableTriple<List<TurboIssue>, String, Date> changes = repo.getUpdatedIssues(model.getRepoId(),
```
dariusf: Could we make this a(n inner) class? Tuple types are a bit verbose and not as clear.

src/main/java/backend/resource/MultiModel.java
```diff
@@ -13,6 +13,9 @@
 import java.util.stream.Collectors;
 
 /**
+ * This represents the true state of local repositories data. Operations meant to change the state
+ * of local repositories data but do not go through the methods in this class i.e. operating on
+ * dangling references its sub-components are considered unsafe
```
dariusf: Hmm, this is implemented already? Doesn't changing labels still mutate the inner objects?

src/main/java/backend/Logic.java
ndt93: Shouldn't we refresh only after the local modification is done. The local modification is also a mutually exclusive `RepoOp` now and is not done immediately in this thread (it was last time since it's too slow to be put into the queue).

src/main/java/backend/control/RepoOpControl.java
ndt93: This is the ugly part of this that I was fighting over. `RepoOpControl` depends on both `RepoIO` and `MultiModel` and so needs to be instantiated in `Logic`. `RepoIO` needs to use the same `RepoOpControl` as Logic (in the `update...` method). And `PMD` thinks it's not a good idea to make `RepoOpControl` as a static instance in `Logic` and instantiate it in `Logic`'s constructor.

One way is to break the `RepoIO`'s `update...` method into smaller methods to be called on a higher level so that `RepoIO` does not depend on `RepoOpControl` but right now the `update` method is doing certain things that make it quite hard to do that (I will explain below).

So any idea to make this nice other than the singleton instance?

src/main/java/backend/control/RepoOpControl.java
ndt93: This is due to how several methods in `RepoIO` works. When opening the model, which is an operation placed on the queue, `RepoIO` will call `updateModel`, which is then executed on the thread for that queue. Then `updateModel` will place this `updateLocalModel` operation on the queue and wait for it to be executed. Consequently there will be a deadlock. So this operation allows to be in two modes, one of which is to bypass putting the operation on the queue if the executing thread is already the thread for that queue.

Now, we can divided this `updateModel` method in `RepoIO` into smaller methods that does each step such as `download` -> `updateLocalModel` -> `writeToJSON`. However, the `writeToJSON` part of the `updateModel` method includes a corrupted JSON explosion detection and recovery that needs to recursively call the whole chain of things again, making it quite hard to separate these steps. So that's the dilemma at the moment.

src/main/java/backend/github/DownloadIssuesUpdatesTask.java
ndt93: This is the returned result from existing methods, but sure I can refactor this.

src/main/java/backend/resource/MultiModel.java
ndt93: We change labels by calling `models.replaceIssueLabels` and passing in the issue's id, the repo's id and the list of labels and no dangly object is modified so I guess that fits into this criteria.

src/main/java/backend/github/DownloadIssuesUpdatesTask.java
dariusf: Oops, sorry. If it's easy, sure, but if it's a lot to change then it's okay to leave it.

src/main/java/backend/Logic.java
dariusf: I was wondering that because right before, we `join` `localReplaceResult`, and `thenRun` on a completed future will execute the action immediately. Not sure if i'm misunderstanding.

src/main/java/backend/control/RepoOpControl.java
dariusf: I see. That's tough. I guess the recursive updating would need to be implemented by having the current task fail, then making whatever is calling it put a new task on the queue.

This recursiveness is the reason for having to synchronise only sometimes, right? We may enter another updateModel and don't want to enqueue it in that case. I think the refactoring you suggested is probably the way to go for the long term, but I'm not sure if we want to do it now.

src/main/java/backend/control/RepoOpControl.java
dariusf: The simplest way (barring making everything singletons) is probably like this:
```java
A a = new A();
B b = new B();
a.setB(b);
b.setA(a);
```
I think the 'right' way is to not have RepoIO depend on RepoOpControl though.

src/main/java/backend/resource/MultiModel.java
dariusf: Okay. Someday we should enforce the latter part by copying or something. :stuck_out_tongue:

src/main/java/backend/Logic.java
ndt93: Ah ok, I think it's a bad idea to call `.join` here since that will block the main thread.

src/main/java/backend/control/RepoOpControl.java
ndt93: All rights, looks like a nicer way to handle the dependencies without the singleton instance.

src/main/java/backend/control/RepoOpControl.java
ndt93: Yup, it's the reason to synchronise only sometimes and also for the reason that `RepoIO` depends on `RepoOpControl`. The flow is kind of convoluted so I think we can leave the refactoring later.

src/main/java/backend/resource/MultiModel.java
ndt93: Yup. To think about it, copying is not that expensive, both space and time wise since all the strings are copied by references.

ndt93:
@dariusf I refactored the singleton stuffs. The tuple types things are the returned types of all the `getUpdated...` methods in the base class `Repo` so I think that should be refactored all together in another issue. You can review this again.

src/main/java/backend/resource/MultiModel.java
```diff
@@ -84,6 +87,11 @@ public synchronized MultiModel replace(List<Model> newModels) {
         return this;
     }
 
+    public synchronized MultiModel replace(Model newModel) {
+        this.add(newModel);
+        return this;
```
dariusf: Is this meant as an alias for `add`?

Also you can `return add(newModel)`.

ndt93: Yup, `add` is a bit ambiguous in the code context where I used it, provided what it actually does, so I created this method since I don't want to break codes in other places that use `add`.

PR #1276, opened 2016-02-20
Generalise ConfigFileHandler to handle more than just the global config file #1272

nicholaslum444:
Fixes #1272 

The GlobalConfig class now inherits from Config.
ConfigFileHandler now only deals with Config type. It doesn't need to know about GlobalConfig.

Updated tests and other code to reflect the changes in the method names and signature.

Also, I have added extra test files that test that the new inheritance framework works with other types of Config subclasses. These can be deleted before the PR is merged, since the tested Config subclass does not exist in current master. It's just as proof of concept.

PR #1334, opened 2016-03-11
Updater - Update Data File #1271

### LINKED ISSUE #1271: Updater - Update Data File

HansNewbie:
Fixes #1271


HansNewbie:
@dariusf This will be the last PR to make download works. This PR adds the capability to check if there is any update to be downloaded. Can you take a look if the implementation is okay?

I added a JSON converter class as well for general use. It might need to be replaced once #1276 is done, or the general implementation I made might need to replace that PR.

Meanwhile, I will add more comments and try to add tests.

HubTurboUpdate.json
```diff
@@ -0,0 +1,12 @@
+{
+  "listOfLinks": [
```
dariusf: We can be a bit more specific about what links these are. Known versions we haven't downloaded?

src/main/java/prefs/UpdateConfig.java
```diff
@@ -0,0 +1,40 @@
...
+    public void setLastUpdateDownloadStatus(boolean status) {
+        this.lastUpdateDownloadStatus = status;
+    }
+
+    public boolean getLastUpdateDownloadStatus() {
+        return this.lastUpdateDownloadStatus;
+    }
+
+    public void addToVersionPreviouslyDownloaded(Version downloadedVersion) {
+        if (listOfVersionsPreviouslyDownloaded == null) {
```
dariusf: To be DRY, maybe we could initialise this in the constructor.

src/main/java/updater/UpdateDownloadLink.java
```diff
@@ -0,0 +1,36 @@
...
+    public int compareTo(UpdateDownloadLink other) {
+        return this.version.compareTo(other.version);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
+        if (!UpdateDownloadLink.class.isAssignableFrom(obj.getClass())) {
```
dariusf: Again, not sure if we need this subtype equality.

dariusf:
Yup, I think it looks good.

HubTurboUpdate.json
HansNewbie: This is a generic list of downloadable HubTurbo link, whether user have downloaded it or not.

Use case:
for incremental data migration, list of links can have 1 link to version major 3, 1 link to version major 4 and 1 link to version major 5. Depending on user's version, they will download one with version major just 1 greater than their current version, e.g. in V3.4.1, user will update to V4.6.1 before V5.2.1.

Will rename the variable accordingly.

HansNewbie:
@dariusf ready for review

HubTurboUpdate.json
```diff
@@ -0,0 +1,12 @@
+{
+  "listOfHTVersionsDownloadLink": [
```
dariusf: Maybe this could be just `HTVersionDownloadLinks`?

docs/gettingStarted.md
```diff
@@ -10,8 +10,7 @@ To use HubTurbo, you'll need:
 
 ## Installation
 
-Once you have the above, grab the installer from [here](https://github.com/HubTurbo/AutoUpdater/releases/download/V0.9.0/HubTurbo.jar) and run it. 
-It will download the latest HubTurbo version and launch it. 
+Once you have the above, grab HubTurbo from [here](https://github.com/HubTurbo/HubTurbo/releases/download/V3.22.0/resource-v3.22.0.jar) and run it.
```
dariusf: Does this need to be updated every time? Maybe we could link to [this](https://github.com/HubTurbo/HubTurbo/releases/latest) page instead.

src/main/java/prefs/UpdateConfig.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Contains configuration information for updating
+ */
+public class UpdateConfig {
+
+    private boolean lastUpdateDownloadStatus;
```
dariusf: What does this field indicate?

src/main/java/prefs/UpdateConfig.java
```diff
@@ -0,0 +1,35 @@
...
+    private List<Version> listOfVersionsPreviouslyDownloaded; // NOPMD
```
dariusf: What's the warning that's being suppressed?

Also, don't think the name needs the `listOf` prefix. Types are metadata/implementation details.

src/main/java/updater/UpdateData.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Represents update data to be put on server for HT to check if there is new update
+ */
+public class UpdateData {
+
+    private List<UpdateDownloadLink> listOfHTVersionsDownloadLink; // NOPMD
```
dariusf: Same here.

src/main/java/updater/UpdateData.java
```diff
@@ -0,0 +1,48 @@
...
+        // List the update link in descending order of version
+        Collections.sort(listOfHTVersionsDownloadLink, Collections.reverseOrder());
+
+        Version current = Version.getCurrentVersion();
+
+        // Get link of version that has same major version or just 1 major version up than current
+        Optional<UpdateDownloadLink> updateLink = listOfHTVersionsDownloadLink.stream()
+                .filter(link ->
+                        link.getVersion().getMajor() == current.getMajor() ||
+                        link.getVersion().getMajor() == current.getMajor() + 1)
```
dariusf: Maybe we could factor this out into a Version method.

src/main/java/util/JsonFileConverter.java
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * Save a given object into JSON file
+     * @param object object to be saved to JSON file
+     */
+    public void saveToFile(Object object) {
+        try {
+            Writer writer = new OutputStreamWriter(new FileOutputStream(jsonFile) , CHARSET);
+            gson.toJson(object, object.getClass(), writer);
+            writer.close();
+        } catch (IOException e) {
```
dariusf: Not sure if the exception should be swallowed here. In general, if this fails to save, the caller should be allowed to react in some way.

src/main/java/util/JsonFileConverter.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    /**
+     * Loads an object from a JSON file
+     * @param objectClass class of the object to be loaded; used to parse the JSON
+     * @return The object from JSON file, or Optional.empty() if the file can't be parsed
+     */
+    public Optional<Object> loadFromFile(Class<?> objectClass) {
+        if (jsonFile.exists()) {
```
dariusf: This can be flipped to put the shorter clause first.

src/main/java/util/JsonFileConverter.java
```diff
@@ -0,0 +1,89 @@
...
+                        })
+                .registerTypeAdapter(LocalDateTime.class,
+                        (JsonDeserializer<LocalDateTime>) (json, typeOfT, context) -> {
+                            Instant instant = Instant.ofEpochMilli(json.getAsJsonPrimitive().getAsLong());
+                            return LocalDateTime.ofInstant(instant, ZoneId.systemDefault());
+                        }
+                ).create();
+    }
+
+    private void createDirectories() {
```
dariusf: This should also have some way for the caller to react to the error. Probably an exception?

src/test/java/tests/UpdateConfigTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+            Field lastUpdateDownloadStatusField = updateConfigClass.getDeclaredField("lastUpdateDownloadStatus");
+            lastUpdateDownloadStatusField.setAccessible(true);
+
+            updateConfig.setLastUpdateDownloadStatus(true);
+            assertTrue((boolean) lastUpdateDownloadStatusField.get(updateConfig));
+
+            updateConfig.setLastUpdateDownloadStatus(false);
+            assertFalse((boolean) lastUpdateDownloadStatusField.get(updateConfig));
+        } catch (NoSuchFieldException e) {
```
dariusf: These exceptions can go into the method signature I think. Makes the test cleaner.

src/main/java/prefs/UpdateConfig.java
HansNewbie: If the last download was successful. So if the last download was cut off halfway (connection lost or user close the HT suddenly), this will be `false`.

src/main/java/prefs/UpdateConfig.java
HansNewbie: The warning is that it can be made `final`, but if it is final, gson can't do conversion. Should I add a comment about it?

HansNewbie:
Made the changes.

src/main/java/prefs/UpdateConfig.java
dariusf: Sure, something brief.

src/main/java/util/JsonFileConverter.java
```diff
@@ -0,0 +1,85 @@
...
+        gson.toJson(object, object.getClass(), writer);
+        writer.close();
+    }
+
+    /**
+     * Loads an object from a JSON file
+     * @param objectClass class of the object to be loaded; used to parse the JSON
+     * @return The object from JSON file, or Optional.empty() if the file can't be parsed
+     */
+    public Optional<Object> loadFromFile(Class<?> objectClass) {
```
dariusf: Could we use `public <T> Optional<T> loadFromFile(Class<T> objectClass) {`? Then we don't have to cast down at use sites.

src/test/java/tests/UpdateDownloadLinkTest.java
```diff
@@ -0,0 +1,196 @@
...
+    private static final String MALFORMED_URL_MESSAGE = "URL is not well formed.";
+
+    @Test
+    public void updateDownloadLinkCompareTo_sameVersionDiffLocation_sameObject() {
+        Version version = new Version(1, 0, 0);
+        UpdateDownloadLink a = new UpdateDownloadLink();
+        a.setVersion(version);
+        try {
+            a.setApplicationFileLocation(new URL("http://google.com"));
+        } catch (MalformedURLException e) {
```
dariusf: These can be `throws` clauses, since the URL is set by us. They'll make the test fail.

src/test/java/tests/UpdateDownloadLinkTest.java
```diff
@@ -0,0 +1,196 @@
...
+
+        try {
+            Class<?> updateDownloadLinkClass = a.getClass();
+
+            Field versionField =
+                    updateDownloadLinkClass.getDeclaredField("version");
+            versionField.setAccessible(true);
+
+            versionField.set(a, version);
+        } catch (NoSuchFieldException e) {
```
dariusf: Same here, these don't have to be caught.

src/main/java/util/JsonFileConverter.java
HansNewbie: Do you mean `public Optional<T> ...`?

dariusf: I think the type parameter needs to be declared before it's used.

HansNewbie: Ok. Will try it out.

HansNewbie:
@dariusf I need another round of review since I changed a behaviour (now JarUpdater will not remove backup).

HubTurboUpdate.json
```diff
@@ -0,0 +1,12 @@
+{
+  "hubturboVersionsDownloadLink": [
```
dariusf: This is really just cosmetic, but can this be something shorter, like `downloadLinks`? We know that it's related to HT versions.

src/main/java/updater/JarUpdater.java
```diff
@@ -177,27 +185,27 @@ private void moveFileToBackup(File file) throws IOException {
...
+     * @return backup filename
      */
-    private void removeBackupOfFileIfExists(File file) throws IOException {
-        log("Removing backup");
-        String backupFilename = getBackupFilename(file.getName());
-        File backup = new File(file.getParent(), backupFilename);
-        if (!Files.deleteIfExists(backup.toPath())) {
-            log(String.format("No backup file %s", file.getName()));
+    private String getJarBackupFilename(String filename) {
+        Pattern jarFilenamePattern = Pattern.compile("^(.*)\\.(jar|JAR)$");
```
dariusf: The case insensitivity can be handled with `Pattern.compile("regex", Pattern.CASE_INSENSITIVE)`.

src/main/java/updater/UpdateData.java
```diff
@@ -0,0 +1,43 @@
...
+
+        // List the update link in descending order of version
+        Collections.sort(hubturboVersionsDownloadLink, Collections.reverseOrder());
+
+        // Get link of version that has same major version or just 1 major version up than current
+        Optional<UpdateDownloadLink> updateLink = hubturboVersionsDownloadLink.stream()
+                .filter(link -> Version.isVersionMajorSameOrJustOneGreaterFromCurrent(link.getVersion())).findFirst();
+
+        if (updateLink.isPresent()) {
+            return updateLink;
```
dariusf: We can just return updateLink here. If it's absent, the result will be empty anyway.

src/main/java/updater/UpdateManager.java
```diff
@@ -123,6 +157,57 @@ private boolean extractJarUpdater() {
...
+            logger.info("Deleting " + allHtBackupFiles.get(i).getName());
+            if (!allHtBackupFiles.get(i).delete()) {
+                logger.warn("Failed to delete old HT backup file " + allHtBackupFiles.get(i).getName());
+            }
+        }
+    }
+
+    private Comparator<File> getHtBackupFileComparatorByVersion() {
+        return (a, b) -> {
+            Pattern htJarFileBackupPattern = Pattern.compile(HT_BACKUP_FILENAME_PATTERN_STRING);
```
dariusf: Maybe the regex matching can be factored out into another method, so it's a bit more SLAP.

damithc: Yes, any regex matching should be extracted out and tested separately, if possible.

src/main/java/updater/UpdateManager.java
```diff
@@ -123,6 +157,63 @@ private boolean extractJarUpdater() {
...
+     * Get version of HubTurbo from Jar backup file (filename is "HubTurbo_V[major].[minor].[patch].jar")
+     * @param filename filename of HT backup JAR
+     * @return version of HT of backup JAR, or Version 0 if filename not matching backup file
+     */
+    private Version getVersionOfHtBackupFileFromFilename(String filename) {
+        Pattern htJarFileBackupPattern = Pattern.compile(HT_BACKUP_FILENAME_PATTERN_STRING);
+        Matcher htJarFileBackupMatcher = htJarFileBackupPattern.matcher(filename);
+        if (!htJarFileBackupMatcher.find()) {
+            return new Version(0, 0, 0);
+        }
```
HansNewbie: I am not sure about this. If the regex fails, should it throw exception? I return Version 0 here because if the backup Jar has non-valid version, might as well treat it as very old (so it will be deleted).

Should I put condition in the header comment instead of the acceptable input, and put assertion failure here instead?

I will add test cases after this.

dariusf: If this should never happen (programmer error), I think it's fine to assert false. We can return version 0 in case, but that behaviour shouldn't be relied on. Maybe we can add to the comment also to say why version 0 is returned.

HansNewbie: I will put assert false.

HansNewbie:
@dariusf ready for review

src/main/java/prefs/UpdateConfig.java
damithc: Then the name should be `isLastUpdateSuccessful` or something like that?

src/main/java/prefs/UpdateConfig.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public UpdateConfig() {
+        versionsPreviouslyDownloaded = new ArrayList<>();
+    }
+
+    /**
+     * Sets last update download status
+     * @param status true if update download is successful, false otherwise
+     */
+    public void setLastUpdateDownloadStatus(boolean status) {
```
damithc: Another boolean variable not properly named.

src/main/java/prefs/UpdateConfig.java
```diff
@@ -0,0 +1,45 @@
...
+        this.lastUpdateDownloadStatus = status;
+    }
+
+    public boolean getLastUpdateDownloadStatus() {
+        return this.lastUpdateDownloadStatus;
+    }
+
+    /**
+     * Add a version to downloaded versions
```
damithc: Adds

src/main/java/prefs/UpdateConfig.java
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public void addToVersionPreviouslyDownloaded(Version downloadedVersion) {
+        versionsPreviouslyDownloaded.add(downloadedVersion);
+    }
+
+    /**
+     * Checks if a version has been downloaded previously
+     */
+    public boolean checkIfVersionWasPreviouslyDownloaded(Version version) {
```
damithc: tsk tsk tsk :-| `wasPreviouslyDownloaded(...)`

src/main/java/updater/JarUpdater.java
```diff
@@ -177,27 +185,27 @@ private void moveFileToBackup(File file) throws IOException {
...
+        String backupFilename = getJarBackupFilename(file.getName());
         File backup = new File(file.getParent(), backupFilename);
         JarUpdaterFileHandler.moveFileWithReplaceExisting(backup.toPath(), file.toPath());
     }
 
     /**
-     * Removes the backup file of a given filename
-     * @param file the file which backup should be removed
-     * @throws IOException if failed to remove backup file
+     * Get backup name of jar file.
```
damithc: Gets

src/main/java/updater/JarUpdater.java
```diff
@@ -177,27 +185,27 @@ private void moveFileToBackup(File file) throws IOException {
...
+     * If backup suffix option is provided, it will be used instead of standard backup suffix.
+     * If filename does not end with ".jar", append standard backup suffix to the filename
```
damithc: appends

src/main/java/updater/UpdateData.java
```diff
@@ -0,0 +1,37 @@
...
+public class UpdateData {
+
+    private List<UpdateDownloadLink> downloadLinks; // NOPMD - not made final for gson
+
+    public UpdateData() {
+        downloadLinks = new ArrayList<>();
+    }
+
+    /**
+     * Get UpdateDownloadLink for update.
```
damithc: Gets

src/main/java/updater/UpdateDownloadLink.java
```diff
@@ -0,0 +1,52 @@
+package updater;
+
+import util.Version;
+
+import java.net.URL;
+
+/**
+ * Represents a HT version with its download link
```
damithc: Not clear what this class represents.

src/main/java/updater/UpdateDownloadLink.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public class UpdateDownloadLink implements Comparable<UpdateDownloadLink> {
```
damithc: Class name sounds like a method.

HansNewbie: Would `HtDownloadLink` be better?

damithc: Yes, that sounds less like a method

damithc:
I reviewed the first part. Revise the rest accordingly and update here when done.

HansNewbie:
Noted.

HansNewbie:
@damithc revised.

src/main/java/prefs/UpdateConfig.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Sets last app update download status
+     * @param isLastAppUpdateDownloadSuccessful true if update download is successful, false otherwise
+     */
+    public void setLastAppUpdateDownloadSuccessful(boolean isLastAppUpdateDownloadSuccessful) {
+        this.isLastAppUpdateDownloadSuccessful = isLastAppUpdateDownloadSuccessful;
+    }
+
+    public boolean getLastAppUpdateDownloadSuccessful() {
```
damithc: This method too can be named in the same way `isLast...`

src/main/java/updater/UpdateManager.java
```diff
@@ -97,12 +131,15 @@ private boolean initUpdate() {
             return false;
         }
 
+        updateConfig.setLastAppUpdateDownloadSuccessful(false);
+        saveUpdateConfig();
+
         return extractJarUpdater();
     }
 
     private boolean extractJarUpdater() {
```
damithc: See if you can exceptions instead of booleans to indicate exceptional situations.

damithc:
Can merge after addressing comments.

src/main/java/updater/UpdateManager.java
```diff
@@ -134,29 +134,30 @@ private boolean initUpdate() {
...
         logger.info("Extracting jarUpdater");
         File jarUpdaterFile = new File(JAR_UPDATER_APP_PATH);
 
-        try {
-            jarUpdaterFile.createNewFile();
-        } catch (IOException e) {
-            logger.error("Can't create empty file for jarUpdater");
-            return false;
+        if (!jarUpdaterFile.exists()) {
+            assert jarUpdaterFile.createNewFile();
```
damithc: I'm not sure assert is the right thing to do here. File system problems are not programmer problems.

damithc: Besides, you should never use asserts in combination with actual work. Asserts can be disabled.

damithc:
Write better commit messages

src/main/java/updater/UpdateManager.java
HansNewbie: Alright. I was trying to prevent findBugs warning that we must process `createNewFile` result (returns true if file was created or false if file already exists, and throws IOException if IO error occurs), which is usually done by asserting. Will think of something else.

src/main/java/updater/UpdateManager.java
```diff
@@ -97,29 +132,93 @@ private boolean initUpdate() {
...
+    }
+
+    private void extractJarUpdater() throws IOException {
+        logger.info("Extracting jarUpdater");
+        File jarUpdaterFile = new File(JAR_UPDATER_APP_PATH);
+
+        if (!jarUpdaterFile.exists() && !jarUpdaterFile.createNewFile()) {
+            logger.error("Failed to create Jar Updater empty file");
+            throw new IOException("Failed to create Jar Updater empty file");
+        }
```
HansNewbie: Made it to throw exception if cannot create file when it does not exist.

damithc: Make sure this exception is caught and handled properly too.

src/test/java/tests/VersionTest.java
```diff
@@ -0,0 +1,117 @@
...
+        // normal values
+        version = new Version(4, 10, 5);
+        assertEquals("V4.10.5", version.toString());
+
+        // big numbers
+        version = new Version(100, 100, 100);
+        assertEquals("V100.100.100", version.toString());
+    }
+
+    //@Test
```
damithc: ?

src/main/java/updater/UpdateManager.java
HansNewbie: Yes, it is caught and handled properly.

src/test/java/tests/VersionTest.java
HansNewbie: I must have mistyped it.

damithc:
Can merge after addressing latest comments.

PR #1340, opened 2016-03-12
F1 key doesn't work when the focus is in a filter box #1167

### LINKED ISSUE #1167: F1 key doesn't work when the focus is in a filter box

YijinL:
Fixes #1167.

**REVIEW**

**Current status**:
When filter text box is in focus, pressing <kbd>F1</kbd> redirects the browser component to the filter docs page.

**Work done**: 
- ~~`FilterPanel` now passes its `UI` reference to `FilterTextField`~~
- ~~`FilterTextField` now maintains a final field to hold the `UI`reference~~
- ~~`FilterTextField` setup method now includes handling for the `KeyboardShortcuts.SHOW_DOCS` key combination (<kbd>F1</kbd>), which calls UI's browser's`BrowserComponent.showDocs()` method.~~
- ~~The above handling code is copied from a [slice](https://github.com/HubTurbo/HubTurbo/blob/880d65c081505c9e8a547c29d188293a56c5de24/src/main/java/ui/listpanel/ListPanel.java#L171-L173) of `ListPanel.setupKeyboardShorcuts()`~~
- Make F1 link to 'filters' section of docs
- Change implementation: expose help key press callback hook on `FilterTextField` to decouple from rest of app
- All instantiations of `FilterTextField` now register the correct callback for help key press
- Regression testing under `guitests.FilterTextFieldTest`

**Future**:
null

### DIFF
--- a/src/main/java/ui/components/FilterTextField.java
+++ b/src/main/java/ui/components/FilterTextField.java
@@ -16,6 +16,8 @@ import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import static ui.components.KeyboardShortcuts.SHOW_DOCS;
+
 /**
  * A field augmented with the ability to revert edits, autocompletion, on-the-fly error
  * checking, and callbacks for various actions.
@@ -32,8 +34,9 @@ public class FilterTextField extends TextField {
     public static final String INVALID_FILTER_STYLE = "-fx-control-inner-background: #EE8993";
 
     // Callback functions
-    private Runnable cancel = () -> {};
-    private Function<String, String> confirm = (s) -> s;
+    private Runnable onCancel = () -> {};
+    private Runnable onShowDocs = () -> {};
+    private Function<String, String> onConfirm = (s) -> s;
 
     // For reverting edits
     private String previousText;
@@ -43,6 +46,7 @@ public class FilterTextField extends TextField {
 
     // For on-the-fly parsing and checking
     private final ValidationSupport validationSupport = new ValidationSupport();
+
 
     public FilterTextField(String initialText) {
         super(initialText);
@@ -81,10 +85,10 @@ public class FilterTextField extends TextField {
             }
         });
         setOnKeyPressed(e -> {
-             if (e.getCode() == KeyCode.TAB) {
+            if (e.getCode() == KeyCode.TAB) {
                  // Disable tab for UI traversal
                  e.consume();
-             }
+            }
         });
         setOnKeyReleased(e -> {
             e.consume();
@@ -92,10 +96,15 @@ public class FilterTextField extends TextField {
                 confirmEdit();
             } else if (e.getCode() == KeyCode.ESCAPE) {
                 if (getText().equals(previousText)) {
-                    cancel.run();
+                    onCancel.run();
                 } else {
                     revertEdit();
                 }
+            }
+        });
+        addEventHandler(KeyEvent.KEY_PRESSED, event -> {
+            if (SHOW_DOCS.match(event)) {
+                onShowDocs.run();
             }
         });
     }
@@ -226,11 +235,11 @@ public class FilterTextField extends TextField {
     }
 
     /**
-     * Commits the current contents of the field. This triggers its 'confirm' callback.
+     * Commits the current contents of the field. This triggers its 'onConfirm' callback.
      */
     private void confirmEdit() {
         previousText = getText();
-        String newText = confirm.apply(getText());
+        String newText = onConfirm.apply(getText());
         int caretPosition = getCaretPosition();
         setText(newText);
         positionCaret(caretPosition);
@@ -245,21 +254,30 @@ public class FilterTextField extends TextField {
     }
 
     /**
-     * Sets the 'cancel' callback, which will be called with Esc is pressed and there is
+     * Sets the 'onCancel' callback, which will be called with Esc is pressed and there is
      * no edit to revert.
      */
-    public FilterTextField setOnCancel(Runnable cancel) {
-        this.cancel = cancel;
+    public FilterTextField setOnCancel(Runnable onCancel) {
+        this.onCancel = onCancel;
         return this;
     }
 
     /**
-     * Sets the 'confirm' callback, which will be called when Enter is pressed, or when the
+     * Sets the 'onConfirm' callback, which will be called when Enter is pressed, or when the
      * contents of the field are manually set with {@link #setFilterText}. The callback will
      * be passed the current contents of the field.
      */
-    public FilterTextField setOnConfirm(Function<String, String> confirm) {
-        this.confirm = confirm;
+    public FilterTextField setOnConfirm(Function<String, String> onConfirm) {
+        this.onConfirm = onConfirm;
+        return this;
+    }
+    
+    /**
+     * Sets the 'onShowDocs' callback, which will be called when the user triggers the show_docs
+     * keyboard shortcut event.
+     */
+    public FilterTextField setOnShowDocs(Runnable onShowDocs) {
+        this.onShowDocs = onShowDocs;
         return this;
     }
 
--- a/src/test/java/guitests/FilterTextFieldTest.java
+++ b/src/test/java/guitests/FilterTextFieldTest.java
@@ -3,15 +3,26 @@ package guitests;
 import static junit.framework.TestCase.assertTrue;
 
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.junit.Test;
 
 import javafx.scene.input.KeyCode;
+import org.loadui.testfx.utils.FXTestUtils;
 import ui.TestController;
+import ui.UI;
 import ui.components.FilterTextField;
 import ui.issuepanel.FilterPanel;
+import util.GitHubURL;
+import util.events.testevents.NavigateToPageEventHandler;
 
 public class FilterTextFieldTest extends UITest {
+
+    @Override
+    public void launchApp() {
+        FXTestUtils.launchApp(
+            TestUI.class, "--test=true", "--bypasslogin=true", "--testchromedriver=true");
+    }
 
     @Test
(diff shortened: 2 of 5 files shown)


dariusf:
Some preliminary comments: this couples FilterTextField to the rest of the app, which isn't so good. A better approach would be to expose some kind of callback which lets us hook into its key event.

```java
FilterTextField field = new FilterTextField();
field.setOnHelp(something::showDocs);
// 'Help' being the high-level name for the F1 key press; can be changed
```

I think linking to the filter docs is a good idea also.

damithc:
@YijinL any updates to this PR?

YijinL:
@damithc 
Pressing F1 key when filter text box is in focus now redirects the browser to the filter docs page.
Now working on refactoring the implementation to reduce coupling

damithc:
Noted. Let us know when it is ready for review.

YijinL:
@dariusf implemented following precedent set by FilterTextField.setOnConfirm

src/main/java/ui/components/FilterTextField.java
```diff
@@ -32,7 +34,8 @@
     private static final String INVALID_FILTER_STYLE = "-fx-control-inner-background: #EE8993";
 
     // Callback functions
-    private Runnable cancel = () -> {};
+    private Runnable onCancel = () -> {};
+    private Runnable onHelp = () -> {};
```
dariusf: This is a good change. Maybe also rename `confirm` while you're at it?

src/main/java/ui/components/FilterTextField.java
```diff
@@ -264,6 +273,15 @@ public FilterTextField setOnConfirm(Function<String, String> confirm) {
     }
 
     /**
+     * Sets the 'onHelp' callback, which will be called when the key combination defined by
+     * KeyboardShortcuts.SHOW_HELP is pressed with this object in focus.
+     */
+    public FilterTextField setOnHelpKeyCombination(Runnable onHelp) {
```
dariusf: Don't think we need to talk about how the method is used in here. The method name can be a bit shorter also, since the help action is not necessarily triggered by a key combination -- that just happens to be the current implementation.

YijinL:
@dariusf Done as suggested, anything else?

src/main/java/ui/components/FilterTextField.java
```diff
@@ -245,21 +254,26 @@ public void setFilterText(String text) {
...
      * be passed the current contents of the field.
      */
-    public FilterTextField setOnConfirm(Function<String, String> confirm) {
-        this.confirm = confirm;
+    public FilterTextField setOnConfirm(Function<String, String> onConfirm) {
+        this.onConfirm = onConfirm;
+        return this;
+    }
+
+    public FilterTextField setOnHelp(Runnable onHelp) {
```
dariusf: This can have a similar javadoc comment as the method above.

src/test/java/guitests/FilterTextFieldTest.java
```diff
@@ -121,6 +128,20 @@ public void cancelTest() {
         awaitCondition(() -> toggle.get() % 2 == 1);
     }
 
+    @Test
+    public void helpTest() {
+        UI.events.registerEvent((NavigateToPageEventHandler) e -> url = e.url);
```
dariusf: Could you use an `AtomicReference<String>` for this instead? So it's kept local to the method.

src/test/java/guitests/FilterTextFieldTest.java
```diff
@@ -121,6 +128,20 @@ public void cancelTest() {
...
+        getFirstPanelField();
+        push(KeyCode.F1);
+        sleep(1000);
```
dariusf: We can't know that this will take exactly one second. We should wait until the condition is true using `waitAndAssertEquals`.

dariusf:
We should try to get coverage back up. Apart from that the implementation is good. Just a few minor changes suggested.

YijinL:
[ChromeDriverTest](https://github.com/HubTurbo/HubTurbo/blob/8f1ae3aed9dc35c2201e07807d5284b7d9e4f9b1/src/test/java/guitests/ChromeDriverTest.java#L66-L74) uses [this 1000ms duration](https://github.com/HubTurbo/HubTurbo/blob/8f1ae3aed9dc35c2201e07807d5284b7d9e4f9b1/src/test/java/guitests/ChromeDriverTest.java#L21) for event delay, should that one be changed too?

dariusf:
Yes, it should. That's an old test. If you want do it now, you can, otherwise we can make a new issue.

The 5% drop is expected. Could you just check that the previous 0.03% drop is fixed? Should be fine after that.

src/main/java/ui/components/FilterTextField.java
```diff
@@ -245,21 +254,29 @@ public void setFilterText(String text) {
...
      */
-    public FilterTextField setOnConfirm(Function<String, String> confirm) {
-        this.confirm = confirm;
+    public FilterTextField setOnConfirm(Function<String, String> onConfirm) {
+        this.onConfirm = onConfirm;
+        return this;
+    }
+    
+    /**
+     * Sets the 'onHelp' callback, which will be called when F1 is pressed.
```
damithc: Perhaps we shouldn't mention when this callback is used? A method should be described independent how it is used.

src/test/java/guitests/FilterTextFieldTest.java
```diff
@@ -121,6 +127,20 @@ public void cancelTest() {
         awaitCondition(() -> toggle.get() % 2 == 1);
     }
 
+    @Test
+    public void helpTest() {
```
damithc: Follow our new test method naming convention (given in java coding standard)

YijinL: OK. I followed the local naming conventions from the other tests in this file, so I will change them to verb form as well.
Should I spawn a new issue for the name refactoring, or should I just do it in this PR?

damithc: Let's write new tests following the new convention. Old ones can be refactored at a later time.

YijinL: Actually it's a really small change, how about I leave this and you get a free `forFirstTimers` issue? I can help create that issue after this PR gets merged.

src/test/java/guitests/FilterTextFieldTest.java
```diff
@@ -121,6 +127,20 @@ public void cancelTest() {
...
+        AtomicReference<String> url = new AtomicReference<>();
+        UI.events.registerEvent((NavigateToPageEventHandler) e -> url.set(e.url));
+        getFirstPanelField();
+        push(KeyCode.F1);
+        waitAndAssertEquals(GitHubURL.FILTERS_PAGE, url::get);
+    }
+    @Override
```
damithc: Add a blank line between methods?

src/main/java/ui/components/FilterTextField.java
YijinL: But if we don't mention what event we are hooking the callback on, new developers will have to trace the source to be sure. Or maybe I misunderstood you?

YijinL:
@dariusf Strange all I did was merge `master` into this branch. When coveralls says `coverage increased/decreased to X`, what baseline are they comparing it to?

src/main/java/ui/components/FilterTextField.java
damithc: This method is about binding `Help` to it's implementation, not binding `F1` to help, right?
Another way to look at it is, if we change the Help short cut to `F2` at some point, it should not require a change to this code, and if the code doesn't change, the comment shouldn't change either.
So, is there a better place to explain `F1` is bound to `Help`?

src/test/java/guitests/FilterTextFieldTest.java
```diff
@@ -121,6 +127,21 @@ public void cancelTest() {
         awaitCondition(() -> toggle.get() % 2 == 1);
     }
 
+    @Test
+    public void testHelp() {
```
damithc: What method or feature are we testing? The method name should be exactly that (i.e. omit `test` part as it is already captured in `@Test`)

src/main/java/ui/components/FilterTextField.java
YijinL: Developers can see the implementation at [`ui.components.KeyboardShortcuts`](https://github.com/HubTurbo/HubTurbo/blob/6e0d10092066fe64fa0415614eba927892fee349/src/main/java/ui/components/KeyboardShortcuts.java#L81-L82).

I see, I will change it to something like:
```java
/**
 * Sets the 'onHelp' callback, which will be called when the user triggers the show_docs keyboard shortcut
 */
```
I suppose line 257: `... called with Esc is pressed ...` 
and line 266: `... called when Enter is pressed ...`
can be left as is since the <kbd>Esc</kbd> and <kbd>Enter</kbd> keybinds are used specially in this class.

src/test/java/guitests/FilterTextFieldTest.java
YijinL: Understood, I will favor the official coding standards over file-local conventions.

YijinL:
Triggering travis rebuild (2 tests failed)

YijinL:
Requesting Travis rebuild, 1 test failed. Seems to be [this](https://travis-ci.org/HubTurbo/HubTurbo/builds/117256446#L3000) one.

dariusf:
I've restarted the build. You can do so yourself by opening/closing the issue, or pushing commits (which you can remove later). If you see specific tests failing frequently, make issues for them.

The baseline is the coverage of the 'merge base' (the common ancestor of this branch and the branch you are merging into).

YijinL:
Noted. Everything seems fine now, thanks!

dariusf:
Do comment when you're done squashing/updating to latest master.

YijinL:
Done.

dariusf:
Thanks!

PR #1342, opened 2016-03-13
Write test for the changing of the background colour of filter text field #1311

### LINKED ISSUE #1311: Write test for the changing of the background colour of filter text field
Tests for #1310

garbanzos:
Fixes #1311.

### DIFF
--- a/src/test/java/guitests/FilterTextFieldTest.java
+++ b/src/test/java/guitests/FilterTextFieldTest.java
@@ -13,34 +13,73 @@ import ui.issuepanel.FilterPanel;
 
 public class FilterTextFieldTest extends UITest {
 
-    private void testCompletions(FilterTextField field) {
-        // Basic completion
-        clearField();
-        type("cou").push(KeyCode.TAB);
-        awaitCondition(() -> field.getText().equals("count"));
+    @Test
+    public void inputsTest() {
+        FilterTextField field = getFirstPanelField();
+        testCompletions(field);
+        testSpaces(field);
+    }
 
-        // Completion does not only work for alternating keys typed
-        clearField();
-        type("c");
-        awaitCondition(() -> field.getSelectedText().equals("losed"));
-        type("l");
-        awaitCondition(() -> field.getSelectedText().equals("osed"));
-        type("o");
-        awaitCondition(() -> field.getSelectedText().equals("sed"));
-        type(KeyCode.TAB);
-        awaitCondition(() -> field.getSelectedText().equals(""));
-        awaitCondition(() -> field.getText().equals("closed"));
+    @Test
+    public void revertTest() {
+        FilterTextField field = getFirstPanelField();
 
-        // Completion with selection
         clearField();
-        type("cou").push(KeyCode.TAB);
-        push(KeyCode.LEFT);
-        for (int i = 0; i < 3; i++) {
-            field.selectBackward();
-        }
-        // c[oun]t
-        type("lo").push(KeyCode.TAB); // 'c' + 'lo' is a prefix of 'closed'
-        awaitCondition(() -> field.getText().equals("closedt"));
+        type("assi").push(KeyCode.TAB);
+        push(KeyCode.ENTER);
+
+        clearField();
+        type("test");
+        awaitCondition(() -> field.getText().equals("test"));
+
+        push(KeyCode.ESCAPE);
+
+        awaitCondition(() -> field.getText().equals("assignee"));
+    }
+
+    @Test
+    public void cancelTest() {
+        FilterTextField field = getFirstPanelField();
+        AtomicInteger toggle = new AtomicInteger(0);
+        field.setOnCancel(toggle::getAndIncrement);
+
+        assertTrue(toggle.get() % 2 == 0);
+
+        type("mile").push(KeyCode.TAB);
+        push(KeyCode.ENTER);
+
+        clearField();
+        type("test");
+        awaitCondition(() -> field.getText().equals("test"));
+
+        push(KeyCode.ESCAPE);
+        awaitCondition(() -> toggle.get() % 2 == 0);
+
+        push(KeyCode.ESCAPE);
+        awaitCondition(() -> toggle.get() % 2 == 1);
+    }
+
+    @Test
+    public void filterTextFieldColor_validFilter_validFilterStyleApplied(){
+        testValidFilterStyleApplied("is:open");
+        testValidFilterStyleApplied("-is:open");
+        testValidFilterStyleApplied("hello");
+        testValidFilterStyleApplied("is:open & has:assignee");
+        testValidFilterStyleApplied("(is:closed)");
+    }
+
+    @Test
+    public void filterTextFieldColor_invalidFilter_invalidFilterStyleApplied(){
+
+        // Tests if parse errors applies invalid filter style to text field
+        testInvalidFilterStyleApplied("is:-open");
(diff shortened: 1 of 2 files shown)


garbanzos:
@dariusf Could you help review this?

src/test/java/guitests/FilterTextFieldTest.java
```diff
@@ -13,68 +13,8 @@
...
-        clearField();
-        type("assignee ");
-        push(KeyCode.LEFT, 2);
-        type(" ");
-        push(KeyCode.LEFT, 8);
-        type(" ");
-        awaitCondition(() -> field.getText().equals(" assigne e "));
-    }
+    private static final String VALID_FILTER_STYLE = "-fx-control-inner-background: white";
+    private static final String INVALID_FILTER_STYLE = "-fx-control-inner-background: #EE8993";
```
dariusf: Rather than redefine these, maybe we could use the existing constants from FilterTextField?

src/test/java/guitests/FilterTextFieldTest.java
```diff
@@ -121,6 +62,96 @@ public void cancelTest() {
         awaitCondition(() -> toggle.get() % 2 == 1);
     }
 
+    @Test
+    public void filterTextFieldColor_validFilter_validFilterStyleApplied(){
+        FilterTextField field = getFirstPanelField();
+
+        clearField();
+        type("is:open");
+        assertTrue(field.getStyle().contains(VALID_FILTER_STYLE));
```
dariusf: Maybe we can factor out the clear, type, assert pattern here, with a function that just takes the string to test.

src/test/java/guitests/FilterTextFieldTest.java
```diff
@@ -121,6 +62,96 @@ public void cancelTest() {
...
+        clearField();
+        type("is:-open");
+        assertTrue(field.getStyle().contains(INVALID_FILTER_STYLE));
+
+        clearField();
+        type("--");
+        assertTrue(field.getStyle().contains(INVALID_FILTER_STYLE));
+
+        clearField();
+        type("is:-open & has:assignee");
```
dariusf: This looks kind of like the first one. Does this cover anything new?

src/test/java/guitests/FilterTextFieldTest.java
```diff
@@ -121,6 +62,96 @@ public void cancelTest() {
...
+        assertTrue(field.getStyle().contains(INVALID_FILTER_STYLE));
+
+        clearField();
+        type("(is:closed");
```
dariusf: Is this the same as just `(`? It helps if we can keep the examples minimal, so it's clear what the problem is.

src/test/java/guitests/FilterTextFieldTest.java
```diff
@@ -121,6 +62,96 @@ public void cancelTest() {
...
+
+        // Tests if semantic errors applies invalid filter style to text field
+        // These invalid filters will cause the invalid filter style be applied after ENTER is pressed
+        clearField();
+        type("is:invalid").push(KeyCode.ENTER);
+        assertTrue(field.getStyle().contains(INVALID_FILTER_STYLE));
+
+        clearField();
+        type("is:invalid & hello").push(KeyCode.ENTER);
+        assertTrue(field.getStyle().contains(INVALID_FILTER_STYLE));
```
dariusf: This looks like the same as the above. Both fail at `is`?

src/test/java/guitests/FilterTextFieldTest.java
```diff
@@ -121,6 +62,96 @@ public void cancelTest() {
...
+    }
+
+    private void testSpaces(FilterTextField field) {
```
dariusf: Were there any changes to this and the one below? Can't tell since the whole thing is in the diff.

src/test/java/guitests/FilterTextFieldTest.java
garbanzos: @dariusf Is `hello & is:valid` supposed to throw an exception?

dariusf: Yes, a semantic exception, because `valid` isn't a valid argument to `is`.

src/test/java/guitests/FilterTextFieldTest.java
garbanzos: Nope. I just moved all the private methods to the end of the class.

src/test/java/guitests/FilterTextFieldTest.java
garbanzos: @dariusf 
`is:invalid & hello` throws an semantic exception but
`hello & is:invalid` doesn't seem to be throwing any semantic exceptions. 

Is this a bug?

garbanzos:
@dariusf 
`is:invalid & hello` throws an semantic exception but
`hello & is:invalid` doesn't seem to be throwing any semantic exceptions. 

Is this a bug?

src/test/java/guitests/FilterTextFieldTest.java
```diff
@@ -121,6 +59,64 @@ public void cancelTest() {
...
+        testValidFilterStyleApplied("(is:closed)");
+    }
+
+    @Test
+    public void filterTextFieldColor_invalidFilter_invalidFilterStyleApplied(){
+
+        // Tests if parse errors applies invalid filter style to text field
+        testInvalidFilterStyleApplied("is:-open");
+        testInvalidFilterStyleApplied("--");
+        testInvalidFilterStyleApplied("is:-open & has:assignee");
```
dariusf: Hmm, think I pointed this out last time. Does this cover anything that the first one doesn't? If not it shouldn't be needed.

We're only testing that a parse error affects the field, so we don't really need that many examples.

src/test/java/guitests/FilterTextFieldTest.java
```diff
@@ -121,6 +59,64 @@ public void cancelTest() {
...
+        testInvalidFilterStyleApplied("(");
+        testInvalidFilterStyleApplied("incomplete &");
+
+        // Tests if semantic errors applies invalid filter style to text field
+        // These invalid filters will cause the invalid filter style be applied after ENTER is pressed
+        testInvalidFilterStyleAppliedAfterEnter("is:invalid");
+        testInvalidFilterStyleAppliedAfterEnter("is:invalid & hello");
```
dariusf: This too. The `& hello` looks unnecessary, as there's no problem there.

garbanzos:
@dariusf I removed the tests you mentioned.

src/test/java/guitests/FilterTextFieldTest.java
```diff
@@ -121,6 +59,62 @@ public void cancelTest() {
...
+        testInvalidFilterStyleApplied("--");
+        testInvalidFilterStyleApplied("(");
+        testInvalidFilterStyleApplied("incomplete &");
+
+        // Tests if semantic errors applies invalid filter style to text field
+        // These invalid filters will cause the invalid filter style be applied after ENTER is pressed
+        testInvalidFilterStyleAppliedAfterEnter("is:invalid");
+    }
+
+    private void testSpaces(FilterTextField field) {
```
damithc: Is this a test? Why is the `@Test` annotation missing?

garbanzos: `testSpaces` is a not a test. It is a helper method for `inputsTest`.

PR #1347, opened 2016-03-14
Create a basic Issue Picker #1317

### LINKED ISSUE #1317: Create Issue Picker to choose issue easily
## UI Design 
- text field for searching 
- suggestion menu that will pop-up when typing

jinified:
Fixes #1317

## How it works
ezgif com-video-to-gif 1

Click [here](https://github.com/jinified/HubTurbo/blob/1317-issue-picker/docs/issue_picker.md) for full documentation

TODO
- [x] stop filtering when if the user input is too short
- [x] sets a fix height for scroll pane to avoid the buttons from getting out of the screen


jinified:
@dariusf Ready for review

damithc:
Can post a screen capture?

src/main/java/backend/resource/TurboIssue.java
```diff
@@ -337,6 +338,28 @@ public TurboIssue combineWithPullRequest(PullRequest pullRequest) {
...
+    /**
+     * Matching is done by matching all words separated by space in query
+     * @param issues
+     * @param query
+     * @return list of issues that contains the query
+     */
+    public static List<TurboIssue> getMatchedIssues(List<TurboIssue> issues, String query){
+        List<String> queries = Arrays.asList(query.split("\\s"));
+        return issues.stream()
+            .filter(i -> Utility.containsIgnoreCaseMultipleWords(i.getTitle() + " " + i.getId(), queries))
```
dariusf: Maybe we could replace this long function name with an overload instead? Kind of like extending the `contains` operation to lists.

src/main/java/ui/components/issuepicker/IssueCard.java
```diff
@@ -0,0 +1,176 @@
...
+import javafx.scene.layout.HBox;
+import javafx.scene.layout.VBox;
+import javafx.scene.paint.Color;
+import javafx.scene.text.Text;
+import ui.GuiElement;
+
+/**
+ * Represents an individual issue card not linked to any panel
+ */
+public class IssueCard extends VBox {
```
dariusf: How much of this is shared with ListPanelCard? Can we factor out the common parts and share as much as possible? So it doesn't get out of sync.

src/main/java/ui/components/issuepicker/IssuePicker.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public IssuePicker(UI ui, Stage stage) {
+        this.stage = stage;
+        ui.registerEvent((ShowIssuePickerEventHandler) e -> 
+            Platform.runLater(() -> showIssuePicker(ui.logic.getModels())));
+    }
+
+    private Optional<List<String>> showIssuePicker(MultiModel models) {
+        IssuePickerDialog issuePickerDialog = new IssuePickerDialog(stage, models);
+        // show IssuePickerDialog and wait for result
```
dariusf: Don't think we need this comment.

src/main/java/ui/components/issuepicker/IssuePickerDialog.java
```diff
@@ -0,0 +1,250 @@
...
+        // Ensures dialog pane resize according to content
+        getDialogPane().getScene().getWindow().sizeToScene();
+    }
+
+    private void populateSelectedIssues(List<TurboIssue> chosenIssues, Optional<TurboIssue> currentSuggestion) {
+        selectedIssues.getChildren().clear();
+        Map<String, FlowPane> repoContent = getRepoContent(chosenIssues, currentSuggestion);
+
+        repoContent.entrySet().forEach(entry -> {
+            entry.getValue().getChildren().add(0, createRepoTitle(entry.getKey()));
```
dariusf: When we constantly add to the beginning this way, do we end up having to shift all elements back? Maybe we could map over the entry set, then use addAll?

src/main/java/ui/components/issuepicker/IssuePickerDialog.java
```diff
@@ -0,0 +1,250 @@
...
+        }
+    }
+
+    private void createButtons() {
+        ButtonType confirmButtonType = new ButtonType("Confirm", ButtonData.OK_DONE);
+        getDialogPane().getButtonTypes().addAll(confirmButtonType, ButtonType.CANCEL);
+
+        // defines what happens when user confirms/presses enter
+        setResultConverter(dialogButton -> {
+            if (dialogButton == confirmButtonType) {
```
dariusf: Is `==` supposed to be used here? Maybe we could just check the type of the button against an enum value?

src/main/java/util/events/ShowIssuePickerEvent.java
```diff
@@ -0,0 +1,23 @@
...
+ * An event to show Issue Picker 
+ */
+public class ShowIssuePickerEvent extends Event {
+
+    public final boolean isStandalone;
+    public final List<TurboIssue> allIssues;
+    
+    /**
+     * @param isStandalone determines if the caller expects a return result
+     * e.g when triggered by other UI element such as the filter text field
```
dariusf: I'm not sure what this means. It doesn't seem to be used either. As opposed to getting a result through a side effect?

jinified: It is not used for now. As in there are plans to interact with `bView` through this property

src/main/java/ui/components/issuepicker/IssuePickerDialog.java
jinified: I think we can do that.

src/main/java/ui/components/issuepicker/IssueCard.java
jinified: I have plans to factor out the UI element into its own component. Do you think I should commit this into the `ui.component` package

src/main/java/backend/resource/TurboIssue.java
jinified: Ok. Noted

src/main/java/ui/components/issuepicker/IssueCard.java
dariusf: That package seems to be for generic JavaFX nodes (though a few exceptions snuck in). Maybe just under `ui`?

src/main/java/ui/components/issuepicker/IssuePickerDialog.java
jinified: Technically we can do something like dialogButton.getButtonData() == ButtonData.OK_DONE but what we want to check for the text of the button also right ?

jinified:
@damithc  Is it ok if we handle issue with long title with ellipsis ? The user can hover to get the full title. 
screenshot from 2016-03-22 11 33 28

src/main/java/ui/components/issuepicker/IssuePickerDialog.java
dariusf: Is it possible that we can have multiple OK_DONE buttons?

But anyway, what I was saying was that you might be able to use `equals` instead of `==`, which seems a bit fragile (depends on implementation details).

damithc:
Actually, I prefer to remove code related to multiple select. Let's try to optimize it for single select and not integrate code that we don't strictly need. In the very rare case of needing to selected multiple issues, it can be a simple matter of using the picker multiple times.

jinified:
I can do that. What about the solution to issue with long title ?

damithc:
For now, yes that's fine. Eventually we can show the entire issue card there right? Now that we don't support multiple select, there's room to show the entire card?

jinified:
Ok noted.

jinified:
Ready for review @dariusf

src/main/java/backend/resource/TurboIssue.java
```diff
@@ -338,6 +339,28 @@ public TurboIssue combineWithPullRequest(PullRequest pullRequest) {
...
+    /**
+     * Matching is done by matching all words separated by space in query
+     * @param issues
+     * @param query
+     * @return list of issues that contains the query
+     */
+    public static List<TurboIssue> contains(List<TurboIssue> issues, String query){
+        List<String> queries = Arrays.asList(query.split("\\s"));
+        return issues.stream()
+            .filter(i -> Utility.containsIgnoreCaseMultipleWords(i.getTitle() + " " + i.getId(), queries))
```
dariusf: What I meant was changing containsIgnoreCaseMultipleWords, and making it an overload of containsIgnoreCase instead. Conceptually the operation would then just be extended to lists.

src/main/java/ui/components/issuepicker/IssuePickerDialog.java
```diff
@@ -0,0 +1,210 @@
...
+        getDialogPane().getScene().getWindow().sizeToScene();
+    }
+
+    private void populateSelectedIssues(List<TurboIssue> chosenIssues) {
+        selectedIssues.getChildren().clear();
+        Map<String, FlowPane> repoContent = getRepoContent(chosenIssues);
+
+        repoContent.entrySet().forEach(entry -> {
+            selectedIssues.getChildren().addAll(createRepoTitle(entry.getKey()), entry.getValue());
+        });
```
dariusf: This isn't what I meant also. We can map createRepoTitle over the entry set to get a new collection, do then addAll on that collection. Here we're (inadvertently?) using the varargs overload, and then doing lots of `add` operations.

src/main/java/ui/components/issuepicker/IssuePickerDialog.java
```diff
@@ -0,0 +1,210 @@
...
+    private void createButtons() {
+        ButtonType confirmButtonType = new ButtonType("Confirm", ButtonData.OK_DONE);
+        getDialogPane().getButtonTypes().addAll(confirmButtonType, ButtonType.CANCEL);
+
+        // defines what happens when user confirms/presses enter
+        setResultConverter(dialogButton -> {
+            if (dialogButton.getButtonData() == ButtonData.OK_DONE) {
+                // Confirms final issue that is suggested to user upon confirmation
+                if (!issuepickerQueryField.isDisabled()) issuepickerQueryField.appendText(";");
+                System.out.println(state.getSelectedIssues());
```
dariusf: Leftover.

src/main/java/ui/components/issuepicker/IssuePickerDialog.java
```diff
@@ -0,0 +1,210 @@
...
+        Label repoName = new Label(name + ": ");
+        repoName.setPadding(new Insets(0, 5, 5, 0));
+        repoName.setMaxWidth(ELEMENT_MAX_WIDTH - 10);
+        repoName.setStyle("-fx-font-size: 110%; -fx-font-weight: bold; ");
+        return repoName;
+    }
+
+    private final Map<String, FlowPane> getRepoContent(List<TurboIssue> chosenIssues) {
+        Map<String, FlowPane> repoContent = new HashMap<>();
+        chosenIssues.stream()
```
dariusf: We can use forEach without the intermediate stream.

But anyway, this looks a like a mapping operation instead. We are transforming the list of chosen issues into a map of nodes.

src/main/java/ui/components/issuepicker/IssuePickerState.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Updates current state based on given user input
+     * @param userInput 
+     */
+    private final void update(String userInput) {
+        String query = userInput.trim();
+        updateSuggestedIssues(allIssues, query);
+        if (!query.isEmpty()) {
```
dariusf: Might want to invert this and return early instead.

src/main/java/ui/components/issuepicker/IssuePickerState.java
```diff
@@ -0,0 +1,81 @@
...
+            currentSuggestion = suggestedIssues.stream().findFirst();
+            TurboIssue.getFirstMatchingIssue(allIssues, query).ifPresent(this::updateSelectedIssues);
+        }
+    }
+
+    /**
+     * Updates selected issues if issue is present
+     * @param issue
+     */
+    public void updateSelectedIssues(TurboIssue selectedIssue) {
```
dariusf: Maybe the method name could be a bit more specific. `toggle`?

src/main/java/ui/components/issuepicker/IssuePickerState.java
```diff
@@ -0,0 +1,81 @@
...
+        } else {
+            selectedIssues.add(selectedIssue);
+        }
+    }
+
+    /**
+     * Updates suggested issues with given query
+     * @param issue
+     */
+    private void updateSuggestedIssues(List<TurboIssue> issues, String query) {
```
dariusf: Same here, can be more specific. Maybe even `setSelectedIssues`.

src/main/java/ui/components/issuepicker/IssuePickerState.java
jinified: But toggle sounds like UI operation. It is technically updating the selected issues based query right ?

dariusf: Chose toggle because it says what the method is doing: if on, turn off; if off, turn on. You can pick any other name though. Was just saying that `update` was a bit too general a term for this.

jinified:
@dariusf  I have made some changes per requested but still not sure about the `update...` method. I think using `setSuggestedIssues` and `setSelectedIssues` would be better than toggle

docs/issue_picker.md
```diff
@@ -0,0 +1,31 @@
...
+A simple picker that allows you to select issues easily on current opened repository.
+
+### Getting started
+Press <kbd>Ctrl</kbd> + <kbd>I</kbd> to `launch` the issue picker. 
+
+![](images/pickers/issue_picker_main.png?raw=true)
+
+The `Issue Picker` is divided into 3 main sections:
+ 1. `Selected issues` : shows issues that will be selected upon confirmation 
+ 2. `Text field`     : enter your query here to filter all issues in current repository
```
damithc: `Text field` --> `Search query`?

src/main/resources/ui/fxml/IssuePickerView.fxml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<?import javafx.scene.control.*?>
+<?import javafx.geometry.*?>
+<?import java.lang.*?>
+<?import javafx.scene.layout.*?>
+
+<VBox fx:id="issuepickerLayout" maxHeight="600.0" maxWidth="-Infinity" minHeight="-Infinity" minWidth="-Infinity" xmlns="http://javafx.com/javafx/8" xmlns:fx="http://javafx.com/fxml/1">
```
damithc: This line is too long. Any chance it can be wrapped?

jinified: Ok noted

docs/issue_picker.md
jinified: It is a description of the container where users can run their queries. Maybe `Query field`

jinified:
@dariusf  Ready for review

src/main/java/ui/components/issuepicker/IssuePickerDialog.java
```diff
@@ -0,0 +1,199 @@
...
+
+        // Ensures dialog pane resize according to content
+        getDialogPane().getScene().getWindow().sizeToScene();
+    }
+
+    private void populateSelectedIssues(List<TurboIssue> chosenIssues) {
+        selectedIssues.getChildren().clear();
+        FlowPane selectedIssueCards = getSelectedIssuesPane(chosenIssues);
+        if (!selectedIssueCards.getChildren().isEmpty()) {
+            selectedIssues.getChildren().addAll(createRepoTitle(models.getDefaultRepo()), selectedIssueCards);
```
dariusf: We're using addAll to add one element here. Is this right?

Also, are we going to remove the code related to multiple selection? Just wondering, since the method signature takes a list.

jinified: We are adding two nodes: a label and a flowpane. I think we can keep the method signature though

dariusf: Oh okay, my mistake. If all the code related to multiple selection is removed, I think this is good to go.

jinified:
Made some changes to structure of IssuePickerState. @dariusf

docs/issue_picker.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Getting started
+Press <kbd>Ctrl</kbd> + <kbd>I</kbd> to `launch` the issue picker. 
+
+![](images/pickers/issue_picker_main.png?raw=true)
+
+The `Issue Picker` is divided into 3 main sections:
+ 1. `Selected issues pane` : shows issues that will be selected upon confirmation 
+ 2. `Query field`     : enter your query here to filter all issues in current repository 
+ 3. `Suggested issues pane`: lists all issues that matches your query.
```
damithc: Not sure if `pane` is a good term for user documentation. `area`? Up to you.

src/main/java/backend/resource/TurboIssue.java
```diff
@@ -338,6 +339,28 @@ public TurboIssue combineWithPullRequest(PullRequest pullRequest) {
         return Optional.empty();
     }
 
+    /**
+     * Matching is done by matching all words separated by space in query
```
damithc: Is this an `AND` match or an `OR` match. Need to specify. Ideally, it should be the latter.

src/main/java/ui/components/issuepicker/IssueCard.java
```diff
@@ -0,0 +1,176 @@
...
+                rightArrow.getStyleClass().addAll("octicon", "pull-request-assign-icon");
+                authorAssigneeBox.getChildren().add(rightArrow);
+            }
+        }
+
+        if (issue.getAssignee().isPresent()) {
+            HBox assigneeBox = createDisplayUserBox(guiElement.getAssignee(), issue.getAssignee().get());
+            authorAssigneeBox.getChildren().add(assigneeBox);
+        }
+    }
```
damithc: Method too long. Can extract smaller `update...` methods?

src/main/java/ui/components/issuepicker/IssuePickerState.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+import backend.resource.TurboIssue;
+
+/**
+ * This class is used to represent the state of the issue picker. 
+ */
```
damithc: ~~`This class is used to represent`~~ `Represents`

src/main/java/ui/components/issuepicker/PickerIssue.java
```diff
@@ -0,0 +1,75 @@
...
+    private final String displayTitle;
+    private boolean isFaded;
+    private boolean isRemoved;
+    
+    public PickerIssue(TurboIssue issue) {
+        super(issue);
+        displayTitle = String.format(LABEL_TITLE, getId(), getTitle());
+    }
+
+    public Node getNode() {
```
damithc: Why a `get` method is `setting` stuff?

src/main/java/ui/components/issuepicker/PickerIssue.java
```diff
@@ -0,0 +1,75 @@
+package ui.components.issuepicker;
+
+import backend.resource.TurboIssue;
+import javafx.scene.Node;
+import javafx.scene.control.Label;
+import javafx.scene.control.Tooltip;
+
+/**
+ * This class is to represent a displayed issue in IssuePickerDialog
```
damithc: As above, get rid of 'deadwood' words like `This class is to`

damithc:

src/main/java/backend/resource/TurboIssue.java
jinified: It is `AND` match because each space delimited word supposed to narrow down the suggestions right ?

src/main/java/ui/components/issuepicker/PickerIssue.java
jinified: I think I will delete `PickerIssue` since we are currently using `IssueCard`

src/main/java/backend/resource/TurboIssue.java
damithc: Rather, each new word potentially brings the target issue closer to the top of the search results. We cannot expect the user to remember the exact words precisely. If you use `AND`, one tiny typo will make the issue impossible to locate. It is fine to use `AND` now, but do create another issue to change it to a smarter search algo.

jinified: Ok noted.

jinified:
@damithc  Some fixes were made according to previous requests

src/main/java/ui/components/issuepicker/IssuePickerState.java
damithc: Still to be fixed.

damithc:
Some of the screenshots may be outdated? Can merge after updating those.

jinified:
Ok noted. Will update the documentation before merging

damithc:
PR title doesn't match issue.

PR #1350, opened 2016-03-14
Label picker: unmatched labels look like disabled #1333

### LINKED ISSUE #1333: Label picker: unmatched labels look like disabled
Unmatched labels look like they are disabled, although they are still clickable. This is not a major problem, just a bit inconsistent with norms. Also note that the black color `discarded` label looks normal all the time, even when it is unmatched.

image

jinified:
Fixes #1333

### DIFF
--- a/src/main/java/ui/components/pickers/PickerLabel.java
+++ b/src/main/java/ui/components/pickers/PickerLabel.java
@@ -37,9 +37,12 @@ public class PickerLabel extends TurboLabel {
         // actual name for labels at the top, add tick for selected labels
         Label label = new Label((canDisplayFullName ? getFullName() : getShortName()));
         label.getStyleClass().add("labels");
-        if (isRemoved) label.getStyleClass().add("labels-removed"); // add strikethrough
-        String style = getStyle() + (isHighlighted ? " -fx-border-color: black;" : ""); // add highlight border
-        style += (isFaded ? " -fx-opacity: 40%;" : ""); // change opacity if needed
+        // add strikethrough
+        if (isRemoved) label.getStyleClass().add("labels-removed"); 
+        // add highlight border
+        String style = getStyle() + (isHighlighted ? " -fx-border-color: black; -fx-border-width: 2px;" : ""); 
+        // change opacity if needed
+        style += (isFaded ? " -fx-opacity: 60%;" : ""); 
         label.setStyle(style);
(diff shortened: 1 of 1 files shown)


jinified:
Effect after decreasing opacity for unmatched labels and increasing thickness of matched label border. 


label_picker_unmatched

damithc:
It's better. But as long as we use opacity, they are going to look 'disabled'. I think the solution is to make the label clickable when the mouse is over it. e.g. change mouse pointer and/or remove opacity

jinified:
I think we just need a better way of highlighting the matched label from the rest e.g making it pop-up using size. In hindsight, using border is not really a good choice. What do you think ?

damithc:
yes, it's nice if we can emphasize matching labels without de-emphasizing unmatched labels. Not sure how hard it is to do so. I don't want to do anything (such as increasing size) that changes the positioning of labels as the user types. In spite of whatever we do there, the labels need to look clickable when hovered over.

jinified:
Ok. So as long as the position of the labels remain the same you are ok with it right ? Coupled that by changing the mouse pointer, I think it will be perfect already.

damithc:
Yup. Let us know when the code is ready for review.

jinified:
Here is the updated version where the cursor icon will change when hover on a clickable label. 

unmatched_label

Ready for review @ndt93

src/main/java/ui/components/pickers/PickerLabel.java
```diff
@@ -37,23 +38,33 @@ public Node getNode() {
...
-        if (isRemoved) label.getStyleClass().add("labels-removed"); // add strikethrough
-        String style = getStyle() + (isHighlighted ? " -fx-border-color: black;" : ""); // add highlight border
-        style += (isFaded ? " -fx-opacity: 40%;" : ""); // change opacity if needed
+        // add strikethrough
+        if (isRemoved) label.getStyleClass().add("labels-removed");
+        // add highlight border
+        String style = getStyle() + (isHighlighted ? " -fx-border-color: black; -fx-border-width: 2px;" : "");
+        // change opacity if needed
+        style += (isFaded ? " -fx-opacity: 40%;" : "");
         label.setStyle(style);
```
ndt93: This style part can be extracted into another method.

jinified: noted

PR #1352, opened 2016-03-14
Create a Milestone Picker that allows the user to add/remove/modify the milestone of an issue #1286

### LINKED ISSUE #1286: Create a Milestone Picker that allows the user to add/remove/modify the milestone of an issue

m133225:
Fixes #1286.

### DIFF
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -274,13 +274,12 @@ public class Logic {
      * Replaces existing milestone with the newMilestone in the issue object, the UI, and the server, in that order.
      * Server update is done after the local update to reduce the lag between the user action and the UI response
      *
-     * @param issue The issue object whose milestone is to be replaced.
+     * @param issue The issue object containing the repoid and issue id of the relevant issue
+     *              whose milestone is to be replaced
      * @param newMilestone The new milestone to be assigned to the issue
-     * @return true if milestone replacement on GitHub was a success, false otherwise.
+     * @return true if milestone replacements locally and on GitHub were successful, false otherwise.
      */
     public CompletableFuture<Boolean> replaceIssueMilestone(TurboIssue issue, Optional<Integer> newMilestone) {
-        Optional<Integer> originalMilestone = issue.getMilestone();
-
         logger.info("Changing milestone for " + issue + " in models");
         CompletableFuture<Optional<TurboIssue>> localMilestoneReplaceFuture =
                 repoOpControl.replaceIssueMilestoneLocally(issue, newMilestone);
@@ -288,9 +287,8 @@ public class Logic {
 
         return updateIssueMilestonesOnServer(issue, newMilestone)
                 .thenCombine(localMilestoneReplaceFuture,
-                        (isUpdateSuccessful, locallyModifiedIssue) ->
-                                handleIssueMilestoneUpdateOnServerResult(isUpdateSuccessful, locallyModifiedIssue,
-                                        originalMilestone));
+                        (isUpdateSuccessful, originalIssue) ->
+                                handleIssueMilestoneUpdateOnServerResult(isUpdateSuccessful, originalIssue));
     }
 
     /**
@@ -343,21 +341,19 @@ public class Logic {
      * Handles the result of updating an issue's milestone on server.
      * Locally reverts back to the original milestone if the server update fails.
      * @param isUpdateSuccessful
-     * @param localModifiedIssue
-     * @param originalMilestone
+     * @param originalIssue
      * @return true if the server update is successful
      */
     private boolean handleIssueMilestoneUpdateOnServerResult(boolean isUpdateSuccessful,
-                                                          Optional<TurboIssue> localModifiedIssue,
-                                                          Optional<Integer> originalMilestone) {
-        if (!localModifiedIssue.isPresent()) {
+                                                          Optional<TurboIssue> originalIssue) {
+        if (!originalIssue.isPresent()) {
             logger.error("Unable to replace issue milestone locally");
             return false;
         }
         if (isUpdateSuccessful) return true;
 
         logger.error("Unable to update model on server");
-        revertLocalMilestoneReplace(localModifiedIssue.get(), originalMilestone);
+        revertLocalMilestoneReplace(originalIssue.get());
         return false;
     }
 
@@ -381,15 +377,14 @@ public class Logic {
     }
 
     /**
-     * Replaces the milestone of the issue in the {@link Logic#models} corresponding to {@code modifiedIssue} with
-     * {@code originalMilestone} if they were both issues have the same last modified LocalDateTime
+     * Replaces the milestone of the issue in the {@link Logic#models} corresponding to {@code originalIssue}
+     * with originalIssue's milestone if both issues have the same last modified LocalDateTime
      *
-     * @param modifiedIssue
-     * @param originalMilestone
+     * @param originalIssue
      */
-    private void revertLocalMilestoneReplace(TurboIssue modifiedIssue, Optional<Integer> originalMilestone) {
-        TurboIssue currentIssue = getIssue(modifiedIssue.getRepoId(), modifiedIssue.getId()).orElse(modifiedIssue);
-        LocalDateTime originalMilestoneModifiedAt = modifiedIssue.getMilestoneLastModifiedAt();
+    private void revertLocalMilestoneReplace(TurboIssue originalIssue) {
+        TurboIssue currentIssue = getIssue(originalIssue.getRepoId(), originalIssue.getId()).orElse(originalIssue);
+        LocalDateTime originalMilestoneModifiedAt = originalIssue.getMilestoneLastModifiedAt();
         LocalDateTime currentMilestoneAssignedAt = currentIssue.getMilestoneLastModifiedAt();
         boolean isCurrentMilestoneModifiedFromOriginalMilestone = originalMilestoneModifiedAt.isEqual(
                 currentMilestoneAssignedAt);
@@ -397,7 +392,7 @@ public class Logic {
         if (!isCurrentMilestoneModifiedFromOriginalMilestone) return;
 
         logger.info("Reverting milestone for issue " + currentIssue);
-        models.replaceIssueMilestone(currentIssue.getRepoId(), currentIssue.getId(), originalMilestone);
+        models.replaceIssueMilestone(currentIssue.getRepoId(), currentIssue.getId(), originalIssue.getMilestone());
         refreshUI();
     }
 
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -4,6 +4,7 @@ import backend.control.RepoOpControl;
 import backend.resource.Model;
 import backend.resource.MultiModel;
 import backend.resource.TurboIssue;
+import backend.resource.TurboMilestone;
 import filter.expression.FilterExpression;
 import filter.expression.Qualifier;
 import filter.expression.QualifierType;
@@ -328,6 +329,18 @@ public class Logic {
             models.replaceIssueLabels(currentIssue.getRepoId(), currentIssue.getId(), originalLabels);
             refreshUI();
         }
+    }
+
+    public CompletableFuture<Boolean> replaceIssueMilestone(TurboIssue issue, Integer milestone) {
+        logger.info("Changing milestone for " + issue + " on GitHub");
+        return repoIO.replaceIssueMilestone(issue, milestone)
+                .thenApply(resultingIssue -> {
+                    logger.info("Changing milestone for " + issue + " on UI");
+                    issue.setMilestone(milestone);
+                    refreshUI();
+                    return true;
+                })
+                .exceptionally(Futures.withResult(false));
     }
 
     /**
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -339,8 +339,8 @@ public class Logic {
     }
 
     /**
-     * Handles the result of updating an issue's milestone on server. Current implementation includes
-     * reverting back to the original labels locally if the server update failed.
+     * Handles the result of updating an issue's milestone on server.
+     * Locally reverts back to the original labels if the server update fails.
      * @param isUpdateSuccessful
      * @param localModifiedIssue
      * @param originalMilestone
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -278,18 +278,19 @@ public class Logic {
      * @param newMilestone The new milestone to be assigned to the issue
      * @return true if milestone replacement on GitHub was a success, false otherwise.
      */
-    public CompletableFuture<Boolean> replaceIssueMilestone(TurboIssue issue, Integer newMilestone) {
+    public CompletableFuture<Boolean> replaceIssueMilestone(TurboIssue issue, Optional<Integer> newMilestone) {
         Optional<Integer> originalMilestone = issue.getMilestone();
 
-        logger.info("Changing milestone for " + issue + " on UI");
+        logger.info("Changing milestone for " + issue + " in models");
         CompletableFuture<Optional<TurboIssue>> localMilestoneReplaceFuture =
                 repoOpControl.replaceIssueMilestoneLocally(issue, newMilestone);
         localMilestoneReplaceFuture.thenRun(this::refreshUI);
 
         return updateIssueMilestonesOnServer(issue, newMilestone)
                 .thenCombine(localMilestoneReplaceFuture,
-                        (isUpdateSuccessful, locallyModifiedIssue) -> handleIssueMilestoneUpdateOnServerResult(
-                        isUpdateSuccessful, locallyModifiedIssue, originalMilestone.orElseGet(() -> null)));
+                        (isUpdateSuccessful, locallyModifiedIssue) ->
+                                handleIssueMilestoneUpdateOnServerResult(isUpdateSuccessful, locallyModifiedIssue,
+                                        originalMilestone));
     }
 
     /**
@@ -310,7 +311,7 @@ public class Logic {
         return repoOpControl.replaceIssueLabelsOnServer(issue, newLabels);
     }
 
-    private CompletableFuture<Boolean> updateIssueMilestonesOnServer(TurboIssue issue, Integer milestone) {
+    private CompletableFuture<Boolean> updateIssueMilestonesOnServer(TurboIssue issue, Optional<Integer> milestone) {
         logger.info("Changing milestone for " + issue + " on GitHub");
         return repoOpControl.replaceIssueMilestoneOnServer(issue, milestone);
     }
@@ -340,7 +341,7 @@ public class Logic {
 
     /**
      * Handles the result of updating an issue's milestone on server.
-     * Locally reverts back to the original labels if the server update fails.
+     * Locally reverts back to the original milestone if the server update fails.
      * @param isUpdateSuccessful
      * @param localModifiedIssue
      * @param originalMilestone
@@ -348,7 +349,7 @@ public class Logic {
      */
     private boolean handleIssueMilestoneUpdateOnServerResult(boolean isUpdateSuccessful,
                                                           Optional<TurboIssue> localModifiedIssue,
-                                                          Integer originalMilestone) {
+                                                          Optional<Integer> originalMilestone) {
         if (!localModifiedIssue.isPresent()) {
             logger.error("Unable to replace issue milestone locally");
             return false;
@@ -386,7 +387,7 @@ public class Logic {
      * @param modifiedIssue
      * @param originalMilestone
      */
-    private void revertLocalMilestoneReplace(TurboIssue modifiedIssue, Integer originalMilestone) {
+    private void revertLocalMilestoneReplace(TurboIssue modifiedIssue, Optional<Integer> originalMilestone) {
         TurboIssue currentIssue = getIssue(modifiedIssue.getRepoId(), modifiedIssue.getId()).orElse(modifiedIssue);
         LocalDateTime originalMilestoneModifiedAt = modifiedIssue.getMilestoneLastModifiedAt();
         LocalDateTime currentMilestoneAssignedAt = currentIssue.getMilestoneLastModifiedAt();
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -10,7 +10,6 @@ import filter.expression.QualifierType;
 import javafx.application.Platform;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.logging.log4j.Logger;
-import org.eclipse.egit.github.core.Issue;
 import prefs.Preferences;
 import ui.GuiElement;
 import ui.TestController;
@@ -281,18 +280,16 @@ public class Logic {
      */
     public CompletableFuture<Boolean> replaceIssueMilestone(TurboIssue issue, Integer newMilestone) {
         Optional<Integer> originalMilestone = issue.getMilestone();
-        logger.info("Changing milestone for " + issue + " on UI");
 
-        /* Calls models to replace the issue's milestone locally since the the reference to the issue here
-           could be invalidated by changes to the models elsewhere */
-        Optional<TurboIssue> localReplaceResult =
-                models.replaceIssueMilestone(issue.getRepoId(), issue.getId(), newMilestone);
-        if (!localReplaceResult.isPresent()) return CompletableFuture.completedFuture(false);
-        refreshUI();
+        logger.info("Changing milestone for " + issue + " on UI");
+        CompletableFuture<Optional<TurboIssue>> localMilestoneReplaceFuture =
+                repoOpControl.replaceIssueMilestoneLocally(issue, newMilestone);
+        localMilestoneReplaceFuture.thenRun(this::refreshUI);
 
         return updateIssueMilestonesOnServer(issue, newMilestone)
-                .thenApply((isUpdateSuccessful) -> handleIssueMilestoneUpdateOnServerResult(
-                        isUpdateSuccessful, localReplaceResult.get(), originalMilestone.orElseGet(() -> null)));
+                .thenCombine(localMilestoneReplaceFuture,
+                        (isUpdateSuccessful, locallyModifiedIssue) -> handleIssueMilestoneUpdateOnServerResult(
+                        isUpdateSuccessful, locallyModifiedIssue, originalMilestone.orElseGet(() -> null)));
     }
 
     /**
@@ -315,7 +312,7 @@ public class Logic {
 
     private CompletableFuture<Boolean> updateIssueMilestonesOnServer(TurboIssue issue, Integer milestone) {
         logger.info("Changing milestone for " + issue + " on GitHub");
-        return repoOpControl.replaceIssueMilestone(issue, milestone);
+        return repoOpControl.replaceIssueMilestoneOnServer(issue, milestone);
     }
 
     /**
@@ -350,11 +347,16 @@ public class Logic {
      * @return true if the server update is successful
      */
     private boolean handleIssueMilestoneUpdateOnServerResult(boolean isUpdateSuccessful,
-                                                          TurboIssue localModifiedIssue,
+                                                          Optional<TurboIssue> localModifiedIssue,
                                                           Integer originalMilestone) {
+        if (!localModifiedIssue.isPresent()) {
+            logger.error("Unable to replace issue milestone locally");
+            return false;
+        }
         if (isUpdateSuccessful) return true;
+
         logger.error("Unable to update model on server");
-        revertLocalMilestoneReplace(localModifiedIssue, originalMilestone);
+        revertLocalMilestoneReplace(localModifiedIssue.get(), originalMilestone);
         return false;
     }
 
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -4,7 +4,6 @@ import backend.control.RepoOpControl;
 import backend.resource.Model;
 import backend.resource.MultiModel;
 import backend.resource.TurboIssue;
-import backend.resource.TurboMilestone;
 import filter.expression.FilterExpression;
 import filter.expression.Qualifier;
 import filter.expression.QualifierType;
@@ -336,7 +335,11 @@ public class Logic {
         return repoIO.replaceIssueMilestone(issue, milestone)
                 .thenApply(resultingIssue -> {
                     logger.info("Changing milestone for " + issue + " on UI");
-                    issue.setMilestone(resultingIssue.getMilestone().getNumber());
+                    if (resultingIssue.getMilestone() != null) {
+                        issue.setMilestoneById(resultingIssue.getMilestone().getNumber());
+                    } else {
+                        issue.setMilestoneById(null);
+                    }
                     refreshUI();
                     return true;
                 })
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -336,7 +336,7 @@ public class Logic {
         return repoIO.replaceIssueMilestone(issue, milestone)
                 .thenApply(resultingIssue -> {
                     logger.info("Changing milestone for " + issue + " on UI");
-                    issue.setMilestone(milestone);
+                    issue.setMilestone(resultingIssue.getMilestone().getNumber());
                     refreshUI();
                     return true;
                 })
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -274,10 +274,9 @@ public class Logic {
      * Replaces existing milestone with the newMilestone in the issue object, the UI, and the server, in that order.
      * Server update is done after the local update to reduce the lag between the user action and the UI response
      *
-     * @param issue The issue object containing the repoid and issue id of the relevant issue
-     *              whose milestone is to be replaced
+     * @param issue The issue object whose milestone is to be replaced
      * @param newMilestone The new milestone to be assigned to the issue
-     * @return true if milestone replacements locally and on GitHub were successful, false otherwise.
+     * @return true if milestone replacements locally and on GitHub were successful
      */
     public CompletableFuture<Boolean> replaceIssueMilestone(TurboIssue issue, Optional<Integer> newMilestone) {
         logger.info("Changing milestone for " + issue + " in models");
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -286,9 +286,7 @@ public class Logic {
         localMilestoneReplaceFuture.thenRun(this::refreshUI);
 
         return updateIssueMilestonesOnServer(issue, newMilestone)
-                .thenCombine(localMilestoneReplaceFuture,
-                        (isUpdateSuccessful, originalIssue) ->
-                                handleIssueMilestoneUpdateOnServerResult(isUpdateSuccessful, originalIssue));
+                .thenCombine(localMilestoneReplaceFuture, this::handleIssueMilestoneUpdateOnServerResult);
     }
 
     /**
@@ -386,8 +384,8 @@ public class Logic {
         TurboIssue currentIssue = getIssue(originalIssue.getRepoId(), originalIssue.getId()).orElse(originalIssue);
         LocalDateTime originalMilestoneModifiedAt = originalIssue.getMilestoneLastModifiedAt();
         LocalDateTime currentMilestoneAssignedAt = currentIssue.getMilestoneLastModifiedAt();
-        boolean isCurrentMilestoneModifiedFromOriginalMilestone = originalMilestoneModifiedAt.isEqual(
-                currentMilestoneAssignedAt);
+        boolean isCurrentMilestoneModifiedFromOriginalMilestone = originalMilestoneModifiedAt
+                                                                    .isEqual(currentMilestoneAssignedAt);
 
         if (!isCurrentMilestoneModifiedFromOriginalMilestone) return;
 
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -271,20 +271,26 @@ public class Logic {
                                      isUpdateSuccessful, locallyModifiedIssue, originalLabels));
     }
 
-    public CompletableFuture<Boolean> replaceIssueMilestone(TurboIssue issue, Integer milestone) {
+    /**
+     * Replaces existing milestone with the newMilestone in the issue object, the UI, and the server, in that order.
+     * Server update is done after the local update to reduce the lag between the user action and the UI response
+     *
+     * @param issue The issue object whose milestone is to be replaced.
+     * @param newMilestone The new milestone to be assigned to the issue
+     * @return true if milestone replacement on GitHub was a success, false otherwise.
+     */
+    public CompletableFuture<Boolean> replaceIssueMilestone(TurboIssue issue, Integer newMilestone) {
         Optional<Integer> originalMilestone = issue.getMilestone();
-        logger.info("Changing milestone for " + issue + " on GitHub");
+        logger.info("Changing milestone for " + issue + " on UI");
 
         /* Calls models to replace the issue's milestone locally since the the reference to the issue here
            could be invalidated by changes to the models elsewhere */
         Optional<TurboIssue> localReplaceResult =
-                models.replaceIssueMilestone(issue.getRepoId(), issue.getId(), milestone);
-        if (!localReplaceResult.isPresent()) {
-            return CompletableFuture.completedFuture(false);
-        }
+                models.replaceIssueMilestone(issue.getRepoId(), issue.getId(), newMilestone);
+        if (!localReplaceResult.isPresent()) return CompletableFuture.completedFuture(false);
         refreshUI();
 
-        return updateIssueMilestonesOnServer(issue, milestone)
+        return updateIssueMilestonesOnServer(issue, newMilestone)
                 .thenApply((isUpdateSuccessful) -> handleIssueMilestoneUpdateOnServerResult(
                         isUpdateSuccessful, localReplaceResult.get(), originalMilestone.orElseGet(() -> null)));
     }
@@ -346,9 +352,7 @@ public class Logic {
     private boolean handleIssueMilestoneUpdateOnServerResult(boolean isUpdateSuccessful,
                                                           TurboIssue localModifiedIssue,
                                                           Integer originalMilestone) {
-        if (isUpdateSuccessful) {
-            return true;
-        }
+        if (isUpdateSuccessful) return true;
         logger.error("Unable to update model on server");
         revertLocalMilestoneReplace(localModifiedIssue, originalMilestone);
         return false;
@@ -375,7 +379,8 @@ public class Logic {
 
     /**
      * Replaces the milestone of the issue in the {@link Logic#models} corresponding to {@code modifiedIssue} with
-     * {@code originalLabels} if the current labels on the issue is assigned at the same time as {@code modifiedIssue}
+     * {@code originalMilestone} if they were both issues have the same last modified LocalDateTime
+     *
      * @param modifiedIssue
      * @param originalMilestone
      */
@@ -386,11 +391,11 @@ public class Logic {
         boolean isCurrentMilestoneModifiedFromOriginalMilestone = originalMilestoneModifiedAt.isEqual(
                 currentMilestoneAssignedAt);
 
-        if (isCurrentMilestoneModifiedFromOriginalMilestone) {
-            logger.info("Reverting milestone for issue " + currentIssue);
-            models.replaceIssueMilestone(currentIssue.getRepoId(), currentIssue.getId(), originalMilestone);
-            refreshUI();
-        }
+        if (!isCurrentMilestoneModifiedFromOriginalMilestone) return;
+
+        logger.info("Reverting milestone for issue " + currentIssue);
+        models.replaceIssueMilestone(currentIssue.getRepoId(), currentIssue.getId(), originalMilestone);
+        refreshUI();
     }
 
     /**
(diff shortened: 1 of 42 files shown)


m133225:
Problem identified. Will be using the old PR again instead.

m133225:
Ready for review~~

ndt93:
@m133225 Can you update to the current master? There are changes in the backend that will affect this, so you might take a look at the flow for labels replacement again.

m133225:
Hmm... Already updated rather recently and resolved conflicts in the backend. Will still update the current master.

ndt93:
The main change is that the local modification to the `TurboIssue` would also need to be wrapped inside a mutually exclusive operation.

m133225:
Updated to incorporate the backend changes.

damithc:
Who's the reviewer for this? @dariusf @ndt93

damithc:
I would like to merge this soon, so let's push it through the phases fast.

src/main/java/backend/Logic.java
```diff
@@ -312,6 +339,28 @@ private boolean handleIssueLabelsUpdateResult(boolean isUpdateSuccessful,
     }
 
     /**
+     * Handles the result of updating an issue's milestone on server. Current implementation includes
+     * reverting back to the original labels locally if the server update failed.
```
damithc: This part is redundant `Current implementation includes`

src/main/java/backend/stub/DummyRepoState.java
```diff
@@ -400,6 +398,32 @@ protected TurboUser deleteUser(String idString) {
         return newLabels.stream().map(new Label()::setName).collect(Collectors.toList());
     }
 
+    protected final Issue setMilestone(int issueId, Integer milestone) {
```
damithc: Header comment?

damithc: The return value and the method name don't seem to match.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -0,0 +1,309 @@
...
+    private static final String DIALOG_TITLE = "Select Milestone";
+
+    private final List<PickerMilestone> originalMilestones = new ArrayList<>();
+    private FlowPane assignedMilestonePane;
+    private VBox matchingMilestonesPane;
+    private TextField inputField;
+    private MilestonePickerState state;
+
+    /**
+     * Constructor to create a MilestonePickerDialog
```
damithc: Redundant statement :-p

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -0,0 +1,123 @@
...
+
+    public MilestonePickerState(List<PickerMilestone> milestones, String userInput) {
+        this(milestones);
+        processInput(userInput);
+    }
+
+    private void cloneList(List<PickerMilestone> sourceList, List<PickerMilestone> destList) {
+        sourceList.stream()
+                .forEach(milestone -> destList.add(new PickerMilestone(milestone)));
+    }
```
damithc: Any reason why a parameter is used for returning the result?

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -0,0 +1,123 @@
...
+     * Gets the list of milestones that matches the current query i.e. not faded
+     * @return
+     */
+    public List<PickerMilestone> getMatchingMilestonesList() {
+        return this.currentMilestonesList.stream()
+                .filter(milestone -> !milestone.isFaded())
+                .collect(Collectors.toList());
+    }
+
+    private void filterMilestones(String query) {
```
damithc: header comment?

damithc: Why the parameter here is `query` but it is `milestoneQuery` in the `toggleMilestone` method?

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -0,0 +1,123 @@
...
+                .forEach(milestone -> destList.add(new PickerMilestone(milestone)));
+    }
+
+    private void processInput(String userInput) {
+        if (userInput.isEmpty()) return;
+
+        String[] userInputWords = userInput.split(" ");
+        for (int i = 0; i < userInputWords.length; i++) {
+            String currentWord = userInputWords[i];
+            if (i < userInputWords.length - 1 || userInput.endsWith(" ")) {
```
damithc: Give this condition an explanatory name. It's not clear what it means.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Gets the name of milestone if it is the only matching milestone to the query
+     * @param query
+     * @return
+     */
+    private Optional<PickerMilestone> getMatchingMilestoneIfOnlyMatch(String query) {
+        if (!hasExactlyOneMatchingMilestone(currentMilestonesList, query)) return Optional.empty();
+
+        return Optional.of(getMatchingMilestone(currentMilestonesList, query));
+    }
```
damithc: This method is odd. It's better if you have a method that returns all matching milestones and let the caller decide if it has 0 or 1, or more. If you have a method like this, the Optional.empty() could mean either there is no matching milestone or there were multiple matching milestones.

damithc: Also keep in mind that the user should not be forced to type until there is only one matching milestone. The picker should suggest one milestone even if there are multiple matches. The suggestion can be based on some algorithm (e.g. closest open milestone that matches) or just pick the first one in the list.

src/main/java/ui/components/pickers/PickerMilestone.java
```diff
@@ -0,0 +1,174 @@
...
+        if (milestone.isHighlighted()) return 1;
+        if (this.getDueDate().equals(milestone.getDueDate())) return 0;
+        if (this.isOpen() != milestone.isOpen()) {
+            return this.isOpen() ? -1 : 1;
+        }
+        if (!this.getDueDate().isPresent()) return -1;
+        if (!milestone.getDueDate().isPresent()) return 1;
+
+        return this.getDueDate().get()
+                .isAfter(milestone.getDueDate().get()) ? -1 : 1;
```
damithc: Does `Milestone` also have a `compareTo` method? In that case, is the logic duplicated?

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -451,6 +471,12 @@ private void stopLoadingAnimation() {
         hideLoadingIndicator();
     }
 
+    private void changeMilestone() {
```
damithc: Method name is not specific enough.

damithc:
Added some comments while we wait for @dariusf to review.

src/main/java/ui/components/pickers/PickerMilestone.java
m133225: No, it does not. Even if there was, the way `PickerMilestone` would probably be sorted using a (mostly) different logic to fit the MilestonePicker's display.

src/main/java/backend/Logic.java
```diff
@@ -271,6 +271,28 @@ public Model getRepo(String repoId) {
...
+     *
+     * @param issue The issue object whose milestone is to be replaced.
+     * @param newMilestone The new milestone to be assigned to the issue
+     * @return true if milestone replacement on GitHub was a success, false otherwise.
+     */
+    public CompletableFuture<Boolean> replaceIssueMilestone(TurboIssue issue, Integer newMilestone) {
+        Optional<Integer> originalMilestone = issue.getMilestone();
+
+        logger.info("Changing milestone for " + issue + " on UI");
+        CompletableFuture<Optional<TurboIssue>> localMilestoneReplaceFuture =
```
dariusf: I don't think the result should be optional. It reflects the error condition of not being able to find an issue in the model, which is very likely to be programmer error. (Being optional means that the value may reasonably be absent, and that the programmer should remember to handle that possibility.)

This is probably better handled by throwing an exception, then using `exceptionally` to log and return `false` to the caller.

src/main/java/backend/Logic.java
```diff
@@ -271,6 +271,28 @@ public Model getRepo(String repoId) {
...
+                repoOpControl.replaceIssueMilestoneLocally(issue, newMilestone);
+        localMilestoneReplaceFuture.thenRun(this::refreshUI);
+
+        return updateIssueMilestonesOnServer(issue, newMilestone)
+                .thenCombine(localMilestoneReplaceFuture,
+                        (isUpdateSuccessful, locallyModifiedIssue) -> handleIssueMilestoneUpdateOnServerResult(
+                        isUpdateSuccessful, locallyModifiedIssue, originalMilestone.orElseGet(() -> null)));
```
dariusf: Hmm, why unwrap the optional only to use null?

Indentation can be improved a bit. A newline after `->` usually helps communicate that what follows is nested.

src/main/java/backend/resource/TurboIssue.java
```diff
@@ -438,13 +443,15 @@ public void addLabel(TurboLabel label) {
     public Optional<Integer> getMilestone() {
         return milestone;
     }
-
-    public void setMilestone(Integer milestone) {
-        this.milestone = Optional.of(milestone);
+    
+    public void setMilestoneById(Integer milestone) {
+        this.milestone = (milestone == null) ? Optional.empty() : Optional.of(milestone);
```
dariusf: Again, this should assume it is given a valid input.

src/main/java/ui/components/pickers/MilestonePicker.java
```diff
@@ -0,0 +1,44 @@
...
+import ui.UI;
+import undo.actions.ChangeMilestoneAction;
+import util.events.ShowMilestonePickerEventHandler;
+
+import java.util.List;
+import java.util.Optional;
+
+public class MilestonePicker {
+    UI ui;
+    Stage stage;
```
dariusf: Can these be `private final`?

src/main/java/ui/components/pickers/MilestonePicker.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public MilestonePicker(UI ui, Stage mainStage) {
+        this.ui = ui;
+        this.stage = mainStage;
+        ui.registerEvent((ShowMilestonePickerEventHandler) e -> Platform.runLater(() -> showMilestonePicker(e.issue)));
+    }
+
+    private void showMilestonePicker(TurboIssue issue) {
+        List<TurboMilestone> milestoneList = ui.logic.getRepo(issue.getRepoId()).getMilestones();
```
dariusf: Types are metadata/implementation details, and can be left out of variable names.

src/main/java/ui/components/pickers/MilestonePicker.java
```diff
@@ -0,0 +1,44 @@
...
+        MilestonePickerDialog dialog = new MilestonePickerDialog(stage, issue, milestoneList);
+        Optional<Pair<ButtonType, Integer>> milestoneDialogResponse = dialog.showAndWait();
```
dariusf: This should be named.

src/main/java/ui/components/pickers/MilestonePicker.java
```diff
@@ -0,0 +1,44 @@
...
+
+        if (!milestoneDialogResponse.isPresent() ||
+                milestoneDialogResponse.get().getKey().equals(ButtonType.CANCEL)) {
+            return;
+        }
+
+        Optional<Integer> newlyAssignedMilestone = Optional.ofNullable(milestoneDialogResponse.get().getValue());
```
dariusf: Optional should be used here to indicate the absence of a result.

So there are 3 cases, right?

- Cancel, nothing selected
- Cancel, something selected
- Confirm, something selected

If it's important that we distinguish the first two, we can use `Optional<ButtonType, Optional<Integer>>`. Otherwise just `Optional<ButtonType, Integer>` will do.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -0,0 +1,308 @@
...
+    private static final String DIALOG_TITLE = "Select Milestone";
+
+    private final List<PickerMilestone> originalMilestones = new ArrayList<>();
+    private FlowPane assignedMilestonePane;
+    private VBox matchingMilestonesPane;
+    private TextField inputField;
+    private MilestonePickerState state;
+
+    /**
+     * The issue and the originalMilestones list provided should come from the same repository
```
dariusf: Maybe we should check this precondition and throw something if it's not met, as it's probably a programmer error.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    private void handleUpdatedInput(String userInput) {
+        state = new MilestonePickerState(originalMilestones, userInput);
+        refreshUI(state);
+    }
+
+    private List<PickerMilestone> convertToPickerMilestones(TurboIssue issue, List<TurboMilestone> milestones) {
+        List<PickerMilestone> originalMilestones = new ArrayList<>();
+        for (int i = 0; i < milestones.size(); i++) {
```
dariusf: Can we iterate with the enhanced for loop?

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -0,0 +1,308 @@
...
+        });
+    }
+
+    private Node setMouseClickForNode(Node node, String milestoneName) {
+        node.setOnMouseClicked(e -> handleMouseClick(milestoneName));
+        return node;
+    }
+
+    /**
+     * This will cause
```
dariusf: Incomplete. :stuck_out_tongue:

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -0,0 +1,308 @@
...
+     * @param milestoneName
+     */
+    private void handleMouseClick(String milestoneName) {
+        // required since clearing inputField will change the reference to the state
```
dariusf: Could be a bit more specific. "We need to save the state because..."

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -0,0 +1,308 @@
...
+
+    private void initUI() {
+        VBox milestoneDialogBox = createMilestoneDialogBox();
+        assignedMilestonePane = createAssignedMilestoneBox();
+        matchingMilestonesPane = createMatchingMilestonesBox();
+        inputField = createInputField();
+
+        milestoneDialogBox.getChildren().add(assignedMilestonePane);
+        milestoneDialogBox.getChildren().add(inputField);
+        milestoneDialogBox.getChildren().add(matchingMilestonesPane);
```
dariusf: Can we use `getChildren().addAll()`?

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -0,0 +1,308 @@
...
+
+    private Optional<PickerMilestone> getExistingMilestone(List<PickerMilestone> milestoneList) {
+        return milestoneList.stream()
+                .filter(milestone -> milestone.isExisting())
+                .findAny();
+    }
+
+    private boolean hasSelectedMilestone(List<PickerMilestone> milestoneList) {
+        return milestoneList.stream()
+                .filter(milestone -> milestone.isSelected())
```
dariusf: These can all use `milestone::isSelected`.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * Toggles the milestone that has milestoneName as its exact name
+     *
+     * This method is case-sensitive
+     * @param milestoneName
+     */
+    public final void toggleExactMatchMilestone(String milestoneName) {
+        Optional<PickerMilestone> onlyMatchingMilestone = getExactMatchMilestone(currentMilestonesList, milestoneName);
+        toggleMilestone(onlyMatchingMilestone);
```
dariusf: Rather than call this with an optional, we should check here if it's present and call/not call it here.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -0,0 +1,145 @@
...
+     */
+    private void toggleMilestone(Optional<PickerMilestone> milestone) {
+        if (!milestone.isPresent()) return;
+        this.currentMilestonesList.stream()
+                .forEach(listMilestone -> listMilestone.setSelected(listMilestone.equals(milestone.get())
+                        && !listMilestone.isSelected()));
+    }
+
+    /**
+     * Gets the current list of milestones
```
dariusf: Don't think we need this comment.

src/main/java/ui/components/pickers/PickerMilestone.java
```diff
@@ -0,0 +1,179 @@
...
+    private static final int BIG_LABEL_FONT = 16;
+    boolean isSelected = false;
+    boolean isHighlighted = false;
+    boolean isFaded = false;
+    boolean isExisting = false;
+
+    public PickerMilestone(TurboMilestone milestone) {
+        super(milestone.getRepoId(), milestone.getId(), milestone.getTitle());
+        setDueDate(milestone.getDueDate());
+        setDescription(milestone.getDescription() == null ? "" : milestone.getDescription());
```
dariusf: Can this ever happen? TurboMilestone is supposed to have eliminated null by this point.

src/main/java/ui/components/pickers/PickerMilestone.java
```diff
@@ -0,0 +1,179 @@
...
+     *
+     * In addition, this treats null milestones with no set due date to be "smaller than"
+     * those which have due dates
+     *
+     * @param milestone
+     * @return
+     */
+    @Override
+    public int compareTo(PickerMilestone milestone) {
+        if (isHighlighted) return -1;
```
dariusf: Hmm, this would make two milestones which are both highlighted smaller than each other, which could cause some weirdness. Probably undefined ordering? Similar problems in the rest of this method.

Instead of documenting all this in the comment, we should probably do it inline. Easier to follow and harder to get out of sync.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -382,6 +391,17 @@ private MenuItem updateChangeLabelsMenuItem() {
         return changeLabelsMenuItem;
     }
 
+    private MenuItem updateChangeMilestoneMenuItem() {
+        Optional<GuiElement> item = listView.getSelectedItem();
+        if (item.isPresent()) {
+            changeMilestoneMenuItem.setDisable(false);
+        } else {
+            changeMilestoneMenuItem.setDisable(true);
```
dariusf: We can use `item.isPresent()` as the argument directly.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -462,6 +482,12 @@ private void changeLabels() {
         }
     }
 
+    private void changeMilestoneOfSelectedIssue() {
+        if (getSelectedElement().isPresent()) {
```
dariusf: Hmm, these no-op-if-absent methods aren't so good. If the value was absent to begin with, why call the method? `getSelectedElement().ifPresent(this::changeMilestoneOfSelectedIssue)` is meant for this case. Not sure if the name can be shortened, as it is a bit verbose.

src/test/java/tests/MilestonePickerStateTests.java
```diff
@@ -0,0 +1,52 @@
...
+        MilestonePickerState state = prepareUnassignedState();
+        state.toggleExactMatchMilestone("milestone1");
+        assertEquals(true, state.getCurrentMilestonesList().get(0).isSelected());
+    }
+
+    @Test
+    public void toggleMilestone_hasMilestone_milestoneReplaced() {
+        MilestonePickerState state = prepareAssignedState();
+        state.toggleExactMatchMilestone("milestone1");
+        assertEquals(true, state.getCurrentMilestonesList().get(0).isSelected());
```
dariusf: You can use `assertFalse` for these.

src/main/java/backend/Logic.java
```diff
@@ -312,6 +339,28 @@ private boolean handleIssueLabelsUpdateResult(boolean isUpdateSuccessful,
     }
 
     /**
+     * Handles the result of updating an issue's milestone on server.
+     * Locally reverts back to the original labels if the server update fails.
```
dariusf: This part is not updated.

src/main/java/backend/github/GitHubRepo.java
```diff
@@ -260,6 +262,23 @@ public boolean login(UserCredentials credentials) {
     }
 
     @Override
+    public Issue setMilestone(String repoId, int issueId, String issueTitle, Integer issueMilestone)
+            throws IOException {
+        Milestone gitHubMilestone = new Milestone();
+        if (issueMilestone != null) {
```
dariusf: We should catch this at a higher level, as early as possible. It's easier to understand if the null case is ruled out closer to where it came from.

src/main/java/backend/github/ReplaceIssueMilestoneTask.java
```diff
@@ -0,0 +1,38 @@
...
+        this.issueTitle = issueTitle;
+    }
+
+    @Override
+    public void run() {
+        try {
+            Issue result = repo.setMilestone(repoId, issueId, issueTitle, issueMilestone);
+
+            if (result.getMilestone() == null) {
+                response.complete(issueMilestone == null);
```
dariusf: Same here. At this point we shouldn't be checking if our input was null.

src/main/java/backend/github/ReplaceIssueMilestoneTask.java
```diff
@@ -0,0 +1,38 @@
...
+            } else {
+                response.complete(result.getMilestone().getNumber() == issueMilestone);
+            }
+        } catch (IOException e) {
```
dariusf: We should try to minimise the scope of this block. No need to wrap everything if only setMilestone will throw.

src/main/java/backend/resource/Model.java
```diff
@@ -211,6 +211,22 @@ private void ______OPERATIONS_____() {}
...
 
+    /**
+     * Replaces the milestone of an issue specified by {@code issueId} with {@code milestone}
+     * @param issueId
+     * @param milestone
+     * @return the modified TurboIssue if successful
+     */
+    public synchronized Optional<TurboIssue> replaceIssueMilestone(int issueId, Integer milestone) {
+        Optional<TurboIssue> issueLookUpResult = getIssueById(issueId);
+        return Utility.safeFlatMapOptional(issueLookUpResult,
```
dariusf: I know this is copied from existing code, but I'm beginning to think flatmapping here is a bad idea, as the issue-not-found case is a programmer error. Could we take the chance to change this to an exception, and make sure it's handled somewhere? Then everything else on top of this won't have to propagate the optional.

src/main/java/backend/resource/MultiModel.java
```diff
@@ -106,6 +106,20 @@ public synchronized MultiModel replace(Model newModel) {
...
+    /**
+     * Replaces the milestone of an issue specified by {@code issueId} in {@code repoId} with {@code labels}
+     * @param repoId
+     * @param issueId
+     * @param milestone
+     * @return the modified TurboIssue if successful
+     */
+    public synchronized Optional<TurboIssue> replaceIssueMilestone(String repoId, int issueId, Integer milestone) {
+        Optional<Model> modelLookUpResult = getModelById(repoId);
+        return Utility.safeFlatMapOptional(modelLookUpResult,
```
dariusf: The exception should propagate out of here.

dariusf:
Nice work with this. :+1:

src/main/java/backend/resource/TurboIssue.java
m133225: Not too sure what you mean... A null milestone is a valid input, when the user wishes to demilestone an issue.

src/main/java/backend/Logic.java
m133225: The milestone variables in our TurboIssue & SerializableIssue are all in the Optional form, while the GitHub API uses an integer or null to indicate the presence or absence of a milestone. After this step, all that matters is the the new value being assigned.

src/main/java/backend/resource/TurboIssue.java
dariusf: The problem with passing null is that it doesn't inherently mean 'demilestone an issue'. Here we're using it to mean that, but it could also be a stray null that someone mistakenly passed in.

To be explicit, maybe we could have another method, `removeMilestone`? Then we can assume that this method will always have a non-null argument, representing a valid milestone.

src/main/java/backend/Logic.java
dariusf: I understand, but code at this level (and at all the levels in between) shouldn't know that deep down, the GitHub API uses null values. For example, if we someday added support for BitBucket and their API didn't use null values, we would have to change everything in between. Ideally we should only use things like null at the very lowest level, and abstract it away otherwise.

src/main/java/ui/components/pickers/MilestonePicker.java
m133225: The 3 cases should be
- Cancel, nothing selected
- Confirm, nothing selected
- Confirm, something selected

So it is important to distinguish whether there is the presence of the result or not. Will change this.

src/main/java/ui/components/pickers/PickerMilestone.java
m133225: There is a mutator method setDescription which may be set it to null if used incorrectly.

dariusf: That is true, but I don't think we need to be defensive here -- we can just assume TurboMilestone follows its contract. We *could* add a null check to `TurboMIlestone.setDescription` and have it fail early there.

src/main/java/backend/github/ReplaceIssueMilestoneTask.java
m133225: This part here is checking the _resulting issue_ returned from GitHub, whether it was correctly/successfully changed.

I think the problem here is the way the GitHub API handles milestones, it either has to be:
- issue.getMilestone() is null if no milestone is assigned to the issue
- issue.getMilestone().getNumber() is the milestone's id if a milestone is assigned to the issue

So the checking of null is because we have to separately handle the case when no milestone is assigned, and we cannot just invoke issue.getMilestone().getNumber() since getMilestone may return null.

src/main/java/backend/resource/Model.java
m133225: Agree. But what should the Logic do to handle this exception? If the milestone change is not reverted, it might cause problems with the models later on (at the moment, this seems to be the case if the label/milestone revert fails.)

m133225: Actually, now that I'm in the midst of trying to change this, I realised that a lot of things will need to be changed, including label change logic flow since it will affect shared methods (e.g. Logic's getIssue). I think this should be done in a separate PR.

src/main/java/ui/components/pickers/PickerMilestone.java
m133225: Hmm, but the way PickerMilestone is used now, there will only be 1 highlighted milestone at all times. Should we simply put this (1 highlighted milestone at all times) as an assumption or change the compareTo to account for the possibility of having 2 highlighted milestones.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
m133225: I think doing that will unnecessarily increase the coupling. MilestonePickerDialog should not need to know of the existing models.

m133225:
Ready for review.
I think most of the coverage drop is due to adding functionalities in DummyRepoState, and some of which are covered in the unstable tests.

The change of making being unable to find an issue/model into an exception will be done in another  issue.

damithc:
Change the label to `toReview` as well?

m133225:
This is the updated milestone picker after today's discussion:
- All matching milestones are shown in a scrollable pane
- Text field is filled with the assigned milestone by default
- Clicking is only used for assigning a milestone i.e. clicking cannot be used to unassign a milestone
- Space no longer means 'confirm', but simply separates different query words for the same desired milestone

src/main/java/backend/resource/Model.java
dariusf: Okay, sure.

src/main/java/backend/github/ReplaceIssueMilestoneTask.java
dariusf: I actually meant comparing the issueMilestone Integer with null. But it seems that's been changed?

src/main/java/ui/components/pickers/PickerMilestone.java
dariusf: You're right that here it will probably never matter, but I'd lean towards a proper compareTo. Mostly so there are no surprises, and so we don't have to go to the trouble of maintaining that invariant (which is outside this class). Seems you've changed this also. :stuck_out_tongue:

src/main/java/ui/components/pickers/MilestonePickerDialog.java
dariusf: We could check if their repo ids are equal maybe? That's info we can access from the data we have here.

src/main/java/backend/Logic.java
```diff
@@ -271,6 +271,27 @@ public Model getRepo(String repoId) {
...
+    public CompletableFuture<Boolean> replaceIssueMilestone(TurboIssue issue, Optional<Integer> newMilestone) {
+        logger.info("Changing milestone for " + issue + " in models");
+        CompletableFuture<Optional<TurboIssue>> localMilestoneReplaceFuture =
+                repoOpControl.replaceIssueMilestoneLocally(issue, newMilestone);
+        localMilestoneReplaceFuture.thenRun(this::refreshUI);
+
+        return updateIssueMilestonesOnServer(issue, newMilestone)
+                .thenCombine(localMilestoneReplaceFuture,
+                        (isUpdateSuccessful, originalIssue) ->
+                                handleIssueMilestoneUpdateOnServerResult(isUpdateSuccessful, originalIssue));
```
dariusf: This bit can be replaced with `this::handleIssueMilestoneUpdateOnServerResult`.

src/main/java/backend/github/GitHubRepo.java
```diff
@@ -260,6 +262,30 @@ public boolean login(UserCredentials credentials) {
...
+        Issue createdIssue = new Issue();
+        createdIssue.setNumber(issueId);
+        createdIssue.setTitle(issueTitle);
+
+        Milestone gitHubMilestone = new Milestone();
+        // set milestone number to the desired milestone id
+        // simply don't set a number to demilestone
+        if (issueMilestone.isPresent()) {
+            gitHubMilestone.setNumber(issueMilestone.get());
+        }
```
dariusf: This part can be `issueMilestone.ifPresent(gitHubMilestone::setNumber);`.

src/main/java/backend/stub/DummyRepo.java
```diff
@@ -183,6 +181,14 @@ private DummyRepoState getRepoState(String repoId) {
     }
 
     @Override
+    public Optional<Integer> setMilestone(String repoId, int issueId, String issueTitle,
+                                          Optional<Integer> issueMilestone) {
+        Issue returnedIssue = getRepoState(repoId).setMilestone(issueId, issueMilestone);
+        if (returnedIssue.getMilestone() == null) return Optional.empty();
+        return Optional.of(returnedIssue.getMilestone().getNumber());
```
dariusf: Same here.

src/main/java/backend/github/GitHubRepo.java
```diff
@@ -260,6 +262,30 @@ public boolean login(UserCredentials credentials) {
...
+        createdIssue.setMilestone(gitHubMilestone);
+
+        Issue returnedIssue = issueService.editIssue(RepositoryId.createFromId(repoId), createdIssue);
+
+        if (returnedIssue.getMilestone() == null) {
+            return Optional.empty();
+        }
+        return Optional.of(returnedIssue.getMilestone().getNumber());
```
dariusf: This is a bit more succinct.

```java
return Optional.ofNullable(returnedIssue.getMilestone())
    .map(Milestone::getNumber);
```

src/main/java/ui/components/pickers/MilestonePicker.java
```diff
@@ -0,0 +1,44 @@
...
+        ui.registerEvent((ShowMilestonePickerEventHandler) e -> Platform.runLater(() -> showMilestonePicker(e.issue)));
+    }
+
+    private void showMilestonePicker(TurboIssue issue) {
+        List<TurboMilestone> milestones = ui.logic.getRepo(issue.getRepoId()).getMilestones();
+        MilestonePickerDialog dialog = new MilestonePickerDialog(stage, issue, milestones);
+        Optional<Pair<ButtonType, Optional<Integer>>> milestoneDialogResponse = dialog.showAndWait();
+
+        if (!milestoneDialogResponse.isPresent() ||
+                milestoneDialogResponse.get().getKey().equals(ButtonType.CANCEL)) {
```
dariusf: This method feels like it could be a bit more SLAP. We can use an inner class instead of the tuple, with a method `wasCancelled` maybe?

The line above can be

```java
Optional<MilestonePickerDialog.Response> response =
    new MilestonePickerDialog(stage, issue, milestones).showAndWait();
```
to cut as much noise as possible.

Registering the undo action also seems like it could be pulled out into a method.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -0,0 +1,287 @@
...
+        initUI();
+        setupKeyEvents();
+
+        fillInputFieldWithExistingMilestone(originalMilestones);
+    }
+
+    private void fillInputFieldWithExistingMilestone(List<PickerMilestone> milestoneList) {
+        Optional<PickerMilestone> existingMilestone = getExistingMilestone(milestoneList);
+        if (!existingMilestone.isPresent()) return;
+        inputField.setText(existingMilestone.get().getTitle());
```
dariusf: We can use ifPresent here also.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -0,0 +1,287 @@
...
+                });
+
+        Collections.sort(originalMilestones);
+
+        return originalMilestones;
+    }
+
+    private boolean isExistingMilestone(TurboIssue issue, PickerMilestone milestone) {
+        if (!issue.getMilestone().isPresent()) return false;
+        return issue.getMilestone().get() == milestone.getId();
```
dariusf: orElse can be used here.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -0,0 +1,111 @@
...
+    public MilestonePickerState(List<PickerMilestone> milestones, String userInput) {
+        this(milestones);
+        processInput(userInput);
+    }
+
+    private List<PickerMilestone> cloneList(List<PickerMilestone> sourceList) {
+        List<PickerMilestone> destList = new ArrayList<>();
+        sourceList.stream()
+                .forEach(milestone -> destList.add(new PickerMilestone(milestone)));
+        return destList;
```
dariusf: Can we use `new ArrayList<>(sourceList)` instead?

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -0,0 +1,111 @@
...
+     * @return
+     */
+    public List<PickerMilestone> getMatchingMilestonesList() {
+        return this.currentMilestonesList.stream()
+                .filter(milestone -> !milestone.isMatching())
+                .collect(Collectors.toList());
+    }
+
+    private void filterMilestones(String query) {
+        currentMilestonesList.stream()
```
dariusf: We can just use `forEach` here, no need for the intermediate `stream`.

dariusf:
That was fast. Anyway, mostly just cosmetics. You might want to squash noisy commits (refactoring, fixing style errors, stuff changed in review) as well.

src/main/java/ui/components/pickers/MilestonePickerState.java
m133225: I don't think so, that will make it refer to the same object right? With each change in input field, the original list of milestones is re-supplied, so we can't mutate them.

dariusf: Oh, this does a deep clone. Missed that.

Instead of the `forEach`, you could also map `PickerMilestone::new` over the list.

src/main/java/backend/Logic.java
```diff
@@ -312,6 +338,26 @@ private boolean handleIssueLabelsUpdateResult(boolean isUpdateSuccessful,
...
+     * @param originalIssue
+     * @return true if the server update is successful
+     */
+    private boolean handleIssueMilestoneUpdateOnServerResult(boolean isUpdateSuccessful,
+                                                          Optional<TurboIssue> originalIssue) {
+        if (!originalIssue.isPresent()) {
+            logger.error("Unable to replace issue milestone locally");
+            return false;
+        }
+        if (isUpdateSuccessful) return true;
```
damithc: This seems to be the lowest hanging fruit, so should be the first one to be picked?

damithc: Ok, never mind, I see why you have it here.

src/main/java/backend/Logic.java
```diff
@@ -312,6 +338,26 @@ private boolean handleIssueLabelsUpdateResult(boolean isUpdateSuccessful,
...
+     * @param isUpdateSuccessful
+     * @param originalIssue
+     * @return true if the server update is successful
+     */
+    private boolean handleIssueMilestoneUpdateOnServerResult(boolean isUpdateSuccessful,
+                                                          Optional<TurboIssue> originalIssue) {
+        if (!originalIssue.isPresent()) {
+            logger.error("Unable to replace issue milestone locally");
+            return false;
+        }
```
damithc: When can this happen? Is this an exceptional situation? or a programmer error?

src/main/java/backend/Logic.java
```diff
@@ -331,6 +377,26 @@ private void revertLocalLabelsReplace(TurboIssue modifiedIssue, List<String> ori
...
+     * with originalIssue's milestone if both issues have the same last modified LocalDateTime
+     *
+     * @param originalIssue
+     */
+    private void revertLocalMilestoneReplace(TurboIssue originalIssue) {
+        TurboIssue currentIssue = getIssue(originalIssue.getRepoId(), originalIssue.getId()).orElse(originalIssue);
+        LocalDateTime originalMilestoneModifiedAt = originalIssue.getMilestoneLastModifiedAt();
+        LocalDateTime currentMilestoneAssignedAt = currentIssue.getMilestoneLastModifiedAt();
+        boolean isCurrentMilestoneModifiedFromOriginalMilestone = originalMilestoneModifiedAt.isEqual(
+                currentMilestoneAssignedAt);
```
damithc: This is more readable (break at the highest level possible):
```java
boolean isCurrentMilestoneModifiedFromOriginalMilestone = originalMilestoneModifiedAt
                                                               .isEqual(currentMilestoneAssignedAt);
```

src/main/java/backend/Logic.java
m133225: I think it is a programmer error. This will be handled in another PR (handle an exception thrown when the original issue cannot be found in the first place).

damithc: OK. programmer error should be handled by an assertion.

damithc:
Let us know after you've done the requested changes.

m133225:
Done!

src/main/java/backend/Logic.java
```diff
@@ -271,6 +271,25 @@ public Model getRepo(String repoId) {
     }
 
     /**
+     * Replaces existing milestone with the newMilestone in the issue object, the UI, and the server, in that order.
+     * Server update is done after the local update to reduce the lag between the user action and the UI response
+     *
+     * @param issue The issue object containing the repoid and issue id of the relevant issue
+     *              whose milestone is to be replaced
```
damithc: This part can be omitted? `containing the repoid and issue id of the relevant issue`

src/main/java/backend/Logic.java
```diff
@@ -271,6 +271,25 @@ public Model getRepo(String repoId) {
...
+     * @param newMilestone The new milestone to be assigned to the issue
+     * @return true if milestone replacements locally and on GitHub were successful, false otherwise.
```
damithc: `false otherwise` is implied?

src/main/java/backend/stub/DummyRepoState.java
```diff
@@ -400,6 +398,38 @@ protected TurboUser deleteUser(String idString) {
...
+        }
+        toSet.setUpdatedAt(LocalDateTime.now());
+
+        markUpdatedEvents(toSet, IssueMetadata.intermediate(eventsOfIssue, metadataOfIssue.getComments(), "", ""));
+
+        Issue newIssue = new Issue();
+        if (milestone.isPresent()) {
+            newIssue.setMilestone(new Milestone().setNumber(milestone.get()));
+        }
+        return newIssue;
```
damithc: The high level steps are not immediately clear. May be add `//comments` to each block of code?

src/main/java/ui/components/pickers/MilestonePickerDialog.java
damithc: @m133225 @m133225 what's the outcome of this comment thread?

src/main/java/ui/components/pickers/MilestonePickerState.java
damithc: @m133225 Need to address  this.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Gets the current list of milestones
+     * @return
+     */
+    public List<PickerMilestone> getCurrentMilestonesList() {
+        return this.currentMilestonesList;
+    }
+
+    /**
+     * Gets the list of milestones that matches the current query i.e. not faded
```
damithc: Does this class know about 'fading' which is a UI thing?

damithc:
Also try to add class header comments to all production (i.e. non-test) classes. As the system gets bigger, even names that sounds obvious now can become not-so-clear later.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
m133225: Just added in the latest commit an assert statement to compare their repoIds.

src/main/java/ui/components/pickers/MilestonePickerState.java
m133225: Removed in the latest commit.

m133225:
Made the changes!

src/main/java/backend/stub/DummyRepoState.java
```diff
@@ -400,6 +398,41 @@ protected TurboUser deleteUser(String idString) {
...
+    }
+
+    private void milestoneIssue(int milestone, TurboIssue toSet, List<TurboIssueEvent> eventsOfIssue) {
+        eventsOfIssue.add(new TurboIssueEvent(new User().setLogin("test"),
+                IssueEventType.Milestoned,
+                new Date()).setMilestoneTitle(milestones.get(milestone).getTitle()));
+        toSet.setMilestoneById(milestone);
+    }
+
+    private void demilestoneIssue(int milestoneOfIssue, TurboIssue toSet, List<TurboIssueEvent> eventsOfIssue) {
```
damithc: Should we use `schedule` and `unschedule` instead of `milestone` and `demilestone`? In any case add header comments to these two methods.

src/main/java/ui/components/pickers/MilestonePicker.java
```diff
@@ -0,0 +1,53 @@
...
+import javafx.stage.Stage;
+import ui.UI;
+import undo.actions.ChangeMilestoneAction;
+import util.events.ShowMilestonePickerEventHandler;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * This class is meant to facilitate creating a dialog when needed for the user to set milestone for an issue
```
damithc: Simplify and shorten. This is probably enough: `A dialog for selecting a milestone`

damithc: Actually, how is this different from `MilestonePickerDialog`?

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -0,0 +1,296 @@
...
+import javafx.stage.Stage;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * This class is meant to handle the display of milestone selection status as well as handle the user's inputs
+ */
```
damithc: Get rid of 'dead wood' words e.g. `This class is meant to handle` --> `Handles`

src/main/java/ui/components/pickers/MilestonePickerDialogResponse.java
```diff
@@ -0,0 +1,27 @@
+package ui.components.pickers;
+
+import javafx.scene.control.ButtonType;
+
+import java.util.Optional;
+
+/**
+ * This class is meant for the MilestonePickerDialog to pass the user's milestone selection back to the
+ * MilestonePicker for processing
```
damithc: Write based on what the class represents rather than how it is used currently. It represents a milestone related button action on a UI. As you can see, it is hard to describe because it is not a very good abstraction. Anyway, we may have to live with it for now.

damithc:
Can merge after addressing the comments. I didn't comment everywhere though. Do the necessary revision in all places a comment applies to.

src/main/java/ui/components/pickers/MilestonePicker.java
m133225: Rather than the dialog itself, it's just the class that registers the handler to open the dialog, as well as trigger the necessary action according to the dialog's result.

PR #1363, opened 2016-03-15
Fix filter field indicating parse errors on incomplete inputs #1360

### LINKED ISSUE #1360: Fix filter field indicating parse errors on incomplete inputs
For example, `id:` is an incomplete but otherwise valid input.

We need this because inputs are validated on the fly, and there are false positives from the parser having too strict a notion of syntactic validity.

dariusf:
Fixes #1360

### DIFF
--- a/src/main/java/filter/Parser.java
+++ b/src/main/java/filter/Parser.java
@@ -17,24 +17,62 @@ import util.Utility;
 
 public final class Parser {
 
-    private Parser(ArrayList<Token> input) {
-        this.input = input;
-    }
-    public static FilterExpression parse(String input) {
-        if (input == null || input.isEmpty()) return Qualifier.EMPTY;
-        return new Parser(new Lexer(input).lex()).parseExpression(0);
-    }
+    private final List<Token> input;
 
-    private final ArrayList<Token> input;
+    /**
+     * Determines if the parser should fail gracefully if it finds the input
+     * in a incomplete state. See {@link #check}.
+     */
+    private final boolean isCheckingSyntax;
+
     private int position = 0;
     private int sourcePosition = 0;
 
+    private Parser(List<Token> input, boolean isCheckingSyntax) {
+        this.input = input;
+        this.isCheckingSyntax = isCheckingSyntax;
+    }
+
+    /**
+     * Parses an input string, returning an abstract syntax tree.
+     * Throws a {@link ParseException} on syntax error.
+     */
+    public static FilterExpression parse(String input) {
+        if (input.isEmpty()) {
+            return Qualifier.EMPTY;
+        }
+        return new Parser(new Lexer(input).lex(), false).parseExpression(0);
+    }
+
+    /**
+     * Checks an input string for errors. Returns true if the string is at
+     * least partially valid, i.e. perfectly valid, or valid but incomplete.
+     */
+    public static boolean check(String input) {
+        if (input.isEmpty()) {
+            return true;
+        }
+
+        try {
+            new Parser(new Lexer(input).lex(), true).parseExpression(0);
+            return true;
+        } catch (IncompleteInputException ignored) {
+            return true;
+        } catch (ParseException ignored) {
+            return false;
+        }
+    }
+
+    private static class IncompleteInputException extends RuntimeException {}
+
     private Token consume(TokenType type) {
-        if (input.get(position).getType() == type) {
-            return input.get(position++);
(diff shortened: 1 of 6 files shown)


dariusf:
@ndt93 could you help to review?

src/main/java/filter/Parser.java
```diff
@@ -17,24 +17,62 @@
...
 
-    private Parser(ArrayList<Token> input) {
+    private final List<Token> input;
+
+    /**
+     * Determines if the parser should fail gracefully if it finds the input
+     * in a incomplete state. See {@link #check}.
+     */
+    private final boolean inCheckMode;
+
```
ndt93: Should a boolean flag be named `is...`?

src/main/java/ui/components/FilterTextField.java
```diff
@@ -55,12 +55,19 @@ private void setup() {
         validationSupport.registerValidator(this, (c, newValue) -> {
             boolean wasError = false;
             try {
-                Parser.parse(getText());
-                setStyleForValidFilter();
+                if (!Parser.check(getText())) {
+                    wasError = true;
+                }
```
ndt93: Naming the variable as `isError` should be better? Also, what is the advantage of having this `check` method instead of catching 2 types of exceptions here?

dariusf: Changed the name. I thought the interface would be nicer. This way the exception is kept within the parser, and callers have to worry about less (syntax is either correct or not).

src/test/java/tests/FilterParserTests.java
```diff
@@ -52,69 +57,77 @@ public void quotes() {
...
+    @Test(expected = ParseException.class)
+    public void parse_nothingAfterInfixOperator_exceptionThrown() {
+        Parser.parse("a(b) ||");
+    }
 
-        try {
-            Parser.parse("-milestone: label:priority.high");
-            fail("'label:' can't be the input to a qualifier => empty qualifier => parse exception");
-        } catch (ParseException ignored) {}
+    @Test(expected = ParseException.class)
```
damithc: I've seen another way to handle expected exceptions in our tests. Should standardize? http://blog.goyello.com/2015/10/01/different-ways-of-testing-exceptions-in-java-and-junit/

src/main/java/filter/Parser.java
```diff
@@ -17,24 +17,62 @@
...
 public final class Parser {
 
-    private Parser(ArrayList<Token> input) {
+    private final List<Token> input;
+
+    /**
+     * Determines if the parser should fail gracefully if it finds the input
+     * in a incomplete state. See {@link #check}.
+     */
+    private final boolean isInCheckMode;
```
damithc: Better name and/or comment? We should explain somewhere what the 'check mode' means.
The comments leans a bit towards 'how it is used' rather than 'what it means'.

src/main/java/filter/Parser.java
```diff
@@ -17,24 +17,62 @@
...
+        }
+
+        try {
+            new Parser(new Lexer(input).lex(), true).parseExpression(0);
+            return true;
+        } catch (IncompleteInputException ignored) {
+            return true;
+        } catch (ParseException ignored) {
+            return false;
+        }
```
damithc: Using exceptions to control 'normal' workflow is not exactly right, but may be there is no choice.

damithc:
Waiting for clarifications

src/test/java/tests/FilterParserTests.java
dariusf: Okay, I can change it to use ExpectedException. Maybe someone can look into AssertJ in future, as that seems to be more generic and doesn't require fields to be defined.

src/main/java/filter/Parser.java
dariusf: Maybe `isCheckingSyntax`? 'Check mode' is defined purely by how the flag is used, so maybe we shouldn't try to make it a thing.

src/main/java/filter/Parser.java
dariusf: Yes, it's used for control flow here. The idea is that if the parser enters certain states, we can return true immediately because we know that the input is incomplete.

The problem is that parseExpression is called recursively from many places, so if we used its return value to indicate this condition, we'd have to change every function signature and add code to propagate it.

This is a bit of an abuse of exceptions but I thought it would be clearer overall.

PR #1367, opened 2016-03-16
'go to file view' keyboard shortcut is broken #1364

### LINKED ISSUE #1364: 'go to file view' keyboard shortcut is broken
`g,f` shortcut used to go to the diff view of a PR seems to be broken, possibly due to a change in the webpage structure.

HansNewbie:
Fixes #1364

### DIFF
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -519,21 +519,24 @@ public class BrowserComponent {
      */
     public void switchToTab(String tabName) {
         if (GitHubURL.isPullRequestLoaded(getCurrentUrl())) {
-            String xpath = "";
+            int tabIndex = 0;
 
             switch(tabName) {
                 case GithubPageElements.DISCUSSION_TAB:
-                    xpath = "//*[@id=\"js-repo-pjax-container\"]/div[2]/div[1]/div/div[2]/div[2]/nav/a[1]";
+                    tabIndex = 1;
                     break;
                 case GithubPageElements.COMMITS_TAB:
-                    xpath = "//*[@id=\"js-repo-pjax-container\"]/div[2]/div[1]/div/div[2]/div[2]/nav/a[2]";
+                    tabIndex = 2;
                     break;
                 case GithubPageElements.FILES_TAB:
-                    xpath = "//*[@id=\"js-repo-pjax-container\"]/div[2]/div[1]/div/div[2]/div[2]/nav/a[3]";
+                    tabIndex = 3;
                     break;
+                default:
+                    return;
             }
 
-            driver.findElement(By.xpath(xpath)).click();
+            String xpath = "//*[@id=\"js-repo-pjax-container\"]/div[2]/div[1]/div/div[2]/div[2]/nav/a[%d]";
+            driver.findElement(By.xpath(String.format(xpath, tabIndex))).click();
         }
     }
 
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -517,7 +517,7 @@ public class BrowserComponent {
     }
 
     /**
-     * Switch to the specified tab in GitHub PR page
+     * Switches to the specified tab in GitHub PR page
      * @param tabName Either GithubPageElements.DISCUSSION_TAB, GithubPageElements.COMMITS_TAB
      *                or GithubPageElements.FILES_TAB
      */
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -50,6 +50,10 @@ public class BrowserComponent {
     private static final int SWP_NOMOVE = 0x0002;
     private static final int SWP_NOACTIVATE = 0x0010;
 
+    /**
+     * The two variables below are used to wait for a webpage to load, where a wait consist of
+     * Thead.sleep() by sleep time, repeated until the webpage is loaded or max retry is reached.
+     */
     private static final int WAIT_PAGE_LOAD_MAX_RETRY = 20;
     private static final int WAIT_PAGE_LOAD_SLEEP_TIME = 10;
 
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -51,10 +51,12 @@ public class BrowserComponent {
     private static final int SWP_NOACTIVATE = 0x0010;
 
     /**
-     * The two variables below are used to wait for a webpage to load, where a wait consist of
-     * Thead.sleep() by sleep time, repeated until the webpage is loaded or max retry is reached.
+     * Specifies how many tries to wait for a web page to load
      */
     private static final int WAIT_PAGE_LOAD_MAX_RETRY = 20;
+    /**
+     * Specifies the duration between tries to wait for a web page to load
+     */
     private static final int WAIT_PAGE_LOAD_SLEEP_TIME = 10;
 
     private static HWND browserWindowHandle;
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -508,12 +508,16 @@ public class BrowserComponent {
         return driver != null && GitHubURL.isUrlIssue(driver.getCurrentUrl());
     }
 
+    public boolean isCurrentUrlPrDiscussion() {
+        return driver != null && GitHubURL.isPullRequestDiscussionPageLoaded(driver.getCurrentUrl());
+    }
+
     public String getCurrentUrl() {
         return driver.getCurrentUrl();
     }
 
     /**
-     * Switch to the tabs in GitHub PR page
+     * Switch to the specified tab in GitHub PR page
      * @param tabName Either GithubPageElements.DISCUSSION_TAB, GithubPageElements.COMMITS_TAB
      *                or GithubPageElements.FILES_TAB
      */
@@ -532,12 +536,17 @@ public class BrowserComponent {
                     tabIndex = 3;
                     break;
                 default:
+                    assert false;
                     return;
             }
 
             String xpath = "//*[@id=\"js-repo-pjax-container\"]/div[2]/div[1]/div/div[2]/div[2]/nav/a[%d]";
-            driver.findElement(By.xpath(String.format(xpath, tabIndex))).click();
+            clickElementByXpath(String.format(xpath, tabIndex));
         }
+    }
+
+    private void clickElementByXpath(String xpath) {
+        driver.findElement(By.xpath(xpath)).click();
     }
 
     public Optional<Integer> getPRNumberFromIssue() {
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -514,7 +514,21 @@ public class BrowserComponent {
 
     public void switchToTab(String tabName) {
         if (GitHubURL.isPullRequestLoaded(getCurrentUrl())) {
-            driver.findElement(By.xpath("//a[@data-container-id='" + tabName + "_bucket']")).click();
+            String xpath = "";
+
+            switch(tabName) {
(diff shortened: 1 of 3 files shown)


HansNewbie:
Ready for review

acruis:
Good to merge after travis passes. I don't think this can be tested.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -512,9 +512,31 @@ public String getCurrentUrl() {
         return driver.getCurrentUrl();
     }
 
+    /**
+     * Switch to the tabs in GitHub PR page
```
damithc: `Switches to the specified tab ...`

src/main/java/browserview/BrowserComponent.java
```diff
@@ -512,9 +512,31 @@ public String getCurrentUrl() {
...
+                    tabIndex = 1;
+                    break;
+                case GithubPageElements.COMMITS_TAB:
+                    tabIndex = 2;
+                    break;
+                case GithubPageElements.FILES_TAB:
+                    tabIndex = 3;
+                    break;
+                default:
+                    return;
```
damithc: Should this be an assertion?

src/main/java/browserview/BrowserComponent.java
```diff
@@ -512,9 +512,31 @@ public String getCurrentUrl() {
...
+            }
+
+            String xpath = "//*[@id=\"js-repo-pjax-container\"]/div[2]/div[1]/div/div[2]/div[2]/nav/a[%d]";
+            driver.findElement(By.xpath(String.format(xpath, tabIndex))).click();
```
damithc: This line can be abstracted to a `click(xpath)` method somewhere (e.g. in BrowserView object)?

damithc:
Some changes requested

damithc:
Looks like the `R` keyboard shortcut (used for 'reply') is broken as well. May be check all the shortcuts and fix any broken ones?

HansNewbie:
Alright.

HansNewbie:
@damithc all the shortcuts in browser seems to work for me, including`R`. As of current implementation, `R` only works for issue; it is not implemented for PR. Do you want me to add it?

HansNewbie:
@damithc so in keyboard shortcut documentation, for PR. `G` `R` will switch to discussion tab without jumping to comment box. There is no `R` shortcut for PR currently.

HansNewbie:
@damithc ignore my comment, I missed some regex. Will look into `R` again.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -212,10 +212,17 @@ private void setupKeyboardShortcuts() {
...
-                } else if (ui.getBrowserComponent().isCurrentUrlIssue()) {
-                    ui.getBrowserComponent().switchToTab(DISCUSSION_TAB);
+                } else {
+                    if (!ui.getBrowserComponent().isCurrentUrlPrDiscussion()) {
+                        ui.getBrowserComponent().switchToTab(DISCUSSION_TAB);
+                    }
+
+                    while (!ui.getBrowserComponent().isCurrentUrlPrDiscussion()) {
+                        // wait
+                    }
```
HansNewbie: @acruis I would like to wait until a url is loaded in browser. Is there any method to do so?

HansNewbie:
@damithc I think `R` for PR is broken in the first place. So the problem is that on pressing `R`, we will open the Discussion tab then jump to comment. 2 cases here:
1. If we are on commit or files tab, we might jump to comment before discussion tab is loaded. Hence it fails.
2. If we are on discussion tab, we will still load the discussion tab. So what happens is that we jump to comment box, *then* the discussion tab is loaded again. Something like: [ask browser to open discussion tab URL] > [jump to comment box] > [discussion tab is loaded].

What I did is make explicit wait until website is opened. But I am using empty while loop which is bad. Waiting for @acruis if he has suggestion how to wait.

src/main/java/ui/listpanel/ListPanel.java
acruis: I think there's no choice here but to busy wait. Maybe on a different thread?

HansNewbie:
Update: All browser keyboard shortcuts work.

The only behavior that may not be as expected is pressing `G` `C` when on PR card, but the browser is on other pages. e.g. user is on PR. Then he pressed `G` `H` to open help page. Then he pressed `G` `C` to open commit page. @damithc I am not sure what the expected output should be here.

Current expected output: the page not loaded since the bView is not in PR url.
Probably expected output: doesn't matter what bView currently is at, commit tab should still be loaded.
The alternative to achieve the probably expected output is to press `Enter` first to load PR url page, then press `G` `C`.

src/main/java/ui/listpanel/ListPanel.java
damithc: Thread.sleep(...)?

src/main/java/browserview/BrowserComponent.java
```diff
@@ -508,16 +508,47 @@ public boolean isCurrentUrlIssue() {
...
+    public boolean isCurrentUrlPrDiscussion() {
+        return driver != null && GitHubURL.isPullRequestDiscussionPageLoaded(driver.getCurrentUrl());
+    }
+
     public String getCurrentUrl() {
         return driver.getCurrentUrl();
     }
 
+    /**
+     * Switch to the specified tab in GitHub PR page
```
damithc: `Switches`

damithc:
Yes, ideally, the shortcut should give the right result even if the right page is not loaded. But it can be done as a separate issue.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -212,10 +212,20 @@ private void setupKeyboardShortcuts() {
...
+                        ui.getBrowserComponent().switchToTab(DISCUSSION_TAB);
+                    }
+
+                    Thread jumpToCommentOncePageLoaded = new Thread(() -> {
+                        while (!ui.getBrowserComponent().isCurrentUrlPrDiscussion()) {
+                        // wait
+                        }
+                        ui.getBrowserComponent().jumpToComment();
+                    });
+                    jumpToCommentOncePageLoaded.start();
```
HansNewbie: @acruis will this be better or would `thread.sleep()` be better? I am thinking of using shortcut to load webpages in quick succession, in which using `thread.sleep` would be better.

Something like:
```
while (//page not loaded) {
  Thread.sleep((100); // low value to not affect responsiveness
}
```

HansNewbie:
I will do it in this issue as well.

src/main/java/ui/listpanel/ListPanel.java
acruis: `Thread.sleep` would be better, I think. @dariusf any comments?

HansNewbie: Met him just now. Will use Thread.sleep.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -530,6 +567,20 @@ public void switchToTab(String tabName) {
...
+            if (!isCurrentUrlDiscussion()) {
+                try {
+                    Thread.sleep(WAIT_PAGE_LOAD_SLEEP_TIME);
+                } catch (IllegalArgumentException | InterruptedException e) {
+                    logger.warn(e);
+                }
+            } else {
+                return;
+            }
+        }
```
HansNewbie: @acruis Here is my waiting. @dariusf suggested that should give limit to how long waiting so it won't get stuck.

HansNewbie:
*I am not sure if this is a good idea, and I can't guess the original intention of the methods.*

I have implemented such that pressing `G``R`, `G``C`, `G``F` and `R` would open the relevant pages from whatever webpage it was on. However, I am still following the original implementation where it switches tab in PR, so if user is opening keyboard shortcut page, pressing `G``C` will open the PR page first before switching tab, which takes time and user can see the steps as opening the PR page first before changing tabs.

I am thinking of reworking the `switchTab(tabsName)` to `openTab(prId, tabsName)` instead, so user won't see the steps of opening PR page before changing tabs. Instead, it will open the relevant PR tab directly. @damithc @acruis Do you think this is a good idea?

damithc:
Yes, that may be better, but may be as a separate issue?

HansNewbie:
@damithc Sure.

HansNewbie:
@acruis ready for review.

src/main/java/browserview/BrowserComponent.java
acruis: OK, the logic looks good. Do you think you can improve SLAP for this method? For example, combine the `for` and `if` into a single `for` or `while` statement.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -530,6 +567,20 @@ public void switchToTab(String tabName) {
...
+        for (int i = 0; i < WAIT_PAGE_LOAD_MAX_RETRY; i++) {
+            if (!isCurrentUrlDiscussion()) {
+                try {
+                    Thread.sleep(WAIT_PAGE_LOAD_SLEEP_TIME);
+                } catch (IllegalArgumentException | InterruptedException e) {
+                    logger.warn(e);
+                }
+            } else {
+                return;
+            }
```
damithc: Use guard clauses to reduce nesting. It's not a good sign when a big chunk of code is nested.

HansNewbie:
@acruis changes requested have been made.

damithc:
Let's merge this soon. I feel handicapped without this shortcut :-p

src/main/java/browserview/BrowserComponent.java
HansNewbie: I changed this as requested by @acruis to put the if clause into the for
condition.

acruis:
Good to merge!

src/main/java/browserview/BrowserComponent.java
```diff
@@ -47,11 +46,13 @@
...
     private static final String CHROME_DRIVER_BINARY_NAME = determineChromeDriverBinaryName();
 
-    private String pageContentOnLoad = "";
-
     private static final int SWP_NOSIZE = 0x0001;
     private static final int SWP_NOMOVE = 0x0002;
     private static final int SWP_NOACTIVATE = 0x0010;
+
+    private static final int WAIT_PAGE_LOAD_MAX_RETRY = 20;
+    private static final int WAIT_PAGE_LOAD_SLEEP_TIME = 10;
```
damithc: Add javadoc comments

damithc:
Can merge after addressing minor comment.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -47,11 +46,17 @@
...
-    private String pageContentOnLoad = "";
-
     private static final int SWP_NOSIZE = 0x0001;
     private static final int SWP_NOMOVE = 0x0002;
     private static final int SWP_NOACTIVATE = 0x0010;
+
+    /**
+     * The two variables below are used to wait for a webpage to load, where a wait consist of
+     * Thead.sleep() by sleep time, repeated until the webpage is loaded or max retry is reached.
+     */
```
HansNewbie: @damithc Just checking if the javadoc is appropriate.

damithc: I don't there there is any need to mention the implementation details. One variable specifies the duration between tries and the other specifies how many tries.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -47,11 +46,19 @@
...
     private static final int SWP_NOACTIVATE = 0x0010;
+
+    /**
+     * Specifies how many tries to wait for a web page to load
+     */
+    private static final int WAIT_PAGE_LOAD_MAX_RETRY = 20;
+    /**
+     * Specifies the duration between tries to wait for a web page to load
+     */
+    private static final int WAIT_PAGE_LOAD_SLEEP_TIME = 10;
```
HansNewbie: Comment changed.

damithc: looks good.

PR #1372, opened 2016-03-17
Add utility method for clicking list items in tests #1313

### LINKED ISSUE #1313: Add utility method for clicking list items in tests
We want to make test code more high-level, so less of the implementation leaks in.

1. We should add a generic way to generate element id strings, so test code isn't filled with ad hoc constants. This should be used to generate the strings for the actual elements as well.
2. After that, we can add more general ways to interacting with UI elements without having to build the id strings manually. A good one to start with is `clickItemAt(index)` which clicks an item in a particular panel.

m133225:
Fixes #1313

### DIFF
--- a/src/main/java/ui/IdGenerator.java
+++ b/src/main/java/ui/IdGenerator.java
@@ -27,6 +27,11 @@ public final class IdGenerator {
         return "queryField";
     }
 
+    public static String getPanelCloseButton(String repoId, int panelIndex) {
+        return repoId + "_col" + panelIndex + "_closeButton";
+    }
+
+
     public static String getPanelCellIdForTest(String repoId, int panelIndex, int issueId) {
         return "#" + getPanelCellId(repoId, panelIndex, issueId);
     }
@@ -49,5 +54,9 @@ public final class IdGenerator {
 
     public static String getLabelPickerTextFieldIdForTest() {
         return "#" + getLabelPickerTextFieldId();
+    }
+
+    public static String getPanelCloseButtonForTest(String repoId, int panelIndex) {
+        return "#" + getPanelCloseButton(repoId, panelIndex);
     }
 }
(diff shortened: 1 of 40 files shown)


m133225:
Ready for review.

The main changes are:
- Id of ALL UI components are now gotten from the IdGenerator's static methods.
- Several clicking and getting utility methods are added to UITests, such as clickPanel, getFilterTextFieldAtPanel, rightClickIssue.
However,
- I have also implemented the methods related to the label picker's text field, but I stopped there since I thought others methods (including the label picker text field's) might be redundant. Would like some review before completely removing it.

src/main/java/ui/IdGenerator.java
```diff
@@ -0,0 +1,125 @@
...
+
+    public static String getBoardNameSaveButtonId() {
+        return "boardsavebutton";
+    }
+
+    public static String getApiBoxId() {
+        return "apiBox";
+    }
+
+    public static String getPanelCellIdForTest(int panelIndex, int issueId) {
```
dariusf: We may not need the `ForTest` suffix here. These may be used in other contexts as well.

dariusf:
Nice! Are we able to get a reference to the label picker field in tests? If so, we don't need to add any related to it. Which are the duplicated methods?

m133225:
@darius Ready for review.
- Some minor changes to the method names `...ForReference` instead of `...ForTest`
- Slight changes to the UI components id e.g. label picker's text field is now `labelPickerTextField` instead of its fxml's `queryField`
- As discussed, the label picker's ids are still dependent on fxml's id unless specfically set

dariusf:
Just curious, what does `ForReference` mean?

m133225:
For reference purposes if others need to find it. :fearful: Or is there a better way to name them? ><

PR #1373, opened 2016-03-17
Move frequently failing tests to unstable #1369

### LINKED ISSUE #1369: Move frequently failing tests to unstable
There are a couple of tests that passes locally but fails rather frequently on Travis.
This causes delay when trying to merge a PR.

For example:
KeyboardShortcutsTest, particularly in line 146.
ContextMenuTests, methods 'test2' and 'test3'.
IssuePanelsTests, guiElementsTest.

m133225:
Fixes #1369.

### DIFF
--- a/src/test/java/unstable/ContextMenuTests.java
+++ b/src/test/java/unstable/ContextMenuTests.java
@@ -1,8 +1,10 @@
-package guitests;
+package unstable;
 
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+
+import guitests.UITest;
 import javafx.application.Platform;
 import javafx.scene.control.ContextMenu;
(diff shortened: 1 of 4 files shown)


m133225:
@dariusf Should RepositorySelectorTest also be moved to unstable? I think it fails somewhat frequently, and also when this PR is first built :stuck_out_tongue: . (not for me, though.)

dariusf:
Sure, I've seen that fail a few times. If there are any that you see fail randomly, you can take the chance to move them.

damithc:
What's the status of this one? From what I remember, this is blocking many of the ongoing pickers, so let's settle this fast.

m133225:
@dariusf Ready for review.

m133225:
@damithc Should be more or less finalised, unless there are problematic tests to be added that others have encountered.

dariusf:
Yeah, I don't have much more to add.

PR #1374, opened 2016-03-17
Refactor KeyboardShortcuts and fix Findbugs violations #874

### LINKED ISSUE #874: Refactor KeyboardShortcuts and fix Findbugs violations
Findbugs flags non-final static fields as problematic. The reasoning is that we have no guarantee that the fields won't be modified and thus shouldn't treat them like constants. Global state should also be avoided in general.

This issue would involve reverting this commit (TODO: add reference to commit in #866 after merge) and converting the fields into method calls. More details in commit message.

While we're at it we might want to check keyboard shortcuts at startup and flag them immediately if they are problematic, instead of giving the option to quit at runtime... This would require changes to tests. Thoughts?

leonardhml:
Fixes #874 

As discussed in #874 and #866, the relevant fields have been changed into method calls. Keyboard shortcut keys are wrapped in an enum for key lookup. This also necessitated updates in the other java files that make use of these shortcut keys.

In line 190 of `KeyboardShortcuts.java`, I returned a `null` value. Is it necessary to replace it with `Optional` instead?

### DIFF
--- a/src/main/java/ui/components/KeyboardShortcuts.java
+++ b/src/main/java/ui/components/KeyboardShortcuts.java
@@ -34,24 +34,11 @@ public final class KeyboardShortcuts {
     private static Map<String, String> keyboardShortcuts = null;
     private static Set<KeyCodeCombination> assignedKeys = null;
 
-    // customizable keyboard shortcuts
-    // ui.listpanel.ListPanel
-    public static KeyCodeCombination markAsRead;
-    public static KeyCodeCombination markAsUnread;
-
-    public static KeyCodeCombination scrollToTop;
-    public static KeyCodeCombination scrollToBottom;
-    public static KeyCodeCombination scrollUp;
-    public static KeyCodeCombination scrollDown;
-
-    //ui.issuepanel.PanelControl
-    public static KeyCodeCombination leftPanel;
-    public static KeyCodeCombination rightPanel;
-
-    // ui.components.NavigableListView
-    static KeyCodeCombination upIssue;
-    static KeyCodeCombination downIssue;
-
+    // key lookup for customizable keyboard shortcuts
+    public enum KeyboardShortcutKey {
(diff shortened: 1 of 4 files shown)


damithc:
Ready for review?

leonardhml:
Hi Prof, 
Ready for review!

damithc:
While we wait for the review, do sync your branch with upstream master. It has fallen behind.

dariusf:
I think an exception would be more appropriate.

Also, you might want to start the PR by reverting c72ef655e60529cd7f27a192c4cab1ab08d35249, which removes the FindBugs exceptions for those fields.

src/main/java/ui/components/KeyboardShortcuts.java
```diff
@@ -177,17 +164,31 @@ private static void addNonCustomizableShortcutKeys() {
...
-        scrollUp = getKeyCodeCombination("SCROLL_UP");
-        scrollDown = getKeyCodeCombination("SCROLL_DOWN");
-        leftPanel = getKeyCodeCombination("LEFT_PANEL");
-        rightPanel = getKeyCodeCombination("RIGHT_PANEL");
-        upIssue = getKeyCodeCombination("UP_ISSUE");
-        downIssue = getKeyCodeCombination("DOWN_ISSUE");
+    public static KeyCodeCombination getKeyboardShortcut(KeyboardShortcutKey key) {
+        switch (key) {
+            case MARK_AS_READ:
+                return getKeyCodeCombination("Mark_AS_READ");
```
dariusf: This repeats ourselves a bit. :stuck_out_tongue: Could we merge this method and `getKeyCodeCombination`?

ndt93:
@leonardhml Are you continuing with this PR? Otherwise we will have to close this by end of the current milestone.

dariusf:
Somewhat related to #900. We may be able to close that when this is done.

PR #1378, opened 2016-03-19
Show breakdown of total issues in panelFooter #1377

### LINKED ISSUE #1377: Show breakdown of total issues in panelFooter
To add a breakdown of the total. e.g. 4+6=10. first number can be green (open issues) and the 2nd one can be red.

yl-coder:
Ready for review
fixes #1377 

image


src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -144,6 +166,9 @@ public final void refreshItems() {
         listView.setItems(null);
         listView.setItems(getElementsList());
         issueCount = getElementsList().size();
+        closedIssueCount = getClosedIssueSize();
+        openIssueCount = issueCount - closedIssueCount;
+        System.out.println("closed = " + closedIssueCount + " open = " + openIssueCount);
```
damithc: May be we use this format instead `T (O + C)`. If Total is 0, we can omit the part inside brackets.

yl-coder: Yes agree, implemented it. it looks better now.

yl-coder:
Ready for review

ndt93:
Can you rebase the current master into this? There are a recent changes that make the tests more stable.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -35,19 +39,29 @@
...
 public class ListPanel extends FilterPanel {
 
     private static final Logger logger = HTLog.get(ListPanel.class);
 
     private final UI ui;
     private final GUIController guiController;
-    private int issueCount;
+    private int issueCount = 0;
+    private int closedIssueCount = 0;
+    private int openIssueCount = 0;
```
ndt93: These `*counts` variables should have plural form? e.g. `issuesCount`

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -12,13 +13,16 @@
...
 import javafx.scene.control.MenuItem;
 import javafx.scene.input.KeyCodeCombination;
 import javafx.scene.input.KeyEvent;
 import javafx.scene.layout.HBox;
 import javafx.scene.layout.Priority;
+import javafx.scene.text.Text;
+import javafx.scene.paint.Color;
+
 
+import javafx.scene.text.TextFlow;
```
ndt93: Not sure if your IDE settings create this but 1 blank line should be enough. Can also your IDE's automatic imports organization for these stuffs.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -35,19 +39,29 @@
...
+    private int openIssueCount = 0;
 
     private final IssueListView listView;
     private final HashMap<Integer, Integer> issueCommentCounts = new HashMap<>();
     private final HashMap<Integer, Integer> issueNonSelfCommentCounts = new HashMap<>();
 
-    private Label totalLabel;
+
+
+    Text openIssueText;
```
ndt93: Same thing there. I think 1 blank line should be readable enough to separate groups of variables.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -152,11 +181,39 @@ public final void refreshItems() {
 
     }
 
+    private int getClosedIssueSize() {
+        int count = 0;
```
ndt93: Should be `getClosedIssuesCount`? To be consistent of the rest of variable names.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -152,11 +181,39 @@ public final void refreshItems() {
...
+        for (GuiElement element: getElementsList()){
+            if (!element.getIssue().isOpen()){
+                count++;
+            }
+        }
+        return count;
+    }
```
ndt93: If you are familiar with Java 8's functional stuffs, you can do this in a more functional way like this:
```java
return getElementsList().stream().filter((element) -> !element.getIssue().isOpen()).count();
```

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -152,11 +181,39 @@ public final void refreshItems() {
...
-        totalLabel.setText(String.format("Total: %d ", issueCount));
+
+        if (issueCount == 0){
+            totalIssueText.setText(String.valueOf(0));
+            hideFooterPanelStatsDetails(true);
+        } else {
+            openIssueText.setText(String.valueOf(openIssueCount));
+            closedIssueText.setText(String.valueOf(closedIssueCount));
+            totalIssueText.setText(String.valueOf(issueCount));
+            hideFooterPanelStatsDetails(false);
```
ndt93: Should be `DRYer` to extract a method to update the texts of all 3 stats text nodes, as they should be updated together (either all or none). So I think it's not a good idea to just update the `totalIssueText` like in the `if` branch.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -152,11 +181,39 @@ public final void refreshItems() {
...
+        }
+
+    }
+
+    private void hideFooterPanelStatsDetails(boolean hide) {
+        openIssueText.setVisible(!hide);
```
ndt93: Not sure if it's a good idea to have a `hideFooter...` method and pass in a boolean to either `hide` or `show`. @damithc

ndt93:
@damithc No more comments from me.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -152,11 +179,36 @@ public final void refreshItems() {
...
+
+    }
+
+    private void updateFooterPanelStatsDetails() {
+        openIssueText.setText(String.valueOf(openIssuesCount));
+        closedIssueText.setText(String.valueOf(closedIssuesCount));
+        totalIssueText.setText(String.valueOf(issuesCount));
+    }
+
+    private void hideFooterPanelStatsDetails(boolean hide) {
```
damithc: boolean variables should be named as such. e.g. `isHidden`

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -152,11 +179,36 @@ public final void refreshItems() {
 
     }
 
+    private int getClosedIssuesCount() {
+        return (int) getElementsList().stream().filter((element) -> !element.getIssue().isOpen()).count();
+    }
```
damithc: @ndt93 is there a better place for this function? Generating the stats should be done closer to the data rather than the UI?

ndt93: There are 2 levels of data.

The lower level data would be in `MultiModel` i.e. `Logic#models`, suitable for stats across repositories or in a `Model` in `models` if the stats concern only 1 repository. This might not be a suitable place for the method as a panel displays mixed subsets of issues across multiple repositories depending on that panel's filter.

The higher level data is a `List` of `GuiElements` that represents the data-state of a panel and is stored within the `ListPanel` object itself. Ideally, we can have a class that encapsulate this, which would become a good place to put this method.

damithc: Yes, that kind of UI-data separation can be useful. We should do that at  some point.

damithc:
@yl-coder one minor change requested.

src/main/java/ui/listpanel/ListPanel.java
yl-coder: I was thinking whether we should move it to FilterPanel.java now, as it is where the List of GuiElements are now.

dariusf:
Looks good. @yl-coder could you squash the noisy commits? Things like fixing violations, things changed in review, etc.

yl-coder:
@dariusf How do i squash the commits when the commits have already been pushed to the remote? After rebase i do a force push? The previous time i did it, after rebase It says there's new commits to be pull, and after pulling everything screw up :p

dariusf:
Yes, because rebasing creates new commits, you will get something like "x ahead, x behind". You can force push after you check that everything went well.

(Just in case, you're at 7cc8721 now.)

yl-coder:
@dariusf Sorry did i do it correctly?

dariusf:
Doesn't look right, since there are others' commits in here now. Have you updated your fork to the latest master? Anyway, you can `git reset --hard 7cc8721` to get back to where you were. No need to make a new PR.

yl-coder:
I updated my fork to the latest master. It doesn't work also. It still show up others commit when i push my squash commits. i have reverted back. Is my first time squashing commits :(

damithc:
@yl-coder you may need to sync your branch with the latest master.

ndt93:
@yl-coder You can try `rebase` master into your branch first next time instead of `merge` since `merge` will interleave other commits and yours and make squashing or cleaning up commits difficult. And we have to `git push --force` after rebase, squash, amend etc. commits which had been pushed to remote since the commits history is rewritten. Ideally that shouldn't be done but since our work flow has only 1 person on a branch so it's still generally acceptable.

@dariusf So can we merge the commits here as it is?

ndt93:
Merged in ed79899

PR #1396, opened 2016-03-22
Change BrowserComponent switchToTab to openTab #1375

### LINKED ISSUE #1375: Change BrowserComponent switchToTab to openTab
Currently, `switchToTab` would switches the tab of PR page (Conversation, Commits and Files). It expects the bView to be in a PR page, otherwise it won't work. This method is problematic when bView iss not originally in a PR page, e.g. bView is in keyboard shortcut then user presses `G``F` to open PR file tab; bView will open PR page first (by default go to conversation tab) before switching tab to file tab. It is slow and user will wonder what's happening since the pages keep changing.

The proposed solution is to have `openPrTab` instead; instead of switching tab by clicking the tab in the page (as `switchToTab` is doing), open the URL to the given PR tab directly instead. May change the method signature to `openPrTab(prId, tabName)`.

As proposed in #1367.

YijinL:
Fixes #1375.

Now goes to URL instead of trying to click DOM tab elements

### DIFF
--- a/src/main/java/browserview/BrowserComponent.java
+++ b/src/main/java/browserview/BrowserComponent.java
@@ -1,5 +1,6 @@
 package browserview;
 
+import backend.resource.TurboIssue;
 import com.sun.jna.platform.win32.User32;
 import com.sun.jna.platform.win32.WinDef.HWND;
 import com.sun.jna.platform.win32.WinUser;
@@ -17,7 +18,7 @@ import ui.UI;
 import util.GitHubURL;
 import util.GithubPageElements;
 import util.PlatformSpecific;
-import util.events.testevents.JumpToCommentEvent;
+import util.events.testevents.JumpToNewCommentBoxEvent;
 import util.events.testevents.SendKeysToBrowserEvent;
 
 import java.awt.Rectangle;
@@ -240,13 +241,17 @@ public class BrowserComponent {
 
     /**
      * Navigates to the GitHub page for the given issue in the currently-active
-     * driver window.
+     * driver window. Ignores current tab if it's a PR.
      * Run on a separate thread.
      */
     public void showIssue(String repoId, int id, boolean isPullRequest, boolean isForceRefresh) {
         if (isPullRequest) {
             logger.info("Showing pull request #" + id);
-            runBrowserOperation(() -> driver.get(GitHubURL.getPathForPullRequest(repoId, id), isForceRefresh));
+            if (!isCurrentUrlPrPageIgnoreTabs(repoId, id)) {
+                runBrowserOperation(() -> driver.get(GitHubURL.getPathForPullRequest(repoId, id), isForceRefresh));
+            } else {
+                logger.info("Already at PR page: " + getCurrentUrl() + " will not reload.");
+            }
         } else {
             logger.info("Showing issue #" + id);
             runBrowserOperation(() -> driver.get(GitHubURL.getPathForIssue(repoId, id), isForceRefresh));
@@ -254,9 +259,34 @@ public class BrowserComponent {
         runBrowserOperation(() -> scrollToBottom());
     }
 
-    public void jumpToComment() {
+    /**
+     * Checks if current URL is the specified PR's page, regardless of the PR tab the URL is in.
+     *
+     * @return true if the url of current page points to the PR identified by the arguments
+     */
+    private boolean isCurrentUrlPrPageIgnoreTabs(String repoId, int id) {
+        return getCurrentUrl().matches(
+            "\\Q" + GitHubURL.getPathForPullRequest(repoId, id) + "\\E(?:/|/commits|/files)?"
+        );
+    }
+
+    /**
+     * Navigates to the main tab of the given issue in the currently-active driver window.
+     * Runs on a separate thread.
+     */
+    public void showIssueMainTab(String repoId, int id, boolean isPullRequest, boolean isForceRefresh) {
+        if (isPullRequest) {
+            logger.info("Showing main tab for pull request #" + id);
+            runBrowserOperation(() -> driver.get(GitHubURL.getPathForPullRequest(repoId, id), isForceRefresh));
+            runBrowserOperation(() -> scrollToBottom());
+        } else {
+            showIssue(repoId, id, isPullRequest, isForceRefresh);
+        }
+    }
+
+    public void jumpToNewCommentBox(){
         if (isTestChromeDriver) {
-            UI.events.triggerEvent(new JumpToCommentEvent());
+            UI.events.triggerEvent(new JumpToNewCommentBoxEvent());
         }
         try {
             WebElement comment = driver.findElementById(GithubPageElements.NEW_COMMENT);
@@ -301,8 +331,7 @@ public class BrowserComponent {
      * Takes care of running it on a separate thread, and normalises error-handling across
      * all types of code.
      */
-
-    private void runBrowserOperation(Runnable operation) {
+    private void runBrowserOperation (Runnable operation) {
         executor.execute(() -> {
             if (isBrowserActive()) {
                 try {
@@ -520,9 +549,16 @@ public class BrowserComponent {
     }
 
     public boolean isCurrentUrlIssue() {
-        return driver != null && GitHubURL.isUrlIssue(driver.getCurrentUrl());
+        return driver != null && GitHubURL.isUrlIssueOrPr(driver.getCurrentUrl());
     }
 
+
+    /**
+     * Checks if current url is a valid pull request page
+     */
+    public boolean isCurrentUrlPr() {
+        return driver != null && GitHubURL.isUrlPullRequest(driver.getCurrentUrl());
+    }
     /**
      * Checks if current URL is GitHub issue or PR discussion page
      */
@@ -535,40 +571,40 @@ public class BrowserComponent {
     }
 
     /**
-     * Switches to the specified tab in GitHub PR page
+     * Opens a specified tab in PR's GitHub page.
+     * Does nothing if specified issue is not a PR.
      *
-     * @param tabName Either GithubPageElements.DISCUSSION_TAB, GithubPageElements.COMMITS_TAB
+     * @param pr TurboIssue of PR whose page is to be opened.
+     * @param tab Either GithubPageElements.DISCUSSION_TAB, GithubPageElements.COMMITS_TAB
      *                or GithubPageElements.FILES_TAB
      */
-    public void switchToTab(String tabName) {
-        if (GitHubURL.isPullRequestLoaded(getCurrentUrl())) {
-            int tabIndex = 0;
-
-            switch (tabName) {
-            case GithubPageElements.DISCUSSION_TAB:
-                tabIndex = 1;
-                break;
-            case GithubPageElements.COMMITS_TAB:
-                tabIndex = 2;
-                break;
-            case GithubPageElements.FILES_TAB:
-                tabIndex = 3;
-                break;
-            default:
-                assert false;
-                return;
-            }
-
-            String xpath = "//*[@id=\"js-repo-pjax-container\"]/div[2]/div[1]/div/div/div[2]/nav/a[%d]";
-            clickElementByXpath(String.format(xpath, tabIndex));
+    public void openPrTab(TurboIssue pr, GithubPageElements.PrTab tab) {
+        if (!pr.isPullRequest()) {
+            return;
         }
+        String suffix;
+        switch (tab) {
+        case DISCUSSION :
+            suffix = GitHubURL.PR_TAB_SUFFIX_DISCUSSION;
+            break;
+        case COMMITS :
+            suffix = GitHubURL.PR_TAB_SUFFIX_COMMITS;
+            break;
+        case FILES :
+            suffix = GitHubURL.PR_TAB_SUFFIX_FILES_CHANGED;
+            break;
+        default:
+            assert false;
+            return;
+        }
+        logger.info("Showing " + tab + " tab for pull request #" + pr.getId());
+        runBrowserOperation(() -> driver.get(
+            GitHubURL.getPathForPullRequest(pr.getRepoId(), pr.getId()) + suffix,
+            false
+        ));
     }
 
-    private void clickElementByXpath(String xpath) {
-        driver.findElement(By.xpath(xpath)).click();
-    }
-
-    public Optional<Integer> getPRNumberFromIssue() {
+    public Optional<Integer> getPrNumberFromIssue() {
         // will only get the first PR number if multiple PR's are mentioned
         try {
             WebElement element = driver.findElement(By.xpath("//div[contains(@id, 'ref-pullrequest')]"));
--- a/src/main/java/browserview/ChromeDriverEx.java
+++ b/src/main/java/browserview/ChromeDriverEx.java
@@ -50,10 +50,7 @@ public class ChromeDriverEx {
             logger.info("Test loading page: " + url);
             testGet();
         } else {
-            if (!isForceRefresh && (driver.getCurrentUrl().equalsIgnoreCase(url) ||
-                    driver.getCurrentUrl().equalsIgnoreCase(url + "/commits") ||
-                    driver.getCurrentUrl().equalsIgnoreCase(url + "/files"))) {
-
+            if (!isForceRefresh && driver.getCurrentUrl().equalsIgnoreCase(url)) {
                 logger.info("Already on page: " + url + " will not load it again. ");
             } else {
                 logger.info("Previous page was: " + driver.getCurrentUrl());
--- a/src/main/java/ui/listpanel/ListPanel.java
+++ b/src/main/java/ui/listpanel/ListPanel.java
@@ -1,10 +1,7 @@
 package ui.listpanel;
 
 import static ui.components.KeyboardShortcuts.*;
-import static util.GithubPageElements.DISCUSSION_TAB;
-import static util.GithubPageElements.COMMITS_TAB;
-import static util.GithubPageElements.FILES_TAB;
-
+import util.GithubPageElements.PrTab;
 
 import java.util.HashMap;
 import java.util.HashSet;
@@ -26,7 +23,6 @@ import javafx.scene.text.TextFlow;
 import javafx.stage.Stage;
 import org.apache.logging.log4j.Logger;
 
-import ui.GUIController;
 import ui.GuiElement;
 import ui.IdGenerator;
 import ui.UI;
@@ -310,39 +306,25 @@ public class ListPanel extends FilterPanel {
             }
             if (NEW_COMMENT.match(event)) {
                 if (KeyPress.isValidKeyCombination(GOTO_MODIFIER.getCode(), event.getCode())) {
-                    ui.getBrowserComponent().switchToTab(DISCUSSION_TAB);
-                } else {
-                    if (!ui.getBrowserComponent().isCurrentUrlIssue()) {
-                        openPageOfCurrentlySelectedIssue();
-                    } else if (!ui.getBrowserComponent().isCurrentUrlDiscussion()) {
-                        ui.getBrowserComponent().switchToTab(DISCUSSION_TAB);
+                    // goto PR discussion tab
+                    ui.getBrowserComponent().openPrTab(getCurrentlySelectedIssue(), PrTab.DISCUSSION);
+                } else { // focus new comment textbox
+                    if (!ui.getBrowserComponent().isCurrentUrlDiscussion()) {
+                        openIssuePageMainTab(getCurrentlySelectedIssue());
                     }
-
                     ui.getBrowserComponent().waitUntilDiscussionPageLoaded();
-                    ui.getBrowserComponent().jumpToComment();
+                    ui.getBrowserComponent().jumpToNewCommentBox();
                 }
 
             }
             if (PR_FILES_CHANGED.match(event)
                     && KeyPress.isValidKeyCombination(GOTO_MODIFIER.getCode(), event.getCode())) {
-
-                if (!ui.getBrowserComponent().isCurrentUrlIssue()) {
-                    openPageOfCurrentlySelectedIssue();
-                    ui.getBrowserComponent().waitUntilDiscussionPageLoaded();
-                }
-
-                ui.getBrowserComponent().switchToTab(FILES_TAB);
+                ui.getBrowserComponent().openPrTab(getCurrentlySelectedIssue(), PrTab.FILES);
                 event.consume();
             }
             if (PR_COMMITS.match(event)
                     && KeyPress.isValidKeyCombination(GOTO_MODIFIER.getCode(), event.getCode())) {
-
-                if (!ui.getBrowserComponent().isCurrentUrlIssue()) {
-                    openPageOfCurrentlySelectedIssue();
-                    ui.getBrowserComponent().waitUntilDiscussionPageLoaded();
-                }
-
-                ui.getBrowserComponent().switchToTab(COMMITS_TAB);
+                ui.getBrowserComponent().openPrTab(getCurrentlySelectedIssue(), PrTab.COMMITS);
                 event.consume();
             }
             if (SHOW_LABELS.match(event)) {
@@ -353,7 +335,8 @@ public class ListPanel extends FilterPanel {
                 }
             }
             if (MANAGE_ASSIGNEES.match(event) && ui.getBrowserComponent().isCurrentUrlIssue()) {
-                ui.getBrowserComponent().switchToTab(DISCUSSION_TAB);
+                ui.getBrowserComponent().openPrTab(getCurrentlySelectedIssue(), PrTab.DISCUSSION);
+                ui.getBrowserComponent().waitUntilDiscussionPageLoaded();
                 ui.getBrowserComponent().manageAssignees(event.getCode().toString());
             }
             if (SHOW_MILESTONES.match(event)) {
@@ -392,7 +375,7 @@ public class ListPanel extends FilterPanel {
         TurboIssue issue = listView.getSelectedItem().get().getIssue();
         Optional<Integer> relatedIssueNumber = listView.getSelectedItem().get().getIssue().isPullRequest()
                 ? GithubPageElements.extractIssueNumber(listView.getSelectedItem().get().getIssue().getDescription())
-                : ui.getBrowserComponent().getPRNumberFromIssue();
+                : ui.getBrowserComponent().getPrNumberFromIssue();
         if (!relatedIssueNumber.isPresent()) return;
         ui.triggerEvent(
                 new IssueSelectedEvent(issue.getRepoId(), relatedIssueNumber.get(), panelIndex, issue.isPullRequest())
@@ -673,9 +656,12 @@ public class ListPanel extends FilterPanel {
         return allReposInFilterExpr.isEmpty();
     }
 
-    private void openPageOfCurrentlySelectedIssue() {
-        TurboIssue issue = getSelectedElement().get().getIssue();
-        ui.getBrowserComponent().showIssue(issue.getRepoId(), issue.getId(), issue.isPullRequest(), false);
+    private TurboIssue getCurrentlySelectedIssue() {
+        return getSelectedElement().get().getIssue();
+    }
+
+    private void openIssuePageMainTab(TurboIssue issue) {
+        ui.getBrowserComponent().showIssueMainTab(issue.getRepoId(), issue.getId(), issue.isPullRequest(), false);
     }
 
     @Override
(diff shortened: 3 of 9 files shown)


YijinL:
Will clean out unused methods and commented code if everything is fine.
@HansNewbie This replaces your hotfix, you can check and see if I missed some edge case you covered in your fix

src/main/java/browserview/BrowserComponent.java
```diff
@@ -252,7 +253,7 @@ public void showIssue(String repoId, int id, boolean isPullRequest, boolean isFo
         runBrowserOperation(() -> scrollToBottom());
     }
 
-    public void jumpToComment(){
+    public void jumpToNewComment(){
```
HansNewbie: Maybe `jumpToCommentBox`? Also, rename the `JumpToCommentEvent` below and its handler.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -235,40 +235,54 @@ private void setupKeyboardShortcuts() {
...
+                if (KeyPress.isValidKeyCombination(GOTO_MODIFIER.getCode(), event.getCode())
+                    && selected.isPullRequest()) { // goto PR discussion tab
+
+                    ui.getBrowserComponent().openPrTab(selected, DISCUSSION_TAB);
+//                    ui.getBrowserComponent().switchToTab(DISCUSSION_TAB);
+                } else { // focus new comment textbox
                     if (!ui.getBrowserComponent().isCurrentUrlIssue()) {
-                        openPageOfCurrentlySelectedIssue();
-                    } else if (!ui.getBrowserComponent().isCurrentUrlDiscussion()) {
+                        openTurboIssuePage(getCurrentlySelectedIssue());
```
HansNewbie: `openIssuePage` would be better. Also, you have `selected` variable in this scope.

HansNewbie:
@YijinL update me when you clean up the codes. Please delete the commented old code; version control helps you to revert to previous state if needed.

YijinL:
@HansNewbie done.

src/main/java/ui/issuepanel/UIBrowserBridge.java
```diff
@@ -1,6 +1,7 @@
 package ui.issuepanel;
 
 import javafx.application.Platform;
+import org.apache.xpath.SourceTree;
```
HansNewbie: remove this import since it is unused.

src/main/java/ui/issuepanel/UIBrowserBridge.java
```diff
@@ -54,6 +55,7 @@ private TickingTimer createTickingTimer(UI ui) {
             // do nothing for each tick
         }, () -> {
             Platform.runLater(() -> {
+                System.out.println("\nREQUEST DELAY TIMER\n");
```
HansNewbie: Is this left out of clean up? Otherwise use logger.

YijinL: Thanks I missed this

src/main/java/util/events/testevents/JumpToCommentEventHandler.java
```diff
@@ -6,5 +6,5 @@
 @FunctionalInterface
 public interface JumpToCommentEventHandler extends EventHandler {
```
HansNewbie: EventHandler name must match Event name. You need to change this as well.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -450,7 +450,7 @@ private void markAsUnread() {
         Optional<GuiElement> item = listView.getSelectedItem();
         if (item.isPresent()) {
             TurboIssue issue = item.get().getIssue();
-            issue.markAsUnread(ui.prefs);
+            issue.markAsUnread(UI.prefs);
```
HansNewbie: Don't change things that's not related to this PR/issue. And I think `ui` is the correct one.

src/main/java/browserview/ChromeDriverEx.java
```diff
@@ -50,9 +50,9 @@ public void get(String url, boolean isForceRefresh) throws WebDriverException {
             logger.info("Test loading page: " + url);
             testGet();
         } else {
-            if (!isForceRefresh && (driver.getCurrentUrl().equalsIgnoreCase(url) ||
+            if (!isForceRefresh && (driver.getCurrentUrl().equalsIgnoreCase(url)/* ||
                 driver.getCurrentUrl().equalsIgnoreCase(url + "/commits") ||
-                driver.getCurrentUrl().equalsIgnoreCase(url + "/files"))) {
+                driver.getCurrentUrl().equalsIgnoreCase(url + "/files")*/)) {
```
HansNewbie: If the commented code not necessary, remove them. Otherwise, don't comment them  :p

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -235,40 +235,36 @@ private void setupKeyboardShortcuts() {
...
+
+                    ui.getBrowserComponent().openPrTab(selected, DISCUSSION_TAB);
+                } else { // focus new comment textbox
                     if (!ui.getBrowserComponent().isCurrentUrlIssue()) {
-                        openPageOfCurrentlySelectedIssue();
-                    } else if (!ui.getBrowserComponent().isCurrentUrlDiscussion()) {
-                        ui.getBrowserComponent().switchToTab(DISCUSSION_TAB);
+                        openIssuePage(selected);
+                    } else if (ui.getBrowserComponent().isCurrentUrlPR()) {
+                        ui.getBrowserComponent().openPrTab(getCurrentlySelectedIssue(), DISCUSSION_TAB);
```
HansNewbie: This whole thing can be changed. Originally, I need to check `isCurrentUrlDiscussion` in case I am in other tabs of PR page. And it was a quick fix, so I didn't bother refactoring.

What you can do here is replace the whole `if else` case to
```
if (!ui.getBrowserComponent().isCurrentUrlIssueDiscussion()) {
  openIssuePage(selected);
}
.
.
.
private boolean isCurrentUrlDiscussion() {
// Use GitHub URL check, but create a new regex where it only matches issues / pull url with no suffix of commit or files.
}
```
This is because `showIssue` in `openIssue` handles both PR and issue, and opening default PR page will direct you to the discussion page, anyway. This leads to cleaner `if` case.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -279,14 +275,18 @@ private void setupKeyboardShortcuts() {
...
+                }
                 ui.getBrowserComponent().manageAssignees(event.getCode().toString());
             }
             if (SHOW_MILESTONES.match(event)) {
                 if (KeyPress.isValidKeyCombination(GOTO_MODIFIER.getCode(), event.getCode())) {
                     ui.getBrowserComponent().showMilestones();
                 } else if (ui.getBrowserComponent().isCurrentUrlIssue()) {
-                    ui.getBrowserComponent().switchToTab(DISCUSSION_TAB);
+                    if (ui.getBrowserComponent().isCurrentUrlPR()) {
+                        ui.getBrowserComponent().openPrTab(getCurrentlySelectedIssue(), DISCUSSION_TAB);
```
HansNewbie: I am not sure about these two. I will get back to you again.

src/main/java/ui/listpanel/ListPanel.java
YijinL: `prefs` is a static field of `UI` though. Yeah it's not related, I'll undo it.

src/main/java/browserview/ChromeDriverEx.java
YijinL: I'm still tracing the effects of this change, but all stable tests pass. I'll remove it in awhile

src/main/java/ui/listpanel/ListPanel.java
HansNewbie: I think it's wrong to use `isCurrentUrlPr` here. You also need to create something new or update `isPullRequestLoaded` accordingly on `GitHubURL`.

So the intention here of switching tab is that if `isCurrentUrlIssue`, it means that we can be in an issue page or PR page in any of the tab. The call `switchToTab(DISCUSSION_TAB)` will not do anything on issue page anyway, so we can just call it here, and if we are on PR page, it ensures that we go to `DISCUSSION_TAB`.

The thing is, `isCurrentUrlPR` checks if we are in PR page, of any tab. Hence the call here is redundant and unnecessary. You should have a new regex (in `GitHubUrl`) that checks if we are in PR page but not discussion tab (by checking the ending of the URL, `files` and `commits`). If we are not, then open the discussion tab.

HansNewbie: More importantly, I think there might be a bug here in the original implementation (you should try it out first, see if my hypothesis is correct) where we might jump to `manageMilestones` before the PR discussion tab is opened. If that is the case, you need to use `waitUntilDiscussionPageLoaded`.

HansNewbie:
Update me again once ready for another review.

src/main/java/ui/listpanel/ListPanel.java
YijinL: I shouldn't change `isPullRequestLoaded` because there are a few tests that specifically test for that behaviour.

Having a new check for whether the current url is in the PR page but not discussion tab also seems redundant, because `openPrTab` already handles the case where the url is in the discussion tab.

One way to make everything neater is to move the `is current page PR` check into `openPrTab`, like in the original `switchToTab` implementation. (there was a PR url check after the `isCurrentUrlIssue` check either way). I'll do that first.

Regarding the bug, I've added `waitUntilDiscussionPageLoaded` accordingly.

YijinL:
@HansNewbie ping for review

HansNewbie:
Your branch has conflict. You need to solve it.

YijinL:
I mean review whether the implementation is fine, I'll resolve the conflict after? Or should I do it now?

src/main/java/util/events/testevents/JumpToCommentBoxEventHandler.java
```diff
@@ -4,7 +4,7 @@
 import util.events.EventHandler;
 
 @FunctionalInterface
-public interface JumpToCommentEventHandler extends EventHandler {
+public interface JumpToCommentBoxEventHandler extends EventHandler {
```
HansNewbie: Need to follow the event class name. `JumpToNewCommentBoxEventHandler`

HansNewbie:
Should resolve in case some other codes have changed whatever you are working on, so that you don't need double work. Also, have you tried running your code? I am not sure if some of the behaviour will change. Still trying to understand the code.

YijinL:
Ok, doing it now

HansNewbie:
Give me a moment to look at your work, so you don't need to solve conflict multiple times.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -244,7 +246,11 @@ public void showFilterDocs() {
     public void showIssue(String repoId, int id, boolean isPullRequest, boolean isForceRefresh) {
         if (isPullRequest) {
             logger.info("Showing pull request #" + id);
-            runBrowserOperation(() -> driver.get(GitHubURL.getPathForPullRequest(repoId, id), isForceRefresh));
+            runBrowserOperation(() -> driver.get(
+                GitHubURL.getPathForPullRequest(repoId, id),
+                isForceRefresh,
+                (currentUrl, targetUrl) -> currentUrl.equalsIgnoreCase(targetUrl + "/commits") || currentUrl.equalsIgnoreCase(targetUrl + "/files")
+            ));
```
HansNewbie: This and the corresponding overloaded `get` is not necessary. Will explain below.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -252,9 +258,23 @@ public void showIssue(String repoId, int id, boolean isPullRequest, boolean isFo
...
+     */
+    public void showIssueMainTab(String repoId, int id, boolean isPullRequest, boolean isForceRefresh) {
+        if (isPullRequest) {
+            logger.info("Showing main tab for pull request #" + id);
+            runBrowserOperation(() -> driver.get(GitHubURL.getPathForPullRequest(repoId, id), isForceRefresh));
+            runBrowserOperation(() -> scrollToBottom());
+        } else {
+            showIssue(repoId, id, isPullRequest, isForceRefresh);
+        }
+    }
```
HansNewbie: This method is also not necessary.

src/main/java/browserview/ChromeDriverEx.java
```diff
@@ -63,6 +63,18 @@ public void get(String url, boolean isForceRefresh) throws WebDriverException {
...
+     * GET, but allows caller to specify custom equivalence rules for the url
+     * @param isConsideredSamePage params (currentUrl, targetUrl)
+    */
+    public void get(String url, boolean isForceRefresh, BiPredicate<String, String> isConsideredSamePage) throws WebDriverException {
+        if (isConsideredSamePage.test(driver.getCurrentUrl(), url)) {
+            logger.info("Already on page: " + url + " will not load it again. ");
+        } else {
+            get(url, isForceRefresh);
+        }
+    }
```
HansNewbie: This is not necessary.

HansNewbie:
For the parts that I mentioned as unnecessary, you can do all the other processing before calling the corresponding methods. Don't forget single responsibility principle; the methods that are in a class of doing basic things should handle more than necessary. The caller should be responsible instead in doing all the checks. I will see where you can put those things in.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -235,40 +235,29 @@ private void setupKeyboardShortcuts() {
...
             if (NEW_COMMENT.match(event)) {
-                if (KeyPress.isValidKeyCombination(GOTO_MODIFIER.getCode(), event.getCode())) {
-                    ui.getBrowserComponent().switchToTab(DISCUSSION_TAB);
-                } else {
-                    if (!ui.getBrowserComponent().isCurrentUrlIssue()) {
-                        openPageOfCurrentlySelectedIssue();
-                    } else if (!ui.getBrowserComponent().isCurrentUrlDiscussion()) {
-                        ui.getBrowserComponent().switchToTab(DISCUSSION_TAB);
+                if (!ui.getBrowserComponent().isCurrentUrlIssue()) {
+                    return;
```
HansNewbie: Why do we return early here? Based on the old code, this `return` will change the behaviour completely, because in the `else` part we will call `jumpToComment` for any case.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -235,40 +235,29 @@ private void setupKeyboardShortcuts() {
...
                 && KeyPress.isValidKeyCombination(GOTO_MODIFIER.getCode(), event.getCode())) {
-
-                if (!ui.getBrowserComponent().isCurrentUrlIssue()) {
-                    openPageOfCurrentlySelectedIssue();
-                    ui.getBrowserComponent().waitUntilDiscussionPageLoaded();
-                }
-
-                ui.getBrowserComponent().switchToTab(FILES_TAB);
+                TurboIssue selected = getCurrentlySelectedIssue();
+                ui.getBrowserComponent().openPrTab(selected, FILES_TAB);
```
HansNewbie: This should follow your commit tab handling, so that your code is consistent.

src/main/java/browserview/ChromeDriverEx.java
```diff
@@ -50,9 +52,7 @@ public void get(String url, boolean isForceRefresh) throws WebDriverException {
             logger.info("Test loading page: " + url);
             testGet();
         } else {
-            if (!isForceRefresh && (driver.getCurrentUrl().equalsIgnoreCase(url) ||
-                driver.getCurrentUrl().equalsIgnoreCase(url + "/commits") ||
-                driver.getCurrentUrl().equalsIgnoreCase(url + "/files"))) {
+            if (!isForceRefresh && (driver.getCurrentUrl().equalsIgnoreCase(url))) {
```
HansNewbie: We don't have a test that covers all the browser action. Changing this here could cause a lot of trouble, unless you have exhaustively test all browser action. You need to change your implementation so that this can stay as it is.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -526,8 +517,11 @@ private boolean hasReposInFilter() {
...
 
-    private void openPageOfCurrentlySelectedIssue() {
-        TurboIssue issue = getSelectedElement().get().getIssue();
-        ui.getBrowserComponent().showIssue(issue.getRepoId(), issue.getId(), issue.isPullRequest(), false);
+    private TurboIssue getCurrentlySelectedIssue() {
+        return getSelectedElement().get().getIssue();
+    }
+
+    private void openIssuePageMainTab(TurboIssue issue) {
+        ui.getBrowserComponent().showIssueMainTab(issue.getRepoId(), issue.getId(), issue.isPullRequest(), false);
```
HansNewbie: What we can do instead is use `isForceRefresh` here, so that no matter which PR tab we are at, it will reload the discussion tab.

HansNewbie: Oh, and can use the normal `showIssue`.

HansNewbie:
Generally the idea of your implementation is correct, but it is too convoluted and given that we don't have a test to cover all browser actions, we can't really change things freely unless you test every possible action exhaustively.

src/main/java/browserview/ChromeDriverEx.java
YijinL: Hmm, I did trace all usages of `ChromeDriverEx.get` before doing this; spent about 3-4 days covering every usage; ran all tests; manually tested each case where this method is involved through the UI. 

It's also weird because you would expect `get` to simply perform `GET` with the refresh check; the way it was, you could never use it to explicitly access other PR tabs when the browser is already in the PR's main tab. Does this violate SRP?

If I keep this old way then I can't use `ChromeDriverEx.get` for switching tabs at all. Implementing a slightly different `get` method just to allow the `*/commits` and `*/files` urls through feels wrong.

YijinL: TL:DR
As long as `get` stays the way it was, it's impossible to navigate to the `files` and `commits` PR tabs without adding a new method to `ChromeDriverEx`

HansNewbie: Makes sense to change `get` to what you do, but I don't think it's a good idea to overload it with the `BiPredicate`. Let me get back to you in a moment.

src/main/java/browserview/BrowserComponent.java
HansNewbie: We can keep this method.

src/main/java/browserview/BrowserComponent.java
HansNewbie: I think rather than overloading `get`, it's better to do the check here. We can abstract it into a method, e.g. `isPrPageIgnoreTab(repoId, id)`, and if it's false, then we call the usual get.

I don't really like the idea of having a `get` with extra parameter to do extra check because as ChromeDriver middle man, it shouldn't be doing this.

Would it work with this?

src/main/java/ui/listpanel/ListPanel.java
YijinL: Just to be clear, you are ok with forcing a refresh? Won't it affect the responsiveness?

HansNewbie: If not necessary, no. But in this case, if we are opening a tab (instead of switching tab), it is equivalent to force refresh, looking at the code (since url is not equal).

If we are already in the PR page, we don't open the tab (there are guard clauses before the most of the calls). But if you can make it work without force refresh, it would be better.

src/main/java/ui/listpanel/ListPanel.java
YijinL: Will fix.
So the intended behaviour is: 
**if current page is issue or pr (does not have to be same as selected card)**, 
    `focus new comment box`
**else**, 
    `go to page of currently selected issue card, then focus new comment box`

HansNewbie: Just follow the original implementation, but change any switching tab to open tab instead (of course, do pre-check if necessary to open tab or whatsoever).

The behaviour of the original is:
- if user press `G`+`R`, it will open discussion tab (if on PR, but since if we are on issue switch to tab won't work, there is no guard clause in the original implementation)
- else (meaning user only press `R`), we should jump to comment box. The original implementation have `if (!ui.getBrowserComponent().isCurrentUrlIssue())` to allow user to go to reply box even if they are on non-issue page, e.g. after they selected an issue, they press `G`+`K` which will open the keyboard shortcut page. We expect pressing `R` will jump to the selected issue page's comment box.

HansNewbie: the go to modifier is the `G` button, btw. And it doesn't jump to comment box :)

src/main/java/browserview/BrowserComponent.java
YijinL: Yes it would work. I will shift this check out of `ChomeDriverEx` into here (`BrowserComponent`)

src/main/java/ui/listpanel/ListPanel.java
YijinL: Since we decided to keep `BrowserComponent.showIssueMainTab`, we won't need to force refresh

HansNewbie: Sure.

YijinL:
Seems like I was working concurrently with someone else, had to merge so many times.

Is there a way to nicely squash my commits? Everyone else's commits are interleaved with mine now.

damithc:
We need this fix soon. The keyboard shortcut broke again. :-p

HansNewbie:
@YijinL you can use `git rebase -i origin/master` after updating master.

HansNewbie:
@damithc which shortcut is broken? I checked the DOM but it hasn't changed.

damithc:
Hans, G,F didn't work in the latest master

-- 
Sent from a mobile device.

damithc:
Looks like the problem was cased by the tip about 'saved responses'. The shortcut started working after I closed it.

YijinL:
@HansNewbie squash-merged, tests passed, review?

src/main/java/browserview/BrowserComponent.java
```diff
@@ -254,9 +259,33 @@ public void showIssue(String repoId, int id, boolean isPullRequest, boolean isFo
         runBrowserOperation(() -> scrollToBottom());
     }
 
-    public void jumpToComment() {
+    /**
+     * true if the url of current page points to the PR identified by the arguments
+     * (different tabs like /commits or /files are considered the same page)
+     */
```
HansNewbie: Follow documentation convention.
```
Checks if current URL is PR page, regardless of the PR tab the URL is in.
@return true <blah blah>
```

src/main/java/browserview/BrowserComponent.java
```diff
@@ -537,35 +568,32 @@ public String getCurrentUrl() {
     /**
      * Switches to the specified tab in GitHub PR page
      *
-     * @param tabName Either GithubPageElements.DISCUSSION_TAB, GithubPageElements.COMMITS_TAB
-     *                or GithubPageElements.FILES_TAB
+     * @param pr
```
HansNewbie: - Give some explanation about `pr`.
- Update the main comment to `Opens a specified tab of a PR` or something similar since we are no longer switching.
- Keep `tabName` verbose to the enums, since we have not really separate them into their own enum.

HansNewbie:
Just a few cosmetic changes needed. Update me once those have been fixed.

YijinL:
@HansNewbie Fixed

src/main/java/browserview/BrowserComponent.java
```diff
@@ -535,37 +567,36 @@ public String getCurrentUrl() {
     }
 
     /**
-     * Switches to the specified tab in GitHub PR page
+     * Opens a specified tab in PR's GitHub page.
+     * Does nothing if specified issue is not a PR.
      *
+     * @param pr TurboIssue associated with the desired PR
```
HansNewbie: More like `TurboIssue of PR which page is to be opened`.

YijinL:
Comment changed accordingly.

HansNewbie:
Will bump status when CI passes.

HansNewbie:
@YijinL I am not sure why GitHub is not showing status properly. You need to rebase your commit to the latest master since some merges happened. This PR is good.

Comment here when you have rebased.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -523,6 +552,9 @@ public boolean isCurrentUrlIssue() {
         return driver != null && GitHubURL.isUrlIssue(driver.getCurrentUrl());
     }
 
+    public boolean isCurrentUrlPR() {
```
damithc: `Pr`?

YijinL:
@HansNewbie Rebased

src/main/java/browserview/BrowserComponent.java
```diff
@@ -523,6 +552,9 @@ public boolean isCurrentUrlIssue() {
         return driver != null && GitHubURL.isUrlIssue(driver.getCurrentUrl());
     }
 
+    public boolean isCurrentUrlPr() {
```
damithc: Add header comment. The method name is a bit ambiguous.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -535,40 +567,39 @@ public String getCurrentUrl() {
...
+            suffix = GitHubURL.PR_TAB_SUFFIX_DISCUSSION;
+            break;
+        case GithubPageElements.COMMITS_TAB :
+            suffix = GitHubURL.PR_TAB_SUFFIX_COMMITS;
+            break;
+        case GithubPageElements.FILES_TAB :
+            suffix = GitHubURL.PR_TAB_SUFFIX_FILES_CHANGED;
+            break;
+        default:
+            throw new IllegalArgumentException("Argument tabName = " + tabName + " is not a valid PR tab identifier");
```
damithc: If this is supposed to be a programmer error, then it should be an assertions.

src/main/java/util/GitHubURL.java
```diff
@@ -46,11 +50,11 @@ public static String getPathForContributors(String repoId) {
     }
 
     public static boolean isUrlIssue(String url) {
-        return url.matches("https://github.com/([^/]+)/([^/]+)/(issues|pull)/([0-9]+)([/commits,/files]*)");
+        return url.matches("https://github\\.com/([^/]+)/([^/]+)/(issues|pull)/([0-9]+)([/commits,/files]*)");
     }
```
damithc: Implementation doesn't seem to match method name

damithc:
Some minor changes requested.

src/main/java/browserview/BrowserComponent.java
YijinL: Assertions can be turned off, should I still throw this exception in case?

src/main/java/util/GitHubURL.java
YijinL: It seems pretty clear? Maybe I can change the name to `isUrlIssuePage`?

src/main/java/browserview/BrowserComponent.java
damithc: By that logic, assertions shouldn't exist at all :-) Exceptions need catching and handling (more code!). Therefore, exceptions should not be used to handle programmer errors - which are not expected to happen and not expected to be recoverable. 
At the moment we don't have assertions enabled, but we should enable them and keep them enabled indefinitely.

src/main/java/util/GitHubURL.java
damithc: Actually, the confusion comes from the word 'issue' and it's not a result of your code. Does 'issues' include PRs? If it does, why do we have a method called `isUrlIssueOrPrDiscussionPage`? If they are mutually exclusive, this method should be `isUrlPrOrIssue`?

src/main/java/util/GithubPageElements.java
```diff
@@ -16,6 +16,10 @@
     public static final String COMMITS_TAB = "commits";
     public static final String FILES_TAB = "files";
 
+    public enum PrTab {
+        DISCUSSION, COMMITS, FILES
+    }
```
HansNewbie: If you are using this, maybe you should to put the String above this enum as enum values? Otherwise don't use the enum at all but rather use the aboves.

HansNewbie: Sorry, I just opened your branch. No need to change this enum, but remove the three unused variables above since you replaced them with the enum.

YijinL:
@HansNewbie done, review?

HansNewbie:
@YijinL yup, seems good. Just waiting for merge approval.

src/main/java/browserview/BrowserComponent.java
```diff
@@ -535,40 +571,40 @@ public String getCurrentUrl() {
...
+            suffix = GitHubURL.PR_TAB_SUFFIX_FILES_CHANGED;
+            break;
+        default:
+            assert false;
+            return;
+        }
+        logger.info("Showing " + tab + " tab for pull request #" + pr.getId());
+        runBrowserOperation(() -> driver.get(
+            GitHubURL.getPathForPullRequest(pr.getRepoId(), pr.getId()) + suffix,
+            false
```
damithc: No need for line break here?

damithc: Never mind. It's fine.

PR #1397, opened 2016-03-22
Can select two panels by confirming/canceling panel rename #1390

### LINKED ISSUE #1390: Can select two panels by confirming/canceling panel rename
UI bug?
[Video here](https://gfycat.com/PepperyFlatBaboon)

Might be linked to #1231?

zhongwei-z:
This fixes #1390. Trigger a `panelClickedEvent` every time we try to close the `renamableTextField`.

### DIFF
--- a/src/main/java/ui/components/PanelMenuBar.java
+++ b/src/main/java/ui/components/PanelMenuBar.java
@@ -20,6 +20,7 @@ import ui.GUIController;
 import ui.IdGenerator;
 import ui.UI;
 import ui.issuepanel.FilterPanel;
+import util.events.PanelClickedEvent;
 import util.events.ShowRenamePanelEvent;
 
 /**
@@ -101,6 +102,7 @@ public class PanelMenuBar extends HBox {
 
     private void activateInplaceRename() {
         ui.triggerEvent(new ShowRenamePanelEvent(panel.panelIndex));
+        clickThisPanel();
     }
 
     private HBox createRenameButton() {
@@ -192,6 +194,14 @@ public class PanelMenuBar extends HBox {
     }
 
     /**
+     * Clears selection on the rest of the panels to avoid multiple
+     * panels being selected at the same time.
+     */
+    private void clickThisPanel() {
+        ui.triggerEvent(new PanelClickedEvent(panel.panelIndex));
+    }
+
+    /**
      * Augments components of PanelMenuBar when the renaming of the panel happens.
      * The confirm button and the undo button are added to the panel.
      */
@@ -216,6 +226,7 @@ public class PanelMenuBar extends HBox {
         menuBarNameArea.getChildren().add(nameBox);
         this.getChildren().addAll(menuBarRenameButton, menuBarCloseButton);
         panel.requestFocus();
+        clickThisPanel();
     }
 
     private void panelNameValidator() {
--- a/src/test/java/guitests/PanelRenameTest.java
+++ b/src/test/java/guitests/PanelRenameTest.java
@@ -4,6 +4,7 @@ import javafx.scene.control.TextField;
 import javafx.scene.input.KeyCode;
 import javafx.scene.text.Text;
 
+import org.junit.Before;
 import org.junit.Test;
 import org.loadui.testfx.exceptions.NoNodesFoundException;
 import org.loadui.testfx.utils.FXTestUtils;
@@ -17,7 +18,10 @@ import ui.issuepanel.PanelControl;
 import util.PlatformEx;
 import util.events.ShowRenamePanelEvent;
 
+import java.util.Optional;
+
 import static org.junit.Assert.assertEquals;
+import static ui.components.KeyboardShortcuts.CLOSE_PANEL;
 import static ui.components.KeyboardShortcuts.CREATE_RIGHT_PANEL;
 import static ui.components.KeyboardShortcuts.MAXIMIZE_WINDOW;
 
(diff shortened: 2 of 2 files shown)


damithc:
Comment when ready to review.

zhongwei-z:
Ready for review!

ndt93:
@acruis Can help review this? Thanks.

acruis:
No problem from me. Do you plan to write any tests for this?

zhongwei-z:
OK, I have written a test for it!

src/test/java/guitests/PanelRenameTest.java
```diff
@@ -130,4 +137,22 @@ public void panelRenameTest() {
...
         sleep(EVENT_DELAY);
     }
+
+    @Test
+    public void panelRenameSelectOnlyOnePanel() {
+        // Test for the currently selected panel when trying to rename
+
+        pushKeys(MAXIMIZE_WINDOW);
+        while (panels.getPanelCount() < 3) {
+            pushKeys(CREATE_RIGHT_PANEL);
```
acruis: This is fragile. The `waitOnFxThread` inside `pushKeys` is not good enough as a guarantee for synchronisation, so `panels.getPanelCount` may execute for the next iteration of the loop before the keypress is even registered.

As a result, sometimes there will be 4-5 panels instead of just 3 panels, which causes this test (as well as the one above it) to break.

src/test/java/guitests/PanelRenameTest.java
```diff
@@ -130,4 +137,22 @@ public void panelRenameTest() {
...
+        }
+        click("#dummy/dummy_col0_renameButton");
```
acruis: Note that the two tests in this class will be done on the same window. This means that whichever test is executed later will have to work on panels created by the test that runs before it, and by JUnit standards [you have no guarantee as to which one will run first](http://stackoverflow.com/questions/3693626/how-to-run-test-methods-in-specific-order-in-junit4).

If `panelRenameTest` runs first, it will create 6 panels, which makes your test break as the first button you want to click is not even visible.

Off the top of my head, there are a couple of things you can do:
* Somehow slot the logic for this test into the currently-existing test: The test will be checking for *two* different things (the pre-existing test, as well as this PR's issue), which will not be obvious to future devs unless you make it clear which part of the test is testing what.
* Make use of Boards functionality to clear panels from the previous tests: You might have to deal with modal dialogs, which does not play well with Travis.
* Run `panelRenameSelectOnlyOnePanel` as a separate class: Not recommended as creating a new window consumes a lot of testing time, but if you find that other solutions are too difficult to implement then go ahead.

zhongwei-z:
Is my current approach okay? I used a local variable to get the panel count, and either open panels or close the panels from the previous test to ensure there are only three panels. At the end of the test I’ll close until there is only one panel left for the next test.

src/test/java/guitests/PanelRenameTest.java
```diff
@@ -130,4 +139,32 @@ public void panelRenameTest() {
         push(KeyCode.ENTER);
         sleep(EVENT_DELAY);
     }
+
+    @Test
+    public void panelRenameSelectOnlyOnePanel() {
```
damithc: Try to use the new naming convention for test methods. It's given in the java coding standard.

acruis:
Do fix the code style so Travis builds can run.

Instead of just checking and modifying panels in your current test, do implement a more general solution to make states consistent between tests (something like a `@Before` or `@After`). It's easier to understand for others reading your code.

zhongwei-z:
Ready for review!

src/test/java/guitests/PanelRenameTest.java
```diff
@@ -24,23 +28,26 @@
...
 
     @Override
     public void launchApp() {
         FXTestUtils.launchApp(TestUI.class, "--bypasslogin=true");
     }
 
+    @Before
+    public void before() {
+        UI ui = TestController.getUI();
+        panels = ui.getPanelControl();
```
acruis: Can be `TestController.getUI().getPanelControl()`?

src/test/java/guitests/PanelRenameTest.java
```diff
@@ -130,4 +137,21 @@ public void panelRenameTest() {
         push(KeyCode.ENTER);
         sleep(EVENT_DELAY);
     }
+
+    @Test
+    public void panelRenameButton_onClick_panelGetsSelected() {
+        // Test for the currently selected panel when trying to rename
+
+        int count = panels.getPanelCount();
+        while (count-- > 1) {
```
acruis: Even this part can be in `@Before`.

acruis:
Almost there!

zhongwei-z:
OK, I’ve fixed those!

acruis:
Okay, no more comments from me.

src/main/java/ui/components/PanelMenuBar.java
```diff
@@ -192,8 +194,13 @@ private void buttonAndKeyboardEventHandler() {
         });
     }
 
-    /**
-     * Augments components of PanelMenuBar when the renaming of the panel happens.
+    /** Use this in closeRenameableTextField() to avoid selecting multiple panels
+     */
```
damithc: Simply explain what the method does, not where to use it.

damithc:
Cosmetic change requested.

zhongwei-z:
Changed the comment

src/main/java/ui/components/PanelMenuBar.java
```diff
@@ -192,8 +194,14 @@ private void buttonAndKeyboardEventHandler() {
         });
     }
 
-    /**
-     * Augments components of PanelMenuBar when the renaming of the panel happens.
+    /** Triggers a {@link PanelClickedEvent} to avoid multiple panels being selected
```
acruis: Add a newline here. Same with `augmentRenameableTextField`'s javadoc comment.

zhongwei-z:
Added new line

acruis:
I edited my comment to add some more information. Do take a look.

zhongwei-z:
OK, reworded it

acruis:
Alright, no more comments from me.

acruis:
We'll see if Travis passes.

src/test/java/guitests/PanelRenameTest.java
```diff
@@ -134,4 +144,17 @@ public void panelRenameTest() {
...
+    public void panelRenameButton_onClick_panelGetsSelected() {
+        // Test for the currently selected panel when trying to rename
+
+        click(IdGenerator.getOcticonButtonIdReference(0, "renameButton"));
+        assertEquals(Optional.of(0), panels.getCurrentlySelectedPanel());
+        pushKeys(CREATE_RIGHT_PANEL);
+        click(IdGenerator.getOcticonButtonIdReference(1, "renameButton"));
+        assertEquals(Optional.of(1), panels.getCurrentlySelectedPanel());
+        pushKeys(CLOSE_PANEL);
+        pushKeys(KeyCode.ESCAPE);
```
acruis: Seems like some UI tests are failing. If it continues to fail, adding another `sleep(EVENT_DELAY)` at the end of this test might help.

zhongwei-z:
Added `sleep`, should be fine now!

acruis:
Okay, good to go. @damithc anything else to add?

damithc:
Nothing more from me. Go ahead and merge.

PR #1400, opened 2016-03-22
User can add issues to a 'watch list' #1389

### LINKED ISSUE #1389: User can add issues to a 'watch list'
... so that users can maintain a 'watch list' of selected issues.

User should be able to right click on an issue card and choose `add to new watch list`. This will create a new panel with the corresponding `id:repo#id` filter. User will be prompted to give a name for the panel.

After that, user should be able to add more issues to the that panel by right-clicking on issues and choosing `add to watch list` --> `panel name`

raisfathin:
Fixes #1389 

[Preview](http://gfycat.com/PleasingTalkativeDartfrog)

TODO: ~~Write tests~~

### DIFF
--- a/docs/panels.md
+++ b/docs/panels.md
@@ -36,6 +36,6 @@ To __abort__ the renaming action any time before step 3, press <kbd>Esc</kbd> or
 Watch list panels are panels whose filter reflects a "watch list". More specifically, a filter reflects a watch list if either of the following conditions is met :
 
 - It is empty (in other words, it watches all issues from the primary repo)
-- It is in the form of `id:repo1#id1;repo2#id2;repo3#id3;...;repoN#idN` (you can refer [here](filters.md) for more explanation about this filter syntax)
+- It is in the form of `id:repo1#id1;repo2#id2;repo3#id3;...;repoN#idN` (you can refer [here](filters.md#operator) and [here](filters.md#id) for more explanation about this filter syntax)
 
-To create a new watch list panel or add an issue to an existing watch list panel, you can refer [here](issuesAndPRs.md).
+To create a new watch list panel or add an issue to an existing watch list panel, you can refer [here](issuesAndPRs.md#watching).
--- a/docs/panels.md
+++ b/docs/panels.md
@@ -2,7 +2,7 @@
 
 ## Adding a New Panel <a name="add"></a>
 
-Panels can be added using the menu or keyboard shortcuts. 
+Panels can be added using the menu or keyboard shortcuts.
 
 - `Panels > Create Panel` to add a panel on the right of the currently selected panel.
     - <kbd>Ctrl</kbd> + <kbd>P</kbd> will do the same.
@@ -30,3 +30,12 @@ Panels can be reordered simply by dragging them into the desired position.
 3. __Save__ the new name by pressing <kbd>Enter</kbd> or clicking ![The tick icon](images/panels/savePanelRename.png?raw=true) at the top right corner of the panel.
 
 To __abort__ the renaming action any time before step 3, press <kbd>Esc</kbd> or click ![The undo arrow icon](images/panels/cancelPanelRename.png?raw=true) at the top right corner of the panel.
+
+## Watch List Panels <a name="watchlist"></a>
+
+Watch list panels are panels whose filter reflects a "watch list". More specifically, a filter reflects a watch list if either of the following conditions is met :
+
+- It is empty (in other words, it watches all issues from the primary repo)
+- It is in the form of `id:repo1#id1;repo2#id2;repo3#id3;...;repoN#idN` (you can refer [here](filters.md) for more explanation about this filter syntax)
+
+To create a new watch list panel or add an issue to an existing watch list panel, you can refer [here](issuesAndPRs.md).
--- a/src/main/java/ui/listpanel/ListPanel.java
+++ b/src/main/java/ui/listpanel/ListPanel.java
@@ -39,7 +39,7 @@ public class ListPanel extends FilterPanel {
 
     private static final Logger logger = HTLog.get(ListPanel.class);
 
-    private static final String WATCH_LIST_FILTER_REGEX = "^id:([^#]+#\\d+;)*([^#]+#\\d+)$";
+    public static final String WATCH_LIST_FILTER_REGEX = "^id:([^#]+#\\d+;)*([^#]+#\\d+)$";
 
     private final UI ui;
     private int issuesCount = 0;
@@ -51,7 +51,6 @@ public class ListPanel extends FilterPanel {
     private final IssueListView listView;
     private final HashMap<Integer, Integer> issueCommentCounts = new HashMap<>();
     private final HashMap<Integer, Integer> issueNonSelfCommentCounts = new HashMap<>();
-
 
     Text openIssueText;
     Text closedIssueText;
@@ -65,9 +64,6 @@ public class ListPanel extends FilterPanel {
     private static final String WATCH_LIST_PANEL_NAME_PROMPT_TITLE = "Watch List Panel Name";
     private static final String WATCH_LIST_PANEL_NAME_PROMPT_HEADER = "Please enter a name for the watch list panel.";
     private static final String DEFAULT_WATCH_LIST_PANEL_NAME = "New Watch List Panel";
-
-    private Label totalLabel;
-
 
     // Context Menu
     private final ContextMenu contextMenu = new ContextMenu();
@@ -442,7 +438,7 @@ public class ListPanel extends FilterPanel {
 
         newWatchListMenuItem.setText(NEW_WATCH_LIST_MENU_ITEM_TEXT);
         newWatchListMenuItem.setOnAction(e -> {
-            createNewWatchListMenuItem();
+            createNewWatchListPanel();
         });
 
         addToWatchListMenu.setText(ADD_TO_WATCH_LIST_MENU_TEXT);
@@ -464,17 +460,20 @@ public class ListPanel extends FilterPanel {
         updateChangeLabelsMenuItem();
         updateChangeMilestoneMenuItem();
         updateChangeAssigneeMenuItem();
-        updateWatchListPanels();
+        updateAvailableWatchListPanels();
 
         return contextMenu;
     }
 
-    /**
-     * Updates the list of available watch list panel
-     */
-    private void updateWatchListPanels() {
-        addToWatchListMenu.getItems().clear();
-        addToWatchListMenu.getItems().add(newWatchListMenuItem);
+    private void updateAvailableWatchListPanels() {
+        restoreDefaultAddToWatchListPanelMenu();
+        Optional<GuiElement> item = listView.getSelectedItem();
+
+        if (item.isPresent()) {
+            addToWatchListMenu.setDisable(false);
+        } else {
+            addToWatchListMenu.setDisable(true);
+        }
 
         parentPanelControl.getChildren().forEach(node -> {
             if (isWatchListPanel(node)) {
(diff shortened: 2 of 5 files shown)


damithc:
Functionality looks good. :+1:

raisfathin:
Ready for review!

damithc:
Try not to have major chunks of code nested. Normal flow of the method should not be nested.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -60,6 +57,12 @@
     Text bracketCloseText;
     Text plusText;
 
+    private static final String APPEND_TO_WATCH_LIST_FORMAT = "%s;%s#%d";
+    private static final String CREATE_WATCH_LIST_FORMAT = "id:%s#%d";
```
dariusf: Hmm, appending strings is a bit fragile, because the filter can be edited by the user. Could we do it on the AST level instead?

- Detection: traverse and find that there are only id qualifiers
- Create: create id qualifier
- Append: nest existing expression in disjunction

The first should be doable with `find`. We don't have a higher-level method yet so we can use that.

src/test/java/unstable/ContextMenuTests.java
```diff
@@ -98,4 +97,57 @@ public void test3() {
...
+        rightClick("#dummy/dummy_col0_11");
+        click("Add to watch list");
+        click("New Panel");
+        assertEquals("id:dummy/dummy#11", createdPanel.getCurrentInfo().getPanelFilter());
+    }
+
+    @Test
+    public void addToWatchListPanel_issueFromDifferentRepos_addedIssuesRetainRepoInfo() {
+        rightClick("#dummy/dummy_col0_9");
+        sleep(EVENT_DELAY);
```
dariusf: Hmm, we should wait for nodes to appear, or synchronise on some explicit condition. We're trying not to use sleep.

dariusf:
We should probably update the docs as well.

src/main/java/ui/listpanel/ListPanel.java
raisfathin: I was thinking of that, initially. But I could not figure out a better solution than using `instanceof` since `id:a#b` will be expanded to a conjunction of repo and id qualifier.

Another way would be to create another method in FilterExpression to check for it, but I feel that it's not what FilterExpression class is supposed to do :s

src/test/java/unstable/ContextMenuTests.java
raisfathin: I see. I was just following the other tests in the unit test file, will fix that.

src/main/java/ui/listpanel/ListPanel.java
dariusf: Ah, I see. Okay, guess we can do this for now, at the risk of things going wrong if the user edits the filter.

I think the long-term solution would be to keep qualified ids around past the parsing phrase and combine them into a conjunction only upon evaluation.

raisfathin: Alright. Should I create an issue for that?

raisfathin:
Ready for another review!

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -423,10 +439,66 @@ private ContextMenu updateContextMenu(ContextMenu contextMenu) {
...
+    private MenuItem createWatchListPanelMenuItem(FilterPanel watchListPanel) {
+        MenuItem watchPanelMenu = new MenuItem();
+        String watchListPanelTitle = watchListPanel.getCurrentInfo().getPanelName();
+        watchPanelMenu.setText(watchListPanelTitle);
+        watchPanelMenu.setOnAction(e -> addSelectedIssueToWatchListPanel(watchListPanel));
+        return watchPanelMenu;
+    }
+
+    /**
+     * A (filter) panel is a watch list panel if its qualifier is empty or its qualifier only has id qualifier.
```
damithc: I think this is enough? Otherwise you end up defining watch list filter in multiple places.
`@return true if the panel's filter is a 'watch list filter'`

src/main/java/ui/listpanel/ListPanel.java
damithc: Eventually, we need to do more manipulations to a watch list filter. e.g. reorder, insert, delete etc.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -492,6 +564,46 @@ private void markAsRead() {
         }
     }
 
+    private void createNewWatchListMenuItem() {
+        Optional<GuiElement> item = listView.getSelectedItem();
+        if (!item.isPresent()) {
+            return;
+        }
```
damithc: Shouldn't this have some error handling rather than failing silently?

src/main/java/ui/listpanel/ListPanel.java
raisfathin: True, that looks better.

src/main/java/ui/listpanel/ListPanel.java
raisfathin: Wait.. I think the method name is wrong, it should be `createNewWatchListPanel`. Will change that in the next push.

Hmm, actually, it should not fail since if there is no selected item, the menu would be disabled (covered in tests already). Maybe an assert would be better?

raisfathin:
Fixed.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -564,11 +564,12 @@ private void markAsRead() {
...
-    private void createNewWatchListMenuItem() {
+    /**
+     * Creates a new watch list panel containing the currently selected issue
+     */
+    private void createNewWatchListPanel() {
         Optional<GuiElement> item = listView.getSelectedItem();
-        if (!item.isPresent()) {
-            return;
-        }
+        assert item.isPresent();
```
damithc: Is it possible that the item is removed from the list between the right click and this method being executed? If that is possible, then it should be an exception.

raisfathin: Hmm, I think it's possible that it's removed when HubTurbo is refreshing. Besides throwing an exception, should we also report this to the user? Perhaps by displaying an alert or error at the status bar?

damithc: Throwing an exception is the first step. The second step is to catch it somewhere and handle it e.g. show an error message to the user. If it is appropriate to handle it here itself, then no need to throw an exception.

raisfathin: Is it okay if we display a dialog to show the error? This way, it would become immediately apparent to the user that an error has occured and we can fail the current operation.

damithc: Yes. However, just because the issue disappeared from the view does not mean it cannot be added to the watch list. So the user will be puzzled. Perhaps we should try to prevent this happening rather than report an error. Is that possible?

raisfathin: So it's fine if the issue disappears before this code gets executed, right?

I've been trying this a few times and couldn't get the assert to fail.

Guess the user has to be so unlucky to encounter this error since the update must occur after the user clicks the corresponding menu but before the first line of the method gets executed (i.e. `Optional<GuiElement> item = listView.getSelectedItem();`) :s 

@dariusf any opinion regarding this issue?

damithc: Yes, the chances are very low. Let's go ahead and show an error message saying the operation failed because the issue disappeared from the view before it can be updated.

raisfathin:
Done.

Screenshot of the error message:
screen shot 2016-03-27 at 18 35 32

src/main/java/ui/listpanel/ListPanel.java
dariusf: Sure, we can make a new issue.

src/main/java/ui/listpanel/ListPanel.java
dariusf: A bit late, but I'm not sure the item can disappear after the right click and before the menu appears. Concurrency is not an issue here, as UI operations all run on the JavaFX thread, i.e. if the items in a list can swapped out before the menu appears, it must be happening by design, which seems unlikely.

The optional there captures the possibility of there not being anything selected. In this case, we've shown the context menu, which is triggered by selection, so there must be a selected item. An assertion seems like the right thing to do then.

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -423,10 +444,66 @@ private ContextMenu updateContextMenu(ContextMenu contextMenu) {
...
+        FilterPanel currentFilterPanel = (FilterPanel) node;
+        return isWatchListFilter(currentFilterPanel.getFilterTextField().getText());
+    }
+
+    /**
+     * A FilterExpression is a valid watch list filter if :
+     *  - it is an empty string, OR
+     *  - it is in the form of : id:repo1#id1;repo2#id2;repo3#id3;...
+     */
+    private boolean isWatchListFilter(String filter) {
```
dariusf: This can be static.

src/test/java/tests/WatchListFilterTest.java
```diff
@@ -0,0 +1,45 @@
...
+        assertFalse(isValidFilter(invalidFilter2));
+
+        String invalidFilter3 = "id:a/a#1 ; c";
+        assertFalse(isValidFilter(invalidFilter3));
+
+        String invalidFilter4 = "repo:xyz";
+        assertFalse(isValidFilter(invalidFilter4));
+    }
+
+    private boolean isValidFilter(String filter) {
```
dariusf: Could we reuse the implementation in FilterPanel here?

src/main/java/ui/listpanel/ListPanel.java
raisfathin: Noted.

src/test/java/tests/WatchListFilterTest.java
raisfathin: Yup.

src/main/java/ui/listpanel/ListPanel.java
raisfathin: I see. Thanks for clarifying.

raisfathin:
Ready for review!

dariusf:
Looks good. Are there enough tests?

docs/panels.md
```diff
@@ -30,3 +30,12 @@ Panels can be reordered simply by dragging them into the desired position.
 3. __Save__ the new name by pressing <kbd>Enter</kbd> or clicking ![The tick icon](images/panels/savePanelRename.png?raw=true) at the top right corner of the panel.
 
 To __abort__ the renaming action any time before step 3, press <kbd>Esc</kbd> or click ![The undo arrow icon](images/panels/cancelPanelRename.png?raw=true) at the top right corner of the panel.
+
+## Watch List Panels <a name="watchlist"></a>
+
+Watch list panels are panels whose filter reflects a "watch list". More specifically, a filter reflects a watch list if either of the following conditions is met :
+
+- It is empty (in other words, it watches all issues from the primary repo)
+- It is in the form of `id:repo1#id1;repo2#id2;repo3#id3;...;repoN#idN` (you can refer [here](filters.md) for more explanation about this filter syntax)
```
damithc: Possible to give a more specific link that goes to the correct heading directly.

raisfathin: I just realised that I haven't added the relevant documentation for shorter form of OR to the documentation :sweat_smile:

Can I do it here? Or should I do it as separate issue?

raisfathin:
I think it's enough. I just added one more.

docs/panels.md
damithc: Better to do as a separate issue. Remember to update this link in that PR too.

PR #1404, opened 2016-03-23
Add 'try some features' section to 'getting started' #825

### LINKED ISSUE #825: Add 'try some features' section to 'getting started'
We can add some things a new user can do with HT

garbanzos:
Fixes #825.


garbanzos:
@dariusf could you help review this?

dariusf:
Looks good. I'm not sure if we want to advertise the work allocation board feature though, since it's kind of half-baked. Maybe we could talk about the milestone board instead? @damithc

damithc:
May be we can start with the `sample board`? After that, simply mention work allocation board and milestone board without going into details.

garbanzos:
Changed last section to "Generate A Sample Board". 
Ready for review!

damithc:
@garbanzos keep in mind that a new user will have the sample board automatically created. So the getting started guide needs to align with that scenario. i.e. playing around with the sample board should come first.

garbanzos:
@damithc 
The purpose of the last section was actually to introduce the users to the auto board creation feature. I guess it is a bit pointless to auto-create Sample Board since a new user will have the sample board automatically created. How about I change the section to **Generate A Milestone Board**? 

And add this section before **Open Another Repo**: 
**Editing the Sample Board (_Draft_)**
When you first boost up HubTurbo, you are presented with a sample board, something like this 
{{Screenshot of sample board}}
Let's try to rename the panels on this board! 
{{steps with pictures to rename panel and save board}}

damithc:
Sounds good. Yes, can use a milestone board. User can also do an edit a filter in the sample board before renaming the panel?

garbanzos:
Ready for review.

PR #1415, opened 2016-03-25
Milestone picker: pick next milestone by default #1408

### LINKED ISSUE #1408: Milestone picker: pick next milestone by default
If the issue is unscheduled, pick the next milestone by default. That means `M` --> `Enter` will assign the next milestone to an unscheduled issue.

garbanzos:
Fixes #1408 

Milestone picker pick next milestone by default only when there is no existing milestone. 

The next milestone picked fulfills the 3 criteria:
- It is open 
- It is not overdue 
- It has the nearest due date 

If there no milestones that fulfills the criteria, no milestone will be picked.

### DIFF
--- a/src/main/java/ui/components/pickers/MilestonePickerDialog.java
+++ b/src/main/java/ui/components/pickers/MilestonePickerDialog.java
@@ -43,10 +43,15 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
         state = new MilestonePickerState(originalMilestones);
         initUI();
         setupKeyEvents();
+        setInputFieldToDefaultMilestone();
+    }
 
-        getExistingMilestone(originalMilestones)
-                .map(PickerMilestone::getTitle)
-                .ifPresent(this::fillInputFieldWithMilestoneName);
+    /**
+     * Fills the input field of milestone picker dialog with default milestone's title
+     */
+    private void setInputFieldToDefaultMilestone() {
+        Optional<PickerMilestone> defaultMilestone = PickerMilestone.getDefaultMilestone(originalMilestones);
+        defaultMilestone.map(PickerMilestone::getTitle).ifPresent(this::fillInputFieldWithMilestoneName);
     }
 
     private boolean areFromSameRepo(TurboIssue issue, List<TurboMilestone> milestones) {
@@ -99,9 +104,9 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
     private void setConfirmResultConverter() {
         setResultConverter((dialogButton) -> {
             List<PickerMilestone> finalList = state.getCurrentMilestonesList();
-            if (hasSelectedMilestone(finalList)) {
-                return new MilestonePickerDialogResponse(dialogButton,
-                                                         Optional.of(getSelectedMilestone(finalList).get().getId()));
+            Optional<PickerMilestone> selectedMilestone = PickerMilestone.getSelectedMilestone(finalList);
+            if (selectedMilestone.isPresent()) {
+                return new MilestonePickerDialogResponse(dialogButton, Optional.of(selectedMilestone.get().getId()));
             }
             return new MilestonePickerDialogResponse(dialogButton, Optional.empty());
         });
@@ -146,9 +151,9 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
 
     private void populateAssignedMilestone(List<PickerMilestone> pickerMilestoneList, FlowPane assignedMilestonePane) {
         assignedMilestonePane.getChildren().clear();
-        updateExistingMilestone(getExistingMilestone(pickerMilestoneList), assignedMilestonePane);
+        updateExistingMilestone(PickerMilestone.getExistingMilestone(pickerMilestoneList), assignedMilestonePane);
         addAssignmentIndicator(assignedMilestonePane);
-        updateNewlyAssignedMilestone(getSelectedMilestone(pickerMilestoneList), assignedMilestonePane);
+        updateNewlyAssignedMilestone(PickerMilestone.getSelectedMilestone(pickerMilestoneList), assignedMilestonePane);
     }
 
     private void populateMatchingMilestones(List<PickerMilestone> matchingMilestoneList, VBox matchingMilestones) {
@@ -272,25 +277,6 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
         milestoneBox.setMaxHeight(30);
         milestoneBox.setStyle("-fx-border-radius: 3;-fx-border-style: dotted;-fx-alignment:center");
         return milestoneBox;
-    }
-
-    private Optional<PickerMilestone> getExistingMilestone(List<PickerMilestone> milestoneList) {
-        return milestoneList.stream()
-                .filter(PickerMilestone::isExisting)
-                .findAny();
-    }
-
-    private boolean hasSelectedMilestone(List<PickerMilestone> milestoneList) {
-        return milestoneList.stream()
-                .filter(PickerMilestone::isSelected)
-                .findAny()
-                .isPresent();
-    }
-
-    private Optional<PickerMilestone> getSelectedMilestone(List<PickerMilestone> milestoneList) {
-        return milestoneList.stream()
-                .filter(PickerMilestone::isSelected)
-                .findAny();
     }
 
 }
--- a/src/main/java/ui/components/pickers/MilestonePickerDialog.java
+++ b/src/main/java/ui/components/pickers/MilestonePickerDialog.java
@@ -105,10 +105,7 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
         setResultConverter((dialogButton) -> {
             List<PickerMilestone> finalList = state.getCurrentMilestonesList();
             Optional<PickerMilestone> selectedMilestone = PickerMilestone.getSelectedMilestone(finalList);
-            if (selectedMilestone.isPresent()) {
-                return new MilestonePickerDialogResponse(dialogButton, Optional.of(selectedMilestone.get().getId()));
-            }
-            return new MilestonePickerDialogResponse(dialogButton, Optional.empty());
+            return new MilestonePickerDialogResponse(dialogButton, selectedMilestone.map(PickerMilestone::getId));
         });
     }
 
--- a/src/main/java/ui/components/pickers/PickerMilestone.java
+++ b/src/main/java/ui/components/pickers/PickerMilestone.java
@@ -100,12 +100,24 @@ public class PickerMilestone extends TurboMilestone implements Comparable<Picker
         milestone.getStyleClass().add("labels-removed"); // add strikethrough
     }
 
+    /**
+     * Gets the existing milestone from the milestoneList
+     *
+     * @param milestoneList
+     * @return Optional of existing milestone
+     */
     public static Optional<PickerMilestone> getExistingMilestone(List<PickerMilestone> milestoneList) {
         return milestoneList.stream()
                 .filter(PickerMilestone::isExisting)
                 .findAny();
     }
 
+    /**
+     * Gets the selected milestone from the milestoneList
+     *
+     * @param milestoneList
+     * @return Optional of selected milestone
+     */
     public static Optional<PickerMilestone> getSelectedMilestone(List<PickerMilestone> milestoneList) {
         return milestoneList.stream()
                 .filter(PickerMilestone::isSelected)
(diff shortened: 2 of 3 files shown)


damithc:
To be clear, pick the next milestone by default if the issue doesn't have a milestone already. If it does have one, the existing one should be picked just like before.

garbanzos:
Ok. The milestone picker selects the next milestone only when there is no existing milestone now.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -43,6 +43,15 @@ public MilestonePickerDialog(Stage stage, TurboIssue issue, List<TurboMilestone>
         state = new MilestonePickerState(originalMilestones);
         initUI();
         setupKeyEvents();
+        selectNextMilestone();
+    }
+
+    private void selectNextMilestone() {
```
damithc: Add a header comment. I have a feeling this method name does not match what it does.

garbanzos:
Ready for review.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -280,6 +294,12 @@ private HBox createPreviouslyAssignedMilestoneBox() {
                 .findAny();
     }
 
+    private Optional<PickerMilestone> getNextOpenMilestone(List<PickerMilestone> milestoneList) {
```
HansNewbie: This should be a `static` method in `PickerMilestone` class. The method signature can remain the same. Put header comment about what this method does, especially that the definition of next open milestone is not widely known (some people might think it refers to the closest milestone, others the next milestone after closest).

Also, I am guessing this method requires the input list to be sorted. Put in the comment, e.g. `Condition: input list must be sorted according to <...>`.

HansNewbie: Should move `getExistingMilestone` to be a `static` method in `PickerMilestone` class as well. (Sorry for the extra work)

src/main/java/ui/components/pickers/PickerMilestone.java
```diff
@@ -128,15 +130,19 @@ public int compareTo(PickerMilestone milestone) {
...
-        return this.getDueDate().get()
-                .isBefore(milestone.getDueDate().get()) ? -1 : 1;
+        LocalDate thisDueDate = this.getDueDate().get();
+        LocalDate otherDueDate = milestone.getDueDate().get();
+        if (!thisDueDate.isEqual(otherDueDate)) {
+            return thisDueDate.isBefore(otherDueDate) ? -1 : 1;
+        }
+
+        // milestones ordered according to their titles if due dates are the same
+        return this.getTitle().compareTo(milestone.getTitle());
```
HansNewbie: Removing equality at the top might introduce a bug, e.g. what if both milestones have no due date - currently, the other milestone will always be treated as smaller. Is this an intended behaviour?

In my opinion, for consistency, if both milestones have same due date or both have no due date, compare by titles.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -43,6 +43,20 @@ public MilestonePickerDialog(Stage stage, TurboIssue issue, List<TurboMilestone>
...
         setupKeyEvents();
+        fillInputField();
+    }
+
+    /**
+     * Fills the input field of milestone picker dialog
+     * If there is an existing milestone, input field is filled with existing milestone's title
+     * Else input field is filled with the next open milestone that is not overdue
+     */
+    private void fillInputField() {
```
HansNewbie: Maybe `setInputFieldToDefaultMilestone` or something similar would be more explanative. Define default milestone (or other better terms you can come up with) on header comment.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -43,6 +43,20 @@ public MilestonePickerDialog(Stage stage, TurboIssue issue, List<TurboMilestone>
...
+        if (!getExistingMilestone(originalMilestones).isPresent()) {
+            getNextOpenMilestone(originalMilestones)
+                    .map(PickerMilestone::getTitle)
+                    .ifPresent(this::fillInputFieldWithMilestoneName);
+        }
```
HansNewbie: This will call `getExistingMilestone` twice (here and below). How about something like
```
// Fill X with the relevant values
Optional<X> defaultMilestone = getExistingMilestone(X).orElse(getNextOpenMilestone(X)).

defaulMilestone.map(X).ifPresent(this::fillInputFieldWithMilestoneName);
```

HansNewbie: To have better SLAP, maybe abstract the first line to another method `getDefaultMilestone`. Again, a `static` method on `PickerMilestone`. If you can, write tests for the methods, at the very least for `getNextOpenMilestone` since it is introduced in this PR.

garbanzos:
Ready for review again!

src/main/java/ui/components/pickers/PickerMilestone.java
```diff
@@ -93,6 +96,47 @@ private void setRemovedInUI(Label milestone) {
...
+    }
+
+    /**
+     * Gets the default milestone
+     * If there is an existing milestone, default milestone is the existing milestone
+     * Else it is the first open milestone that is not overdue
+     * Precondition: sortedMilestoneList needs to be sorted in its natural order
+     *
+     * @param sortedMilestoneList
+     * @return
```
HansNewbie: Remove `@return` since it is not used anyway.

src/test/java/tests/PickerMilestoneTest.java
```diff
@@ -0,0 +1,92 @@
...
+        listWithSelected.add(selectedMilestone);
+        listWithSelected.add(openMilestone);
+
+        listWithNextMilestone = new ArrayList<>();
+        listWithNextMilestone.add(openMilestone);
+        listWithNextMilestone.add(closedMilestone);
+    }
+
+    @Test
+    public void getExistingMilestone_returnEmpty() {
```
HansNewbie: Forgot the condition here. `noExistingAndNoSelectedMilestone`?

src/main/java/ui/components/pickers/PickerMilestone.java
garbanzos: @HansNewbie It is better if I fill it up? 
`@return Optional of default milestone`

src/main/java/ui/components/pickers/PickerMilestone.java
```diff
@@ -93,6 +96,47 @@ private void setRemovedInUI(Label milestone) {
         milestone.getStyleClass().add("labels-removed"); // add strikethrough
     }
 
+    public static Optional<PickerMilestone> getExistingMilestone(List<PickerMilestone> milestoneList) {
+        return milestoneList.stream()
+                .filter(PickerMilestone::isExisting)
+                .findAny();
+    }
+
+    public static Optional<PickerMilestone> getSelectedMilestone(List<PickerMilestone> milestoneList) {
```
HansNewbie: Add header comment to this and above, since now they are public :p

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -99,9 +104,9 @@ private void setupButtons(DialogPane milestonePickerDialogPane) {
     private void setConfirmResultConverter() {
         setResultConverter((dialogButton) -> {
             List<PickerMilestone> finalList = state.getCurrentMilestonesList();
-            if (hasSelectedMilestone(finalList)) {
-                return new MilestonePickerDialogResponse(dialogButton,
-                                                         Optional.of(getSelectedMilestone(finalList).get().getId()));
+            Optional<PickerMilestone> selectedMilestone = PickerMilestone.getSelectedMilestone(finalList);
+            if (selectedMilestone.isPresent()) {
+                return new MilestonePickerDialogResponse(dialogButton, Optional.of(selectedMilestone.get().getId()));
```
HansNewbie: Can we use `selectedMilestone.map(PickerMilestone::getId)` here? Then I think we can get rid of the `Optional.of()`.

src/test/java/tests/PickerMilestoneTest.java
HansNewbie: I meant `getExistingMilestone_noExistingAndNoSelectedMilestone_returnEmpty`.

src/main/java/ui/components/pickers/PickerMilestone.java
HansNewbie: Sure. to be more complete.

garbanzos:
Ready for review.

src/main/java/ui/components/pickers/PickerMilestone.java
```diff
@@ -93,6 +96,47 @@ private void setRemovedInUI(Label milestone) {
         milestone.getStyleClass().add("labels-removed"); // add strikethrough
     }
 
+    public static Optional<PickerMilestone> getExistingMilestone(List<PickerMilestone> milestoneList) {
+        return milestoneList.stream()
```
HansNewbie: `Gets...`

src/main/java/ui/components/pickers/PickerMilestone.java
```diff
@@ -128,15 +172,17 @@ public int compareTo(PickerMilestone milestone) {
...
 
         // milestones with due dates are smaller
         if (!this.getDueDate().isPresent()) return 1;
         if (!milestone.getDueDate().isPresent()) return -1;
 
         // milestones with earlier due dates are smaller
-        return this.getDueDate().get()
-                .isBefore(milestone.getDueDate().get()) ? -1 : 1;
+        return this.getDueDate().get().isBefore(milestone.getDueDate().get()) ? -1 : 1;
     }
```
HansNewbie: Sorry, I missed this one. You need a test to cover this as well, for picker milestone in which both has no due date and both has same due date.

HansNewbie:
Other than the last requested test case, this looks good. Update me again once ready for review.

garbanzos:
@HansNewbie 
Just added the tests for the cases you mentioned. 
But I'm not very sure how to name the test methods and they ended up really long. 
Is the name below acceptable? :sweat_smile: `getDefaultMilestone_haveMilestonesWithSameDueDates_returnMilestoneWithLexicographicalSmallerName`

HansNewbie:
Yep. Sounds good.

garbanzos:
@HansNewbie ready for review!

HansNewbie:
Do correct your style (checkstyleMain is not passing). You can run the gradle task on your machine and see where the violation is. This is so that the Travis build is triggered.

HansNewbie:
Also, you might want to squash the several last commits so that the commits are clean.

garbanzos:
@HansNewbie I think the checkstyleMain failed because I did not rebase my branch. (I just rebased it)
It this PR good to go after I squash the commits and the rest of the tests pass?

HansNewbie:
I will do one last review then I will change the PR status. Should be good.

src/main/java/ui/components/pickers/PickerMilestone.java
```diff
@@ -97,6 +100,59 @@ private void setRemovedInUI(Label milestone) {
         milestone.getStyleClass().add("labels-removed"); // add strikethrough
     }
 
+    /**
+     * Gets the existing milestone from the milestoneList
```
HansNewbie: `Gets a ...` since there could be more than one but we just want one.

src/main/java/ui/components/pickers/PickerMilestone.java
```diff
@@ -97,6 +100,59 @@ private void setRemovedInUI(Label milestone) {
...
+     * @return Optional of existing milestone
+     */
+    public static Optional<PickerMilestone> getExistingMilestone(List<PickerMilestone> milestoneList) {
+        return milestoneList.stream()
+                .filter(PickerMilestone::isExisting)
+                .findAny();
+    }
+
+    /**
+     * Gets the selected milestone from the milestoneList
```
HansNewbie: Same here

HansNewbie:
Just 2 cosmetic changes. Squashed it afterwards. Good to go afterwards.

src/main/java/ui/components/pickers/PickerMilestone.java
garbanzos: @HansNewbie 
The should only be one existing and selected milestone at any point in time actually. 
Because each issue can only be assigned to one milestone.

HansNewbie: Alright then. Let's leave it this way.

HansNewbie:
@garbanzos you can merge this already.

garbanzos:
Ok :smile:

PR #1417, opened 2016-03-25
Milestone picker: use lighter colors for open/close #1409

### LINKED ISSUE #1409: Milestone picker: use lighter colors for open/close
We can use lighter red/green to make them easier to read

garbanzos:
Fixes #1409. 

image

### DIFF
--- a/src/main/java/ui/components/pickers/PickerMilestone.java
+++ b/src/main/java/ui/components/pickers/PickerMilestone.java
@@ -11,8 +11,8 @@ import javafx.scene.text.Font;
  * This class handles the statuses and appearance of the milestones in MilestonePickerDialog
  */
 public class PickerMilestone extends TurboMilestone implements Comparable<PickerMilestone> {
-    public static final String OPEN_COLOUR = "#84BE54";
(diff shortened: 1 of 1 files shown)


garbanzos:
@damithc Are these colours ok?

damithc:
They are ok. A tiny bit lighter is even better.

garbanzos:
@damithc It this better?
image

damithc:
Yes, colors look good.

garbanzos:
@damithc Is this merge approved?

damithc:
@m133225 can you review this?

m133225:
@damithc Actually, would making the font white(which is what label picker does when the background is too dark) be better than making the colours lighter ? Since the only purpose for the change is to make it easier to read. The colour will then be consistent with GitHub's ones too.

damithc:
I prefer lighter colors because brighter colors attract attention unnecessarily. In this UI, the open/closed/overdue status is not the most important information and hence should not be shown in the brightest colors.

m133225:
If that's the case, then I have no further comments for this PR since it is rather straightforward. :stuck_out_tongue:

Issue #1127, opened 2015-11-18
Issue filter will be wrong when user click on other panel while typing


HansNewbie:
Say I am typing "milestone:V3.4" on a panel then without pressing `Enter` or `Esc` I click on another panel, the filter will stay at "milestone:V3.4" but the issues are not filtered on that.

We should either return the filter text to the last filter text or filter the issue according to the new (possibly unfinished) filter.

damithc:
We can pop up a dialog to ask user whether to apply the new filter or revert to previous.

PR #1420, opened 2016-03-25
Issue filter will be wrong when user click on other panel while typing #1127

lowjenhui:
From a UX perspective, it is quite weird/annoying to have dialog boxes popping up to ask the user whether he wants to change the panel or not. For example, when you search in google and then click elsewhere. There isn't any search performed unless enter is pressed.

This is a 'prototype' for #1127

### DIFF
--- a/src/main/java/ui/components/FilterTextField.java
+++ b/src/main/java/ui/components/FilterTextField.java
@@ -1,17 +1,21 @@
 package ui.components;
 
 import javafx.application.Platform;
+import javafx.beans.value.ChangeListener;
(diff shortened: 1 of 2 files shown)


damithc:
Actually, Google search results are auto loaded as you type to match the search box so the search page remains consistent. Our problem is that clicking outside makes the filter text inconsistent with the results.  I don't mind reverting to original filter if the user clicks outside but there is a risk that the user might lose her edits by an accidental click. I thought it is a common UX practice to ask 'the content has changed, are you sure want to navigate away'?

PR #1422, opened 2016-03-25
TurboUser's getRealName always return an empty string. #1384

### LINKED ISSUE #1384: TurboUser's getRealName always return an empty string.
Probably need to extend egit.

garbanzos:
Fixes #1384. 

**Problem**
The response for `GET /repos/:owner/:repo/collaborators` does not contain real names of users.

untitled

**Solution**
Obtained the list of collaborators using `getCollaborators` from `CollaboratorService` 
Then iterate through the list and get each collaborator's full details using `getUser` from `UserService`.


garbanzos:
@dariusf Could you help review this?

src/main/java/github/CollaboratorServiceEx.java
```diff
@@ -0,0 +1,50 @@
...
+     * @param repository
+     * @return list of collaborators
+     * @throws IOException
+     */
+    @Override
+    public List<User> getCollaborators(IRepositoryIdProvider repository) throws IOException {
+        List<User> collaborators = super.getCollaborators(repository);
+        List<User> collaboratorsWithFullDetails = new ArrayList<>();
+
+        collaborators.forEach(user -> {
```
dariusf: This is a mapping operation, so you can `.stream().map()` over `collaborators` to get the new list.

src/main/java/github/CollaboratorServiceEx.java
```diff
@@ -0,0 +1,50 @@
...
+            try {
+                collaboratorsWithFullDetails.add(userService.getUser(user.getLogin()));
+            } catch (IOException e) {
+                logger.error("Unable to get full details for user " + user.getLogin());
```
dariusf: If we fail to get the details, we should probably still return the incomplete user. It can still be used.

dariusf:
Nice work!

@ndt93 can we test this using MockServer?

garbanzos:
@dariusf Ready for review again. 
`getCollaborators` uses `map` and uses the incomplete user if `UserService` fails to get the user.

src/main/java/github/CollaboratorServiceEx.java
```diff
@@ -0,0 +1,47 @@
...
+import org.eclipse.egit.github.core.client.GitHubClient;
+import org.eclipse.egit.github.core.service.CollaboratorService;
+import org.eclipse.egit.github.core.service.UserService;
+import util.HTLog;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class CollaboratorServiceEx extends CollaboratorService {
```
damithc: W.r.t. `CollaboratorServiceEx`, are we following some naming convention?

ndt93: I think so. All the services that we extend from existing services in `egit.github` have a `Ex` suffix.

ndt93:
@dariusf Yep, we can use `MockServer` to test this.
@garbanzos You can refer to `PullRequestServiceExTests` or `MilestoneUpdateServiceTests` as examples.

garbanzos:
@dariusf I just added the tests!

src/main/java/github/CollaboratorServiceEx.java
damithc: I see. does it mean `Extension`?

ndt93: Yup.

src/main/java/github/CollaboratorServiceEx.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public CollaboratorServiceEx(GitHubClient client) {
+        super(client);
+    }
+
+    /**
+     * Gets the list of collaborators using CollaboratorService
+     * Then iterate through the list to get each collaborator using USerService
+     * This is done because the GitHub response for GET /repos/:owner/:repo/collaborators
+     * does not return a User with all the attributes of the user
```
dariusf: Maybe this part could be a bit clearer. It returns a user without all normal attributes/with attributes missing?

src/main/java/github/CollaboratorServiceEx.java
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * @param repository
+     * @return list of collaborators
+     * @throws IOException
+     */
+    @Override
+    public List<User> getCollaborators(IRepositoryIdProvider repository) throws IOException {
+        return super.getCollaborators(repository).stream()
+                .map( user -> {
```
dariusf: Extra whitespace after the `(` here. :smiling_imp:

src/test/java/tests/CollaboratorServiceExTests.java
```diff
@@ -0,0 +1,62 @@
...
+import java.lang.reflect.Type;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockserver.model.HttpRequest.request;
+import static org.mockserver.model.HttpResponse.response;
+
+public class CollaboratorServiceExTests {
+
+    @Test(expected = IllegalArgumentException.class)
```
dariusf: You can use ExpectedException for this instead, see [here](https://github.com/HubTurbo/HubTurbo/blob/master/src/test/java/tests/FilterParserTests.java#L19-L26).

src/test/java/tests/CollaboratorServiceExTests.java
```diff
@@ -0,0 +1,62 @@
...
+    public void getCollaborators_invalidRepoId() throws IOException {
+        GitHubClient client = new GitHubClient();
+        CollaboratorServiceEx service = new CollaboratorServiceEx(client);
+
+        service.getCollaborators(null);
+        service.getCollaborators(RepositoryId.createFromId(""));
```
dariusf: This part and the line below should probably be in separate test methods? Since the null would cause control flow to stop before this line.

src/test/java/tests/CollaboratorServiceExTests.java
```diff
@@ -0,0 +1,62 @@
...
+
+        Type listOfUsers = new TypeToken<List<User>>() {}.getType();
+        List<User> expectedCollaborators = new Gson().fromJson(sampleCollaborators, listOfUsers);
+        List<User> actualCollaborators = service.getCollaborators(RepositoryId.createFromId("hubturbo/tests"));
+
+        assertEquals(expectedCollaborators.size(), actualCollaborators.size());
+
+        for (int i = 0; i < expectedCollaborators.size(); i++) {
+            assertEquals(expectedCollaborators.get(i).getLogin(), actualCollaborators.get(i).getLogin());
+            assertEquals(expectedCollaborators.get(i).getName(), actualCollaborators.get(i).getName());
```
dariusf: Maybe we could check here that the names of the actual collaborators aren't empty.

src/test/resources/tests/CollaboratorsSample.json
```diff
@@ -0,0 +1,98 @@
...
+    "gists_url": "https://api.github.com/users/acrius/gists{/gist_id}",
+    "starred_url": "https://api.github.com/users/acrius/starred{/owner}{/repo}",
+    "subscriptions_url": "https://api.github.com/users/acrius/subscriptions",
+    "organizations_url": "https://api.github.com/users/acrius/orgs",
+    "repos_url": "https://api.github.com/users/acrius/repos",
+    "events_url": "https://api.github.com/users/acrius/events{/privacy}",
+    "received_events_url": "https://api.github.com/users/acrius/received_events",
+    "type": "User",
+    "site_admin": false,
+    "name": null,
```
dariusf: Is this intended? @acruis looks like he has a real name set.

garbanzos: @dariusf Nope, I spelled acruis wrongly when I did the curl. :sweat_smile:

acruis: !!!!!!

garbanzos:
Ready for review.

Issue #1413, opened 2016-03-25
Milestone picker: show all milestones in a scrollable panel


damithc:
Below the best 5 matches panel, we can show the full list in a scrollable panel. It can be a static panel (no highlighting etc), but it should be clickable.

m133225:
Should this be done only after #1411? Since there is no _best 5 matches panel_ yet.

damithc:
The one we have already is the 'best 5  matches' panel right? The only problem is it sometimes show fewer than 4. 
The one requested in this issue is somewhat independent. It's just a clickable panel of milestones sorted by due date.

m133225:
No, it currently shows all matches, fitted into a scrollable panel.

damithc:
Actually, I don't mind if it shows more than 5 and is scrollable. Just ensure that it shows at least 5 and height is fixed to 5 milestones.

PR #1439, opened 2016-03-28
Milestone picker: Always show 5 best matching milestones #1411

### LINKED ISSUE #1411: Milestone picker: Always show 5 best matching milestones
We can show the 5 best matching milestones, as long as there are at least 5 milestones.
If there aren't 5 milestones that match the user input, we can show the next best matches (e.g. those that match the text without the last letter; assuming the last letter was a typo). Over time, we can improve the algorithm used to find the '5 best matching' so that the 5 almost always contain the one user is looking for.

m133225:
Fixes #1411.

Needs refactoring since the MilestonePickerState has become rather messy with the new changes.

### DIFF
--- a/src/main/java/ui/components/pickers/MilestonePickerDialog.java
+++ b/src/main/java/ui/components/pickers/MilestonePickerDialog.java
@@ -23,6 +23,8 @@ import java.util.Optional;
 public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse> {
     private static final String OCTICON_ARROW = "\uf03e";
     private static final String DIALOG_TITLE = "Select Milestone";
+    private static final int DETAILED_MILESTONE_HEIGHT = 30;
+    private static final int PREV_ASSIGNED_MILESTONE_HEIGHT = 30;
 
     private final List<PickerMilestone> originalMilestones = new ArrayList<>();
     private FlowPane assignedMilestoneBox;
@@ -40,7 +42,7 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
         setupButtons(getDialogPane());
         setConfirmResultConverter();
         originalMilestones.addAll(convertToPickerMilestones(issue, milestones));
-        state = new MilestonePickerState(originalMilestones, "");
+        state = new MilestonePickerState(originalMilestones);
         initUI();
         setupKeyEvents();
         setInputFieldToDefaultMilestone();
@@ -103,7 +105,7 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
 
     private void setConfirmResultConverter() {
         setResultConverter((dialogButton) -> {
-            List<PickerMilestone> finalList = state.getCurrentMilestones();
+            List<PickerMilestone> finalList = state.getAllMilestones();
             Optional<PickerMilestone> selectedMilestone = PickerMilestone.getSelectedMilestone(finalList);
             return new MilestonePickerDialogResponse(dialogButton, selectedMilestone.map(PickerMilestone::getId));
         });
@@ -134,7 +136,7 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
     }
 
     private void refreshUI(MilestonePickerState state) {
-        populateAssignedMilestone(state.getCurrentMilestones(), assignedMilestoneBox);
+        populateAssignedMilestone(state.getAllMilestones(), assignedMilestoneBox);
         populateMatchingMilestones(state.getBestMatchingMilestones(), matchingMilestonesBox);
     }
 
@@ -183,7 +185,7 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
 
     private HBox createDetailedMilestoneBox(PickerMilestone milestone) {
         HBox detailedMilestoneBox = new HBox();
-        detailedMilestoneBox.setPrefHeight(30);
+        detailedMilestoneBox.setPrefHeight(DETAILED_MILESTONE_HEIGHT);
         detailedMilestoneBox.setSpacing(3);
         detailedMilestoneBox.setPadding(new Insets(3, 3, 3, 3));
         detailedMilestoneBox.setStyle("-fx-border-width: 0 0 1 0; -fx-border-style: solid;");
@@ -248,7 +250,7 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
 
     private VBox createMatchingMilestonesBox() {
         VBox milestoneGroup = new VBox();
-        milestoneGroup.setPrefHeight(150);
+        milestoneGroup.setPrefHeight(DETAILED_MILESTONE_HEIGHT * MilestonePickerState.BEST_MATCHING_LIMIT);
         milestoneGroup.setStyle("-fx-border-radius: 3;-fx-background-color: white;-fx-border-color: black;");
         return milestoneGroup;
     }
@@ -264,8 +266,8 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
     private HBox createPreviouslyAssignedMilestoneBox() {
         HBox milestoneBox = new HBox();
         milestoneBox.setPrefWidth(120);
-        milestoneBox.setPrefHeight(30);
-        milestoneBox.setMaxHeight(30);
+        milestoneBox.setPrefHeight(PREV_ASSIGNED_MILESTONE_HEIGHT);
+        milestoneBox.setMaxHeight(PREV_ASSIGNED_MILESTONE_HEIGHT);
         milestoneBox.setStyle("-fx-border-radius: 3;-fx-border-style: dotted;-fx-alignment:center");
         return milestoneBox;
     }
--- a/src/main/java/ui/components/pickers/MilestonePickerDialog.java
+++ b/src/main/java/ui/components/pickers/MilestonePickerDialog.java
@@ -40,7 +40,7 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
         setupButtons(getDialogPane());
         setConfirmResultConverter();
         originalMilestones.addAll(convertToPickerMilestones(issue, milestones));
-        state = new MilestonePickerState(originalMilestones);
+        state = new MilestonePickerState(originalMilestones, "");
         initUI();
         setupKeyEvents();
         setInputFieldToDefaultMilestone();
@@ -103,7 +103,7 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
 
     private void setConfirmResultConverter() {
         setResultConverter((dialogButton) -> {
-            List<PickerMilestone> finalList = state.getCurrentMilestonesList();
+            List<PickerMilestone> finalList = state.getCurrentMilestones();
             Optional<PickerMilestone> selectedMilestone = PickerMilestone.getSelectedMilestone(finalList);
             return new MilestonePickerDialogResponse(dialogButton, selectedMilestone.map(PickerMilestone::getId));
         });
@@ -115,7 +115,8 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
     }
 
     private void handleMouseClick(String milestoneName) {
-        inputField.setText(milestoneName);
+        inputField.setDisable(true);
+        state.toggleExactMatchMilestone(milestoneName);
         refreshUI(state);
     }
 
@@ -123,27 +124,18 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
         VBox milestoneDialogBox = createMilestoneDialogBox();
         assignedMilestoneBox = createAssignedMilestoneBox();
         matchingMilestonesBox = createMatchingMilestonesBox();
-        ScrollPane matchingMilestonesScrollPane = coverWithScrollPane(matchingMilestonesBox);
         inputField = createInputField();
 
-        milestoneDialogBox.getChildren().addAll(assignedMilestoneBox, inputField, matchingMilestonesScrollPane);
+        milestoneDialogBox.getChildren().addAll(assignedMilestoneBox, inputField, matchingMilestonesBox);
 
         getDialogPane().setContent(milestoneDialogBox);
         refreshUI(state);
         Platform.runLater(inputField::requestFocus);
     }
 
-    private ScrollPane coverWithScrollPane(VBox matchingMilestonesBox) {
-        ScrollPane matchingMilestonesScrollPane = new ScrollPane();
-        matchingMilestonesScrollPane.setMinHeight(200);
-        matchingMilestonesScrollPane.setMaxHeight(200);
-        matchingMilestonesScrollPane.setContent(matchingMilestonesBox);
-        return matchingMilestonesScrollPane;
-    }
-
     private void refreshUI(MilestonePickerState state) {
-        populateAssignedMilestone(state.getCurrentMilestonesList(), assignedMilestoneBox);
-        populateMatchingMilestones(state.getMatchingMilestonesList(), matchingMilestonesBox);
+        populateAssignedMilestone(state.getCurrentMilestones(), assignedMilestoneBox);
+        populateMatchingMilestones(state.getBestMatchingMilestones(), matchingMilestonesBox);
     }
 
     private void populateAssignedMilestone(List<PickerMilestone> pickerMilestoneList, FlowPane assignedMilestonePane) {
@@ -191,6 +183,7 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
 
     private HBox createDetailedMilestoneBox(PickerMilestone milestone) {
         HBox detailedMilestoneBox = new HBox();
+        detailedMilestoneBox.setPrefHeight(30);
         detailedMilestoneBox.setSpacing(3);
         detailedMilestoneBox.setPadding(new Insets(3, 3, 3, 3));
         detailedMilestoneBox.setStyle("-fx-border-width: 0 0 1 0; -fx-border-style: solid;");
@@ -255,6 +248,7 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
 
     private VBox createMatchingMilestonesBox() {
         VBox milestoneGroup = new VBox();
+        milestoneGroup.setPrefHeight(150);
         milestoneGroup.setStyle("-fx-border-radius: 3;-fx-background-color: white;-fx-border-color: black;");
         return milestoneGroup;
     }
--- a/src/main/java/ui/components/pickers/MilestonePickerState.java
+++ b/src/main/java/ui/components/pickers/MilestonePickerState.java
@@ -30,11 +30,8 @@ public class MilestonePickerState {
     }
 
     private void processInput(String userInput) {
-        if (userInput.isEmpty()) {
-            bestMatchingMilestones.addAll(currentMilestones);
-            return;
-        }
         determineBestMatchingMilestones(userInput);
+        if (userInput.isEmpty()) return;
         toggleFirstMilestone(bestMatchingMilestones);
     }
 
@@ -71,6 +68,9 @@ public class MilestonePickerState {
         this.currentMilestones.stream()
                 .forEach(listMilestone -> listMilestone.setSelected(listMilestone.equals(milestone)
                         && !listMilestone.isSelected()));
+        this.bestMatchingMilestones.stream()
+                .forEach(listMilestone -> listMilestone.setSelected(listMilestone.equals(milestone)
+                        && !listMilestone.isSelected()));
     }
 
     public List<PickerMilestone> getCurrentMilestones() {
@@ -82,7 +82,13 @@ public class MilestonePickerState {
     }
 
     private void determineBestMatchingMilestones(String userInput) {
-        bestMatchingMilestones.addAll(getLimitedMatchingMilestones(userInput, BEST_MATCHING_LIMIT));
+        if (userInput.isEmpty()) {
+            bestMatchingMilestones.addAll(currentMilestones);
+            return;
+        }
+        List<PickerMilestone> matchingMilestones = getLimitedMatchingMilestones(userInput, BEST_MATCHING_LIMIT);
+        System.out.println(matchingMilestones.toString());
+        bestMatchingMilestones.addAll(matchingMilestones);
         if (bestMatchingMilestones.size() == BEST_MATCHING_LIMIT) return;
         bestMatchingMilestones.addAll(getLimitedSuggestedMilestones(userInput, bestMatchingMilestones,
                 BEST_MATCHING_LIMIT - bestMatchingMilestones.size()));
@@ -93,8 +99,15 @@ public class MilestonePickerState {
                                                                 int noToAdd) {
         List<PickerMilestone> suggestedMilestones = new ArrayList<>();
         suggestedMilestones.addAll(getLimitedMatchesFromPreviousInput(userInput, matchingMilestones, noToAdd));
-        suggestedMilestones.addAll(getLimitedSortedMilestones(matchingMilestones,
+
+        List<PickerMilestone> curMilestones = new ArrayList<>(matchingMilestones);
+        curMilestones.addAll(suggestedMilestones);
+
+        suggestedMilestones.addAll(getLimitedSortedMilestones(curMilestones,
                                                               noToAdd - suggestedMilestones.size()));
+
+        suggestedMilestones.stream()
+                .forEach(milestone -> milestone.setMatching(false));
         return suggestedMilestones;
     }
 
--- a/src/main/java/ui/components/pickers/MilestonePickerState.java
+++ b/src/main/java/ui/components/pickers/MilestonePickerState.java
@@ -2,6 +2,7 @@ package ui.components.pickers;
 
 import util.Utility;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -12,14 +13,13 @@ import java.util.stream.Stream;
  * which determine the result of the user's input
  */
 public class MilestonePickerState {
-    private List<PickerMilestone> currentMilestonesList;
-
-    public MilestonePickerState(List<PickerMilestone> milestones) {
-        currentMilestonesList = cloneList(milestones);
-    }
+    public static final int BEST_MATCHING_LIMIT = 5;
+    private final List<PickerMilestone> currentMilestones;
+    private final List<PickerMilestone> bestMatchingMilestones;
 
     public MilestonePickerState(List<PickerMilestone> milestones, String userInput) {
-        this(milestones);
+        currentMilestones = cloneList(milestones);
+        bestMatchingMilestones = new ArrayList<>();
         processInput(userInput);
     }
 
@@ -30,13 +30,12 @@ public class MilestonePickerState {
     }
 
     private void processInput(String userInput) {
-        if (userInput.isEmpty()) return;
-
-        String[] userInputWords = userInput.split(" ");
-        Stream.of(userInputWords)
-                .forEach(this::filterMilestones);
-
-        toggleFirstMatchingMilestone(currentMilestonesList);
+        if (userInput.isEmpty()) {
+            bestMatchingMilestones.addAll(currentMilestones);
+            return;
+        }
+        determineBestMatchingMilestones(userInput);
+        toggleFirstMilestone(bestMatchingMilestones);
     }
 
     /**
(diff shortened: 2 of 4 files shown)


m133225:
Ready for review.
A significant change in MilestonePickerState.
It will now maintain a list of `bestMatchingMilestones` after processInput is called:
- Milestones that match the user input will be added to this list
- If there are still less than the limit (of 5), get all milestones that match subsets of the user input (e.g. "3.22" will check for milestones matching "3.2", "3.", "3"... in that order) until the input is empty. Add the first required number of milestones to the list (milestones added must not already be in the list)
- If the limit of 5 is still not reached, it will simply add the first required number of milestones in the given order in currentMilestones (milestones added must not already be in the list)

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -47,61 +57,206 @@ private void processInput(String userInput) {
...
+    public List<PickerMilestone> getCurrentMilestones() {
+        return currentMilestones;
+    }
+
+    public List<PickerMilestone> getBestMatchingMilestones() {
+        return bestMatchingMilestones;
+    }
+
+    /**
+     * Contains the algorithm to generate the best matching milestones
```
damithc: Avoid redundant words. e.g. `Contains the algorithm to`

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,13 +30,23 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
+                .forEach(query -> filterMilestonesByQuery(milestones, query));
+    }
 
-        toggleFirstMatchingMilestone(currentMilestonesList);
+    private void filterMilestonesByQuery(List<PickerMilestone> milestones, String query) {
+        milestones.stream()
+                .forEach(milestone -> {
+                    if (milestone.isMatching()) milestone.setMatching(isMatchingQuery(milestone, query));
+                });
     }
```
damithc: The difference between this method and the one above is the number of words in the query? In that case names can reflect that.
e.g. `filterMilestonesByQuerySentence` `filterMilestonesByQueryWord`

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -47,61 +57,206 @@ private void processInput(String userInput) {
...
+     * @param bestMatchingMilestones
+     * @param limit
+     * @param userInput
+     * @param milestones
+     */
+    private void addPreviouslyMatchingMilestones(List<PickerMilestone> bestMatchingMilestones, int limit,
+                                                 String userInput, List<PickerMilestone> milestones) {
+        if (limit <= 0) return;
+        List<PickerMilestone> previouslyMatchingMilestones = getLimitedMatchesFromPreviousInput(userInput,
+                bestMatchingMilestones, limit, milestones);
```
damithc: Line break after `=`?

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -47,61 +57,204 @@ private void processInput(String userInput) {
...
+     *
+     * The added milestones are references to the actual elements in currentMilestones, and thus should not be
+     * unnecessarily mutated
+     * @param userInput
+     */
+    private void determineBestMatchingMilestones(String userInput) {
+        addMatchingMilestones(bestMatchingMilestones, getNumberOfMilestonesToAdd(), currentMilestones);
+        addPreviouslyMatchingMilestones(bestMatchingMilestones, getNumberOfMilestonesToAdd(), userInput,
+                                        currentMilestones);
+        addLikelyUnmatchedMilestones(bestMatchingMilestones, getNumberOfMilestonesToAdd(), currentMilestones);
```
damithc: I think the level of abstraction here is high (which is good) but it hides the true logic. Consider this:
```
bestMatching = empty list;
extractBestMatching(bestMatching, all milestones, query);
if(bestMatching count fewer than expected length) 
    addPartialMatches(bestMatching, all milestones, query, expectedCount-actualCount);
if(bestMatching count fewer than expected length)
   addLikelyUnmatches(bestMatching, all milestones, expectedCount-actualCount);
```

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -12,14 +13,13 @@
  * which determine the result of the user's input
  */
 public class MilestonePickerState {
-    private List<PickerMilestone> currentMilestonesList;
-
-    public MilestonePickerState(List<PickerMilestone> milestones) {
-        currentMilestonesList = cloneList(milestones);
-    }
+    public static final int BEST_MATCHING_LIMIT = 5;
```
HansNewbie: Does this need to be public?

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -12,14 +13,13 @@
  * which determine the result of the user's input
  */
 public class MilestonePickerState {
-    private List<PickerMilestone> currentMilestonesList;
-
-    public MilestonePickerState(List<PickerMilestone> milestones) {
-        currentMilestonesList = cloneList(milestones);
-    }
```
HansNewbie: Why is this constructor removed?

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -47,61 +57,204 @@ private void processInput(String userInput) {
...
+     * @param userInput
+     */
+    private void determineBestMatchingMilestones(String userInput) {
+        addMatchingMilestones(bestMatchingMilestones, getNumberOfMilestonesToAdd(), currentMilestones);
+        addPreviouslyMatchingMilestones(bestMatchingMilestones, getNumberOfMilestonesToAdd(), userInput,
+                                        currentMilestones);
+        addLikelyUnmatchedMilestones(bestMatchingMilestones, getNumberOfMilestonesToAdd(), currentMilestones);
+    }
+
+    private int getNumberOfMilestonesToAdd() {
```
HansNewbie: Method name unclear. `...ToAdd` to what?

src/main/java/ui/components/pickers/MilestonePickerState.java
m133225: Oops, I was supposed to make the height of the dialog relative to this variable value. If that's the case then yes.

src/main/java/ui/components/pickers/MilestonePickerState.java
m133225: It won't be used anymore since the initialisation will pass the list of milestones together with an empty string instead.

src/main/java/ui/components/pickers/PickerMilestone.java
```diff
@@ -153,20 +154,25 @@ private void setRemovedInUI(Label milestone) {
...
 
-    public void setSelected(boolean isSelected) {
-        this.isSelected = isSelected;
+    private void setNotMatchingInUI(Label milestone) {
+        milestone.setStyle(milestone.getStyle() + " -fx-opacity: 60%;");
     }
 
-    public boolean isSelected() {
-        return this.isSelected;
+    public PickerMilestone setMatching(boolean isMatching) {
```
HansNewbie: Does it need the return value here? Seems the return value is not used anywhere.

src/main/java/ui/components/pickers/MilestonePickerState.java
HansNewbie: Can't we keep it? then we can call `this(milestones, "")` in this contructor.

src/test/java/unstable/MilestonePickerTests.java
```diff
@@ -13,21 +13,31 @@
 public class MilestonePickerTests extends UITest {
     @Test
     public void showMilestonePicker() {
-        click("#dummy/dummy_col0_9");
-        press(KeyCode.M);
+        clickIssue(0, 9);
+        push(KeyCode.M);
```
HansNewbie: I think we should use `press` here.

src/main/java/ui/components/pickers/PickerMilestone.java
```diff
@@ -23,8 +23,8 @@
     private static final int BIG_LABEL_FONT = 16;
 
     private boolean isSelected = false;
-    private boolean isMatching = false;
     private boolean isExisting = false;
+    private boolean isMatching = true;
```
HansNewbie: Just wondering if it makes a difference for `isMatching` here to be `true` or `false`.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -115,7 +115,8 @@ private Node setMouseClickForNode(Node node, String milestoneName) {
     }
 
     private void handleMouseClick(String milestoneName) {
-        inputField.setText(milestoneName);
+        inputField.setDisable(true);
```
HansNewbie: Do we need to disable the inputField? Just wondering if suddenly user wants to change the milestone query.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -123,27 +124,18 @@ private void initUI() {
         VBox milestoneDialogBox = createMilestoneDialogBox();
         assignedMilestoneBox = createAssignedMilestoneBox();
         matchingMilestonesBox = createMatchingMilestonesBox();
-        ScrollPane matchingMilestonesScrollPane = coverWithScrollPane(matchingMilestonesBox);
         inputField = createInputField();
 
-        milestoneDialogBox.getChildren().addAll(assignedMilestoneBox, inputField, matchingMilestonesScrollPane);
+        milestoneDialogBox.getChildren().addAll(assignedMilestoneBox, inputField, matchingMilestonesBox);
```
HansNewbie: Did you remove the scroll pane because now it should show only 5 best matching milestones? So even if input field is empty, it will still just show 5 milestones?

src/test/java/unstable/MilestonePickerTests.java
m133225: I think the `press` function doesn't let go of the key which causes the test to fail (the milestone picker dialog only appears _occasionally_). Also, the typing of `milestones` later will miss out the letter `m`. Had these problems previously which were solved by changing it to `push`.

src/main/java/ui/components/pickers/MilestonePickerState.java
m133225: Ok, will add it back.

src/test/java/unstable/MilestonePickerTests.java
HansNewbie: Right. I didn't see we don't have an overloaded method for `press`. My bad.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
m133225: Yes. This is in consideration of #1413, where a scrollable pane filled with all the milestones will be added. It will seem redundant/weird if the matching milestones pane is also scrollable. Or should we leave it as scrollable until #1413 is done?

src/main/java/ui/components/pickers/MilestonePickerDialog.java
m133225: Yes, this has been discussed with @damithc. This was decided mainly due to 2 reasons:
- consistent with label picker
- the best matching milestones should already contain the milestone that the user wishes to pick

src/main/java/ui/components/pickers/MilestonePickerDialog.java
damithc: Can remove. No point keeping to scrollable if we limit the list to 5 items.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
HansNewbie: Noted.

src/main/java/ui/components/pickers/PickerMilestone.java
m133225: Technically it shouldn't make a difference, as we can always set it later. But `MilestonePickerState` now assumes that they all start with `isMatching`, and does its filtering from there.

I guess `MilestonePickerState` should manually set `isMatching` to true before filtering? Then this initialisation can be false to be consistent with others.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
HansNewbie: Sure.

src/main/java/ui/components/pickers/MilestonePickerState.java
m133225: This was the initial implementation but it felt a bit overwhelming for a top-level method (with all the braces, variable names, etc). Will change it back.

src/main/java/ui/components/pickers/PickerMilestone.java
HansNewbie: Rather than to be consistent with others, I am thinking more on how the behaviour should be especially that things might change in the future. Setting to `false` makes sense to me because nothing should be matching at the start unless you do some preprocessing (say the bestMatching algorithm).

That being said, if the intended behaviour is for all `PickerMilestone` to match at the start (that is general to any algorithm other than best matching, say #1413), then this is fine.

HansNewbie:
Comment again when ready for another review.

m133225:
@HansNewbie Ready for review!

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,78 +34,236 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
+     * if they are not already in {@code bestMatchingMilestones}
+     *
+     * This method will do nothing if {@code limit} is not a positive integer
+     * @param bestMatchingMilestones
+     * @param limit
+     * @param milestones
+     */
+    private void addLikelyUnmatchedMilestones(List<PickerMilestone> bestMatchingMilestones, int limit,
+                                              List<PickerMilestone> milestones) {
+        if (limit <= 0) return;
```
damithc: This method is not supposed to be called with `limit <= 0` right? This kind of 'just in case' coding can hide programmer errors. I'd rather you state the assumption in the method comment and assert it here.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,78 +34,236 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
-     * <p>
-     * This method is NOT case-sensitive
+     * Toggles the selection status of the first milestone in the list
      *
      * @param milestoneList
      */
-    private final void toggleFirstMatchingMilestone(List<PickerMilestone> milestoneList) {
-        Optional<PickerMilestone> firstMatchingMilestone = getFirstMatchingMilestone(milestoneList);
+    private final void toggleFirstMilestone(List<PickerMilestone> milestoneList) {
+        Optional<PickerMilestone> firstMatchingMilestone = milestoneList.stream()
```
HansNewbie: Variable name is misleading. There is no other mention of matching in this whole method.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,78 +34,236 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
+    }
+
+    public List<PickerMilestone> getBestMatchingMilestones() {
+        return bestMatchingMilestones;
     }
 
-    public List<PickerMilestone> getCurrentMilestonesList() {
-        return this.currentMilestonesList;
+    /**
+     * Generates the best matching milestones which will then be added to bestMatchingMilestones
```
HansNewbie: Maybe `Populates bestMatchingMilestones from list of milestones which best matches user input`?

HansNewbie: Method name also can be better. `determine` sounds like you would like to get a result back, while in this case you are populating some internal variable.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,78 +34,236 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
+        if (limit <= 0) return;
+        List<PickerMilestone> likelyUnmatchedMilestones =
+                getLimitedFirstMilestones(bestMatchingMilestones, limit, milestones);
+        bestMatchingMilestones.addAll(likelyUnmatchedMilestones);
     }
 
     /**
-     * Gets the list of milestones that matches the current query
+     * Adds up to {@code limit} number of milestones to {@code bestMatchingMilestones}
+     * These milestones are taken from {@code milestones} if they match subsets of {@code userInput}, and are
```
HansNewbie: The `, and` is confusing; `,` can be removed.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,78 +34,236 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
+     * This method will do nothing if {@code limit} is not a positive integer
+     * @param bestMatchingMilestones
+     * @param limit
+     * @param userInput
+     * @param milestones
      */
-    public List<PickerMilestone> getMatchingMilestonesList() {
-        return this.currentMilestonesList.stream()
-                .filter(milestone -> !milestone.isMatching())
+    private void addPreviouslyMatchingMilestones(List<PickerMilestone> bestMatchingMilestones, int limit,
```
HansNewbie: Method name here is quite confusing. I thought it means from previous query, while it actually means user input that excludes last few characters. Maybe `LikelyMatchedMilestones` as opposite to `LikelyUnmatchedMilestones`?

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,78 +34,236 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
+    /**
+     * Gets first {@code limit} milestones from {@code milestones} which are not already in {@code allMilestones}
+     *
+     * Assumption: {@code limit} is not negative
+     * @param currentMilestones
+     * @param limit
+     * @param milestones
+     */
+    private List<PickerMilestone> getLimitedFirstMilestones(List<PickerMilestone> currentMilestones,
+                                                            int limit, List<PickerMilestone> milestones) {
```
HansNewbie: I don't feel comfortable calling `First` in this method, because it suggests that the input list is ordered that you get the first few. I can't think of a better method name, though.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,78 +34,236 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
 
-    private void filterMilestones(String query) {
-        currentMilestonesList
-                .forEach(milestone -> {
-                    boolean matchQuery = Utility.containsIgnoreCase(milestone.getTitle(), query);
-                    if (!milestone.isMatching()) milestone.setMatching(!matchQuery);
-                });
+    /**
+     * Concatenates all the matching milestones for previous userInput(s), then gets the first {@code limit}
+     * milestones which are not already in {@code currentMilestones}
```
HansNewbie: I think the comment can be better. The implementation of doing concatenation is not important to be mentioned in the comment. Maybe `Gets a limited number of milestones matching previous userInput(s) which are not already in {@code currentMilestones}.`

HansNewbie:
Mostly commenting on method names and header comments.

For header comments, I don't think it's a good practice to use `{@code sampleVar}` in describing the method. `@code` is just a formatting in JavaDoc to make HTML tag `<code>`. While it looks neat, mentioning parameter with `@code` in describing a method has some problems:
- It makes it hard to read the header comment in editing the code. It also makes the header comment archaic in a sense that it needs to fit the parameter name.
- We can always use `@param` to describe what a parameter is for. This also enables automatic refactoring in IDE if the parameter name is changed in the future.

For example:
```
/**
* Adds up to {@code limit} number of milestones to {@code bestMatchingMilestones}
* These milestones are taken from {@code milestones}, in the same order given,
* if they are not already in {@code bestMatchingMilestones}
*
* This method will do nothing if {@code limit} is not a positive integer
* @param bestMatchingMilestones
* @param limit
* @param milestones
*/
private void addLikelyUnmatchedMilestones(List<PickerMilestone> bestMatchingMilestones, int limit, List<PickerMilestone> milestones)
```

can become (_just for example, maybe this method can be kept the same_):
```
/**
* Adds a limited number of milestones from a given milestones to bestMatchingMilestones in the same order given if they are not already in bestMatchingMilestones.
*
* @param bestMatchingMilestones
* @param limit the number of milestones to be added to bestMatchingMilestones; nothing is done if this is not a positive integer
* @param milestones
*/
private void addLikelyUnmatchedMilestones(List<PickerMilestone> bestMatchingMilestones, int limit, List<PickerMilestone> milestones)
```
Even better, the parameter name can be more general but descriptive enough, say `milestonesToBePopulated` as opposed to `bestMatchingMilestones`. This is because the idea of `bestMatchingMilestones` may not be relevant in such method. Of course, this may not apply to all the methods; some might be relevant only when we are talking about `bestMatchingMilestones` in which we can keep the parameter name.

src/main/java/ui/components/pickers/MilestonePickerState.java
m133225: Sorry, don't get what you mean here... It is true that this assumes that the input list is already ordered.

HansNewbie: List it under assumption, then :)

m133225: Isn't it redundant though? Since the method already states that it is getting the first few milestones from the list.. :stuck_out_tongue:

HansNewbie: It's more like it's a required condition as the parameter. This method might be used by someone else in the future, in which if they don't know the input needs to be ordered, the output might be wrong.

If the input need not be ordered, saying `first` is not accurate, say the implementation can be changed to `Set` instead of `List`, in which there is no first few, just few (as it is random).

m133225:
@HansNewbie Made the changes!

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,82 +34,244 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
-        String[] userInputWords = userInput.split(" ");
+    private void filterMilestonesByQuerySentence(List<PickerMilestone> milestones, String querySentence) {
+        String[] userInputWords = querySentence.split(" ");
         Stream.of(userInputWords)
-                .forEach(this::filterMilestones);
+                .forEach(queryWord -> filterMilestonesByQueryWord(milestones, queryWord));
+    }
 
-        toggleFirstMatchingMilestone(currentMilestonesList);
+    private void filterMilestonesByQueryWord(List<PickerMilestone> milestones, String queryWord) {
```
HansNewbie: For this and above, `filter` does not seem accurate. If I understand correctly, it's more of `preprocessMatchingMilestonesBy...`. By `filter`, I assume I will get something back that is filtered, whether on the variable itself or in return value. In this case, it's more of processing each milestone to set matching.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,82 +34,244 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
+                .forEach(milestone -> {
+                    if (milestone.isMatching()) milestone.setMatching(isMatchingQuery(milestone, queryWord));
+                });
     }
 
     /**
-     * Toggles the milestone that has milestoneName as its exact name
-     * <p>
-     * This method is case-sensitive
+     * Toggles the selection status of the milestone that has milestoneName as its exact name
```
HansNewbie: Let's reword this. `Toggles the selection of the milestone which name is exactly milestoneName. This method is case-sensitive` (don't give line break as it may be skipped).

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,82 +34,244 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
+        Optional<PickerMilestone> firstMilestone = milestones.stream()
+                                                            .findFirst();
+        firstMilestone.ifPresent(this::toggleMilestone);
     }
 
     /**
-     * Changes the selection statuses of all milestones in the milestone list,
-     * such that the new given milestone is toggled, and the rest are not selected
+     * Changes the selection statuses of all milestones in allMilestones,
+     * such that the given milestone is toggled, and the rest are not selected
```
HansNewbie: Remove the `,`. It breaks the sentence flow.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,82 +34,244 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
+        return bestMatchingMilestones;
+    }
+
+    /**
+     * Populates bestMatchingMilestones with milestones from allMilestones that best match
+     * {@code userInput}.
+     *
+     * The added milestones are references to the actual elements in allMilestones, and thus should not be
+     * unnecessarily mutated
+     * @param userInput
```
HansNewbie: Put explanation on param, what is `userInput`. I think it's query sentence? `userInput` is very general and might become ambiguous.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,82 +34,244 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
+        if (isBelowLimit(bestMatchingMilestones, BEST_MATCHING_LIMIT)) {
+            addPartiallyMatchingMilestones(bestMatchingMilestones, getRemainingBestMatchesToLimit(), userInput,
+                                            allMilestones);
+        }
+        if (isBelowLimit(bestMatchingMilestones, BEST_MATCHING_LIMIT)) {
+            addLikelyUnmatchedMilestones(bestMatchingMilestones, getRemainingBestMatchesToLimit(), allMilestones);
+        }
+    }
+
+    private boolean isBelowLimit(List<PickerMilestone> milestones, int limit) {
```
HansNewbie: Rename to `isMilestonesSizeBelowLimit` or something similar. Can't tell what limit this method checks just from current signature.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,82 +34,244 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
+     * @param currentMilestones
+     * @param limit
+     * @param allMilestones
+     */
+    private List<PickerMilestone> getLimitedMatchesFromPreviousInput(String userInput,
+                                                                     List<PickerMilestone> currentMilestones,
+                                                                     int limit, List<PickerMilestone> allMilestones) {
+        List<PickerMilestone> matchesForPreviousInput = new ArrayList<>();
+        String curUserInput = userInput;
+        while (matchesForPreviousInput.size() < limit && curUserInput.length() > 0) {
```
HansNewbie: Do we need to assert limit in this method or return empty if limit is negative in this method?

src/test/java/tests/MilestonePickerStateTests.java
```diff
@@ -9,46 +9,55 @@
...
-        label2.setSelected(true);
-        labelList.add(label);
-        labelList.add(label2);
-        return new MilestonePickerState(labelList);
+        List<PickerMilestone> milestoneList = new ArrayList<>();
+        PickerMilestone milestone1 = new PickerMilestone(new TurboMilestone("testrepo", 1, "milestone1"));
+        PickerMilestone milestone2 = new PickerMilestone(new TurboMilestone("testrepo", 2, "milestone2"));
+        milestoneList.add(milestone1);
+        milestoneList.add(milestone2);
+        return new MilestonePickerState(milestoneList, milestone2.getTitle());
```
HansNewbie: Put a header comment to this method to tell that `userInput is set to title of milestone 2` or something similar.

src/test/java/tests/MilestonePickerStateTests.java
```diff
@@ -9,46 +9,55 @@
...
-        assertFalse(state.getCurrentMilestonesList().get(1).isSelected());
+        assertFalse(state.getAllMilestones().get(0).isSelected());
+        assertFalse(state.getAllMilestones().get(1).isSelected());
+    }
+
+    @Test
+    public void bestMatches_withUserInput_previousMatchingMilestoneAppears() {
+        MilestonePickerState state = prepareAssignedState();
+        assertEquals("milestone2", state.getBestMatchingMilestones().get(0).getTitle());
+        assertTrue(state.getBestMatchingMilestones().get(0).isSelected());
```
HansNewbie: Will this milestone also be `isMatching()`? If yes, should assert it as well.

HansNewbie:
Some more cosmetic changes requested. Maybe you can squash some of your commits as well.

src/main/java/ui/components/pickers/MilestonePickerState.java
m133225: This method already throws an `IllegalArgumentException` if limit is negative (since we can't limit a stream size to a negative number).

HansNewbie: Alright.

m133225:
@HansNewbie Ready for review.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -255,6 +248,7 @@ private FlowPane createAssignedMilestoneBox() {
 
     private VBox createMatchingMilestonesBox() {
         VBox milestoneGroup = new VBox();
+        milestoneGroup.setPrefHeight(30 * MilestonePickerState.BEST_MATCHING_LIMIT);
```
HansNewbie: Is `30` here the same as above (`detailedMilestoneBox.setPrefHeight(30)`)? If yes, we should abstract it out into a constant.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,82 +34,244 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
     private void processInput(String userInput) {
+        processMatchingMilestonesByQuerySentence(allMilestones, userInput);
+        populateBestMatchingMilestones(userInput);
         if (userInput.isEmpty()) return;
+        toggleFirstMilestone(bestMatchingMilestones);
+    }
 
-        String[] userInputWords = userInput.split(" ");
+    private void processMatchingMilestonesByQuerySentence(List<PickerMilestone> milestones, String querySentence) {
+        String[] userInputWords = querySentence.split(" ");
```
HansNewbie: `userInputWords` should be renamed. `queryWords`?

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,82 +34,244 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
-    public List<PickerMilestone> getCurrentMilestonesList() {
-        return this.currentMilestonesList;
+    public List<PickerMilestone> getBestMatchingMilestones() {
+        return bestMatchingMilestones;
     }
 
     /**
-     * Gets the list of milestones that matches the current query
+     * Populates bestMatchingMilestones with milestones from allMilestones that best match
+     * {@code userInput}.
```
HansNewbie: `userInput` -> `querySentence`?

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,82 +34,244 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
     }
 
-    private void filterMilestones(String query) {
-        currentMilestonesList
-                .forEach(milestone -> {
-                    boolean matchQuery = Utility.containsIgnoreCase(milestone.getTitle(), query);
-                    if (!milestone.isMatching()) milestone.setMatching(!matchQuery);
-                });
+    /**
+     * Gets a limited number of milestones from {@code allMilestones} matching previous userInputs(s)
```
HansNewbie: `userInput(s)`?

m133225: If the current userInput is of length 1 then there is only one previous userInput :stuck_out_tongue:

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -30,82 +34,244 @@ public MilestonePickerState(List<PickerMilestone> milestones, String userInput)
...
+    private List<PickerMilestone> getNewMatchingMilestones(String userInput, List<PickerMilestone> currentMilestones,
+                                                           List<PickerMilestone> allMilestones) {
+        String[] curUserInputWords = userInput.split(" ");
+
+        List<PickerMilestone> matchingMilestones = allMilestones;
+        for (String word : curUserInputWords) {
+            matchingMilestones = matchingMilestones.stream()
+                    .filter(milestone -> isMatchingQuery(milestone, word))
+                    .collect(Collectors.toList());
+        }
```
HansNewbie: Isn't there a bug here? If there is a space in the query, e.g. "Ver 1.3", this will only get milestones matching `1.3

HansNewbie: Sorry, mispressed a key. I updated the comment above.

src/test/java/tests/MilestonePickerStateTests.java
```diff
@@ -9,46 +9,57 @@
...
         MilestonePickerState state = prepareAssignedState();
         state.toggleExactMatchMilestone("milestone2");
-        assertFalse(state.getCurrentMilestonesList().get(0).isSelected());
-        assertFalse(state.getCurrentMilestonesList().get(1).isSelected());
+        assertFalse(state.getAllMilestones().get(0).isSelected());
+        assertFalse(state.getAllMilestones().get(1).isSelected());
+    }
+
+    @Test
+    public void bestMatches_withUserInput_previousMatchingMilestoneAppears() {
```
HansNewbie: `previousMatchingMilestoneIncluded` sounds more accurate. `Appears` suggests this is some UI state test, which isn't.

HansNewbie:
Some cosmetic changes requested. There seems to be a bug; awaiting for confirmation.

Otherwise, this PR is good to merge.

src/main/java/ui/components/pickers/MilestonePickerState.java
m133225: I don't think that is a bug and is intended. Multiple query words will only match milestones that match all query words (this allows the user to progressively narrow down the list of milestones). In the above example, it will match milestones that has `Ver` and `1.3` e.g. `Ver 1.3` and `Ver 1.3.99`.

m133225: Sorry, I think I misunderstood what you meant. This will remove unmatching milestones as we iterate through the words i.e. only milestones that match all queries will remain. So yes, "Ver 1.3", will only get milestones matching `Ver` __AND__ `1.3`.

HansNewbie: Alright. I misread. I thought `isMatchingQuery` is an exact match when it's a substring matching :stuck_out_tongue:

m133225:
@HansNewbie Made the requested changes and renamed all `userInput...` to `query...`.

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -23,6 +23,8 @@
 public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse> {
     private static final String OCTICON_ARROW = "\uf03e";
     private static final String DIALOG_TITLE = "Select Milestone";
+    public static final int DETAILED_MILESTONE_HEIGHT = 30;
+    public static final int PREV_ASSIGNED_MILESTONE_HEIGHT = 30;
```
HansNewbie: `private`?

HansNewbie:
Okay. Looks good. Just one change for constant to `private`. Will bump status to `toMerge` when it passes CI.

PR #1453, opened 2016-03-30
Fix KeyboardShortcutTest #1452

### LINKED ISSUE #1452: Fix KeyboardShortcutTest
There is a line using `pushKeys` instead of `press`. Seems it causes the test to fail.

HansNewbie:
Fixes #1452


HansNewbie:
@ndt93 just a one liner but seems that the bug causes the mentioned test to fail consistently (tested on Mac).

ndt93:
I'm not entirely sure about the rationale for the failure but seems to pass on Travis also, so should be good then.

HansNewbie:
@ndt93 the failure was because `pushKeys` doesn't use a reliable `release` key method while `press` does, which makes the test to pass more reliably. In the whole test, only this line uses `pushKeys`; the rest uses `press`.

PR #1458, opened 2016-04-03
Add login username to auto-create panel filters

### LINKED ISSUE #1457: Add login username to auto-create panel filters
`Recently updated issues` -> `assignee:codemanmeet updated:<48` obtained from the Auto-create Panels menu would have codemanmeet appear based on the logged in user.

codemanmeet:
Fixes #1457.


src/main/java/ui/PanelMenuCreator.java
```diff
@@ -48,15 +45,24 @@ public Menu generatePanelMenu() {
...
+        for (Map.Entry<String, String> entry : generateCustomizedPanelDetails().entrySet()) {
+            autoCreateItems.add(createCustomizedPanelMenuItem(entry.getKey(), entry.getValue()));
+        }
         autoCreatePanelMenu.getItems().addAll(autoCreateItems);
         panelMenu.getItems().addAll(items);
         panelMenu.getItems().add(autoCreatePanelMenu);
         return panelMenu;
     }
 
+    public Map<String, String> generateCustomizedPanelDetails(){
```
dariusf: - Can be static
- You can use Collections.unmodifiableMap before returning
- Use String.format instead of concatenating strings

src/main/java/ui/UI.java
```diff
@@ -221,6 +221,9 @@ private void initPreApplicationState() {
 
         TestController.setUI(this, getParameters());
         prefs = TestController.loadApplicationPreferences();
+        if(TestController.isBypassLogin()){
+            prefs.setLastLoginCredentials("test", "test");
+        }
```
dariusf: This doesn't look like it belongs here. Can we put this inside `loadApplicationPreferences` instead?

src/test/java/guitests/PanelMenuCreatorTest.java
```diff
@@ -5,44 +5,36 @@
...
-import static ui.PanelMenuCreator.ASSIGNEE_PANEL_NAME;
-import static ui.PanelMenuCreator.MILESTONE_FILTER_NAME;
-import static ui.PanelMenuCreator.MILESTONE_PANEL_NAME;
-import static ui.PanelMenuCreator.UPDATED_FILTER_NAME;
-import static ui.PanelMenuCreator.UPDATED_PANEL_NAME;
-
 public class PanelMenuCreatorTest extends UITest {
 
     private PanelControl panelControl;
+    private MenuControl menuBar;
```
dariusf: If we don't use this anywhere else, it doesn't have to be a field.

src/main/java/ui/UI.java
codemanmeet: Yup sure!

src/main/java/ui/PanelMenuCreator.java
codemanmeet: Non static field "prefs" cannot be referenced from a static context. The method cannot be static. I now return an unmodifiable map and have changed the strings to String.format.

dariusf: You can move the last login username into a parameter. That makes the function signature a bit clearer -- instead of taking the whole context of this object only to call one method, it takes the result of that method directly.

codemanmeet: Sure!

codemanmeet:
@dariusf ready for review!

src/main/java/ui/PanelMenuCreator.java
```diff
@@ -21,23 +21,17 @@
...
 
-    public static final String UPDATED_FILTER_NAME = "assignee:me updated:<48";
-    public static final String UPDATED_PANEL_NAME = "Recently Updated issues";
+    private static final Logger logger = LogManager.getLogger(MenuControl.class.getName());
 
     private final PanelControl panelControl;
     private final ScrollPane panelsScrollPane;
 
-    public PanelMenuCreator(PanelControl panelControl, ScrollPane panelsScrollPane) {
+    public PanelMenuCreator(PanelControl panelControl, ScrollPane panelsScrollPane, Preferences prefs) {
```
dariusf: Actually, if the last login username is all we need here, we can pass that instead of all of prefs. Unless you foresee us needing other fields?

src/main/java/ui/PanelMenuCreator.java
```diff
@@ -48,15 +42,33 @@ public Menu generatePanelMenu() {
         items.add(createLeftPanelMenuItem());
         items.add(createRightPanelMenuItem());
         items.add(closePanelMenuItem());
-        autoCreateItems.add(createCustomizedPanelMenuItem(ASSIGNEE_PANEL_NAME, ASSIGNEE_FILTER_NAME));
-        autoCreateItems.add(createCustomizedPanelMenuItem(MILESTONE_PANEL_NAME, MILESTONE_FILTER_NAME));
-        autoCreateItems.add(createCustomizedPanelMenuItem(UPDATED_PANEL_NAME, UPDATED_FILTER_NAME));
+        for (Map.Entry<String, String> entry :
+                generateCustomizedPanelDetails(prefs.getLastLoginUsername()).entrySet()) {
```
dariusf: This part can be made less verbose. Method names can be shortened (`customPanelDetails`?). If needed, you can also give names to intermediate values.

src/main/java/ui/PanelMenuCreator.java
```diff
@@ -48,15 +42,33 @@ public Menu generatePanelMenu() {
...
+            autoCreateItems.add(createCustomizedPanelMenuItem(entry.getKey(), entry.getValue()));
```
dariusf: `createCustomizedPanelMenuItem` is also pretty long. Is there a need for the `Customized` qualifier? Seems like all panels would be customised.

src/main/java/ui/PanelMenuCreator.java
```diff
@@ -48,15 +42,33 @@ public Menu generatePanelMenu() {
...
     }
 
+    /**
+     * Returns a map of custom panels that can be created using the auto-create menu
+     * with the key of the map as the panel name and the value as the corresponding filter name.
+     * Uses the username as a parameter to construct the filter names.
+     * @param lastLoginUsername
+     */
+    public static Map<String, String> generateCustomizedPanelDetails(String lastLoginUsername){
+        Map<String, String> customPanelMap = new LinkedHashMap<>();
```
dariusf: This doesn't need the `Map` suffix.

src/main/java/ui/PanelMenuCreator.java
```diff
@@ -48,15 +42,33 @@ public Menu generatePanelMenu() {
...
+        customPanelMap.put("Open issues and PR's",
+                String.format("is:open ((is:issue assignee:%s) OR (is:pr author:%s))", lastLoginUsername,
```
dariusf: We can use a positional argument in the format string here.

src/test/java/guitests/PanelMenuCreatorTest.java
```diff
@@ -6,43 +6,35 @@
...
-    }
-
-    @Test
-    public void recentlyUpdatedPanelMenuItemTest() {
-        customizedPanelMenuItemTest(UPDATED_PANEL_NAME, UPDATED_FILTER_NAME);
+    public void autoCreatePanels_createCustomPanelsFromMenu_panelsCreatedWithAppropriatePanelNameAndFilter()
+            throws NoSuchFieldException, IllegalAccessException {
+        Field panelMenuCreatorField = MenuControl.class.getDeclaredField("panelMenuCreator");
+        panelMenuCreatorField.setAccessible(true);
+        PanelMenuCreator value = (PanelMenuCreator) panelMenuCreatorField.get(ui.getMenuControl());
```
dariusf: Could we factor out the reflection? So we deal only with the `value` in here.

dariusf:
The PR title should also be changed.

A good way to structure this would be to have one commit for the `TestController` changes, and another to add the username to PanelMenuCreator.

codemanmeet:
@dariusf Ready for review!

src/main/java/ui/PanelMenuCreator.java
```diff
@@ -12,32 +12,25 @@
...
 public class PanelMenuCreator {
 
-    private static final Logger logger = LogManager.getLogger(MenuControl.class.getName());
-
-    public static final String MILESTONE_FILTER_NAME = "milestone:curr sort:status";
-    public static final String MILESTONE_PANEL_NAME = "Current Milestone";
-
-    public static final String ASSIGNEE_FILTER_NAME = "is:open ((is:issue assignee:me) OR (is:pr author:me))";
-    public static final String ASSIGNEE_PANEL_NAME = "Open issues and PR's";
+    private final String lastLoginUsername;
```
dariusf: Can we name this something more specific? It's the username of the current user.

src/main/java/ui/PanelMenuCreator.java
```diff
@@ -48,15 +41,33 @@ public Menu generatePanelMenu() {
         items.add(createLeftPanelMenuItem());
         items.add(createRightPanelMenuItem());
         items.add(closePanelMenuItem());
-        autoCreateItems.add(createCustomizedPanelMenuItem(ASSIGNEE_PANEL_NAME, ASSIGNEE_FILTER_NAME));
-        autoCreateItems.add(createCustomizedPanelMenuItem(MILESTONE_PANEL_NAME, MILESTONE_FILTER_NAME));
-        autoCreateItems.add(createCustomizedPanelMenuItem(UPDATED_PANEL_NAME, UPDATED_FILTER_NAME));
+        for (Map.Entry<String, String> entry :
```
dariusf: We won't need the `Map.` prefix if we qualify `Entry` when it is imported.

src/main/java/ui/PanelMenuCreator.java
```diff
@@ -48,15 +41,33 @@ public Menu generatePanelMenu() {
...
     }
 
+    /**
+     * Returns a map of custom panels that can be created using the auto-create menu
+     * with the key of the map as the panel name and the value as the corresponding filter name.
+     * Uses the username as a parameter to construct the filter names.
+     * @param lastLoginUsername
+     */
+    public static Map<String, String> generatePanelDetails(String lastLoginUsername){
+        Map<String, String> customPanel = new LinkedHashMap<>();
```
dariusf: customPanels?

src/main/java/ui/PanelMenuCreator.java
```diff
@@ -48,15 +41,33 @@ public Menu generatePanelMenu() {
...
+        customPanel.put("Open issues and PR's",
+                String.format("is:open ((is:issue assignee:%1$s) OR (is:pr author:%2$s))", lastLoginUsername,
+                        lastLoginUsername));
```
dariusf: I suggested the positional arguments so we don't have to repeat the variable here.

src/test/java/guitests/PanelMenuCreatorTest.java
```diff
@@ -6,43 +6,39 @@
...
-        customizedPanelMenuItemTest(ASSIGNEE_PANEL_NAME, ASSIGNEE_FILTER_NAME);
-    }
-
-    @Test
-    public void milestonePanelMenuItemTest() {
-        customizedPanelMenuItemTest(MILESTONE_PANEL_NAME, MILESTONE_FILTER_NAME);
+    public void autoCreatePanels_createCustomPanelsFromMenu_panelsCreatedWithAppropriatePanelNameAndFilter()
+            throws NoSuchFieldException, IllegalAccessException {
+        PanelMenuCreator value = (PanelMenuCreator) getPanelMenuCreatorField().get(ui.getMenuControl());
+        for (Map.Entry<String, String> entry :
```
dariusf: Same here.

dariusf:
Do take note of the comments on PR/issue title and commits.

codemanmeet:
Sure, will do!

codemanmeet:
@dariusf ready for review again!

src/main/java/ui/PanelMenuCreator.java
```diff
@@ -12,32 +12,26 @@
...
 
-    public static final String UPDATED_FILTER_NAME = "assignee:me updated:<48";
-    public static final String UPDATED_PANEL_NAME = "Recently Updated issues";
+    private static final Logger logger = LogManager.getLogger(MenuControl.class.getName());
 
     private final PanelControl panelControl;
     private final ScrollPane panelsScrollPane;
 
-    public PanelMenuCreator(PanelControl panelControl, ScrollPane panelsScrollPane) {
+    public PanelMenuCreator(PanelControl panelControl, ScrollPane panelsScrollPane, String lastLoginUsername) {
```
dariusf: The name of this parameter should be changed as well.

dariusf:
PR/issue title isn't changed.

codemanmeet:
Done!

PR #1459, opened 2016-04-03
Unexpected write error #1300

### LINKED ISSUE #1300: Unexpected write error
image

Quota seems to be ok and I have write permissions.
On a related note, the icon in the dialog looks like a close button.

ndt93:
Fixes #1300 

Propagates exceptions when replace labels, replace milestones or edit issue state failed so that more informative error message can be displayed.

### DIFF
--- a/src/main/java/backend/Logic.java
+++ b/src/main/java/backend/Logic.java
@@ -272,11 +272,7 @@ public class Logic {
                 })
                 .thenCombine(localLabelsReplaceFuture, (isUpdateSuccessful, locallyModifiedIssue) ->
                         handleIssueLabelsUpdateResult(isUpdateSuccessful, locallyModifiedIssue, originalLabels))
-                .thenAccept((isReplacementSuccessful) -> {
-                    if (!result.isCompletedExceptionally()) {
-                        result.complete(isReplacementSuccessful);
-                    }
-                });
+                .thenAccept(result::complete);
 
         return result;
     }
@@ -306,11 +302,7 @@ public class Logic {
                 .thenCombine(localMilestoneReplaceFuture, (isUpdateSuccessful, locallyModifiedIssue) ->
                         handleIssueMilestoneUpdateOnServerResult(
                                 isUpdateSuccessful, locallyModifiedIssue, oldMilestone))
-                .thenAccept((isReplacementSuccessful) -> {
-                    if (!result.isCompletedExceptionally()) {
-                        result.complete(isReplacementSuccessful);
-                    }
-                });
+                .thenAccept(result::complete);
 
         return result;
     }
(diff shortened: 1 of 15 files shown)


ndt93:
@dariusf Can you help review this? Thanks.

src/main/java/backend/Logic.java
```diff
@@ -264,9 +265,20 @@ public Model getRepo(String repoId) {
...
+        updateIssueLabelsOnServer(issue, newLabels)
+                .exceptionally((e) -> {
+                    result.completeExceptionally(e);
+                    return false;
+                })
                 .thenCombine(localLabelsReplaceFuture, (isUpdateSuccessful, locallyModifiedIssue) ->
-                        handleIssueLabelsUpdateResult(isUpdateSuccessful, locallyModifiedIssue, originalLabels));
+                        handleIssueLabelsUpdateResult(isUpdateSuccessful, locallyModifiedIssue, originalLabels))
+                .thenAccept((isReplacementSuccessful) -> {
+                    if (!result.isCompletedExceptionally()) {
```
dariusf: This part is a bit confusing -- we propagate failure using `completeExceptionally` above as well as with a boolean value. It's not clear what the difference between the two is from just reading this part. Is there a way to simplify it, like maybe `completeExceptionally` only at the end if `isReplacementSuccessful` is false?

src/main/java/undo/UndoController.java
```diff
@@ -73,6 +77,11 @@ private boolean handleActionResult(TurboIssue issue, Action<TurboIssue> action,
         return success;
     }
 
+    private boolean handleActionException(TurboIssue issue, Action<TurboIssue> action, Throwable exception) {
+        showErrorDialog(issue, action, Utility.removeFirstWord(exception.getMessage()));
```
dariusf: Maybe we can add a comment here saying what exception messages look like, and what the intended output is. Also, I'm not sure if messages are consistent across exception types. Might need a regex.

src/test/java/tests/LogicTests.java
```diff
@@ -88,6 +88,18 @@ public void replaceIssueMilestone_repoIOFails_unsuccessful() throws ExecutionExc
     }
 
     /**
+     * Tests that replaceIssueMilestone propagates exceptions when repoIO encounter exceptions
+     */
+    @Test(expected = ExecutionException.class)
```
dariusf: We can use [ExpectedException](https://github.com/HubTurbo/HubTurbo/blob/b20957b3c5f6d99901ab38c8ff21e67092c3464f/src/test/java/tests/CollaboratorServiceExTests.java#L38-L42) here. Think that's what we're supposed to standardise on.

src/main/java/backend/Logic.java
ndt93: There are two `Futures` here because we want to handle the exception both in `Logic` (reverting to the previous labels) and then still propagate it to the `UI` (`exceptionally` will end the exception propagation). That's the also the reason why the `exceptionally` is put in the middle. I will see how this can be made clearer.

ndt93:
@dariusf I've made some changes. You can review this again.

dariusf:
Nice work. Last thing: `handleActionException` isn't covered, and it seems like it needs a UI test. 😛 Is it worth covering?

ndt93:
Yeah, we should probably cover that T.T

ndt93:
@dariusf Covered the method with a UI test. You can check this again.

PR #1462, opened 2016-04-03
`BoardAutoCreator` should not prompt for save if there is no need to #1461

### LINKED ISSUE #1461: BoardAutoCreator should not prompt for save if current board is not dirty
Actually I don't understand why `BoardAutoCreator` would want to trigger a `Save as` when current board is not dirty. Even if current board is dirty, as long as it was a saved board, not a "nameless" board, we should trigger `Save` instead of `Save as`.

Steps to reproduce:
1. Open a saved board, do not make any changes to it. 

2. Choose `Menu > Boards > Auto-create > Sample Board! It is prompting you to save even there is no need to.
screen shot 2016-04-03 at 15 45 35

3. Choose `Yes`. It is prompting you to enter new board name instead of just saving the current board.
screen shot 2016-04-03 at 15 46 00

xinan:
Fixes #1461

Now it will only prompt for save if current board is dirty, or current board has not been saved before. 

If current board is a dirty saved board, it will trigger `Save`. 

If current board has not been saved before, it will trigger `Save as`.

### DIFF
--- a/src/main/java/prefs/PanelInfo.java
+++ b/src/main/java/prefs/PanelInfo.java
@@ -2,6 +2,9 @@ package prefs;
 
 import java.util.Objects;
 
+/**
+ * Represents the information about a Panel. Contains two String attributes {@code name} and {@code filter}.
+ */
 public class PanelInfo {
 
     private final String name;
--- a/src/main/java/prefs/PanelInfo.java
+++ b/src/main/java/prefs/PanelInfo.java
@@ -3,7 +3,7 @@ package prefs;
 import java.util.Objects;
 
 /**
- * Represents the information about a Panel. Contains two String attributes {@code name} and {@code filter}.
+ * Represents the information about a Panel.
  */
 public class PanelInfo {
 
--- a/src/main/java/prefs/PanelInfo.java
+++ b/src/main/java/prefs/PanelInfo.java
@@ -1,5 +1,7 @@
 package prefs;
 
+import java.util.Objects;
+
 public class PanelInfo {
 
     private final String name;
@@ -23,4 +25,16 @@ public class PanelInfo {
         return this.filter;
     }
 
+    @Override
+    public boolean equals(Object object) {
+        if (this == object) return true;
+        if (object == null || getClass() != object.getClass()) return false;
+        PanelInfo panelInfo = (PanelInfo) object;
+        return Objects.equals(name, panelInfo.name) && Objects.equals(filter, panelInfo.filter);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(name, filter);
+    }
 }
--- a/src/main/java/ui/BoardAutoCreator.java
+++ b/src/main/java/ui/BoardAutoCreator.java
@@ -62,21 +62,21 @@ public class BoardAutoCreator {
         Menu autoCreate = new Menu("Auto-create");
         MenuItem sample = new MenuItem(SAMPLE_BOARD);
         sample.setOnAction(e -> {
-            saveBoardAfterUserConfirmation(SAMPLE_BOARD);
+            saveBoardAfterUserConfirmationIfNeeded(SAMPLE_BOARD);
             createSampleBoard(true);
         });
         autoCreate.getItems().add(sample);
 
         MenuItem milestone = new MenuItem(MILESTONES);
         milestone.setOnAction(e -> {
-            saveBoardAfterUserConfirmation(MILESTONES);
+            saveBoardAfterUserConfirmationIfNeeded(MILESTONES);
             createMilestoneBoard();
         });
         autoCreate.getItems().add(milestone);
 
         MenuItem workAllocation = new MenuItem(WORK_ALLOCATION);
         workAllocation.setOnAction(e -> {
-            saveBoardAfterUserConfirmation(WORK_ALLOCATION);
+            saveBoardAfterUserConfirmationIfNeeded(WORK_ALLOCATION);
             createWorkAllocationBoard();
         });
         autoCreate.getItems().add(workAllocation);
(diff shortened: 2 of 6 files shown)


ndt93:
@xinan Hey xinan, can you resolve the conflicts with mater branch?

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -84,8 +84,11 @@ public Menu generateBoardAutoCreateMenu() {
         return autoCreate;
     }
 
-    private void saveBoardAfterUserConfirmation(String boardName) {
-        if (isSaveBoardDialogResponsePositive(boardName)) ui.getMenuControl().saveBoardAs();
+    private void saveBoardAfterUserConfirmationIfNeeded(String boardName) {
+        boolean isNeeded = !prefs.getLastOpenBoardPanelInfos().isPresent()
+                        || !prefs.getLastOpenBoardPanelInfos().get().equals(panelControl.getCurrentPanelInfos());
```
damithc: I think a more accurate name is `isOpenBoardDifferentFromSavedBoard` or something like that, because that's what the condition seems to be checking.

src/test/java/guitests/BoardAutoCreatorTest.java
```diff
@@ -65,6 +71,45 @@ public void boardAutoCreator_clickYesInSavePrompt_currentBoardSaved() {
...
+        // create 3 new panels
+        pushKeys(KeyboardShortcuts.CREATE_RIGHT_PANEL);
+        pushKeys(KeyboardShortcuts.CREATE_RIGHT_PANEL);
+        pushKeys(KeyboardShortcuts.CREATE_RIGHT_PANEL);
+        PlatformEx.waitOnFxThread();
+        assertEquals(panelCount + 3, panelControl.getPanelCount());
+
+        // create milestones board
+        traverseMenu("Boards", "Auto-create", "Milestones");
+        PlatformEx.waitOnFxThread();
```
damithc: @ndt93 @xinan what's intent of this statement? The rest of the code is written at a higher level except this statement. Perhaps we should abstract it into a method (but probably not in this PR).

ndt93: From the docs in `PlatformEx`, the `waitOnFxThread` is used to block until the java fx events queue become empty. We should rename it so that the purpose is clearer?

src/main/java/prefs/PanelInfo.java
```diff
@@ -1,5 +1,7 @@
 package prefs;
 
+import java.util.Objects;
+
 public class PanelInfo {
```
ndt93: Not a change in this PR, but can you help adding the header comment for this class?

src/main/java/ui/BoardAutoCreator.java
```diff
@@ -84,8 +84,13 @@ public Menu generateBoardAutoCreateMenu() {
         return autoCreate;
     }
 
-    private void saveBoardAfterUserConfirmation(String boardName) {
-        if (isSaveBoardDialogResponsePositive(boardName)) ui.getMenuControl().saveBoardAs();
+    private void saveBoardAfterUserConfirmationIfNeeded(String boardName) {
```
ndt93: The confusing name of this method might be a call to refactor some methods or maybe add a comment explaining the `IfNeeded` and `UserConfirmation` part.

src/main/java/ui/MenuControl.java
```diff
@@ -89,19 +89,25 @@ private Menu createFileMenu() {
     }
 
     /**
-     * Creates an empty board. User will be prompted to
-     * confirm the action if there are unclosed panels.
+     * Called upon the Boards > New being clicked
```
ndt93: Comments should better explain what it does rather then how it's called.

src/main/java/ui/MenuControl.java
```diff
@@ -126,11 +132,20 @@ private boolean isNewBoardCreationDialogConfirmed() {
                 response.get().getButtonData() == ButtonData.OK_DONE;
     }
 
+    /**
+     * Called upon the Boards > Save being clicked
+     */
```
ndt93: Same here.

src/main/java/ui/MenuControl.java
```diff
@@ -126,11 +132,20 @@ private boolean isNewBoardCreationDialogConfirmed() {
...
     private void onBoardSave() {
         logger.info("Menu: Boards > Save");
+        saveBoard();
+    }
 
+    /**
+     * Tries to save current board. If current board is dangling, it calls {@code saveBoardAs()}
+     */
+    public final void saveBoard() {
```
ndt93: Not very clear what's `dangling` is referring to here.

src/main/java/ui/MenuControl.java
```diff
@@ -193,7 +220,15 @@ private void onBoardOpen(String boardName, List<PanelInfo> panelInfo) {
      */
     private void onBoardDelete(String boardName) {
         logger.info("Menu: Boards > Delete > " + boardName);
+        deleteBoard(boardName);
+    }
 
+    /**
+     * Delete the board named {@boardName}
+     *
```
ndt93: Should be `Deletes`

src/test/java/guitests/BoardAutoCreatorTest.java
damithc: Yes, we should wrap it in a method that states the high level intent. The current statement doesn't match the abstraction level of the code around it. But can be done as a separate issue.

src/main/java/prefs/PanelInfo.java
```diff
@@ -1,5 +1,10 @@
 package prefs;
 
+import java.util.Objects;
+
+/**
+ * Represents the information about a Panel. Contains two String attributes {@code name} and {@code filter}.
```
damithc: 2nd sentence is unnecessary and easily outdated.

damithc:
Can merge after removing the comment text mentioned.

PR #1463, opened 2016-04-03
Milestone picker: make whole row clickable #1425

### LINKED ISSUE #1425: Milestone picker: make whole row clickable
We can make the whole row clickable, not just the milestone name. Furthermore, the mouse pointer should indicate the area to be clickable. 
image

garbanzos:
Fixes #1425.

### DIFF
--- a/src/main/java/ui/components/pickers/PickerMilestone.java
+++ b/src/main/java/ui/components/pickers/PickerMilestone.java
@@ -3,8 +3,11 @@ package ui.components.pickers;
 import backend.resource.TurboMilestone;
 import com.sun.javafx.tk.FontLoader;
 import com.sun.javafx.tk.Toolkit;
+import javafx.geometry.Insets;
+import javafx.geometry.Pos;
 import javafx.scene.Node;
 import javafx.scene.control.Label;
+import javafx.scene.layout.HBox;
 import javafx.scene.text.Font;
 
 import java.util.List;
@@ -26,6 +29,8 @@ public class PickerMilestone extends TurboMilestone implements Comparable<Picker
     private boolean isExisting = false;
     private boolean isMatching = true;
 
+    private double progress = 0;
+
     public PickerMilestone(TurboMilestone milestone) {
         super(milestone.getRepoId(), milestone.getId(), milestone.getTitle());
         setDueDate(milestone.getDueDate());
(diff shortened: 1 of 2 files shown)


garbanzos:
@ndt93 Could you assigned a review for this?

ndt93:
@m133225 Hi Martin, can you help to review this PR? Thanks.

src/main/java/ui/components/pickers/PickerMilestone.java
```diff
@@ -63,6 +59,56 @@ public Node getNewlyAssignedMilestoneNode() {
...
+        milestoneDetailsBox.setAlignment(Pos.CENTER_RIGHT);
+
+        if (getDueDate().isPresent()) {
+            Label dueDate = new Label("Due on: " + getDueDate().get().toString());
+            dueDate.setPrefWidth(150);
+            milestoneDetailsBox.getChildren().add(dueDate);
+        }
+        int totalIssues = getOpenIssues() + getClosedIssues();
+        double progressValue = totalIssues > 0
+                ? (double) getClosedIssues() / (getOpenIssues() + getClosedIssues()) : 0;
```
m133225: Now that this is being done in `PickerMilestone`, I think we can extract getting `progressValue` into a separate method and compute it in the constructor instead. We can store its result in a class variable. This is so that it does not need to re-compute `progressValue` every time this is called.

Also, do change the `(getOpenIssues() + getClosedIssues())` to `totalIssues` which I missed out the last time. :stuck_out_tongue:

m133225:
Other than the mentioned change, I think this PR is good to go.

src/main/java/ui/components/pickers/PickerMilestone.java
```diff
@@ -64,6 +62,60 @@ public Node getNewlyAssignedMilestoneNode() {
...
+
+        HBox milestoneNodeBox = createMilestoneNodeBox();
+        Label separator = new Label("|");
+        HBox milestoneDetailsBox = createMilestoneDetailsBox();
+
+        matchingMilestone.getChildren().setAll(milestoneNodeBox, separator, milestoneDetailsBox);
+        return matchingMilestone;
+    }
+
+    private double calculateProgress(TurboMilestone milestone) {
```
m133225: Should only pass the required parameters here instead of the whole milestone object.

src/main/java/ui/components/pickers/PickerMilestone.java
```diff
@@ -64,6 +62,60 @@ public Node getNewlyAssignedMilestoneNode() {
...
+        int totalIssues = milestone.getOpenIssues() + milestone.getClosedIssues();
+        return totalIssues > 0 ? (double) milestone.getClosedIssues() / (totalIssues) : 0;
```
m133225: Brackets not required for `totalIssues`.

PR #1473, opened 2016-04-04
Error message: invalid keyboard shortcuts #1432

### LINKED ISSUE #1432: Error message: invalid keyboard shortcuts
I get this error when I open HT with latest master 
image

m133225:
Fixes #1432.

Removed customised shortcuts entirely.


m133225:
Ready for review.

ndt93:
@dariusf Hi Darius, can help review this? Thanks.

src/main/java/ui/components/KeyboardShortcuts.java
```diff
@@ -1,61 +1,49 @@
...
- * <p>
- * Classes that currently have keyboard shortcut code:
- * ui.components.NavigableListView
- * ui.issuepanel.PanelControl
- * ui.listpanel.ListPanel
- * ui.MenuControl
- * <p>
- * Utility Class:
- * util.KeyPress
+ * This class handles the keyboard shortcuts used by the different UI components
```
dariusf: Hmm, this class seems quite self-explanatory now. Don't think we need this any more.

src/main/java/ui/components/KeyboardShortcuts.java
```diff
@@ -1,61 +1,49 @@
...
-    private static final Logger logger = LogManager.getLogger(KeyboardShortcuts.class.getName());
-
-    private static Map<String, String> keyboardShortcuts = null;
-    private static Set<KeyCodeCombination> assignedKeys = null;
-
-    // customizable keyboard shortcuts
     // ui.listpanel.ListPanel
-    public static KeyCodeCombination markAsRead;
-    public static KeyCodeCombination markAsUnread;
+    public static final KeyCodeCombination MARK_AS_READ =
```
dariusf: We don't need the comments saying where each of these is used either.

dariusf:
Nice work. Can we also remove [these exceptions](https://github.com/HubTurbo/HubTurbo/blob/master/config/findbugs/excludeFilter.xml#L63)?

src/main/java/ui/components/KeyboardShortcuts.java
m133225: Actually, wouldn't it be better to keep it? This comment will explain to new contributors that this class is meant for all UI components instead of some subset of them. Moreover, it is suggested by @damithc to add header comments to all production classes. :stuck_out_tongue:

dariusf: I feel like that's clear from everything in the class being static and final, but I guess it's subjective.

PR #1481, opened 2016-04-05
assignee picker: previous and new assignee should not be the same #1479

### LINKED ISSUE #1479: assignee picker: previous and new assignee should not be the same
In this situation, the new assignee should be blank as the use intent is to remove the existing assignee

image

garbanzos:
Fixes #1479 

The assignee picker's text field is now initally blank. Pressing enter at this point will unassign the existing assignee.

image

~~**EDIT**~~
~~The existing assignee will not be struck out initially.~~ 
~~If the user presses backspace when the text field is empty, the existing assignee will be struck out. 
If enter is pressed at this point, the existing assignee will be unassigned.~~

### DIFF
--- a/src/test/java/guitests/AssigneePickerTests.java
+++ b/src/test/java/guitests/AssigneePickerTests.java
@@ -18,8 +18,6 @@ public class AssigneePickerTests extends UITest {
     public void showAssigneePicker_typeQuery_displaysCorrectly() {
         triggerAssigneePicker(getIssueCell(0, 9).getIssue());
         clickAssigneePickerTextField();
-        selectAll();
-        push(KeyCode.BACK_SPACE);
         type("world");
         assertEquals("world", getAssigneePickerTextField().getText());
         exitCleanly();
@@ -29,8 +27,6 @@ public class AssigneePickerTests extends UITest {
     public void showAssigneePicker_noAssignee_assigneeAssigned() {
         TurboIssue issue = getIssueCell(0, 9).getIssue();
         triggerAssigneePicker(issue);
-        selectAll();
-        push(KeyCode.BACK_SPACE);
         type("User");
         push(KeyCode.ENTER);
         assertEquals(true, issue.getAssignee().isPresent());
--- a/src/test/java/guitests/AssigneePickerTests.java
+++ b/src/test/java/guitests/AssigneePickerTests.java
@@ -3,6 +3,7 @@ package guitests;
 import backend.resource.TurboIssue;
 import javafx.application.Platform;
 import javafx.scene.input.KeyCode;
+import org.junit.After;
 import org.junit.Test;
 import ui.IdGenerator;
 import ui.UI;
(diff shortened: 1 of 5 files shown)


damithc:
It should be blank, but the cursor should be in the text box. Furthermore, the current assignee should not be stricken out at the beginning. It will be stricken out if the user press backspace or delete. does it work that way?

garbanzos:
@damithc Actually, initially the assignee picker was filled with the existing assignee because the milestone picker was filled with the existing milestone. 

I thought the assignee picker's interactions should be the similar to the milestone picker? If we change  the assignee picker to start off with a blank text field, it will be different from the milestone picker?

damithc:
I didn't notice it before. Both need to change. When there is no text in the text box, the 'best matches' panel should be filled with a suitable 'likely' targets. e.g. the next 4 milestone + recently closed milestone.

garbanzos:
Ok. 

So for the assignee picker: 
When the user opens the assignee picker, the text field should be empty and the existing assignee is not struck out. 

When the text field is empty: 
If the user presses backspace, the existing assignee is struck out. (If enter is pressed that this point, the existing assignee will be unassigned). 
If the user presses backspace another time, the existing assignee will not be struck out again. (If enter is pressed, no changes will be made).

damithc:
I don't think toggling is needed. User can simply use esc to dismiss the picker rather than toggle the struck out assignee.

acruis:
Ready to review?

Also, do write some tests for this!

garbanzos:
@acruis Ready for review.

damithc:
@garbanzos wait a bit before we decide the behavior as per discussion in #1487 
Over there I'm suggesting the current assignee to be stricken out if the text box is empty. That means there is no need for backspace.

garbanzos:
@damithc  

So, the assignee picker will have this behaviour now: 

The assignee starts off with an empty text field and the existing assignee is struck out. 
If the user presses enter when the text field is empty, the existing assignee is unassigned. 

Is this correct?

garbanzos:
@acruis Ready for review!

acruis:
Maybe add a test into `AssigneePickerTests`?

garbanzos:
@acruis Ready for review!

src/test/java/guitests/AssigneePickerTests.java
```diff
@@ -12,28 +12,51 @@
...
+    public void showAssigneePicker_pickAUser_userPickedAssigned() {
         TurboIssue issue = getIssueCell(0, 9).getIssue();
         triggerAssigneePicker(issue);
-        selectAll();
-        push(KeyCode.BACK_SPACE);
-        type("User");
+        type("User 1");
         push(KeyCode.ENTER);
+        assertEquals("User 1", issue.getAssignee().get());
+        exitCleanly();
```
acruis: You can put a `@Before` for `exitCleanly()`; that means you don't have to manually call it in every `@Test` method.

garbanzos: Shouldn't it be a `@After`?

garbanzos:
@acruis Ready for review again!

src/test/java/guitests/AssigneePickerTests.java
acruis: Oops, yup, it's `@After`.

acruis:
OK, looks good!

PR #1485, opened 2016-04-05
`Menu > Boards > New` should restore current board if save is aborted #1460

### LINKED ISSUE #1460: `Menu > Boards > New` should restore current board if save is aborted
The way we are doing it now is a bit hack-ish. It clear the current board, and calls `Menu > Boards > Save as`. So what happens if I abort the `Save as`?

Steps to reproduce the bug:
1. Choose `Menu > Boards > New` and click `OK`.
screen shot 2016-04-03 at 15 35 20
2. Click `Cancel` to abort the `Save as` dialog.
screen shot 2016-04-03 at 15 35 30
3. Voila! You remain at your current board but all panels are closed.
screen shot 2016-04-03 at 15 35 37

xinan:
Fixes #1460 

* Now it will prompt for save if the current board is dirty, instead of alerting all the time but does not provide option to save. 
* It also does not clear the current board before the `New` action is confirmed. So it is fine to cancel a `New` action. 
* Instead of opening an empty board without any panels. `Boards > New` now opens up a fresh board with a single empty panel. (Default panel name and no filter.)


HansNewbie:
@xinan no worries, close & reopen to trigger travis

src/main/java/ui/ConfirmChangesDialog.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public class ConfirmChangesDialog extends Alert {
+
+    private static final String DIALOG_TITLE = "Save Changes?";
+    private static final String DIALOG_HEADER = "There are unsaved changes to your current board.";
+    private static final String DIALOG_CONTENT = "If you choose not to save, all unsaved changes will be discarded.\n\n"
+                                               + "Do you want to save them?";
+
+    public ConfirmChangesDialog(Stage stage) {
+        super(AlertType.CONFIRMATION);
```
HansNewbie: We will need to merge this whole class to `Util.DialogMessage`. Either use the existing `Warning` dialog (not recommended) or create a new confirmation dialog in that class. We would like any dialog to be available for general use.

src/main/java/ui/MenuControl.java
```diff
@@ -89,47 +86,69 @@ private Menu createFileMenu() {
     }
 
     /**
-     * Logs the user action and creates a new board
+     * Checks whether the current open board has unsaved changes
+     * @return {@code true} if current open board is dirty, {@code false} otherwise.
```
HansNewbie: Don't use `{@code}` for `true` and `false`. Also, should define in comment about what is dirty, e.g. `... board is dirty, i.e. has unsaved changes`.

src/main/java/ui/MenuControl.java
```diff
@@ -89,47 +86,69 @@ private Menu createFileMenu() {
...
+            return true;
+        }
+        return !panels.getCurrentPanelInfos().equals(prefs.getLastOpenBoardPanelInfos().get());
     }
 
     /**
-     * Creates an empty board. User will be prompted to
-     * confirm the action if there are unclosed panels
+     * Prompts a {@code ConfirmChangesDialog} to ask the user whether or not to save the unsaved changes
+     * in the current board.
```
HansNewbie: If you want to refer to the class (need to be changed though), use `@link`. Also, this method does not need to know if the current board has unsaved changes. `Prompts ... whether or not to save the current board` should be sufficient.

src/main/java/ui/MenuControl.java
```diff
@@ -89,47 +86,69 @@ private Menu createFileMenu() {
...
      */
-    private boolean isNewBoardCreationDialogConfirmed() {
-        List<PanelInfo> panelList = panels.getCurrentPanelInfos();
-        if (panelList.isEmpty()) {
-            return true;
+    public final Optional<String> promptForNewBoardName() {
+        BoardNameDialog dlg = new BoardNameDialog(prefs, mainStage);
+        Optional<String> response = dlg.showAndWait();
+        ui.showMainStage();
+        panels.selectFirstPanel();
```
HansNewbie: I don't think `selectFirstPanel` is relevant to this method name. Should be done by the caller.

src/main/java/ui/MenuControl.java
```diff
@@ -89,47 +86,69 @@ private Menu createFileMenu() {
...
+        return response;
+    }
+
+    /**
+     * Logs the user action and creates a new board
+     */
+    private void onBoardNew() {
```
HansNewbie: Can it be renamed to `createNewBoard`? Remove `Logs the user action` from the header comment as it's not about the method.

src/main/java/ui/MenuControl.java
```diff
@@ -89,47 +86,69 @@ private Menu createFileMenu() {
...
-        return response.isPresent() &&
-                response.get().getButtonData() == ButtonData.OK_DONE;
+        newBoard();
+    }
+
+    /**
+     * Creates an empty board. User will be prompted to
+     * confirm the action if there are unclosed panels
+     */
+    public final void newBoard() {
```
HansNewbie: Rename method to `createEmptyBoard`. Also, update header comment.

src/main/java/ui/MenuControl.java
```diff
@@ -173,21 +192,27 @@ private void onBoardSaveAs() {
      * Prompts a BoardNameDialog and saves the current board based on its response
      */
     public final void saveBoardAs() {
```
HansNewbie: Can you help to update header comment? Should not be tied to implementation. `Prompts for board name and save the current board as per the name` or something similar.

src/main/java/ui/ConfirmChangesDialog.java
xinan: Why use of the existing `showYesNoWarningDialog` is not recommended?

src/main/java/ui/MenuControl.java
```diff
@@ -173,21 +192,27 @@ private void onBoardSaveAs() {
...
-            prefs.addBoard(boardName, panelList);
-            prefs.setLastOpenBoard(boardName);
-            prefs.setLastOpenBoardPanelInfos(panelList);
-            ui.triggerEvent(new BoardSavedEvent());
-            logger.info("New board " + boardName + " saved");
-            ui.updateTitle();
-        }
+        response.ifPresent(boardName -> {
+            createAndOpenNewBoard(boardName, panels.getCurrentPanelInfos());
+        });
```
HansNewbie: Shouldn't this have `ui.triggerEvent(new BoardSavedEvent());`?

HansNewbie: Wait nevermind.

src/main/java/ui/MenuControl.java
```diff
@@ -173,21 +192,27 @@ private void onBoardSaveAs() {
...
+    }
+
+    /**
+     * Creates a new board with the given board name and panel infos, then open that new board
+     * @param boardName
+     * @param panelList
+     */
+    public final void createAndOpenNewBoard(String boardName, List<PanelInfo> panelList) {
```
HansNewbie: This method should be separated to:
- `saveBoard()` - for all the `prefs` and `triggerEvent`
- `showBoard()` - for `openBoard`. `ui.updateTitle()` should actually be called in `openBoard` I believe.

src/main/java/ui/MenuControl.java
xinan: This is the method that will be the called when user clicked `Boards > New`. It is named similarly as other user click actions. 

I abstract all the actual `creating`, `deleting` and `saving` actions out of the UI event handlers in #1462 to make them available as APIs. 

The only arguable thing here is whether I should put the prompt in the UI event handler (i.e. `onBoardNew`, it is not really an event handler but it acts like an event handler), or in the actual action. I put it in `newBoard` instead of `onBoardNew` because I think some other classes that will be calling `newBoard` might not want that prompt.

src/main/java/ui/MenuControl.java
xinan: Actually it is not really an "empty" board. It has an empty panel by default. Coz I think it is not useful to create just an empty board, the first thing that the user will do is to add an empty board to it. So why don't we just do that for the user?

src/main/java/ui/ConfirmChangesDialog.java
HansNewbie: Since you use `confirmation` type here. I think there will be some UI
difference, also. Unless if your dialog is also a warning.

src/main/java/ui/MenuControl.java
HansNewbie: Sure. Let's match the header comment then.

src/main/java/ui/MenuControl.java
xinan: I just realised that these two lines are actually not needed.

src/main/java/ui/MenuControl.java
```diff
@@ -89,47 +87,71 @@ private Menu createFileMenu() {
...
     /**
-     * Logs the user action and creates a new board
+     * Checks whether the current open board is dirty, i.e. has unsaved changes.
+     *
+     * @return true if current open board is dirty, false otherwise
      */
-    private void onBoardNew() {
-        logger.info("Menu: Boards > New");
-        newBoard();
+    public final boolean isCurrentOpenBoardDirty() {
```
HansNewbie: I think `isCurrentBoardDirty` is sufficient as we only have 1 open board. Update header comment as well.

src/test/java/guitests/BoardTests.java
```diff
@@ -265,6 +265,8 @@ public void boards_panelCount_nothingHappensWhenNewBoardIsCancelled() {
         assertEquals(1, panelControl.getPanelCount());
 
         traverseMenu("Boards", "New");
+        waitUntilNodeAppears("No");
+        clickOn("No");
         press(KeyCode.ESCAPE);
```
HansNewbie: Is escape still needed here?

src/test/java/guitests/BoardTests.java
```diff
@@ -277,15 +279,15 @@ public void boards_panelCount_newBoardCreated() {
         PanelControl panelControl = ui.getPanelControl();
 
         traverseMenu("Boards", "New");
-        waitUntilNodeAppears(hasText("OK"));
-        clickOn("OK");
+        waitUntilNodeAppears(hasText("No"));
+        clickOn("No");
```
HansNewbie: Should put a comment why we click `No` because it is not obvious. When you want a new board to be created, when we click `New` it seems more intuitive to choose `Yes`, e.g. the prompt is `Your current board is not saved. Open a new board anyway?`.

src/main/java/ui/MenuControl.java
```diff
@@ -89,47 +87,71 @@ private Menu createFileMenu() {
...
-     * Creates an empty board. User will be prompted to
-     * confirm the action if there are unclosed panels
+     * Prompts a dialog to ask the user whether or not to save the current board.
      */
-    public final void newBoard() {
-        if (!isNewBoardCreationDialogConfirmed()) {
-            logger.info("New board creation cancelled");
+    public final void promptToSaveCurrentBoard() {
+        boolean shouldSave = DialogMessage.showYesNoConfirmationDialog("Save Changes?",
+                "There are unsaved changes to your current board.",
```
HansNewbie: I don't think the message should be this.

While it might be true, this will need us to check if board is dirty before calling this method. The line below this (All unsaved changes will be discarded) should be enough.

src/main/java/ui/MenuControl.java
```diff
@@ -89,47 +87,71 @@ private Menu createFileMenu() {
...
-        saveBoardAs();
+        saveBoard();
+        logger.info("Changes to the current board saved.");
     }
 
     /**
-     * Returns a boolean indicating whether we can proceed with
-     * the new board creation. If there are no panels open, it
-     * returns true; else a dialog is shown for the user to confirm.
+     * Prompts a {@code BoardNameDialog} to ask the user for a new board name
```
HansNewbie: Better to use `@link`.

Reference on how to use:
- http://stackoverflow.com/questions/5915992/javadoc-writing-links-to-methods
- http://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html#CHDDIECH

src/main/java/ui/MenuControl.java
```diff
@@ -89,47 +87,71 @@ private Menu createFileMenu() {
...
+     * @return the user's response
      */
-    private boolean isNewBoardCreationDialogConfirmed() {
-        List<PanelInfo> panelList = panels.getCurrentPanelInfos();
-        if (panelList.isEmpty()) {
-            return true;
+    public final Optional<String> promptForNewBoardName() {
```
HansNewbie: Maybe `promptForBoardName`? If we use this method for renaming board in the future, the `New` will be irrelevant.

src/main/java/ui/MenuControl.java
HansNewbie: Okay for the name. Let me see first about the implementation thingy.

HansNewbie: Okay for implementation as well.

HansNewbie:
Mostly cosmetic change.

1 thing needed: test for the confirmation dialog. Look at `guitests>DialogMessageTests`.

src/test/java/guitests/BoardTests.java
xinan: The prompt is actually something like `Do you want to save the current board before open a new one?` instead of `Open a new board anyway?`. In other words, no matter whether `Yes` or `No` is chosen, a new board will be created. Just the matter of whether the current board is saved.

HansNewbie: Sure. Jut put that as a comment for readers to understand.

HansNewbie:
Update me when ready for another review.

xinan:
@HansNewbie Yes, it is ready!

src/main/java/ui/MenuControl.java
```diff
@@ -89,47 +87,71 @@ private Menu createFileMenu() {
...
-        saveBoardAs();
+        saveBoard();
+        logger.info("Changes to the current board saved.");
     }
 
     /**
-     * Returns a boolean indicating whether we can proceed with
-     * the new board creation. If there are no panels open, it
-     * returns true; else a dialog is shown for the user to confirm.
+     * Prompts a {@link BoardNameDialog} to ask the user for a new board name
```
HansNewbie: Update comment to `... for a board name`.

src/main/java/ui/MenuControl.java
```diff
@@ -89,47 +87,71 @@ private Menu createFileMenu() {
...
+     * @return the user's response
      */
-    private boolean isNewBoardCreationDialogConfirmed() {
-        List<PanelInfo> panelList = panels.getCurrentPanelInfos();
-        if (panelList.isEmpty()) {
-            return true;
+    public final Optional<String> promptForBoardName() {
+        BoardNameDialog dlg = new BoardNameDialog(prefs, mainStage);
+        Optional<String> response = dlg.showAndWait();
```
HansNewbie: Can we do `return dlg.showAndWait()`

src/main/java/ui/MenuControl.java
```diff
@@ -89,47 +87,71 @@ private Menu createFileMenu() {
...
+
+        return response;
+    }
+
+    /**
+     * Logs the user action and creates a new board
```
HansNewbie: Remove `Logs the user action and...`. Logging is irrelevant to method description.

src/main/java/ui/MenuControl.java
```diff
@@ -170,24 +193,25 @@ private void onBoardSaveAs() {
     }
 
     /**
-     * Prompts a BoardNameDialog and saves the current board based on its response
+     * Prompts for board name and save the current board as per the name
```
HansNewbie: `saves` is correct :p

src/test/java/guitests/DialogMessageTests.java
```diff
@@ -40,6 +40,25 @@ public void showYesNoWarningDialogTest() throws ExecutionException, InterruptedE
     }
 
     @Test
+    public void showYesNoConfirmationDialogTest() throws ExecutionException, InterruptedException {
```
HansNewbie: Change this test name to follow the convention. `showYesNoConfirmationDialog_tryYesAndNo_getCorrectValue` or something similar.

src/main/java/ui/MenuControl.java
```diff
@@ -91,59 +89,82 @@ private Menu createFileMenu() {
...
+        if (!prefs.getLastOpenBoardPanelInfos().isPresent()) {
+            return true;
+        }
+        return !panels.getCurrentPanelInfos().equals(prefs.getLastOpenBoardPanelInfos().get());
     }
 
     /**
-     * Creates an empty board. User will be prompted to
-     * confirm the action if there are unclosed panels
+     * Prompts a dialog to ask the user whether or not to save the current board.
```
damithc: The method is doing more than just prompting. Both method name and comment misleading.

src/test/java/guitests/DialogMessageTests.java
```diff
@@ -40,6 +40,25 @@ public void showYesNoWarningDialogTest() throws ExecutionException, InterruptedE
     }
 
     @Test
+    public void showYesNoConfirmationDialog_tryYesAndNo_getCorrectValue() throws ExecutionException, InterruptedException {
+        FutureTask<Boolean> yesTask = new FutureTask<>(() ->
+                DialogMessage.showYesNoConfirmationDialog("Confirm", "Confirm Header", "Confirm Message", "yEs", "nO"));
+        Platform.runLater(yesTask);
+        waitUntilNodeAppears(hasText("Confirm Header"));
+        assertNodeExists(hasText("Confirm Header"));
+        clickOn("yEs");
```
damithc: Any idea why this weird capitalization in `yEs`?

HansNewbie: I think it was to test that we can get the result that the button was
clicked no matter the content of the button (like more than just Yes or No)

damithc: Looks weird. Should we keep it that way?

xinan: I think it also tests that the button text are not hardcoded as `Yes` and `No` but can vary according to the arguments?

HansNewbie: I guess we can change.

src/main/java/ui/MenuControl.java
xinan: I can't think of a better name. Do you have any suggestion how to name such methods?

HansNewbie: We can save after the prompt? Don't save in this method, I mean.

src/test/java/guitests/DialogMessageTests.java
damithc: If it serves a purpose, we should at least make it clear (add a comment?)

HansNewbie: I was thinking usin something more readable, say 1 test is `yes/no`,
another is `Non-Standard-Yes/Non-Standard-No`. But a comment works, too.

src/main/java/ui/MenuControl.java
xinan: So I make it return a `boolean`?

HansNewbie: That's what I have in mind.

src/main/java/ui/MenuControl.java
```diff
@@ -91,59 +89,87 @@ private Menu createFileMenu() {
...
+     *
+     * @return true if user choose to save, false otherwise
      */
-    public final void newBoard() {
-        if (!isNewBoardCreationDialogConfirmed()) {
-            logger.info("New board creation cancelled");
-            return;
-        }
+    public final boolean promptToSaveCurrentBoard() {
```
damithc: A better name is `isUserAgreeableToSavingBorad()`

src/main/java/ui/MenuControl.java
```diff
@@ -91,59 +89,87 @@ private Menu createFileMenu() {
...
-        Alert dlg = new Alert(AlertType.CONFIRMATION, "");
-        dlg.initModality(Modality.APPLICATION_MODAL);
-        dlg.setTitle("Confirmation");
-        dlg.getDialogPane().setHeaderText("Create a new board");
-        dlg.getDialogPane().setContentText("Are you sure you want to create a new board?"
-                                           + " All unsaved changes to the current board will be lost.");
-        Optional<ButtonType> response = dlg.showAndWait();
-        return response.isPresent() &&
-                response.get().getButtonData() == ButtonData.OK_DONE;
+        if (shouldSave) {
```
damithc: Can be `if (isCurrentBoardDirty() && isUserAgreeableToSave())`

damithc:
@HansNewbie any more comments?

HansNewbie:
Looks good. No more comments.

PR #1486, opened 2016-04-06
repo picker: move 'add repo' functionality to a menu item or a separate button. #1468

### LINKED ISSUE #1468: repo picker: move 'add repo' functionality to a menu item or a separate button.
I think pickers should not facilitate a 'add' function, at least not the way it does now. Either we should move the 'add repo' function to a new menu item or provide a 3rd button to add a new repo within the repo picker. i.e. type new repo name and click 'add'.
The current implementation prevents us from picking a repo by typing a few letters only.

raisfathin:


raisfathin:
Ready for review!

damithc:
Can post a screenshot?

raisfathin:
Dialog asking for new repository:
screen shot 2016-04-06 at 21 50 39

When the user enters an invalid GitHub repo:
screen shot 2016-04-06 at 21 51 11

src/main/java/ui/components/pickers/RepositoryPicker.java
```diff
@@ -15,23 +15,23 @@
...
+        new RepositoryPickerDialog(storedRepos, this::pickRepository, (repoId) -> ui.logic.isRepositoryValid(repoId));
+    }
+
+    private void pickRepository(String repoId) {
         String defaultRepo = ui.logic.getDefaultRepo();
-        RepositoryPickerDialog dialog = new RepositoryPickerDialog(storedRepos, stage);
-        Optional<String> chosenRepo = dialog.showAndWait();
-        if (chosenRepo.isPresent() && !defaultRepo.equals(chosenRepo.get())) {
-            onValueChangeCallback.accept(chosenRepo.get());
+        if (repoId != null && !defaultRepo.equals(repoId)) {
```
dariusf: Just realised the repo picker isn't a Dialog subclass, and so doesn't have access to `showAndWait`.

Instead of using null, could we pass an optional instead? That's a bit of an abuse of optionals (their use case is as return values, not parameters), but since we're not using `showAndWait` I guess it's fine.

src/main/java/ui/components/pickers/RepositoryPickerDialog.java
```diff
@@ -136,18 +165,76 @@ private void createUserInputTextField() {
...
+        dialog.getEditor().textProperty().addListener((observable, oldValue, newValue) -> {
+            String repositoryId = Utility.removeAllWhitespace(newValue);
+            if (!repositoryId.equals(newValue)) {
+                dialog.getEditor().setText(repositoryId);
+                return;
+            }
+        });
+        Optional<String> userInput = dialog.showAndWait();
+        if (!userInput.isPresent()) {
+            return Futures.unit(false);
```
dariusf: We can use `CompletableFuture.completedFuture` (from the standard library) here. This `unit` method does the same and is a bit of a legacy thing.

src/main/java/ui/components/pickers/RepositoryPicker.java
raisfathin: It was, but apparently I couldn't get it to spawn another dialog (the addRepository dialog one). So I had to make it not subclassing Dialog. Is it okay?

raisfathin:
Done. Ready for another review!

src/main/java/ui/components/pickers/RepositoryPicker.java
dariusf: Yeah, it's fine.

PR #1493, opened 2016-04-07
Change search and ranking algorithm for Issue Picker #1436

### LINKED ISSUE #1436: Changes search and ranking algorithm for Issue Picker
## Current matching algorithm 
Performs a `AND` filter over space-delimited query e.g `this is a test issue and`.

## Proposed changes
Performs a `OR` filter and rank matching issues based on number of matches with the query. 
Given a query `this is a test issue` and the following issue titles `this`, `this is` and `a is issue`. This will be the ranking:
- `a is issue`
- `this is`
- `this`

jinified:
Fixes #1436

### DIFF
--- a/src/main/java/backend/resource/TurboIssue.java
+++ b/src/main/java/backend/resource/TurboIssue.java
@@ -380,7 +380,7 @@ public class TurboIssue {
         List<String> queries = Arrays.asList(query.split("\\s"));
         return issues.stream()
                 .filter(i -> Utility.containsIgnoreCase(i.getMatchableText(), queries))
-                .sorted((i1, i2) -> i2.getNumberOfMatchingQuery(queries) - i1.getNumberOfMatchingQuery(queries))
+                .sorted((i1, i2) -> i2.getMatchingQueryCount(queries) - i1.getMatchingQueryCount(queries))
                 .collect(Collectors.toList());
     }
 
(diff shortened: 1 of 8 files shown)


jinified:
Ready for review @dariusf

src/main/java/backend/resource/TurboIssue.java
```diff
@@ -390,6 +392,23 @@ public TurboIssue combineWithPullRequest(PullRequest pullRequest) {
         return getMatchedIssues(issues, query).stream().findFirst();
     }
 
+    /**
+     * @param queries
+     * @return number of matches to a list of query
+     */
+    private int getNumberOfMatchingQuery(List<String> queries) {
+        return (int) queries.stream()
```
dariusf: We can use [toIntExact](http://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#toIntExact-long-) here.

Just realised we have `Utility. safeLongToInt` also. That can probably be removed now if it does the same thing.

jinified:
@dariusf Changes were made and managed to remove `Utility.safeLongToInt` method from codebase

src/main/java/backend/resource/TurboIssue.java
```diff
@@ -391,6 +393,22 @@ public TurboIssue combineWithPullRequest(PullRequest pullRequest) {
         return getMatchedIssues(issues, query).stream().findFirst();
     }
 
+    /**
+     * @param queries
+     * @return number of matches to a list of query
+     */
+    private int getNumberOfMatchingQuery(List<String> queries) {
```
dariusf: Maybe this could be named `getMatchingQueryCount`? It's the number of queries which would match this issue. That can be in the comment as well possibly.

src/main/java/backend/resource/TurboIssue.java
```diff
@@ -391,6 +393,22 @@ public TurboIssue combineWithPullRequest(PullRequest pullRequest) {
...
+        return Math.toIntExact(
+            queries.stream().filter(query -> Utility.containsIgnoreCase(getMatchableText(), query)).count());
+    }
+
+    /**
+     * @return string that can be matched by any query
```
dariusf: I think we can explain this a little more. This is like a string representation of issues that all queries are matched against.

jinified:
@dariusf  Ready for another round of review

jinified:
Merged with master.

dariusf:
Merged in f540ba8893689e489c1ccd0dba2a935b809b280f.

PR #1504, opened 2016-04-09
Author qualifier does not consider real name #1489

### LINKED ISSUE #1489: Author qualifier does not consider real name

raisfathin:
Fixes #1489

### DIFF
--- a/src/main/java/filter/expression/Qualifier.java
+++ b/src/main/java/filter/expression/Qualifier.java
@@ -364,7 +364,7 @@ public class Qualifier implements FilterExpression {
         case LABEL:
             return labelsSatisfy(model, issue);
         case AUTHOR:
-            return authorSatisfies(issue);
+            return authorSatisfies(model, issue);
         case ASSIGNEE:
             return assigneeSatisfies(model, issue);
         case INVOLVES:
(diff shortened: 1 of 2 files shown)


raisfathin:
@dariusf currently, this PR assumes that in the future, the real name of non collaborators will be stored inside `Model`. There is another alternative which creates a `TurboUser` field inside `TurboIssue` to store the information of the creator instead of just their username.

Which one do you think is better?

dariusf:
The first one. The second will duplicate information, since TurboIssue is meant to be serialised.

The way we store data is (kind of) relational -- we store keys instead of nesting objects. It isn't the most pleasant to use, as we have to join the data explicitly now, and this should ideally be done behind the scenes. Hopefully we can look into improving the interface soon.

raisfathin:
I see. Is there anything else that I should add besides looking for author in `Model`?

dariusf:
Don't think so, at least for this issue.

src/main/java/filter/expression/Qualifier.java
```diff
@@ -873,15 +873,23 @@ private boolean assigneeSatisfies(IModel model, TurboIssue issue) {
...
 
-        String creator = issue.getCreator();
-        return creator.toLowerCase().contains(content.get().toLowerCase());
+        String content = this.content.get().toLowerCase();
+        String login = issue.getCreator().toLowerCase();
+        String name = ""; // default value, since TurboIssue does not keep track of creator's real name
+
+        if (author.isPresent()) {
+            name = author.get().getRealName() == null ? "" : author.get().getRealName().toLowerCase();
+        }
```
dariusf: We can simplify this part a bit:

```java
String name = author.map(TurboUser::getRealName)
    .orElse("").toLowerCase();
```

raisfathin:
Done! Ready for review.

PR #1506, opened 2016-04-10
repo picker: 'Add repo' dialog can be autofilled with text in the query box #1505

### LINKED ISSUE #1505: repo picker: 'Add repo' dialog can be autofilled with text in the query box
For example, if the user types a repo name in the query box, realizes that repo does not exist, and click 'add repository' button, we can autofill the text box in the dialog box with the same text user has entered in the query box. 
The text should be auto-selected for easy deletion in case that is not what the user wanted.

raisfathin:
Fixes #1505


ndt93:
@acruis Hey acruis, can help review this?

acruis:
Good tests.

PR #1507, opened 2016-04-10
Milestone picker: support up/down arrows #1412

### LINKED ISSUE #1412: Milestone picker: support up/down arrows
Let's say we want to reschedule an issue from current milestone to the next one. `M` `down` `enter` should accomplish that.

m133225:
Fixes #1412 .

### DIFF
--- a/src/main/java/ui/components/pickers/MilestonePickerDialog.java
+++ b/src/main/java/ui/components/pickers/MilestonePickerDialog.java
@@ -77,6 +77,22 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
         inputField.textProperty().addListener((observable, oldValue, newValue) -> {
             handleUpdatedInput(newValue);
         });
+        inputField.setOnKeyPressed(e -> {
+            switch (e.getCode()) {
+            case DOWN:
+                state.selectNextBestMatchingMilestone();
+                refreshUI(state);
+                e.consume();
+                break;
+            case UP:
+                state.selectPreviousBestMatchingMilestone();
+                refreshUI(state);
+                e.consume();
+                break;
+            default:
+                break;
+            }
+        });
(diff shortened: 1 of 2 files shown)


src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -77,6 +78,19 @@ private void setupKeyEvents() {
...
             handleUpdatedInput(newValue);
         });
+        this.
+        inputField.setOnKeyPressed(e -> {
+            if (e.getCode() == KeyCode.DOWN) {
+                e.consume();
+                state.selectNextBestMatchingMilestone();
+                refreshUI(state);
+            }
+            if (e.getCode() == KeyCode.UP) {
```
damithc: `else if`? or `switch`?

m133225:
Ready for review.

ndt93:
@dariusf Hi Darius, can help review this?

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -54,6 +54,50 @@ private void processMatchingMilestonesByQueryWord(List<PickerMilestone> mileston
...
+     * @param milestones
+     * @return -1 if there are no selected milestones
+     */
+    private int getSelectedIndex(List<PickerMilestone> milestones) {
+        for (int i = 0; i < milestones.size(); i++) {
+            if (milestones.get(i).isSelected()) {
+                return i;
+            }
+        }
+        return -1;
```
dariusf: Is it possible that nothing is selected? If so, the return type should be optional (-1 is kind of ad hoc). If it's not possible, there should be an assertion failure.

m133225:
@dariusf Changed the method to return an optional!

PR #1508, opened 2016-04-10
Show loading indicator when there is a change of board #1233

### LINKED ISSUE #1233: Show loading indicator when there is a change of board

m133225:
Fixes #1233.


m133225:
Ready for review.

PR #1512, opened 2016-04-11
issue picker: issues shown are from a different repo #1511

### LINKED ISSUE #1511: issue picker: issues shown are from a different repo
image

The default repo at the time was hubturbo, but the issue shown is from teammates

jinified:
Fixes #1511 

Simple fix by using issues only from default repository. Here is the result after applying the fix:

correct_repo

### DIFF
--- a/src/main/java/ui/components/issuepicker/IssuePickerDialog.java
+++ b/src/main/java/ui/components/issuepicker/IssuePickerDialog.java
@@ -116,9 +116,8 @@ public class IssuePickerDialog extends Dialog<String> {
     private void populateSelectedIssues(Optional<TurboIssue> selectedIssue) {
         selectedIssues.getChildren().clear();
         FlowPane selectedIssueCards = getSelectedIssuesPane(selectedIssue);
-        if (!selectedIssueCards.getChildren().isEmpty()) {
-            selectedIssues.getChildren().addAll(createRepoTitle(models.getDefaultRepo()), selectedIssueCards);
-        }
+        selectedIssue.ifPresent(issue ->
+            selectedIssues.getChildren().addAll(createRepoTitle(issue.getRepoId()), selectedIssueCards));
     }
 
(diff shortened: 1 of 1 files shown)


jinified:
Ready for review

dariusf:
Looks good. Are tests needed?

jinified:
I don't think we will need tests for this since it is guaranteed that the issues are from default repo programmatically.

damithc:
I know this started as a bug, but can we make it a feature? :-p
i.e. make it search all repos involved in the current board?

jinified:
@damithc Technically that is what I wanted to do. Just need to fix the wrong repository title right in this case but same behaviour as the one mentioned in the issue

damithc:
Yes, let's fix the repo title shown in the issue picker.

jinified:
Requested changes were made

src/main/java/ui/components/issuepicker/IssuePickerDialog.java
```diff
@@ -117,7 +117,7 @@ private void populateSelectedIssues(Optional<TurboIssue> selectedIssue) {
         selectedIssues.getChildren().clear();
         FlowPane selectedIssueCards = getSelectedIssuesPane(selectedIssue);
         if (!selectedIssueCards.getChildren().isEmpty()) {
-            selectedIssues.getChildren().addAll(createRepoTitle(models.getDefaultRepo()), selectedIssueCards);
+            selectedIssues.getChildren().addAll(createRepoTitle(selectedIssue.get().getRepoId()), selectedIssueCards);
```
dariusf: If the selected issue is guaranteed to be here at this point, the method should take it directly instead of an optional.

dariusf: Okay.

jinified:
Ready for review

damithc:
Does this fix the problem of repo shown in the dialog not matching the repo the issue is from?

jinified:
It fixes the displayed repo id to match repo id of selected issue. Previously, it constantly shows the default repo

PR #1515, opened 2016-04-11
Issue count is not showing up on all panels #1501

### LINKED ISSUE #1501: Issue count is not showing up on all panels
image

yl-coder:
Fixes #1501 

Exact scenario to reproduce the problem:
1) On a panel with issues, click on any of them.
2) On the same panel, key in a filter expression that will result in zero issues showing up.
3) Click on the same panel listview again for a few times (like you are clicking an empty issue in the empty panel)
4) key in a filter expression that will result in non-empty panel.
5) issue count not shown.

NavigableListView.java: Line 113
` int nextIndex = Math.min(selectedIndex.get(), lastIndex);`

selectedIndex.get() throws an exception as the value is null. Therefore the issue count is not updated.
I believe the selectedIndex was set to null when there is an empty selection on the panel listview.

### DIFF
--- a/src/main/java/ui/components/NavigableListView.java
+++ b/src/main/java/ui/components/NavigableListView.java
@@ -96,19 +96,19 @@ public abstract class NavigableListView<T> extends ScrollableListView<T> {
         }
         boolean itemFound = index > -1;
 
+        // Figure out what is selected based on previous and current state
         if (itemFound) {
-            // Select that item
+            // Maintain selection
             getSelectionModel().clearAndSelect(index);
             selectedIndex = Optional.of(index);
             // Do not trigger event; selection did not conceptually change
         } else {
             // The item disappeared
-            if (getItems().size() == 0) {
-                // No more items in the list
+            if (getItems().isEmpty()) {
+                // Nothing can be selected
                 selectedIndex = Optional.empty();
-            } else {
-                // The list is non-empty, so we can be sure that we're selecting something
-                // The current index is the same as the next, due to the item disappearing
+            } else if (selectedIndex.isPresent()){
+                // Consider what is currently selected to be the new selection
                 int lastIndex = getItems().size() - 1;
                 int nextIndex = Math.min(selectedIndex.get(), lastIndex);
 


yl-coder:
Ready for review.

dariusf:
Hmm, I can't seem to reproduce this. Step 3 doesn't change the value of `selectedIndex`. Where does the empty `selectedIndex` in that block come from?

yl-coder:
apr-14-2016 12-52-08

Right, Step 3 doesn't change the value of selectedIndex. I think the selectedIndex was empty when switching to another repo.

yl-coder:
This assumption is not true i guess.

`Line 110 : // The list is non-empty, so we can be sure that we're selecting something`

dariusf:
Yes, I think you're right.

src/main/java/ui/components/NavigableListView.java
```diff
@@ -103,7 +103,7 @@ public void restoreSelection() {
             // Do not trigger event; selection did not conceptually change
         } else {
             // The item disappeared
-            if (getItems().size() == 0) {
+            if (getItems().size() == 0 || !selectedIndex.isPresent()) {
```
dariusf: Rather than do this, which is a bit odd given the body (if index is empty, set index to empty), we can move this to the `else` branch instead.

The comments here can also be improved, to say what we are doing.
```java
// Figure out what is selected based on previous and current state
if (itemFound) {
    // Maintain selection
} else {
    if (getItems().isEmpty()) {
        // Nothing can be selected
    } else if (selectedIndex.isPresent()) {
        // Consider what is currently selected to be the new selection
    }
}
```

src/main/java/ui/listpanel/ListPanel.java
```diff
@@ -190,14 +190,14 @@ public final void refreshItems() {
...
+
         issuesCount = getElementsList().size();
         closedIssuesCount = getClosedIssuesCount();
         openIssuesCount = issuesCount - closedIssuesCount;
-        listView.restoreSelection();
-        this.setId(IdGenerator.getPanelId(panelIndex));
         updateFooter();
 
-
+        listView.restoreSelection();
```
dariusf: Does the relative order of this and `updateFooter` matter? Feels like it shouldn't.

src/main/java/ui/components/NavigableListView.java
yl-coder: ok noted. changed.

yl-coder:
Ready for review

src/main/java/ui/listpanel/ListPanel.java
yl-coder: yes right. It doesn't matter.

dariusf:
Thanks for fixing this. Could you squash that last commit? Will merge after that.

yl-coder:
Np. :) done

PR #1521, opened 2016-04-13
watch list: canceling 'add to new watch list' results in new panel #1519

### LINKED ISSUE #1519: watch list: canceling 'add to new watch list' results in new panel
Steps:
1. Select an issue and choose 'add to watch list --> new watch list'
2. Choose 'cancel' in the confirmation dialog

Expected: no new panel created
Actual: a new panel with the default name is created

raisfathin:
Fixes #1519

### DIFF
--- a/src/test/java/unstable/ContextMenuTests.java
+++ b/src/test/java/unstable/ContextMenuTests.java
@@ -169,6 +169,26 @@ public class ContextMenuTests extends UITest {
     }
 
     @Test
+    public void addToWatchListPanel_cancelOperation_noNewPanelCreated() {
+        ListPanel listPanel = getPanel(0);
+        assertEquals(1, listPanel.getParentPanelControl().getPanelCount());
+
+        clickIssue(0, 9);
+        traverseContextMenu(listPanel.getContextMenu(), "Add to watch list", "New watch list");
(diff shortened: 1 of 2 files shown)


src/test/java/unstable/ContextMenuTests.java
```diff
@@ -169,6 +169,25 @@ public void addToWatchListPanel_createNewWatchListPanel_issueAddedToNewPanel() {
     }
 
     @Test
+    public void addToWatchListPanel_cancelOperation_operationCancelled() {
```
HansNewbie: Can we rename the last part to `noNewPanelCreated()`.

src/test/java/unstable/ContextMenuTests.java
```diff
@@ -169,6 +169,25 @@ public void addToWatchListPanel_createNewWatchListPanel_issueAddedToNewPanel() {
...
+        ListPanel listPanel = getPanel(0);
+
```
HansNewbie: We should `assert` here that the panel count is 1 in the first place.

raisfathin:
Done!

HansNewbie:
Looks good.

PR #1532, opened 2016-04-15
milestone picker: up/down keys don't work if the query is empty #1531

### LINKED ISSUE #1531: milestone picker: up/down keys don't work if the query is empty
up/down keys cannot be used to navigate the top 5 matches if the query box is empty

m133225:
Fixes #1531.

Note that the new behaviour does not allow unselection of a milestone using up/down arrow keys. The user has to type and backspace again to do so (or simply close the dialog using `esc`).


m133225:
Ready for review.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -55,11 +55,14 @@ private void processMatchingMilestonesByQueryWord(List<PickerMilestone> mileston
...
      * Selects the next milestone in the list of best matching milestones
+     *
+     * If there is no currently selected milestone, the next milestone will be the first one in
+     * the list
      */
     public void selectNextBestMatchingMilestone() {
         Optional<Integer> curSelectedIndex = getSelectedIndex(bestMatchingMilestones);
         if (!canIncreaseIndex(curSelectedIndex, bestMatchingMilestones)) return;
-        selectMilestoneAtIndex(bestMatchingMilestones, curSelectedIndex.get() + 1);
+        selectMilestoneAtIndex(bestMatchingMilestones, curSelectedIndex.isPresent() ? curSelectedIndex.get() + 1 : 0);
```
ndt93: Can use `curSelectedIndex.orElse(-1) + 1` here.

src/main/java/ui/components/pickers/MilestonePickerState.java
```diff
@@ -72,7 +75,7 @@ public void selectPreviousBestMatchingMilestone() {
     }
 
     private boolean canIncreaseIndex(Optional<Integer> curSelectedIndex, List<PickerMilestone> list) {
-        return curSelectedIndex.isPresent() && curSelectedIndex.get() < list.size() - 1;
+        return !curSelectedIndex.isPresent() || curSelectedIndex.get() < list.size() - 1;
```
ndt93: Can use `Optional.filter` or `Optional.flatMap` here.

ndt93:
Is there any test to cover this case?

src/main/java/ui/components/pickers/MilestonePickerState.java
m133225: Hmm, not too sure on how to use `filter` or `flatMap` here? The resulting value would not be a boolean for returning.

ndt93: We can use `filter` like: `return curSelectedIndex.filter((index) -> index < list.size() - 1).isPresent()`. `flatMap` can also be used in similar way.

But you can keep it as it is if you think the current way is more readable.

m133225:
@ndt93 Ready for review. Made some changes and added a test to cover this!

Issue #1525, opened 2016-04-13
milestone picker: align right edge


damithc:
As per screenshot
![image](https://cloud.githubusercontent.com/assets/1673303/14496025/6e14e738-01c4-11e6-9dee-09d0a4967a88.png)

PR #1535, opened 2016-05-05
milestone picker: use lighter lines #1527

### LINKED ISSUE #1527: milestone picker: use lighter lines
These lines are not very important, so should be in a lighter shade.

image

garbanzos:
Fixes #1527. 

image

### DIFF
--- a/src/main/java/ui/components/pickers/MilestonePickerDialog.java
+++ b/src/main/java/ui/components/pickers/MilestonePickerDialog.java
@@ -167,14 +167,15 @@ public class MilestonePickerDialog extends Dialog<MilestonePickerDialogResponse>
 
     private ScrollPane createAllMilestonesScrollPane() {
         VBox milestonesBox = new VBox();
-        milestonesBox.setStyle("-fx-border-radius: 3;-fx-background-color: white;-fx-border-color: black;");
+        milestonesBox.setPrefWidth(398);
+        milestonesBox.setStyle("-fx-background-color: white;");
         originalMilestones.forEach(milestone ->
                 milestonesBox.getChildren().add(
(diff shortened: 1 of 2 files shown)


src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -167,14 +167,15 @@ private void initUI() {
...
+        milestonesBox.setPrefWidth(398);
+        milestonesBox.setStyle("-fx-background-color: white;");
         originalMilestones.forEach(milestone ->
                 milestonesBox.getChildren().add(
                 setMouseClickForNode(milestone.getDetailedMilestoneNode(), milestone.getTitle()))
         );
 
         ScrollPane milestonesScrollPane = new ScrollPane();
-        milestonesScrollPane.setMaxHeight(163);
+        milestonesScrollPane.setMaxHeight(160);
```
m133225: Actually, could this value be the same as the `Search results`'s i.e. set to `height * no. of milestones`?

src/main/java/ui/components/pickers/MilestonePickerDialog.java
```diff
@@ -167,14 +167,15 @@ private void initUI() {
 
     private ScrollPane createAllMilestonesScrollPane() {
         VBox milestonesBox = new VBox();
-        milestonesBox.setStyle("-fx-border-radius: 3;-fx-background-color: white;-fx-border-color: black;");
+        milestonesBox.setPrefWidth(398);
```
m133225: Does this currently affect the visibility of the milestones? If not, I think width changes should be handled in #1525 instead.

m133225:
Some clarifications and changes requested.

m133225:
Seems good. Not sure if the border radii should be removed though.

